PR #10, opened 2017-08-14
[WIP] swig python interface

mrpollo:
adds the cmake build step for the interface file that exposes `DroneLink` into python

```
make clean
make default
```

TODO:
- [x] Figure out why the generated output is missing bindings from Python.h
- [x] build dronelink.py
- [x] install dronelink py
- [ ] add all plugins
- [ ] write a simple python test for dronelink.py (takeoff / goto / land perhaps)
- [ ] docs


mrpollo:
we won't be using swig for python

PR #34, opened 2017-08-22
Initial changes to generate Java bindings

avinash-palleti:
It is currently generating JAVA bindings for `android_x86` arch. For other arch it is having an issue as `dronecode.i` is having hardcoded build directory name as below,

`%include "../build/android_x86/include/device_plugin_container.h";`


mrpollo:
no more swig :*(

Thanks for your help @avinash-palleti

PR #35, opened 2017-08-22
Add travis CI build

julianoes:
This should enable the default build using tavis-ci.


TSC21:
@julianoes should we get this in?

PR #37, opened 2017-08-24
Python with Pybind11

mrpollo:
all credit here goes to @julianoes I only cleaned up the Make files

## Build
```
make python
```

## Install (to site-packages)
```
make python install
```

## Tests
**NOTE:** Tests will only run if you have SITL running
```
make python_test
```


python_pybind_test.py
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env python
```
hamishwillee: @mrpollo @julianoes This is functionally the same example as for C++. We're going to want it as an example, in a sensible folder structure that identifies the test programming language and the test name (allowing us to have multi-file tests). Ideally this should match the C++ test structure.

julianoes:
Note to self: need to upload the new Docker image for the Python dependency.

python_pybind_test.py
mrpollo: agreed, I added a commit to move it

mrpollo:
Updated this PR, python directory will hold all python files from now on, we only have a test for now, this is no longer a WIP.

**NOTE** if we decide to move on with gRPC we should figure out what it means for pybind.

mrpollo:
@julianoes we need to decide if we are going to be using pybind11 or not, I think it's a great feature to have, but if we are not going to be able to properly maintain it we shouldn't add it, I also think gRPC would deprecate this, whats your take?

julianoes:
@mrpollo you are right. It looks like gRPC will take care of this.

Issue #21, opened 2017-08-21
Add support for TCP connections


julianoes:
It would be nice to support for TCP connections.

The implementation should be done in a similar way as it is done for UDP connections.
The skeleton/stub is here:
https://github.com/dronecore/DroneCore/blob/master/core/tcp_connection.h
https://github.com/dronecore/DroneCore/blob/master/core/tcp_connection.cpp

julianoes:
Added in #59.

Issue #20, opened 2017-08-21
Add support for serial connections


julianoes:
It would be nice to support serial connections.

The implementation should be done in a similar way as it is done for UDP connections.
The skeleton/stub is here:
https://github.com/dronecore/DroneCore/blob/master/core/serial_connection.h
https://github.com/dronecore/DroneCore/blob/master/core/serial_connection.cpp

julianoes:
Added in #59.

PR #38, opened 2017-08-24
Support for Tcp Connection

Rjasuja:
Issue : #21

### DIFF
--- a/core/dronecore.cpp
+++ b/core/dronecore.cpp
@@ -4,6 +4,7 @@
 #include "connection.h"
 #include "udp_connection.h"
 #include "tcp_connection.h"
+#include "serial_connection.h"
 
 
 namespace dronecore {
@@ -41,12 +42,13 @@ DroneCore::ConnectionResult DroneCore::add_udp_connection(int local_port_number)
 
 DroneCore::ConnectionResult DroneCore::add_tcp_connection()
 {
-    return add_tcp_connection("127.0.0.1",0);
+    return add_tcp_connection("", 0);
 }
 
-DroneCore::ConnectionResult DroneCore::add_tcp_connection(std::string ipAddress,int local_port_number)
+DroneCore::ConnectionResult DroneCore::add_tcp_connection(std::string remote_ip,
+                                                          int remote_port)
 {
-    Connection *new_connection = new TcpConnection(_impl,ipAddress, local_port_number);
+    Connection *new_connection = new TcpConnection(_impl, remote_ip, remote_port);
     DroneCore::ConnectionResult ret = new_connection->start();
 
     if (ret != DroneCore::ConnectionResult::SUCCESS) {
@@ -58,6 +60,25 @@ DroneCore::ConnectionResult DroneCore::add_tcp_connection(std::string ipAddress,
     return DroneCore::ConnectionResult::SUCCESS;
 }
 
+DroneCore::ConnectionResult DroneCore::add_serial_connection()
+{
+    return add_serial_connection("",0);
+}
+
+DroneCore::ConnectionResult DroneCore::add_serial_connection(std::string dev_path,
+                                                          int baudrate)
+{
+    Connection *new_connection = new SerialConnection(_impl, dev_path, baudrate);
+    DroneCore::ConnectionResult ret = new_connection->start();
+
+    if (ret != DroneCore::ConnectionResult::SUCCESS) {
+        delete new_connection;
+        return ret;
+    }
+
+    _impl->add_connection(new_connection);
+    return DroneCore::ConnectionResult::SUCCESS;
+}
 const std::vector<uint64_t> &DroneCore::device_uuids() const
 {
     return _impl->get_device_uuids();
--- a/core/dronecore.cpp
+++ b/core/dronecore.cpp
@@ -62,11 +62,11 @@ DroneCore::ConnectionResult DroneCore::add_tcp_connection(std::string remote_ip,
 
 DroneCore::ConnectionResult DroneCore::add_serial_connection()
 {
-    return add_serial_connection("",0);
+    return add_serial_connection("", 0);
 }
 
 DroneCore::ConnectionResult DroneCore::add_serial_connection(std::string dev_path,
-                                                          int baudrate)
+                                                             int baudrate)
 {
     Connection *new_connection = new SerialConnection(_impl, dev_path, baudrate);
     DroneCore::ConnectionResult ret = new_connection->start();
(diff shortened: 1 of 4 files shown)


Rjasuja:
As Dronecore uses Discovery mechanism, i could not test it. Please let me know the test procedure.

core/dronecore.cpp
```diff
@@ -38,6 +39,26 @@ DroneCore::ConnectionResult DroneCore::add_udp_connection(int local_port_number)
     return DroneCore::ConnectionResult::SUCCESS;
 }
 
+DroneCore::ConnectionResult DroneCore::add_tcp_connection()
+{
+    return add_tcp_connection("127.0.0.1", 0);
+}
+
+DroneCore::ConnectionResult DroneCore::add_tcp_connection(std::string ipAddress,
```
julianoes: I vote to keep variable naming consistent with underscores, so I suggest `ip_address` instead of `ipAddress`.

core/dronecore.cpp
```diff
@@ -38,6 +39,26 @@ DroneCore::ConnectionResult DroneCore::add_udp_connection(int local_port_number)
     return DroneCore::ConnectionResult::SUCCESS;
 }
 
+DroneCore::ConnectionResult DroneCore::add_tcp_connection()
+{
+    return add_tcp_connection("127.0.0.1", 0);
```
julianoes: Do we ever need not localhost? What is the idea of exposing this as an argument?

core/tcp_connection.cpp
```diff
@@ -1,36 +1,279 @@
...
+#endif
+
+    _socket_fd = socket(AF_INET, SOCK_STREAM, 0);
+
+    if (_socket_fd < 0) {
+        Debug() << "socket error" << GET_ERROR(errno);
+        return DroneCore::ConnectionResult::SOCKET_ERROR;
+    }
+
+    sockaddr_in local_addr;
```
julianoes: In C++ you can just initialize to 0 without memset doing: `sockaddr_in local_addr {};`

core/tcp_connection.cpp
```diff
@@ -1,36 +1,279 @@
...
+
+    std::string portName = std::to_string(port);
+    memset(&hints, 0, sizeof(hints));
+    hints.ai_family = AF_UNSPEC;
+    hints.ai_socktype = SOCK_STREAM;
+    hints.ai_protocol = IPPROTO_TCP;
+
+    int rc = getaddrinfo(ipAddress.c_str(), portName.c_str(), &hints, &result);
+    if (rc != 0) {
+        throw std::runtime_error("TcpClientPort getaddrinfo failed with error:\n");
```
julianoes: The idea was not to throw exceptions but rather return a `ConnectionError`.

core/tcp_connection.cpp
```diff
@@ -1,36 +1,279 @@
...
+            addr->sin_family = temp_ptr->sin_family;
+            addr->sin_addr.s_addr = temp_ptr->sin_addr.s_addr;
+            addr->sin_port = temp_ptr->sin_port;
+            ip_found = true;
+            break;
+        }
+    }
+    freeaddrinfo(result);
+    if (!ip_found) {
+        throw std::runtime_error("TcpClientPort could not resolve ip address for:\n");
```
julianoes: Same here.

core/tcp_connection.h
```diff
@@ -3,20 +3,42 @@
 #include "dronecore.h"
 #include "dronecore_impl.h"
 #include "connection.h"
+#include<sys/types.h>
```
julianoes: nitpick: space after `#include`.

hamishwillee: @julianoes Shouldn't `make fix_style` catch this sort of thing?

include/dronecore.h
```diff
@@ -34,6 +35,8 @@ class DroneCore
 
     ConnectionResult add_udp_connection();
     ConnectionResult add_udp_connection(int local_port_number);
+    ConnectionResult add_tcp_connection();
```
hamishwillee: @julianoes When you've worked out the doc standard, can you please make sure it gets added to this. Specifically, the default ip and port need to be documented.  Also if there is any internal timeout, what that is, and whether it can be configured.

hamishwillee:
@julianoes When will this go in? There will be a need to update the places where we say "currently only UDP connections are supported".

anitha-suresh:
@hamishwillee We plan to get this in by 28th/29th (early next week)

core/dronecore.cpp
shakthi-prashanth-m: @julianoes : Are we not going to deprecate **discover device** method and replace it with connecting to a specific IP/PORT ? If so, then we need to make similar changes to existing UDP solution too.. Is that right ?

julianoes: No we are not going to deprecate "discover device" as far as I know. Why would we do that? The thing is that you need to add a connection and then wait until you actually receive mavlink.

julianoes:
@Rjasuja please try to be a bit more specific in the git commit messages and actually describe the change that you are doing instead of just the topic this is about.

screen shot 2017-08-28 at 09 55 12

Also, check out: https://chris.beams.io/posts/git-commit/

include/dronecore.h
```diff
@@ -34,6 +35,8 @@ class DroneCore
...
+    ConnectionResult add_tcp_connection(std::string ipAddress, int local_port_number);
```
julianoes: Here, we're missing the `add_serial_connection()`, right?

Rjasuja: i'll add that as part of #20 .

julianoes: @Rjasuja I don't understand. Shouldn't it be part of this PR?

shakthi-prashanth-m: As there 2 issues raised for TCP & Serial connection support, @Rjasuja may be submitting different PR's for each. Both can be merged in the same PR.

julianoes: I only see this PR though 😄

Rjasuja: @julianoes I was in notion that Serial connection is another issue, all changes related can be submitted with the same. Do you want me to update here ?
I'll create PR for Serial connection support.

julianoes: Yes just update this PR until it's good to go in.

julianoes:
@Rjasuja thanks for adding the `add_serial_connection`.

Please fix the style issues, see: https://travis-ci.org/dronecore/DroneCore/jobs/269502432#L664-L689

Make sure you have astyle and run `make fix_style`

julianoes:
I'm now confused if this should be merged because the serial stuff is also added in #45.

julianoes:
Also, were you able to test this?

julianoes:
@Rjasuja what is the state here? And could you please resolve the conflicts with master?

Rjasuja:
Closing this, created a new PR : #52

PR #46, opened 2017-08-30
Sample application to explore gRPC

avinash-palleti:
Sample python application using gRPC to arm, takeoff and land.


julianoes:
Continued in #48, closing.

PR #48, opened 2017-09-03
Sample application to explore gRPC [2]

julianoes:
This continues @avinash-palleti's work from #46 to explore GRPC.

Todos:
- [x] Working prototype
- [x] Try sync interface for action
- [ ] Try async interface for action
- [ ] Improve mission interface
- [ ] Mabe split dronecore.proto into multiple proto files
- [ ] Integrate GRPC build steps into cmake build


julianoes:
@avinash-palleti feel free to continue working on this and pushing more commits to this branch. You should have access rights for it.

grpc/python_client/fly_mission.py
```diff
@@ -42,13 +42,16 @@ def run():
                                         yaw=-60,
                                         camera_action=0))
 
-    stub.AddMission(dc.Mission(mission_items=mission_items))
+    stub.SendMission(dc.Mission(mission_items=mission_items))
```
avinash-palleti: This will be a blocking call, as server is  waiting for call back from dronecore to write the result.

avinash-palleti: we can have similar async call as below.

grpc/server/dronecore_server.cpp
```diff
@@ -93,169 +75,47 @@ class DroneCoreRPCImpl final : public DroneCoreRPC::Service
...
-        }
-        if (_mission_state == MissionState::DONE) {
-            response->set_ret(true);
-        } else {
-            response->set_ret(false);
+        // TODO: there has to be a beter way than using std::future.
+        auto prom = std::make_shared<std::promise<void>>();
+        auto future_result = prom->get_future();
+
+        dc.device().mission().start_mission_async(
```
avinash-palleti: We need to provide another sync api for client to get status of Mission. That will help once we have mission update apis in place.

grpc/server/dronecore_server.cpp
```diff
@@ -93,169 +75,47 @@ class DroneCoreRPCImpl final : public DroneCoreRPC::Service
...
+        [prom, response](Mission::Result result) {
+            response->set_result(static_cast<dronecorerpc::MissionResult::Result>(result));
+            response->set_result_str(Mission::result_str(result));
```
avinash-palleti: we are directly writing result to rpc client, I think we need to keep track of return status in server as well, in case if user calls other mission apis asynchronously

mrpollo:
@julianoes are we ready to merge this in?

grpc/python_client/fly_mission.py
julianoes: Hm, ok so what is your suggestion? I'm not sure I understand your concern.

grpc/server/dronecore_server.cpp
julianoes: Why? Because you don't want to use `std::future`?

grpc/server/dronecore_server.cpp
julianoes: How would these return stati mix up?

julianoes:
@mrpollo if someone wants to do more reviewing, now is the time :).

grpc/python_client/fly_mission.py
```diff
@@ -0,0 +1,61 @@
...
+    thread_status = False
+
+
+def run():
+    global thread_status
+    channel = grpc.insecure_channel('0.0.0.0:50051')
+    stub = dronecore_pb2_grpc.DroneCoreRPCStub(channel)
+
+    mission_items = []
+    mission_items.append(dc.MissionItem(latitude=47.398170327054473,
```
hamishwillee: @julianoes Might be worth updating to more closely match our C++ example.

julianoes: Ok let me copy that over.

PR #51, opened 2017-09-06
Update build.md

Rjasuja:
1) Update build instructions for Library on Windows
2) Update build instructions for Example on Windows


hamishwillee:
FWIW I think we should instead update https://dronecore.gitbooks.io/dronecore-guide/content/en/contributing/build.html and link to it everywhere (ie delete this file)
@julianoes

julianoes:
Ok, we can delete this file. We don't want to have it double I agree.

hamishwillee:
OK, closing then. @Rjasuja this info is now in the guide. I'll delete this file and redirect the readme.

PR #52, opened 2017-09-06
Tcp and Serial connection Support, Fix Install Errors on Windows

Rjasuja:
Tcp Connection Support : Default Ip as 127.0.0.1 and port as 5760
Serial Connection Support(only for Linux) : Default Port as ttyS0 and Baudrate as 9600.

Fix Install Errors For Dronecore Library and Example
Fix install path for Dronecore library on Windows in CMakeLists
Fix Error flags for Takeoff_and_land Example on Windows

### DIFF
--- a/include/dronecore.h
+++ b/include/dronecore.h
@@ -45,6 +45,7 @@ public:
         TIMEOUT, /**< @brief %Connection timed out. */
         SOCKET_ERROR, /**< @brief Socket error. */
         BIND_ERROR, /**< @brief Bind error. */
+        SOCKET_CONNECTION_ERROR,
         CONNECTION_ERROR, /**< @brief %Connection error. */
         NOT_IMPLEMENTED, /**< @brief %Connection type not implemented. */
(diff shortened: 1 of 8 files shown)


include/dronecore.h
```diff
@@ -77,6 +78,40 @@ class DroneCore
     ConnectionResult add_udp_connection(int local_port_number);
 
     /**
+     * @brief Adds a TCP connection with the default arguments.
+     *
+     * This will connect to local TCP port 5760.
+     *
+     * @return The result of adding the connection.
+     */
+    ConnectionResult add_tcp_connection();
```
hamishwillee: Just For my information, why don't we have single function with default arguments for these connection APIs?

hamishwillee:
Docs text looks good to me (made some minor updates). Just for my information - why don't we have single function with default arguments for these connection APIs (rather than two methods).

hamishwillee:
When is this planned to be merged?

julianoes:
Rebased and fixed in #59.

PR #62, opened 2017-09-13
Update action docs to make clear how action commands work

hamishwillee:
@julianoes As discussed, this now makes it clear that the action commands send a message and complete with result of acknowledgement from vehicle - ie the actual "action" does not complete when a function returns.


hamishwillee:
Thanks very much. Done. Merged.

PR #64, opened 2017-09-13
Colour code errors in example

hamishwillee:
This makes errors show up a little more clear in the output. 

I also added a check for the vehicle being in an armable state, as I think this is a useful thing to show users. Note that none of the checks are necessarily realistic but good for the example (e.g. in real code I might way for armable state rather than exiting).


hamishwillee:
@julianoes Good idea. Defines added - better?

PR #68, opened 2017-09-13
grpc: WIP for async server

julianoes:
Just an ugly try really, not sure how to go from here.

FYI: @avinash-palleti


julianoes:
It turns out that the clients can also do async calls, so the server doesn't necessarily need to be async. Closing for now.

PR #69, opened 2017-09-14
Make examples have consistent buildfiles

hamishwillee:
The fly mission example was linux specific - this adds the required windows compatibility (copied from takeoff_and_land example). It also removes a spurious message for the cmake source directory from the takeoff example. 

1. If we have installed the DroneCore artifacts system wide will the example pick these up instead of the relative path to the include directory that I see defined mid-way up?
2. The instructions we given in  [Install Artifacts](http://localhost:4000/en/contributing/build.html#install-artifacts) state that to install files system wide you do the following when building DroneCore:
   ```
   INSTALL_PREFIX=/usr/local make default install
   ```
   - Won't this only work on linux?
   - Does cmake look for this location when creating make files for examples?
   - Is there a way to get this to work on windows or should I add note "not supported on Windows".
3. Do we have an intention to deploy DroneCore library as a binary?

### DIFF
--- a/example/fly_mission/CMakeLists.txt
+++ b/example/fly_mission/CMakeLists.txt
@@ -2,7 +2,12 @@ cmake_minimum_required(VERSION 2.8.12)
 
 project(fly_mission)
 
-add_definitions("-std=c++11 -Wall -Wextra -Werror")
+if(NOT MSVC)
+    add_definitions("-std=c++11 -Wall -Wextra -Werror")
+else()
+    add_definitions("-std=c++11 -WX -W2")
+    set(platform_libs "Ws2_32.lib")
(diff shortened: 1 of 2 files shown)


julianoes:
1. Yes it should.
2. We should support this, see #70.
    - Yes this only works on Linux. I don't know how these things work in Windows (and MSVC)
    - I think so.
    - I would say "currently not supported on Windows". Do we have a developer that could take on Windows support?
3. Yes, why not. Releases using github and travis should be easy.  #71

hamishwillee:
Thanks. Sounds great. I'll add request for Windows contributors to docs as well - https://github.com/dronecore/docs/issues/26

PR #74, opened 2017-09-17
Improve timeout functionality

julianoes:
This improves the timeout checker: 

- supports multiple timeouts per plugin/class
- is unit tested

Along with the change:
- a param set bugfix because of shadowing
- shadowing warnings

### DIFF
--- a/core/timeout_handler.cpp
+++ b/core/timeout_handler.cpp
@@ -1,0 +1,83 @@
+#include "timeout_handler.h"
+
+namespace dronecore {
+
+TimeoutHandler::TimeoutHandler()
+{
+}
+
+TimeoutHandler::~TimeoutHandler()
+{
+}
+
+void TimeoutHandler::add(std::function<void()> callback, double duration_s, void **cookie)
+{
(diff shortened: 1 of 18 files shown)


mrpollo:
is time using a monotonic clock?

julianoes:
> is time using a monotonic clock?

Yep, see https://github.com/dronecore/DroneCore/blob/master/core/global_include.cpp#L11
http://en.cppreference.com/w/cpp/chrono/steady_clock

core/timeout_handler.cpp
```diff
@@ -0,0 +1,86 @@
...
+{
+    std::lock_guard<std::mutex> lock(_timeouts_mutex);
+
+    auto it = _timeouts.find((void *)cookie);
+    if (it != _timeouts.end()) {
+        _timeouts.erase((void *)cookie);
+    }
+}
+
+void TimeoutHandler::run_once()
```
darioxz: Nice and testable code :+1:

mrpollo:
thanks @julianoes @darioxz

PR #75, opened 2017-09-17
Improve debugging printfs

julianoes:
This improves the log output. Instead of only one `Debug()` printf for everything, there are now proper log levels: `LogDebug()`, `LogInfo()`, `LogWarn()`, `LogErr()`.

It also adds file name and file number to the printfs which should help to figure out where stuff went wrong.

### DIFF
--- a/core/log.h
+++ b/core/log.h
@@ -1,0 +1,52 @@
+#pragma once
+
+#include <sstream>
+
+#if ANDROID
+#include <android/log.h>
+#else
+#include <iostream>
+#endif
+
+namespace dronecore {
+
+class Debug
+{
+// TODO: This needs to be split into Debug(), Info(), Warn(), Err().
+//       For now, we just print all Debug() even in Release mode.
+//#ifdef DEBUG
+#if 1
+public:
+    Debug() : _s() {}
+
+    template <typename T>
+    Debug &operator << (const T &x)
+    {
+        _s << x;
+        return *this;
+    }
+
+    ~Debug()
+    {
+#if ANDROID
+        __android_log_print(ANDROID_LOG_DEBUG, "DroneCore", "%s", _s.str().c_str());
+#else
+        std::cout << _s.str() << std::endl;
+#endif
+    }
+
+private:
+    std::stringstream _s;
+
+#else
+public:
+    template <typename T>
+    Debug &operator << (const T &x)
+    {
+        UNUSED(x);
+        return *this;
+    }
+#endif
+};
+
+} // namespace dronecore
--- a/core/log.h
+++ b/core/log.h
@@ -8,45 +8,162 @@
 #include <iostream>
 #endif
 
+#define ANSI_COLOR_RED     "\x1b[31m"
+#define ANSI_COLOR_YELLOW  "\x1b[33m"
+#define ANSI_COLOR_GRAY    "\x1b[37m"
+#define ANSI_COLOR_RESET   "\x1b[0m"
+
+// Remove path and extract only filename.
+#define __FILENAME__ (__builtin_strrchr(__FILE__, '/') ? __builtin_strrchr(__FILE__, '/') + 1 : __FILE__)
+
+// For release builds, don't show debug printfs, and just discard it into the NullStream.
+class NullStream
+{
+public:
+    template<typename TPrintable>
+    NullStream &operator<<(TPrintable const &)
+    {
+        /* no-op */
+        static NullStream nothing;
+        return nothing;
+    }
+};
+
+#if DEBUG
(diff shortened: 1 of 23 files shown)


darioxz:
Awesome! What about adding a timestamp? This could be helpful in many cases.

core/log.h
```diff
@@ -0,0 +1,169 @@
...
+    enum LogLevel {
+        Debug,
+        Info,
+        Warn,
+        Err
+    } _log_level;
+
+private:
+    std::stringstream _s;
+    const char *_filename;
```
darioxz: `_filename` is the filename of the source file which called the log printer, right? 

I initially thought that it's the filename of a log file.

julianoes:
Good idea, will add a timestamp. Local time or seconds since the library is running?

darioxz:
I'd vote for local time.

core/log.h
julianoes: Yes, it's the source file name `__FILENAME__` determined above. Better name suggestion?

mrpollo: _callername?

julianoes: I like `_caller_filename`, thx.

julianoes: Fixed in 9e6d97a.

julianoes:
Time added in f063fae. The resolution is seconds. What do you think?

mrpollo:
I think its fine for now, lets test this out for a few days

mrpollo:
@julianoes  Can you rebase?

core/log.h
```diff
@@ -0,0 +1,186 @@
...
+                break;
+        }
+#else
+
+        // Time output taken from:
+        // https://stackoverflow.com/questions/16357999#answer-16358264
+
+
+        time_t rawtime;
+        time(&rawtime);
```
mrpollo: is this using a monotonic clock? I think its important for logging and real time systems to use a monotonic clock so every part of the system is using the same time, I've seen lots of operating systems time warp at boot crashing timers, timeouts or any time dependent loop, not to mention the effects on logging and the integrity of logged data

julianoes: The thing is that `monotonic` clock is not actually the local time as I understand this:

http://en.cppreference.com/w/cpp/chrono/steady_clock
> This clock is not related to wall clock time (for example, it can be time since last reboot), and is most suitable for measuring intervals.

mrpollo: If we can't get a reliable clock I guess we are stuck with this, @hamishwillee can you point this out in our docs please?, logs are not using a monotonic clock hence are susceptible to time warping

julianoes: @mrpollo unless you just want seconds since you started the executable, that's fine with a monotonic clock.

hamishwillee: Er, I haven't been tracking this, but would I be correct in thinking that this discussion has nothing to do with the proposed public API for logging - `dronecore::Logging`?

What I THINK - please confirm:
- this  is an API for people *developing* DroneCore to see what is going on. 
- There doesn't appear to be an actual log - everything is going out to the standard cout console? 
- There doesn't appear to be any way to control what output is displayed from a user point of view - this is just nice colour coding to help people looking at the debug output work out what is critical and what is less important?
- This is not exported, so it can't be used by some third party writing DroneCore apps - only plugin developers?
- I presume you want some docs in the "Contributing" section?
- As Julian says, the fact that the clock is monotonic not very relevant - it is very unlikely to ever overflow, and will always be local time since start of exe right?

julianoes:
> this is an API for people developing DroneCore to see what is going on.

correct.

> There doesn't appear to be an actual log - everything is going out to the standard cout console?

yes.

> There doesn't appear to be any way to control what output is displayed from a user point of view - this is just nice colour coding to help people looking at the debug output work out what is critical and what is less important?

There is one thing. If you build for `Debug`, you see the `LogDebug` printfs (and all others), if you build for `Release`, you do not see the `LogDebug` printfs.

> This is not exported, so it can't be used by some third party writing DroneCore apps - only plugin developers?

It's internal for now, yes.

> I presume you want some docs in the "Contributing" section?

Yes please, especially the log levels and the fact that `LogDebug` is ignored for `Release` builds.

> As Julian says, the fact that the clock is monotonic not very relevant - it is very unlikely to ever overflow, and will always be local time since start of exe right?

Yes, it's just to have some time context.

hamishwillee:
@julianoes @mrpollo 
Thanks. As we just discussed, the fact that this is internal means that you can't use it in our example code - where it would be very useful.
Would it make sense to take this a step further and finish developing the public Logging API? Essentially this would be the same, except that you could redirect the output to console or file or both.

PR #77, opened 2017-09-18
mavlink system_id instead of UUID for devices

julianoes:
The problem is that not all mavlink devices support the
AUTOPILOT_VERSION.uid field and therefore won't connect and be
addressable properly unless we switch from UUIDs to the mavlink system
ID.

Also, the UUID in fact gave us a wrong sense of security because when
communicating with a device we still use the system ID and not the UUID
anyway.

### DIFF
--- a/core/dronecore_impl.cpp
+++ b/core/dronecore_impl.cpp
@@ -115,27 +115,27 @@ void DroneCoreImpl::add_connection(Connection *new_connection)
     _connections.push_back(new_connection);
 }
 
-const std::vector<uint64_t> &DroneCoreImpl::get_device_uuids() const
+const std::vector<uint8_t> &DroneCoreImpl::get_device_system_ids() const
 {
     // This needs to survive the scope but we need to clean it up.
-    static std::vector<uint64_t> uuids = {};
-    uuids.clear();
+    static std::vector<uint8_t> system_ids = {};
+    system_ids.clear();
 
     for (auto it = _device_impls.begin(); it != _device_impls.end(); ++it) {
-        uint64_t uuid = it->second->get_target_uuid();
-        if (uuid != 0) {
-            uuids.push_back(uuid);
+        uint8_t system_id = it->second->get_target_system_id();
+        if (system_id != 0) {
+            system_ids.push_back(system_id);
         }
     }
 
-    return uuids;
+    return system_ids;
 }
 
 Device &DroneCoreImpl::get_device()
 {
     {
         std::lock_guard<std::recursive_mutex> lock(_devices_mutex);
-        // In get_device withoiut uuid, we expect to have only
+        // In get_device without system ID, we expect to have only
         // one device conneted.
         if (_device_impls.size() == 1) {
             return *(_devices.at(_device_impls.begin()->first));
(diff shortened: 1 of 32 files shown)


core/dronecore_impl.cpp
```diff
@@ -115,27 +115,27 @@ void DroneCoreImpl::add_connection(Connection *new_connection)
...
+    return system_ids;
 }
 
 Device &DroneCoreImpl::get_device()
 {
     {
         std::lock_guard<std::recursive_mutex> lock(_devices_mutex);
-        // In get_device withoiut uuid, we expect to have only
+        // In get_device without system ID, we expect to have only
         // one device conneted.
```
hamishwillee: one device connected (typo)

hamishwillee:
Apologies @julianoes . I know I suggested it, but I'm now leaning towards using UUID rather than SYSID. 

On a *well configured system* (with one vehicle mapped to each SYSID) then it doesn't make much difference. On a system that is not well configured the UUID approach does allow a more robust client API. Specifically DroneCore can abstract clients from some MAVLink network changes and make usage more robust and client friendly.

Consider these cases:
- What if multiple vehicles share a SYSID
- What if a vehicle changes its SYSID
- What if a vehicle changes its SYSID to a shared SYSID
- How we might report the above errors to the client. 

With a UUID we can detect the above changes:
- If multiple vehicles share a SYSID then clients cannot safely send commands due to SYSID clash.
  - BUT We might prevent sending and return a new error. This allows sensible handling in line with our current "error handling semantics". It is clear to user what is happening. 
- If multiple vehicles share a SYSID then using UUID clients can still receive telemetry. For some use cases this might still be really useful (i.e. monitoring apps).
- If a vehicle changes its SYSID and the SYSID is *not shared*, then using a UUID DroneCore can transparently switch to the new SYSID and the client does not need to know about this. 

By contrast, using a SYSID we only have one handle to possibly multiple vehicles. In particular this means that:
- if a vehicle changes SYSID it looks like a new vehicle to clients.  Clients will lose any "history" they might be tracking. Logs might be spread across SYSID rather than being associated with the one vehicle. 
- if a vehicle shares a SYSID we won't be able to get telemetry for either vehicle.

If **all systems** have a UUID then I think that the current implementation is clearly more robust (provided we properly inform users when they try to send).

The original question (and problem) is what to do if a system does not report a UUID - can can we return a SYSID if a system doesn't have a UUID?
- Is there other information (other than UUID) that is *likely* to be more unique than just SYSID - for example, a hash of [this other info](https://docs.dronecore.io/en/api_reference/structdronecore_1_1_info_1_1_version.html) and possibly the SYSID and port used? I know it is a placebo, but might reduce "risk".
- Either way, can we inform the user sensibly that this device is "at risk" - ie allow a successful connection, but report with a different type of code "SYSID"? Provide an API that you can check to determine if UUID is "real"

As I say, I am now leaning back to UUIDs - accepting the risk that we might get clashes if we SYSID for devices do not have UUIDs - even though it is a little deceptive. Would appreciate @mrpollo point of view on this. 

If we do go with this PR

1. All looks sane to me, but perhaps @mrpollo better person to review the C++.
2. There are possibly some docs I will need to update as well.
3. We need to add error reporting for case where we detect multiple UUIDs sharing one SYSID.

mrpollo:
I agree with the notion that UUIDs will help us avoid conflicts in a system, I think we should also consider the cases @hamishwillee describes here

core/dronecore_impl.cpp
julianoes: Fixed in 02187b0.

julianoes:
Ok, I think I'm convinced to stay with UUID. However, I need the system ID as a fallback.

New PR in #80.

PR #78, opened 2017-09-18
various clang fixes

julianoes:
With this flag, the clang build lead to many linking errors on a Linux
such as Fedora.


core/serial_connection.cpp
```diff
@@ -160,7 +160,7 @@ void SerialConnection::receive(SerialConnection *parent)
         if (recv_len < -1) {
             Debug() << "read failure: " << GET_ERROR(errno);
         }
-        if (recv_len > sizeof(buffer) || recv_len == 0) {
+        if (recv_len > (int)sizeof(buffer) || recv_len == 0) {
```
darioxz: We could discuss about using c++ casts instead of c-casts.

PR #80, opened 2017-09-19
Fallback to mavlink system ID if no UUID available

julianoes:
This changes the logic so that you the UUID is used if available but the
mavlink system ID is used as a fallback option. The fallback happens if
3 requests for the autopilot_version are unsuccessful or if the
AUTOPILOT_VERSION.uid field is 0.

This replaces #77.

### DIFF
--- a/core/device_impl.cpp
+++ b/core/device_impl.cpp
@@ -196,6 +196,11 @@ bool DeviceImpl::send_message(const mavlink_message_t &message)
 
 void DeviceImpl::request_autopilot_version()
 {
+    if (_target_uuid_initialized) {
+        // Already initialized, we can exit.
+        return;
+    }
+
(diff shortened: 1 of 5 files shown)


hamishwillee:
Should the user be warned that they're working on a SYSID and not a real UUID? I guess the chances are that if the UUID  < 255 they can tell that it is not "real".

hamishwillee:
I added #81 for discussion of the cases that come up due to shared SYSID, and in particular when the UUID is the SYSID.

julianoes:
> Should the user be warned that they're working on a SYSID and not a real UUID? I guess the chances are that if the UUID < 255 they can tell that it is not "real".

Ok I'll add that to the docs, thanks @hamishwillee.

PR #83, opened 2017-09-20
integration_tests: improve mission example

julianoes:
This removes the clunky state machine and instead uses futures/promises
to wrap around async functions.

The test currently exits too early because the landed/in-air state seems
to be wrong.

@hamishwillee I did not get to the example yet. However, it is pretty much the same code. Also, I need to investigate what is wrong with the landed state.
At least it's a push in the direction that it could go.


integration_tests/waypoints.cpp
```diff
@@ -64,276 +37,171 @@ TEST_F(SitlTest, MissionAddWaypointsAndFly)
...
-void receive_disarm_result(Action::Result result)
-{
-    EXPECT_EQ(result, Action::Result::SUCCESS);
-
-    if (result == Action::Result::SUCCESS) {
-    } else {
-        std::cerr << "Error: disarming result: " << unsigned(result) << std::endl;
+    if (current == total) {
+        // We are done, and can do RTL to go home.
+        const Action::Result result = device->action().arm();
```
hamishwillee: This should be `.rtl()` right?

integration_tests/waypoints.cpp
```diff
@@ -64,276 +37,171 @@ TEST_F(SitlTest, MissionAddWaypointsAndFly)
...
-    if (result == Action::Result::SUCCESS) {
-    } else {
-        std::cerr << "Error: return to land result: " << unsigned(result) << std::endl;
-        _mission_state = MissionState::ERROR;
+            future_result.wait();
+            future_result.get();
+        }
+        // We don't want to pause muptiple times in case we get the progress notification
+        // several times.
+        _break_done = true;
```
hamishwillee: Not used, see above

integration_tests/waypoints.cpp
```diff
@@ -64,276 +37,171 @@ TEST_F(SitlTest, MissionAddWaypointsAndFly)
...
+void receive_mission_progress(int current, int total, Device *device)
 {
-    std::cout << "Mission status update: " << current << " / " << total << std::endl;
+    LogInfo() << "Mission status update: " << current << " / " << total;
 
-    if (current > 0 && current == total) {
-        _mission_state = MissionState::MISSION_DONE;
-    }
-}
+    if (current == 2) {
```
hamishwillee: I think this needs to also check `_break_done `

integration_tests/waypoints.cpp
```diff
@@ -64,276 +37,171 @@ TEST_F(SitlTest, MissionAddWaypointsAndFly)
...
-        std::cerr << "Error: mission start result: " << unsigned(result) << std::endl;
-        _mission_state = MissionState::ERROR;
-    }
-}
-
-void receive_continue_mission_result(Mission::Result result)
-{
-    EXPECT_EQ(result, Mission::Result::SUCCESS);
+    // FIXME: this check does currently not work in SITL.
+    while (device.telemetry().in_air()) {
```
hamishwillee: You could try check on `disarmed()`. Worked on last example.

integration_tests/waypoints.cpp
```diff
@@ -64,276 +37,171 @@ TEST_F(SitlTest, MissionAddWaypointsAndFly)
     Device &device = dc.device();
 
     while (!device.telemetry().health_all_ok()) {
-        std::cout << "waiting for device to be ready" << std::endl;
+        LogInfo() << "waiting for device to be ready";
```
hamishwillee: - Capitalise sentence (prep for the example). 
- Do we need to think about case where health is never OK?

integration_tests/waypoints.cpp
```diff
@@ -64,276 +37,171 @@ TEST_F(SitlTest, MissionAddWaypointsAndFly)
...
         std::this_thread::sleep_for(std::chrono::seconds(1));
     }
 
-    std::cout << "Device ready, let's start" << std::endl;
+    LogInfo() << "Device ready, let's start";
```
hamishwillee: Probably just make this text "Device ready"

integration_tests/waypoints.cpp
```diff
@@ -64,276 +37,171 @@ TEST_F(SitlTest, MissionAddWaypointsAndFly)
...
```
hamishwillee: Perhaps here add line "Creating and uploading mission"

integration_tests/waypoints.cpp
```diff
@@ -64,276 +37,171 @@ TEST_F(SitlTest, MissionAddWaypointsAndFly)
...
+        auto prom = std::make_shared<std::promise<void>>();
+        auto future_result = prom->get_future();
+        device.mission().send_mission_async(
+        mission_items, [prom](Mission::Result result) {
+            EXPECT_EQ(result, Mission::Result::SUCCESS);
+            prom->set_value();
+        });
+
+        future_result.wait();
+        future_result.get();
```
hamishwillee: After this line, add add debug "Mission uploaded". Not sure if that is best done in the lambda callback or here.

integration_tests/waypoints.cpp
```diff
@@ -64,276 +37,171 @@ TEST_F(SitlTest, MissionAddWaypointsAndFly)
...
+                         MissionItem::CameraAction::STOP_PHOTO_INTERVAL));
+
+    {
+        // We only have the send_mission function asynchronous for now, so we wrap it using
+        // std::future.
+        auto prom = std::make_shared<std::promise<void>>();
+        auto future_result = prom->get_future();
+        device.mission().send_mission_async(
+        mission_items, [prom](Mission::Result result) {
+            EXPECT_EQ(result, Mission::Result::SUCCESS);
```
hamishwillee: What happens if they aren't equal inside async request? Does this propagate out of the promise? 
For test code I guess you want this to exit/crash but I'd hope for example code we'd get sensible handling.

integration_tests/waypoints.cpp
```diff
@@ -64,276 +37,171 @@ TEST_F(SitlTest, MissionAddWaypointsAndFly)
...
+            prom->set_value();
+        });
+
+        future_result.wait();
```
hamishwillee: I'm don't think we have to `wait()` on a single future - `get()` handles this.  Could be wrong.

integration_tests/waypoints.cpp
```diff
@@ -64,276 +37,171 @@ TEST_F(SitlTest, MissionAddWaypointsAndFly)
...
+        future_result.get();
     }
 
-    EXPECT_EQ(_mission_state, MissionState::DONE);
-}
+    const Action::Result arm_result = device.action().arm();
```
hamishwillee: Maybe here, and at least for example, info for "Arming" and "Armed"

integration_tests/waypoints.cpp
```diff
@@ -64,276 +37,171 @@ TEST_F(SitlTest, MissionAddWaypointsAndFly)
...
+    EXPECT_EQ(arm_result, Action::Result::SUCCESS);
 
-void receive_send_mission_result(Mission::Result result)
-{
-    EXPECT_EQ(result, Mission::Result::SUCCESS);
+    // Before starting the mission, we want to be sure to subscribe to the mission progress.
+    // We pass on device to receive_mission_progress because we need it in the callback.
+    device.mission().subscribe_progress(std::bind(&receive_mission_progress, _1, _2, &device));
```
hamishwillee: For my understanding, do we use `std::bind` here to allow us to send a three argument function even though our calback signature only takes 2 arguments? - and is that the reason we're not using a lambda? I'm not very familiar with `std:bind`.

integration_tests/waypoints.cpp
```diff
@@ -64,276 +37,171 @@ TEST_F(SitlTest, MissionAddWaypointsAndFly)
...
+            EXPECT_EQ(result, Mission::Result::SUCCESS);
+            prom->set_value();
+        });
 
-    if (result == Mission::Result::SUCCESS) {
-        _mission_state = MissionState::STARTING_DONE;
-    } else {
-        std::cerr << "Error: mission start result: " << unsigned(result) << std::endl;
-        _mission_state = MissionState::ERROR;
+        future_result.wait();
```
hamishwillee: As above, not sure is needed (and everywhere)

integration_tests/waypoints.cpp
```diff
@@ -64,276 +37,171 @@ TEST_F(SitlTest, MissionAddWaypointsAndFly)
...
-    }
-}
-
-void receive_start_mission_result(Mission::Result result)
-{
-    EXPECT_EQ(result, Mission::Result::SUCCESS);
+    {
+        auto prom = std::make_shared<std::promise<void>>();
+        auto future_result = prom->get_future();
+        device.mission().start_mission_async(
```
hamishwillee: For example at least suggest we have info "Starting mission". IN the callback "Mission started".

integration_tests/waypoints.cpp
```diff
@@ -64,276 +37,171 @@ TEST_F(SitlTest, MissionAddWaypointsAndFly)
...
 
     std::vector<std::shared_ptr<MissionItem>> mission_items;
 
     mission_items.push_back(
-        add_waypoint(47.398170327054473,
-                     8.5456490218639658,
-                     10.0f, 5.0f, false,
-                     20.0f, 60.0f,
-                     MissionItem::CameraAction::NONE));
+        add_mission_item(47.398170327054473,
```
hamishwillee: - Is this a takeoff item? 
- Would we be better off having an explicit takeoff and checking for success?

The reason I ask is that this will work for VTOL and MC, but could well fail for Plane, since takeoff is not automatically added for plane missions.

hamishwillee:
Great improvements. Some possible issues highlighted inline. 

Other things:
1. I can't build this. I synced with latest master and installed system wide, but I get error:
`error: integration_test_helper.h: No such file ...`
2.  Can we rely on SITL to always start at the same home location? If possible, I would set the home and vehicle location to be the default location so this works if someone has SITL set differently.
   OR I would dynamically create my waypoint positions relative to whatever home position is set.
3. On this same note, one thing I learned in DK-Python is that it is always easier for example code (and real world testing) to define positions relative to NED or local body frame). That way you can move the "behaviour" (fly  a square etc) without having to manually work out all the positions. It is extremely desirable to have a library that can give you the GPS position at a specified location away from the vehicle at both a NED frame (go north 20 metres say) and a body frame (e.g. go forward 20 metres).

julianoes:
@hamishwillee thanks for your review. There is a lot still wrong :). Let me fix up these things.
Any reason you closed this?

hamishwillee:
Closed by accident! Sorry :-0

hamishwillee:
Re

> I can't build this. I synced with latest master and installed system wide, but I get error:
error: integration_test_helper.h: No such file ...

I installed this as an example, not test code. That explains why no access to header, but does highlight the value in getting that functionality exposed.

julianoes:
>  error: integration_test_helper.h: No such file ...

Oh, got it. I'll fix it.

integration_tests/waypoints.cpp
julianoes: Thx, fixed.

integration_tests/waypoints.cpp
julianoes: Oh thanks.

integration_tests/waypoints.cpp
julianoes: Good idea, thanks.

integration_tests/waypoints.cpp
julianoes: Done, and yes we should add a timeout ad give up.

integration_tests/waypoints.cpp
julianoes: Sure.

integration_tests/waypoints.cpp
julianoes: You are right that we should crash because there is no point in continuing if that fails.

For the integration_test what I we should do is make it an `ASSERT_EQ` which exits the test if it fails.

For the example we can either `abort()`, or propagate the result out of the promise and check for it.

integration_tests/waypoints.cpp
julianoes: I think you're right.

integration_tests/waypoints.cpp
julianoes: Ok.

integration_tests/waypoints.cpp
julianoes: You can add more arguments with `std::bind` but also with lambdas. I just took it out because it's more than a few lines in that function.

integration_tests/waypoints.cpp
julianoes: Ok.

integration_tests/waypoints.cpp
julianoes: Looks like we don't support fixedwing yet then. 😞

hamishwillee: Fair enough. Should we add a test for vehicle type and exit. Or we could add a takeoff mission item?

integration_tests/waypoints.cpp
```diff
@@ -64,276 +37,168 @@ TEST_F(SitlTest, MissionAddWaypointsAndFly)
...
+void receive_mission_progress(int current, int total, Device *device)
 {
-    std::cout << "Mission status update: " << current << " / " << total << std::endl;
+    LogInfo() << "Mission status update: " << current << " / " << total;
 
-    if (current > 0 && current == total) {
-        _mission_state = MissionState::MISSION_DONE;
-    }
-}
+    if (current == 2 && !_break_done) {
```
hamishwillee: Maybe >=2, if there is any chance that the loop is not run on 2nd waypoint.

integration_tests/waypoints.cpp
```diff
@@ -64,276 +37,168 @@ TEST_F(SitlTest, MissionAddWaypointsAndFly)
...
-    }
-}
+        {
+            auto prom = std::make_shared<std::promise<void>>();
+            auto future_result = prom->get_future();
 
-void receive_return_to_launch_result(Action::Result result)
-{
-    EXPECT_EQ(result, Action::Result::SUCCESS);
+            device->mission().pause_mission_async(
```
hamishwillee: Maybe add debug info for the pause states.

hamishwillee:
@julianoes I added a couple more comments, but fixes look good by inspection. I assume your logging comes from integration_test_helper.h? That wouldn't be a good name for when we use in example - better to have a common file for that?
I haven't actually run this.

integration_tests/waypoints.cpp
julianoes: We could try that but I'm a bit reluctant to change it without more testing in place.

julianoes:
> 2. Can we rely on SITL to always start at the same home location? If possible, I would set the home and vehicle location to be the default location so this works if someone has SITL set differently.
OR I would dynamically create my waypoint positions relative to whatever home position is set.

It is set up for the default PX4 SITL location. If someone uses another home location in SITL, the mission will fail and the autopilot will complain "1st waypoint too far".

> 3. On this same note, one thing I learned in DK-Python is that it is always easier for example code (and real world testing) to define positions relative to NED or local body frame). That way you can move the "behaviour" (fly a square etc) without having to manually work out all the positions. It is extremely desirable to have a library that can give you the GPS position at a specified location away from the vehicle at both a NED frame (go north 20 metres say) and a body frame (e.g. go forward 20 metres).

Right. I don't think missions in NED are supported (yet) though.

hamishwillee:
Hamish

julianoes:
> 1. I can't build this. I synced with latest master and installed system wide, but I get error:
error: integration_test_helper.h: No such file ...

See bb73b9286a2c41a3bbaa64ced479c72a64b055e9

hamishwillee:
Damn, I really wanted to be able to use your debug output to have pretty output of the examples. :-( 

I'll try run that new version Monday.

hamishwillee:
@julianoes Example testing:
1. If there is a simulator running the example proceeds through to completion. In this case the only suggestion I have is that perhaps we should only send updates when the mission state changes. I say *suggestion* because it is useful to have the extra notification of state 2 after the pause resumes.

2. There is something wrong with our timeout/error handling code
  * If you start the example without the simulator it gets stuck in "Waiting for device" forever.
  * If you stop the simulator after things have started there is no notification of disconnect. I think that we probably need to add a watcher for that. 

Cheers
H

hamishwillee:
PS. I also got debug output for info and debug ..... 
```
[07:20:32|Debug] Send mission item 17 (mission_impl.cpp:506)
[07:20:32|Debug] Send mission item 18 (mission_impl.cpp:506)
[07:20:32|Debug] Send mission item 19 (mission_impl.cpp:506)
[07:20:32|Debug] Send mission item 20 (mission_impl.cpp:506)
[07:20:32|Debug] Send mission item 21 (mission_impl.cpp:506)
Arming...
Mission uploaded.
[07:20:32|Info ] Sucess, done (mission_impl.cpp:128)
Armed.
```
1. This was all in white. Aren't they supposed to have different colours?
2. I just ran this from my build directory on linux using ./fly_mission . Where/how is debug vs release built?

example/fly_mission/fly_mission.cpp
```diff
@@ -0,0 +1,258 @@
...
+static void receive_mission_progress(int current, int total);
+
+static std::atomic<bool> _want_to_pause {false};
+
+int main(int /*argc*/, char ** /*argv*/)
+{
+    DroneCore dc;
+
+    {
+        auto prom = std::make_shared<std::promise<void>>();
```
hamishwillee: Should this promise include the connection error? We're using this to pass a UUID and I don't believe we handle the case where we get a timeout or no UUID.

hamishwillee: Arguably this could also be a synchronous fail case for the whole example - since we're never going to recover from not being able to connect. 
Unless we can presumably recover, in which case it would be nice to think about what sort of recovery options users have, and how they might implement them.

julianoes:
> Not by MAVLink, but trivially by us. In DK-Python we just implemented the well known frame conversion functions - [see here](http://python.dronekit.io/guide/copter/guided_mode.html). All the DK-Python examples can be used anywhere, because they dynamically calculate movement with respect to the current vehicle position on startup. In an ideal world this would be a set of static helper functions provided by the core API, that we could then use to provide support for a NED mission API. if we wanted. (In DK-Python we just had it as some example code that people could re-use, but that was a bit poor. ) Anyway, I think this is worth doing at least for the examples, and I'd like to have the discussion about whether this could be in the API.

I did not actually think of putting the logic into DroneCore, it's an idea.

However, I would like to avoid workarounds in DroneCore if PX4 does not support something yet. There is actually support for local NED missions in mavlink (http://mavlink.org/messages/common#MISSION_ITEM), so I think the right approach would be to add it to PX4 as soon as possible, since it's nice (or essential?) feature to have anyway.

example/fly_mission/fly_mission.cpp
julianoes: Yes, we don't handle a timeout, we're just stuck forever. However, there won't ever be no UUID, and since we don't need the UUID outside, we don't need to include it in the promise.

julianoes:
> If there is a simulator running the example proceeds through to completion. In this case the only suggestion I have is that perhaps we should only send updates when the mission state changes. I say suggestion because it is useful to have the extra notification of state 2 after the pause resumes.

Ok makes sense.

> There is something wrong with our timeout/error handling code
> 
>  1. If you start the example without the simulator it gets stuck in "Waiting for device" forever.
>  2. If you stop the simulator after things have started there is no notification of disconnect. I think that we probably need to add a watcher for that.

1. Yes that's kind of expected.
2. Sure we can add that.

hamishwillee:
>> There is something wrong with our timeout/error handling code

>> If you start the example without the simulator it gets stuck in "Waiting for device" forever.

> 1. Yes that's kind of expected.

I was confused by the previous example/takeoff, which fails if nothing is detected in 2 seconds. This makes perfect sense. 

>> If you stop the simulator after things have started there is no notification of disconnect. I think that we probably need to add a watcher for that.

> Sure we can add that.

Great.

example/fly_mission/fly_mission.cpp
hamishwillee: OK. As for other response, this makes sense - was just not what I expected based on previous example.

hamishwillee:
>> Not by MAVLink, but trivially by us. In DK-Python we just implemented the well known frame conversion functions - see here. All the DK-Python examples can be used anywhere, because they dynamically calculate movement with respect to the current vehicle position on startup. In an ideal world this would be a set of static helper functions provided by the core API, that we could then use to provide support for a NED mission API. if we wanted. (In DK-Python we just had it as some example code that people could re-use, but that was a bit poor. ) Anyway, I think this is worth doing at least for the examples, and I'd like to have the discussion about whether this could be in the API.

> I did not actually think of putting the logic into DroneCore, it's an idea.

> However, I would like to avoid workarounds in DroneCore if PX4 does not support something yet. There is actually support for local NED missions in mavlink (http://mavlink.org/messages/common#MISSION_ITEM), so I think the right approach would be to add it to PX4 as soon as possible, since it's nice (or essential?) feature to have anyway.

Depends on how long we're talking. If it "might" happen "one day" I'd say suggest we implement our own NED API - it would not be hard and we could pull out our implementation without breaking clients.

If it is definitely going to happen but maybe in 6 months then I'd update the examples with the code as done in DK-PYthon. Much easier for people to reuse our code if we provide something that does this.

YOur call though!

julianoes:
> Depends on how long we're talking. If it "might" happen "one day" I'd say suggest we implement our own NED API - it would not be hard and we could pull out our implementation without breaking clients.

Let me find out how hard/far away that would be.

julianoes:
> 1. This was all in white. Aren't they supposed to have different colours?

I've changed `LogInfo` to blue but it's not in this pull request yet. See #86.

> 2. I just ran this from my build directory on linux using ./fly_mission . Where/how is debug vs release built?

To have a release build instead of debug for the library, you need to do:
```
cd DroneCore
make clean
BUILD_TYPE=Release make
```

mrpollo:
@hamishwillee thanks for the extensive review (as always), I agree with your assessment that having a NED interface would allow us to be more flexible, I think we should find a way to meet halfway with PX4 and come up with a solution that would enable developers to use NED now but won't break our implementation once PX4 adds support, I think we can add a thin layer that gives us what we need, this will even help add NED to older supported PX4 versions (older to whenever we add a NED interface in PX4)

example/fly_mission/fly_mission.cpp
```diff
@@ -0,0 +1,264 @@
...
+
+        const Mission::Result result = future_result.get();
+        if (result != Mission::Result::SUCCESS) {
+            std::cout << "Mission upload failed (" << Mission::result_str(result) << "), exiting." << std::endl;
+            return 1;
+        }
+    }
+
+    std::cout << "Arming..." << std::endl;
+    const Action::Result arm_result = device.action().arm();
```
mrpollo: would have been cool to have the arm_async here

mrpollo:
@julianoes can you rebase (master) please? you can also go ahead to rebase/merge this back to master once you are done

example/fly_mission/fly_mission.cpp
hamishwillee: I don't think using async particularly matters here - in the case where you have a single command that has to be completed in sequence to progress.

hamishwillee:
Agreed re taking NED to a separate thread. Lets get any changes @julianoes agrees with in and merge this.

example/fly_mission/fly_mission.cpp
```diff
@@ -0,0 +1,264 @@
...
+    {
+        std::cout << "Uploading mission..." << std::endl;
+        // We only have the send_mission function asynchronous for now, so we wrap it using
+        // std::future.
+        auto prom = std::make_shared<std::promise<Mission::Result>>();
+        auto future_result = prom->get_future();
+        device.mission().send_mission_async(
+        mission_items, [prom](Mission::Result result) {
+            prom->set_value(result);
+            std::cout << "Mission uploaded." << std::endl;
```
hamishwillee: @julianoes I think at this point we don't know if the mission uploaded. This should be "Mission uploa**ding**"

julianoes: You're right, I moved it in c67e45c.

hamishwillee:
Nice one! FYI, I created https://github.com/dronecore/DroneCore/issues/100 and https://github.com/dronecore/DroneCore/issues/101 for the NED mission discussion.

PR #85, opened 2017-09-21
Support for commands in parallel

julianoes:
This is a big refactor of the MavlinkCommands class in order to support
multiple commands in parallel. It won't do multiple commands with the
same command ID at the same time in order to prevent conflicts/races
with acks that can't be assigned uniquely.

### DIFF
--- a/.ycm_extra_conf.py
+++ b/.ycm_extra_conf.py
@@ -1,0 +1,206 @@
+# This file is NOT licensed under the GPLv3, which is the license for the rest
+# of YouCompleteMe.
+#
+# Here's the license text for this file:
+#
+# This is free and unencumbered software released into the public domain.
+#
+# Anyone is free to copy, modify, publish, use, compile, sell, or
+# distribute this software, either in source code form or as a compiled
+# binary, for any purpose, commercial or non-commercial, and by any
+# means.
+#
+# In jurisdictions that recognize copyright laws, the author or authors
+# of this software dedicate any and all copyright interest in the
+# software to the public domain. We make this dedication for the benefit
+# of the public at large and to the detriment of our heirs and
+# successors. We intend this dedication to be an overt act of
+# relinquishment in perpetuity of all present and future rights to this
+# software under copyright law.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+# EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+# MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+# IN NO EVENT SHALL THE AUTHORS BE LIABLE FOR ANY CLAIM, DAMAGES OR
+# OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+# ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+# OTHER DEALINGS IN THE SOFTWARE.
+#
+# For more information, please refer to <http://unlicense.org/>
+
+import os
+import ycm_core
+
+# These are the compilation flags that will be used in case there's no
+# compilation database set (by default, one is not set).
+# CHANGE THIS LIST OF FLAGS. YES, THIS IS THE DROID YOU HAVE BEEN LOOKING FOR.
+flags = [
+'-Wall',
+'-Wextra',
+'-Werror',
+'-fno-exceptions',
+'-Wshadow',
+'-Wno-strict-aliasing',
+# You 100% do NOT need -DUSE_CLANG_COMPLETER in your flags; only the YCM
+# source code needs it.
+'-DUSE_CLANG_COMPLETER',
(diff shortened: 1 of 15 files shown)


.ycm_extra_conf.py
```diff
@@ -0,0 +1,206 @@
+# This file is NOT licensed under the GPLv3, which is the license for the rest
+# of YouCompleteMe.
```
hamishwillee: See YouCompleteMe

julianoes: 😄 what?

hamishwillee: Exactly - what!!

The text says "This file is NOT licensed under the GPLv3, which is the license for the rest of YouCompleteMe."

That looks like a placeholder to me.

And it should be obvious I am not actually reviewing this - I just wanted to see what the code looked like :-)

julianoes: I can remove it and say it is licensed as BSD 3-clause like the rest of the repo.

hamishwillee: Thanks.

julianoes: Actually this file says it is public domain, so I'm not sure what I should do.

hamishwillee: I thought this was our file. If YouComleteMe is an actual real name of a product called YouCompleteMe then we have to leave that alone - it isn't ours to modify.

core/log.h
```diff
@@ -101,6 +102,7 @@ class LogDetailed
                 std::cout << "|Debug] ";
                 break;
             case LogLevel::Info:
+                std::cout << ANSI_COLOR_BLUE;
```
darioxz: This is really not critical, but #86 should get merged before #86.

.ycm_extra_conf.py
mrpollo: Yep this file was copied from the YouCompleteMe project, we should leave as is

core/log.h
mrpollo: Is #86 rebased on to of #85?, can we remove this so we can merge w/o blocking?

core/locked_list.h
```diff
@@ -0,0 +1,88 @@
...
+public:
+    LockedList() :
+        _list(),
+        _mutex()
+    {};
+
+    class iterator: public std::list<T>::iterator
+    {
+    public:
+        iterator(typename std::list<T>::iterator c, std::mutex &mutex) :
```
mrpollo: I know we are using single char variables elsewhere but we should move away from this, can we name this variable something more descriptive, please?

core/mavlink_commands.h
```diff
@@ -53,30 +54,27 @@ class MavlinkCommands
...
 private:
-    enum class State {
-        NONE,
-        WAITING,
-        IN_PROGRESS,
-        DONE,
-        FAILED
-    } _state = State::NONE;
-    std::mutex _state_mutex {};
+    static constexpr double DEFAULT_TIMEOUT_NORMAL_S = 0.5;
```
mrpollo: We need to find a way to get the default timeouts into the documentation (if they aren't already there), this is important behavior that should be visible to external developers

core/mavlink_commands.h
```diff
@@ -53,30 +54,27 @@ class MavlinkCommands
...
+    static constexpr double DEFAULT_TIMEOUT_IN_PROGRESS_S = 3.0;
+    static constexpr int RETRIES = 3;
 
     struct Work {
-        int retries_to_do = 3;
-        double timeout_s = 0.5;
+        int num_sent = 0;
```
mrpollo: can we use a more descriptive name here? is `num_set` the number of times a command has been set after retries? I tried following the code and got a bit confused

core/timeout_handler.cpp
```diff
@@ -35,8 +35,17 @@ void TimeoutHandler::refresh(const void *cookie)
 
     auto it = _timeouts.find((void *)(cookie));
     if (it != _timeouts.end()) {
-        dl_time_t future_time = steady_time_in_future(it->second->duration_s);
-        it->second->time = future_time;
+        it->second->start_time = steady_time();
+    }
+}
+
+void TimeoutHandler::update(double new_duration_s, const void *cookie)
```
mrpollo: what about `updated_duration_s` instead of `new_duration_s`?

core/mavlink_commands.h
hamishwillee: @julianoes @mrpollo Arguably some timeouts should actually be part of the settable public API - and if so they could then be documented in the same way as everything else (?).

@julianoes Are all timeouts named consistently so we can find them? (i.e. can I have a list)

core/locked_list.h
julianoes: Ok 173ac1a.

core/mavlink_commands.h
julianoes: Ok, I agree. Let's check this later.

core/timeout_handler.cpp
julianoes: Sure, thanks. See ea4cdb8.

core/mavlink_commands.h
julianoes: It is `num_sent`, not `num_set` but I've changed it to `num_command_sent` in a746eab.

PR #86, opened 2017-09-22
core: make Info printfs blue

julianoes:
In order to distinguish Info and Debug, Info is now made blue.

Also, the color is now properly set at the beginning of the line instead
of after the time.

### DIFF
--- a/core/log.h
+++ b/core/log.h
@@ -12,6 +12,7 @@
 
 #define ANSI_COLOR_RED     "\x1b[31m"
 #define ANSI_COLOR_YELLOW  "\x1b[33m"
+#define ANSI_COLOR_BLUE    "\x1b[34m"
 #define ANSI_COLOR_GRAY    "\x1b[37m"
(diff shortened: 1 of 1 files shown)


hamishwillee:
Code-wise this is fine, but what is the semantic difference between info and debug? ie when would you use one and not the other?

The most common output should be white I'd think?

julianoes:
> Code-wise this is fine, but what is the semantic difference between info and debug? ie when would you use one and not the other?
> 
> The most common output should be white I'd think?

Ok, so you would vote to have debug output blue? I tried grey but that was not really visible in my console.

mrpollo:
I agree with @darioxz we should only have the `[Time | Debug]` part in color.

I like the blue!
https://media.giphy.com/media/tRx1Zf2jF2N8I/giphy.gif

julianoes:
Ok, changed in b21c983.

mrpollo:
@julianoes Can you please rebase master and then make rebase/merge back ? this is now approved 👍

PR #88, opened 2017-09-25
telemetry: use available instead of lost for RC

julianoes:
It is more intuitive to have the flag `available` signalling connected
then `not lost` like previously. Also, the initialization with `false`
makes more sense this way.


hamishwillee:
Thanks!

PR #89, opened 2017-09-25
grpc: split dronecore.proto into <plugin>.proto

julianoes:
This is a try to split up the dronecore service into multiple services
with, so that each plugin can have its proto definition and server
(wrapper) implementation.

This is fairly ugly still. I'm not sure if we should go down that route (yet).


julianoes:
This makes the gRPC API ugly and non-intuitive. We need to do more research how gRPC can work together with the plugin architecture. Closing for now.

PR #93, opened 2017-09-26
UDP connection: allow remote IP and port to change

julianoes:
It does not really make sense that a device always needs to keep its IP
and port because a wifi connection might drop for a bit and therefore
the IP and port can change. Therefore, we no longer ignore messages
after the IP/port have changed and instead just accept it.

@darioxz this is for you, please review.


julianoes:
Yay, thanks for testing.

PR #97, opened 2017-09-26
core: add is_connected interface

### LINKED ISSUE #56: We should be able to track connection status of Device objects other than by callback
The suggestion is that it should be possible to detect whether Device is connected from the object itself. This would allow a client to iterate the vector of devices and only handle those that are connected - without having to separately manage tracking connection/and disconnection.

julianoes:
This adds a sync interface for the connection status of a device. It is
basically the same information as also covered by the async methods
on_discover and on_timeout.

Closes #56.

### DIFF
--- a/include/dronecore.h
+++ b/include/dronecore.h
@@ -148,6 +148,29 @@ public:
     typedef std::function<void(uint64_t uuid)> event_callback_t;
 
     /**
+     * @brief Returns true if exactly one device is currently connected.
+     *
+     * Connected means we are receiving heartbeats from this device.
+     * It means the same as "discovered" and "not timed out".
+     *
+     * If multiple devices have connected, this will return false.
+     *
+     * @return true if exactly one device is connected.
+     */
+    bool is_connected() const;
+
+    /**
+     * @brief Returns true if a device is currently connected.
+     *
+     * Connected means we are receiving heartbeats from this device.
+     * It means the same as "discovered" and "not timed out".
+     *
+     * @param uuid UUID of device to check.
+     * @return true if device is connected.
+     */
+    bool is_connected(uint64_t uuid) const;
+
+    /**
      * @brief Register callback for device discovery.
      *
      * This sets a callback that will be notified if a new device is discovered.
--- a/core/device_impl.cpp
+++ b/core/device_impl.cpp
@@ -40,6 +40,11 @@ DeviceImpl::~DeviceImpl()
     }
 }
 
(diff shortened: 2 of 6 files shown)


include/dronecore.h
```diff
@@ -148,6 +148,29 @@ class DroneCore
...
+     * It means the same as "discovered" and "not timed out".
+     *
+     * If multiple devices have connected, this will return false.
+     *
+     * @return true if exactly one device is connected.
+     */
+    bool is_connected() const;
+
+    /**
+     * @brief Returns true if a device is currently connected.
```
hamishwillee: should be "Returns true if the specified device is currently connected."

hamishwillee:
This allows you to test from a DroneCore object whether you have exactly one object connected or if a particular UUID is connected. It does what it says.

1.  When/why would you want to check you have just one device connected? 
2. I can imagine you might want to  test whether a call to `Device &device()` will succeed but this API won't do that since it returns false if you have multiple connected devices. 
3. The API to get the connection status for a particular UUID should be useful, because we could choose to not get handles to disconnected objects and monitor the connection API to start handling them as needed. 

My problem with this is that I don't know if it allows us to sensibly handle the case where we disconnect and reconnect (both with 1 and multiple devices). We need to work out how we would do that in order to decide whether connection information needs to be testable from the `Device` itself, and whether we need to be able to register for connection/timeout events from the `Device` itself.  Just FYI, added https://github.com/dronecore/DroneCore/issues/99 for that example.

julianoes:
> 1. When/why would you want to check you have just one device connected?

Whenever you only expect one device, so most of the time I'd say.

> 2. I can imagine you might want to test whether a call to Device &device() will succeed but this API won't do that since it returns false if you have multiple connected devices.

I agree that `device()` and `is_connected()` are not behaving the same way since `device()` returns the first device for the case of multiple connected devices. This makes me wonder if we should go back to the discussion whether `device()` should return a pointer which can be `nullptr`. That could make it consistent.

hamishwillee:
>> When/why would you want to check you have just one device connected?

> Whenever you only expect one device, so most of the time I'd say.

If I know I'm only ever going to get one device, test seems a little pointless.

julianoes:
> If I know I'm only ever going to get one device, test seems a little pointless.

Just to be sure and to know it is connected.

julianoes:
I'm merging this for now. We can change it again later.

hamishwillee:
OK. Thanks!

PR #98, opened 2017-09-26
integration_tests: add checks to takeoff/land test

julianoes:
This adds some checks to the ActionSimpleHover tests to actually check
what PX4 SITL is executing rather than just checking if we get the
correct responses.

This could be an example for CI tests for PX4 using DroneCore.

@dagar have a look

To test this:
- Start PX4 SITL
- Then start the test:
    ```
    make && build/default/integration_tests_runner --gtest_filter="SitlTest.ActionSimpleHover*"
    ```
    
Or if you have Firmware and DroneCore next to each other in a folder:

```
make && AUTOSTART_SITL=1 build/default/integration_tests_runner --gtest_filter="SitlTest.ActionSimpleHover*"
```

### DIFF
--- a/integration_tests/simple_hover.cpp
+++ b/integration_tests/simple_hover.cpp
@@ -4,7 +4,27 @@
 
 using namespace dronecore;
 
-TEST_F(SitlTest, ActionSimpleHover)
+
+static void takeoff_and_hover_at_altitude(float altitude_m = NAN);
+
+
+TEST_F(SitlTest, ActionSimpleHoverDefault)
+{
+    takeoff_and_hover_at_altitude();
+}
+
+TEST_F(SitlTest, ActionSimpleHoverHigher)
+{
+    takeoff_and_hover_at_altitude(5.0f);
+}
+
+TEST_F(SitlTest, ActionSimpleHoverLower)
+{
+    takeoff_and_hover_at_altitude(1.0f);
+}
+
+
+void takeoff_and_hover_at_altitude(float altitude_m)
 {
     DroneCore dc;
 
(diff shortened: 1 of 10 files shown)


integration_tests/simple_hover.cpp
```diff
@@ -13,31 +33,55 @@ TEST_F(SitlTest, ActionSimpleHover)
 
     // Wait for device to connect via heartbeat.
     std::this_thread::sleep_for(std::chrono::seconds(2));
+    ASSERT_TRUE(dc.is_connected());
 
     Device &device = dc.device();
 
-    while (!device.telemetry().health_all_ok()) {
+    for (int i = 0; ; ++i) {
```
hamishwillee: Just FMI, why not a while loop? Seem odd to see an infinite for loop :-)

hamishwillee: Ignore that, I see you want to count out 10 seconds.

julianoes: Yer it's a bit ugly but I haven't found a prettier way yet.

An alternative is to use the async interface and then do a `wait_for` on the promise.

hamishwillee: Not prettier, but perhaps a little more intuitive
```
int itter=0;
while (!device.telemetry().health_all_ok()) {
   std::cout << "waiting for device to be ready" << std::endl;
   ASSERT_LT(++itter, 10);
}
```

julianoes: Yer, you're right. That's nicer. I'll change it.

julianoes:
Replaced by #110.

PR #104, opened 2017-09-27
fix comment, use lambdas for mission progress

julianoes:
Thanks for spotting this @hamishwillee.

- Fixed a wrong comment.
- Changed mission progress handling into lambda because it is more readable.


hamishwillee:
@julianoes Looks good to me, but is out of date with base. Can you merge when you have a chance?

PR #108, opened 2017-09-29
Various small fixes

julianoes:
A few printfs and a param fix.


core/mavlink_parameters.h
```diff
@@ -64,7 +64,15 @@ class MavlinkParameters
         void set_from_mavlink_param_ext_value(mavlink_param_ext_value_t mavlink_ext_value)
         {
             switch (mavlink_ext_value.param_type) {
-                // FIXME: the camera params are UINT32 but we treat them as INT32.
+                // FIXME: the camera params are others but we treat them as INT32.
```
darioxz: What about adding an assert in the default case? Could be useful in the integration tests.

julianoes: Good idea, done.

PR #116, opened 2017-10-13
Offboard example

shakthi-prashanth-m:
Adding Offboard velocity control app.

Offboard velocity control is demonstrated using
* Local NED
* Body NED


.gitignore
```diff
@@ -3,3 +3,8 @@
 **/install
 **/logs
 .vscode
+DroneCore.config
```
julianoes: Can you make a comment where these are from?

example/offboard_velocity/offboard_velocity.cpp
```diff
@@ -0,0 +1,229 @@
+//
+// Example that demonstrates Offboard velocity control using LOCAL_NED & BODY_NED
```
julianoes: What about?
`Example that demonstrates offboard velocity control in local NED and body coordinates`

example/offboard_velocity/offboard_velocity.cpp
```diff
@@ -0,0 +1,229 @@
...
+
+using namespace dronecore;
+
+#define ERROR_CONSOLE_TEXT "\033[31m" //Turn text on console red
+#define TELEMETRY_CONSOLE_TEXT "\033[34m" //Turn text on console blue
+#define NORMAL_CONSOLE_TEXT "\033[0m"  //Restore normal console colour
+
+void wait_msec(int msec)
+{
+    std::this_thread::sleep_for(std::chrono::milliseconds(msec));
```
julianoes: You could alternatively do this:
```
using std::this_thread::sleep_for;
using std::chrono::milliseconds;

sleep_for(milliseconds(100));
```

example/offboard_velocity/offboard_velocity.cpp
```diff
@@ -0,0 +1,229 @@
...
+}
+
+void wait_sec(int sec)
+{
+    std::this_thread::sleep_for(std::chrono::seconds(sec));
```
julianoes: Same here.

example/offboard_velocity/offboard_velocity.cpp
```diff
@@ -0,0 +1,229 @@
...
+    if (ret != DroneCore::ConnectionResult::SUCCESS) {
+        std::cout << ERROR_CONSOLE_TEXT << "Connection failed: "
+                  << DroneCore::connection_result_str(ret)
+                  << NORMAL_CONSOLE_TEXT << std::endl;
+        return EXIT_FAILURE;
+    }
+
+    // Wait for device to connect via heartbeat.
+    wait_sec(2);
+    Device &device = dc.device();
```
julianoes: You can now use `is_connected()`, see https://github.com/dronecore/DroneCore/blob/develop/include/dronecore.h#L161.

example/offboard_velocity/offboard_velocity.cpp
```diff
@@ -0,0 +1,229 @@
...
+        }
+        std::cout << "[" << offb_mode << "] Done..." << std::endl;
+        wait_sec(5);
+
+        {
+            const float step_size = 0.01f;
+            const float one_cycle = 2.0f * (float)M_PI;
+            const unsigned steps = (unsigned)(one_cycle / step_size);
+
+            std::cout << "[" << offb_mode << "]" << " Go right & oscilate" << std::endl;
```
julianoes: `oscilate` -> `oscillate`

example/offboard_velocity/offboard_velocity.cpp
```diff
@@ -0,0 +1,229 @@
...
+        std::cout << ERROR_CONSOLE_TEXT << "Takeoff failed:" << Action::result_str(
+                      takeoff_result) << NORMAL_CONSOLE_TEXT << std::endl;
+        return EXIT_FAILURE;
+    }
+
+    std::cout << "In Air" << std::endl;
+    wait_sec(5);
+
+    //////////////////////////////////////////////////////////////////////////////
+    if (offb_mode == "local-ned") {
```
julianoes: What about having two functions for the two if cases. That would make it easier to read and not require `/////...`.

.gitignore
shakthi-prashanth-m: ok, I'll add it. It comes from Qt  Creator

example/offboard_velocity/offboard_velocity.cpp
shakthi-prashanth-m: right..

mrpollo:
@shakthi-prashanth-m can you please rebase this and push to upstream? I think Travis needs a little push

example/offboard_velocity/offboard_velocity.cpp
```diff
@@ -0,0 +1,229 @@
...
+{
+    std::this_thread::sleep_for(std::chrono::seconds(sec));
+}
+
+int main(int argc, char **argv)
+{
+    DroneCore dc;
+    std::string offb_mode = (argc > 1) ? argv[1] : "";
+
+    if (argc == 1 || (offb_mode != "local-ned" && offb_mode != "body-ned")) {
```
mrpollo: can we have this default to local? this would simplify the main routine, and perhaps use `"body"` instead of `"body-ned"` for the argument, just to simplify this

example/offboard_velocity/offboard_velocity.cpp
```diff
@@ -0,0 +1,229 @@
...
+        std::cerr << "Usage: offboard <coordinate-frame>\n" <<
+                  "coordinate-frame: \"local-ned\", or \"body-ned\""
```
mrpollo: if we default to local then this would explain that we are using local by default and explain how to enable body ned

example/offboard_velocity/offboard_velocity.cpp
```diff
@@ -0,0 +1,229 @@
...
+                  << DroneCore::connection_result_str(ret)
+                  << NORMAL_CONSOLE_TEXT << std::endl;
+        return EXIT_FAILURE;
+    }
+
+    // Wait for device to connect via heartbeat.
+    wait_sec(2);
+    Device &device = dc.device();
+
+    while (!device.telemetry().health_all_ok()) {
```
mrpollo: what about [`health_all_ok_async`](https://github.com/dronecore/DroneCore/blob/develop/plugins/telemetry/telemetry_impl.h#L76)?

although this would force you to split your code even further

example/offboard_velocity/offboard_velocity.cpp
```diff
@@ -0,0 +1,229 @@
...
+
+    Action::Result takeoff_result = device.action().takeoff();
+    if (takeoff_result != Action::Result::SUCCESS) {
+        std::cout << ERROR_CONSOLE_TEXT << "Takeoff failed:" << Action::result_str(
+                      takeoff_result) << NORMAL_CONSOLE_TEXT << std::endl;
+        return EXIT_FAILURE;
+    }
+
+    std::cout << "In Air" << std::endl;
+    wait_sec(5);
```
mrpollo: you could use `[in_air_async](https://github.com/dronecore/DroneCore/blob/develop/plugins/telemetry/telemetry_impl.h#L65)` instead of waiting

example/offboard_velocity/offboard_velocity.cpp
mrpollo: It would certainly allow for more modular code

example/offboard_velocity/offboard_velocity.cpp
```diff
@@ -0,0 +1,229 @@
...
+            device.offboard().set_velocity_ned({0.0f, 0.0f, 0.0f, 270.0f});
+            wait_msec(10);
+        }
+        std::cout << "[" << offb_mode << "] Done..." << std::endl;
+        wait_sec(2);
+
+        std::cout << "[" << offb_mode << "]" << " Go UP 2 m/s, Turn clock-wise 180 deg" << std::endl;
+        for (unsigned i = 0; i < 400; ++i) {
+            device.offboard().set_velocity_ned({0.0f, 0.0f, -2.0f, 180.0f});
+            wait_msec(10);
```
mrpollo: @julianoes I loathe waiting using timers to catch up with movement it can get unpredictable and unstable, this is something I always wanted to fix in DroneKit, we should come up with a way to either block for N time by an argument on set_velocity or find a way to have a callback

This isn't something you should worry on this PR @shakthi-prashanth-m, I'm just venting here, I would like to break this pattern in UAV development, we should be able to be more precise

example/offboard_velocity/offboard_velocity.cpp
hamishwillee: Generally I don't see much use for the async APIs for managing a single vehicle. 

In any case though we might use this as an opportunity to show off some "comprehensive" takeoff code, as per [this PR](https://github.com/dronecore/docs/pull/42)
- Close app if vehicle not calibrated
- Print what we're waiting on (maybe)
- Check that arming occurred before calling takeoff
- Check that we reached takeoff height rather than using timer. This is a bit better than in_air() because you actually know you have completed the takeoff.

example/offboard_velocity/offboard_velocity.cpp
hamishwillee: Don't know what you mean by "catch up with movement". 

Generally for velocity control examples using timers makes sense because they are something like "travel this far for X time". If we wanted to control on distance we'd use the position variants of the functions. If we do want to control on distance anyway, then we might add telemetry to measure that distance and block on meeting targets.

example/offboard_velocity/offboard_velocity.cpp
shakthi-prashanth-m: ok.. Do you mean when the example is launched with no arguments, it should start offboard with local NED coordinates mode ?

example/offboard_velocity/offboard_velocity.cpp
shakthi-prashanth-m: We need to report usage of the example anyway even if we default to local. So, I thought we shall leave this to user.

example/offboard_velocity/offboard_velocity.cpp
shakthi-prashanth-m: @mrpollo, thanks. I will rebase with changes suggested and push new changes may be tomorrow.

example/offboard_velocity/offboard_velocity.cpp
hamishwillee: IMO code should show both modes at work - not an either/or choice determined by user. That way you can show off the nuances of using each approach.

example/offboard_velocity/offboard_velocity.cpp
hamishwillee: Disagree - this synchronous is better. 

@mrpollo The code can't progress until health is achieved so if you were to do this aysnc style you'd end up using promises to effectively create a synchronous solution. The only reason to use async here would be if the code was a library potentially running in a context that could block other code (e.g. UI).  There are very few cases with single-vehicle drones where using the async approach makes - because we're only every waiting on one instruction, and instructions are mostly daisychained. 
It might be worth using async to manage multiple vehicles, or if we can find a case where we need code to run in parallel.

example/offboard_velocity/offboard_velocity.cpp
hamishwillee: I tend to think that the two frame cases are separate, and users will tend to copy the code for the frame they are interested in using. Furthermore, what if PX4 adds additional frame support? So I would tend to have completely separate code blocks - and even only use functions if it aids the explanation.

example/offboard_velocity/offboard_velocity.cpp
shakthi-prashanth-m: @hamishwillee that is fair enough..

example/offboard_velocity/offboard_velocity.cpp
shakthi-prashanth-m: Done.

example/offboard_velocity/offboard_velocity.cpp
shakthi-prashanth-m: As of now, we're demonstrating both NED & Body co-ordinates. So, removed command-line handling.

example/offboard_velocity/offboard_velocity.cpp
shakthi-prashanth-m: @hamishwillee: I agree. I feel its fine to use sync API. There is no vital task to perform until `device.telemetry().health_all_ok()` return true. @mrpollo is that ok ?

example/offboard_velocity/offboard_velocity.cpp
shakthi-prashanth-m: @mrpollo, this is basically to let drone reach takeoff altitude before we start issuing OFFBOARD commands. I think it should be fine to sleep for 5 seconds, similar to the way done in [takeoff land](https://github.com/dronecore/DroneCore/blob/master/example/takeoff_land/takeoff_and_land.cpp#L96) example.

example/offboard_velocity/offboard_velocity.cpp
```diff
@@ -0,0 +1,266 @@
...
+#define TELEMETRY_CONSOLE_TEXT "\033[34m" //Turn text on console blue
+#define NORMAL_CONSOLE_TEXT "\033[0m"  //Restore normal console colour
+
+// Handles Action's result
+inline void action_error_exit(Action::Result result, const std::string &message)
+{
+    if (result != Action::Result::SUCCESS) {
+        std::cerr << ERROR_CONSOLE_TEXT << message << Action::result_str(
+                      result) << NORMAL_CONSOLE_TEXT << std::endl;
+        exit(EXIT_FAILURE);
```
hamishwillee: Just for me, where is EXIT_FAILURE defined?

example/offboard_velocity/offboard_velocity.cpp
```diff
@@ -0,0 +1,266 @@
...
+    Offboard::Result offboard_result = device.offboard().start();
+    offboard_error_exit(offboard_result, "Offboard start failed");
+    offboard_log(offb_mode, " OFFBOARD started");
+    sleep_for(seconds(1));
+
+    // Let yaw settle.
+    offboard_log(offb_mode, " Let yaw settle...");
+    for (unsigned i = 0; i < 100; ++i) {
+        device.offboard().set_velocity_ned({0.0f, 0.0f, 0.0f, 90.0f});
+        sleep_for(milliseconds(10));
```
hamishwillee: @julianoes Are we expecting the API client to maintain the resend rate? Not saying we shouldn't, just that we should think about whether there is any way to make this easier for developers. 
I don't think it matters too much for velocity control, but definitely for position control it would be much better if the developer could send a message "move to position" and not have to control resends.

Perhaps I am overthinking this.

hamishwillee:
@shakthi-prashanth-m Great job!

hamishwillee:
@shakthi-prashanth-m Would be good to add [the calibration checks](https://docs.dronecore.io/en/guide/taking_off_landing.html#health-check) (at least) before the health check.

example/offboard_velocity/offboard_velocity.cpp
shakthi-prashanth-m: `stdlib.h`

shakthi-prashanth-m:
@hamishwillee  Thanks!
Will look into adding callbacks. Thanks for the suggestion.

example/offboard_velocity/offboard_velocity.cpp
julianoes: I agree that we might want to handle that internally.

PR #118, opened 2017-10-16
info: add vendor and product ID and name

julianoes:
This adds the vendor and product ID and name back.

### DIFF
--- a/integration_tests/info.cpp
+++ b/integration_tests/info.cpp
@@ -44,6 +44,12 @@ TEST_F(SitlTest, Info)
         // FIXME: This is currently 0.
         //EXPECT_NE(version.os_sw_major, 0);
 
+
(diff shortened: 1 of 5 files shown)


julianoes:
I guess it could also live in the same struct, you're right. But version is misleading for a vendor and product name.

julianoes:
I'll merge this for now, we can always change it.

hamishwillee:
One bugger about this develop codeline approach is that it isn't clear when the changes are available in mainline/release - so nothing to trigger me to regenerate our API ref.

julianoes:
I vote to get rid of develop and work on master and have releases of master.

hamishwillee:
Me too. @mrpollo any objections to getting rid of develop and using releases of master? We can always create a release branch off master or just create a tag. If we find that managing development gets tough we can always add it back in again.

PR #119, opened 2017-10-16
core: add mavlink statustexts as debug printfs

julianoes:
By printing all mavlink statustexts we should get a bit more context
when something goes wrong on the firmware side.

@bkueng do you have a better suggestion how to make sure the `statustext.text` string is null terminated

### DIFF
--- a/core/device_impl.cpp
+++ b/core/device_impl.cpp
@@ -24,6 +24,10 @@ DeviceImpl::DeviceImpl(DroneCoreImpl *parent,
     register_mavlink_message_handler(
         MAVLINK_MSG_ID_AUTOPILOT_VERSION,
         std::bind(&DeviceImpl::process_autopilot_version, this, _1), this);
+
+    register_mavlink_message_handler(
+        MAVLINK_MSG_ID_STATUSTEXT,
+        std::bind(&DeviceImpl::process_statustext, this, _1), this);
 }
 
 DeviceImpl::~DeviceImpl()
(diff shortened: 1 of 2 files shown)


core/device_impl.cpp
```diff
@@ -155,6 +159,50 @@ void DeviceImpl::process_autopilot_version(const mavlink_message_t &message)
     unregister_timeout_handler(_autopilot_version_timed_out_cookie);
 }
 
+void DeviceImpl::process_statustext(const mavlink_message_t &message)
+{
+    mavlink_statustext_t statustext;
+    mavlink_msg_statustext_decode(&message, &statustext);
+
+    std::string debug_str = "mavlink ";
```
hamishwillee: Is this enough to differentiate device as opposed to DroneCore originated warnings? Maybe text:
`MAVLink: emergency` or `Device: emergency`. 
Might even be worth having a special class of debug for these. Just a thought.

mrpollo: Agreed @hamishwillee, it would be good to understand when are these called and if we should be raising exceptions or alerting the users rather than just using them for debugging.

That said this PR is to have more debugging data, so we should merge and get someone to research onto what the messages mean, I want to know if we should be taking more critical actions

core/device_impl.cpp
```diff
@@ -155,6 +159,50 @@ void DeviceImpl::process_autopilot_version(const mavlink_message_t &message)
...
+            debug_str += "debug";
+            break;
+        default:
+            break;
+    }
+
+    // statustext.text is not null terminated, therefore we copy it first to
+    // an array big enough that is zeroed.
+    char text_with_null[sizeof(statustext.text) + 1] {};
+    memcpy(text_with_null, statustext.text, sizeof(statustext.text));
```
mrpollo: dirty indeed, but I don't see any other way around this if we can't control the input, and now that I think about it, can this lead to memory leaks? can someone use this to DOS an app using DroneCore by crafting big malicious messages?

julianoes: I think all that a malicious drone can do here is to print a bunch of memory that follows after `statustext.text` until `0` appears.

core/device_impl.cpp
julianoes: @hamishwillee I can change it to `MAVLink: emergency`.

@mrpollo to catch errors, it might make sense that the plugins check for errors and inform the API user if it is relevant.

hamishwillee: > @mrpollo That said this PR is to have more debugging data, so we should merge and get someone to research onto what the messages mean, I want to know if we should be taking more critical actions

> @mrpollo to catch errors, it might make sense that the plugins check for errors and inform the API user if it is relevant.

> @hamishwillee I can change it to MAVLink: emergency.

Agree with all above - yes, ideally plugins should check for errors and work out what is relevant but we need to do more research on this. IMO we should do change suggested (MAVLink: xxxx) for now, accept PR, and raise issue to look at errors as a separate task.

julianoes:
So, does anyone feel like this can be approved and be merged now? 😄

core/device_impl.cpp
mrpollo: In conclusion, yes we make ourselves open to a crafty mavlink attack, but I'm sure this isn't the only place where this occurs

mrpollo:
@julianoes ready be to merged as soon as you rebase

core/device_impl.cpp
hamishwillee: Yeah, merge whenever. I have created https://github.com/dronecore/DroneCore/issues/127 to add eventual task to look at these messages.

hamishwillee:
When do we plan to merge all this stuff into master. I don't want to document stuff that isn't merged, since currently people build off master.

julianoes:
> When do we plan to merge all this stuff into master. I don't want to document stuff that isn't merged, since currently people build off master.

I think the docs need to reference to a version. I thought you could do that?

hamishwillee:
We can do versions using [this plugin](https://plugins.gitbook.com/plugin/versions). I have not tried it because none of the other projects wanted real versioning. 

I guess I could try creating a develop branch and merge in docs that way. I'll look into that today.

PR #124, opened 2017-10-20
Added gimbal plugin

julianoes:
This adds a plugin which can set pitch and yaw angle of a gimbal.
Currently, the commands are sent to component ID 1, so the assumption is
that the gimbal is connected to the autopilot and not talking mavlink
itself. This could change or be made adaptable in the future.

### DIFF
--- a/plugins/gimbal/gimbal.h
+++ b/plugins/gimbal/gimbal.h
@@ -6,33 +6,87 @@ namespace dronecore {
 
 class GimbalImpl;
 
+/**
+ * @brief The Gimbal class enables to control a gimbal.
+ *
+ * Synchronous and asynchronous variants of the gimbal methods are supplied.
+ */
 class Gimbal
 {
 public:
+    /**
+     * @brief Constructor (internal use only).
+     *
+     * @param impl Private internal implementation.
+     */
     explicit Gimbal(GimbalImpl *impl);
+
+    /**
+     * @brief Destructor (internal use only).
+     */
     ~Gimbal();
 
+    /**
+     * @brief Possible results returned for gimbal commands.
+     */
     enum class Result {
-        SUCCESS = 0,
-        ERROR,
-        TIMEOUT,
-        UNKNOWN
+        SUCCESS = 0, /**< @brief Success. The gimbal command was accepted. */
+        ERROR, /**< @brief Error. An error occured sending the command. */
+        TIMEOUT, /**< @brief Timeout. A timeout occured sending the command. */
+        UNKNOWN /**< @brief Unspecified error. */
     };
 
+    /**
+     * @brief Returns a human-readable English string for Gimbal::Result.
+     *
+     * @param Result The enum value for which a human readable string is required.
+     * @return Human readable string for the Gimbal::Result.
+     */
     static const char *result_str(Result result);
 
+    /**
+     * @brief Callback type for asynchronous Gimbal calls.
+     */
     typedef std::function<void(Result)> result_callback_t;
 
+    /**
+     * @brief Set gimbal pitch and yaw angles (synchronous).
+     *
+     * This sets the desired pitch and yaw angles of a gimbal.
+     * The function will return when the command is accepted, however, it might
+     * take the gimbal longer to actually be set to the new angles.
+     *
+     * @param pitch_deg The pitch angle in degrees. Negative to point down.
+     * @param yaw_deg The yaw angle in degrees. Positive for clock-wise, range -180..180 or 0..360.
+     * @return Result of request.
+     */
     Result set_pitch_and_yaw(float pitch_deg, float yaw_deg);
 
+    /**
+     * @brief Set gimbal pitch and yaw angles (asynchronous).
+     *
+     * This sets the desired pitch and yaw angles of a gimbal.
+     * The callback will be called when the command is accepted, however, it might
+     * take the gimbal longer to actually be set to the new angles.
+     *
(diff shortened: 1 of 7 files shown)


plugins/gimbal/gimbal.h
```diff
@@ -0,0 +1,39 @@
...
+    ~Gimbal();
+
+    enum class Result {
+        SUCCESS = 0,
+        ERROR,
+        TIMEOUT,
+        UNKNOWN
+    };
+
+    static const char *result_str(Result result);
```
hamishwillee: No docs on any of this.

plugins/gimbal/gimbal.h
```diff
@@ -0,0 +1,39 @@
...
+
+    typedef std::function<void(Result)> result_callback_t;
+
```
hamishwillee: Need to add docs and make clear co-ordinate frame used.

plugins/gimbal/gimbal.h
```diff
@@ -0,0 +1,39 @@
...
+class GimbalImpl;
+
+class Gimbal
+{
+public:
+    explicit Gimbal(GimbalImpl *impl);
+    ~Gimbal();
+
+    enum class Result {
```
hamishwillee: Why only this set? Can a camera not return CommandDenied, Busy or whatever?

hamishwillee:
> Currently, the commands are sent to component ID 1, so the assumption is that the gimbal is connected to the autopilot and not talking mavlink itself. This could change or be made adaptable in the future.

So is component ID = 1 the autopilot? - ie we're assuming that the gimbal is part of the autopilot? 

Don't Yuneec use a MAVLink camera? I.e. don't we already have a case of a camera that might have a separate component ID? I'd be tempted to add a mechanism to support that now. If not, then this "limitation" should be part of the documentation.

A few other thoughts:
- Is it possible for a vehicle to have multiple gimbals?
- Is it possible that a gimbal will have connection semantics?
- What else might we want to do other than control pitch and yaw? 
  - Any other axis?
  - Power up/down? Retract?
 - What if a camera only allows control over a single axis? Errors reported?

julianoes:
> So is component ID = 1 the autopilot? - ie we're assuming that the gimbal is part of the autopilot?

> Don't Yuneec use a MAVLink camera? I.e. don't we already have a case of a camera that might have a separate component ID? I'd be tempted to add a mechanism to support that now. If not, then this "limitation" should be part of the documentation.

Both cases are possible. I'm not sure which one makes sense to implement, and an automatic mechanism needs a bit more work. For instance, you can check if you have received heartbeats from a gimbal component ID. If yes, you send the commands there, otherwise to the autopilot.

> A few other thoughts:
> 
>     - Is it possible for a vehicle to have multiple gimbals?

If you have different component IDs for the multiple gimbals, yes. Right now, only one exists [MAV_COMP_ID_GIMBAL](http://mavlink.org/messages/common#MAV_COMP_ID_GIMBAL).

>     - Is it possible that a gimbal will have connection semantics?

It would be heartbeats being sent from the gimbal component ID, as suggested above.

>     - What else might we want to do other than control pitch and yaw?
>         - Any other axis?

There is roll but I have never seen a drone that wants to have a roll axis different from horizontal 😄 

>         - Power up/down? Retract?

This is all hypothetical. We can add/implement this when there is an actual use case.

>     - What if a camera only allows control over a single axis? Errors reported?

No idea. I have to ignore anything that does not have a real use case right now. There are endless possibilities.

mrpollo:
@julianoes please rebase

hamishwillee:
OK, so re that whole discussion, sure - no point implementing anything that has no real use case.

I still lean towards proper support for MAVLink camera though - on the basis that it sounds like we do have real use cases.

julianoes:
Support for a mavlink camera? Absolutely! But this is just for a gimbal, and I don't know of a mavlink gimbal yet.

hamishwillee:
OK, good point. I'll got back to my corner now. 

Happy for this to go out. Just FMI, is it possible to get confirmation that the gimbal has reached its target attitude?

julianoes:
> Happy for this to go out. Just FMI, is it possible to get confirmation that the gimbal has reached its target attitude?

So if a gimbal sends [MOUNT_ORIENTATION](http://mavlink.org/messages/common#MOUNT_ORIENTATION) that's what you can use to track the actual angle.

hamishwillee:
>> Just FMI, is it possible to get confirmation that the gimbal has reached its target attitude?

> So if a gimbal sends MOUNT_ORIENTATION that's what you can use to track the actual angle.

Should we expose the current orientation through this API?
I guess you could argue that people know what orientation they want so they can set it and it will change if it is different, or nothing will happen if it is already on that orientation. I lean towards exposing the information, but then you end up with "nothing" if the gimbal isn't sending information. Hmm.

julianoes:
@hamishwillee yes you're right, we should expose it. I'm just wondering if we should expose it in `telemetry` or `gimbal`?

julianoes:
I think we can add that in `telemetry` sometime. I'll merge this for now.

hamishwillee:
>> @hamishwillee yes you're right, we should expose it. I'm just wondering if we should expose it in telemetry or gimbal?

> I think we can add that in telemetry sometime. I'll merge this for now.

Sounds good. I added https://github.com/dronecore/DroneCore/issues/141 to track this. I tend to agree "Telemetry" is the right place for this, but it isn't a strong belief and you could easily convince me otherwise. Either way, if it goes in Telemetry we should add cross link/info in Gimbal docs.

PR #125, opened 2017-10-20
Improve debug printfs

### LINKED ISSUE #103: Make "Debug" output coloured
The output of the debug build with the fly_mission example is shown below. As you can see, the **Info** output is very clear (blue) but the **Debug** is white, and hence not so easy to differentiate from the app's output. 

I think we can assume that, since we aren't making our debug output available, users will almost always output white.

image

I am not sure what colour would be best/we have already used. But see the ubuntu green/blue above are pretty easy to discern!

julianoes:
This makes debug printfs green as requested in #103 (closes #103).
Also, the color is now only applied for `[time|type]` instead of the whole line as agreed upon in #86.

### DIFF
--- a/core/log.h
+++ b/core/log.h
@@ -10,6 +10,7 @@
 #endif
 
 #define ANSI_COLOR_RED     "\x1b[31m"
+#define ANSI_COLOR_GREEN   "\x1b[32m"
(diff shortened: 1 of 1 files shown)


hamishwillee:
Cool. @julianoes , when you do you plan on merging this into master?

PR #130, opened 2017-10-23
Support to generate .deb packages

julianoes:
This is a step towards integration tests for github.com/PX4/Firmware by generating .deb packages on releases.

In order to streamline the .deb generation, I've changed the build to a shared library for Linux and macOS, and made the default install location `/usr/local`.

While doing that I actually realized that libdronecore was not being linked to curl and pthread which required all executable to link against them. This should now be fixed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -76,15 +76,16 @@ install:
 
 script:
 - if [[ "${BUILD_TARGET}" = "gcc_build" ]]; then
-    make EXTERNAL_DIR=external_example;
+    make INSTALL_PREFIX=. EXTERNAL_DIR=external_example;
     make default install;
     make docs;
     make run_unit_tests;
     (cd example/takeoff_land && mkdir build && cd build && cmake .. && make);
     (cd example/fly_mission && mkdir build && cd build && cmake .. && make);
+    (cd example/offboard_velocity && mkdir build && cd build && cmake .. && make);
     make clean;
-    make BUILD_TYPE=Release;
-    make default install;
+    make INSTALL_PREFIX=. BUILD_TYPE=Release;
+    make install;
     make run_unit_tests;
   fi
 - if [[ "${BUILD_TARGET}" = "coverage_build" ]]; then
--- a/.travis.yml
+++ b/.travis.yml
@@ -76,15 +76,15 @@ install:
 
 script:
 - if [[ "${BUILD_TARGET}" = "gcc_build" ]]; then
-    make INSTALL_PREFIX=. EXTERNAL_DIR=external_example;
-    make default install;
+    make INSTALL_PREFIX=/usr EXTERNAL_DIR=external_example;
+    make install;
     make docs;
     make run_unit_tests;
     (cd example/takeoff_land && mkdir build && cd build && cmake .. && make);
     (cd example/fly_mission && mkdir build && cd build && cmake .. && make);
     (cd example/offboard_velocity && mkdir build && cd build && cmake .. && make);
     make clean;
-    make INSTALL_PREFIX=. BUILD_TYPE=Release;
+    make INSTALL_PREFIX=/usr BUILD_TYPE=Release;
     make install;
     make run_unit_tests;
   fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -120,3 +120,19 @@ after_success:
     --exclude="debug_helpers" --exclude="example" --exclude="external_example"
     --exclude-pattern=".*CMakeFiles.*" --exclude-pattern=".*_test.*";
   fi
+
+before_deploy:
+- export RELEASE_PKG_FILE=$(ls *.deb)
+- echo "Deploying $RELEASE_PKG_FILE to GitHub releases"
+
+deploy:
+  provider: releases
+  skip_cleanup: true
+  api_key:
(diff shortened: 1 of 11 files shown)


.travis.yml
```diff
@@ -117,3 +117,19 @@ after_success:
...
+
+before_deploy:
+- export RELEASE_PKG_FILE=$(ls *.deb)
+- echo "Deploying $RELEASE_PKG_FILE to GitHub releases"
+
+deploy:
+  provider: releases
+  skip_cleanup: true
+  api_key:
+    secure: hBX3pFWNZiDbz4yKnOjhLg3QS9Ubn1XePxSeIt2Btq5GzbomOPDCgpIFijBppliwj9oKc302EMnZSg2QWeAzFKn9UnmIflJ0E4iymYgwWdTJv+bSnYALJEmO8F6gF9FgRlPk8FCtZiECoTsa75w5TrEZKZpFpmzVYRiDu0eo6sEjW7UJPC0A2KSTXLrBCHSIZy/iasbGmuur4brG7NO0QdMOvDXvhsYfkXDRJFMTtTHvLiKJcqiunPfqARzf1H4x4iczRYscKu5Vn8Kmw3NANGkcIDvEj4ooih831EXxACRZw0VgycgNHOKRXKC9pZ4hLQMon+jxpQX+X8k/K5161oEkF/gCVKyFb31Pk/4Uwe81p1GJY2lAC7MDUxA98RKXhdvVYF2Cp44+IbF0YVoWRUtVAhknXRQ3Weg25kyVSu83q2nN2nZq2qGTnpNIbdN56s/F+uaFtipGEh+vmiv8rNUz+Z5MFrY2FQaSvBTFw9K4tNs9uc+VQd1bE7X5wh0yywEqUEw2nzqTB2xR+OubygUASbk2GLNdc254P0lrzCHbNM62Y7sRX06CM7hPlwhELEkVtUXZWJ0KuhQyLvRh3aPJ3Jj30EswTt/FGT1gzSP1FjjHBRZCK4P2D2rwJ5TMn2JrZKfPxmEd3kVmn6h80+gBbKgonGmZspd2SvPEI5g=
```
hamishwillee: Probably a stupid question, but if this key is secure, why is it public here?

julianoes: You only see the encrypted string. It's a common way with travis.
https://docs.travis-ci.com/user/environment-variables/#Defining-encrypted-variables-in-.travis.yml

example/fly_mission/CMakeLists.txt
```diff
@@ -6,34 +6,12 @@ if(NOT MSVC)
...
-# Not needed if DroneCore installed system-wide
-include_directories(
-    ${CMAKE_SOURCE_DIR}/../../install/include
-)
-
 add_executable(fly_mission
     fly_mission.cpp
 )
 
-# Not needed if DroneCore installed system-wide
```
hamishwillee: Does the fact this is removed mean that you have sorted out Windows case for installing system wide?

example/fly_mission/CMakeLists.txt
```diff
@@ -6,34 +6,12 @@ if(NOT MSVC)
     add_definitions("-std=c++11 -Wall -Wextra -Werror")
 else()
     add_definitions("-std=c++11 -WX -W2")
-    set(platform_libs "Ws2_32.lib")
 endif()
 
-# Add DEBUG define for Debug target
-set(CMAKE_CXX_FLAGS_DEBUG "-DDEBUG")
```
hamishwillee: Release is built by default. Is this removed because we're happy for release build to be the default?

example/takeoff_land/CMakeLists.txt
```diff
@@ -6,34 +6,12 @@ if(NOT MSVC)
...
 else()
     add_definitions("-std=c++11 -WX -W2")
-    set(platform_libs "Ws2_32.lib")
 endif()
 
-# Add DEBUG define for Debug target
-set(CMAKE_CXX_FLAGS_DEBUG "-DDEBUG")
-
-# This finds thread libs on Linux, Mac, and Windows.
-find_package(Threads REQUIRED)
```
hamishwillee: So this isn't needed because now the underlying dronecore library links to the thread library?

hamishwillee:
So looks good to me, but I haven't actually run it. Main questions are:
1. What happens on windows? Previously we supported only local install, but most of the local install stuff now pulled out of examples.
2. If system wide is supported as default, is there still a case for local install? ie is this something we need to document?
3. How do other things that previously installed to local system work - e.g. building docs.

example/fly_mission/CMakeLists.txt
julianoes: I have currently no access to a Windows machine. I'll come back to the Windows build once I do.

example/fly_mission/CMakeLists.txt
julianoes: This is not actually about the default. All this does is it adds a `#define DEBUG` in case it is a debug build. This is not actually used or needed in the examples, so I removed it.

example/takeoff_land/CMakeLists.txt
julianoes: Exactly.

julianoes:
> 1. What happens on windows? Previously we supported only local install, but most of the local install stuff now pulled out of examples.

I'll figure that out in the next days, probably as part of this PR even.

> 2. If system wide is supported as default, is there still a case for local install? ie is this something we need to document?

I think we can document that you can set the prefix where to install it. I'd leave it to devs to figure out how to link to it for that case.

> 3. How do other things that previously installed to local system work - e.g. building docs.

`make docs` works if you run it with `sudo` which is a little awkward, or we could have a script that builds locally and generates the docs there.

Makefile
```diff
@@ -75,9 +75,10 @@ install: ARGS += install
 install: default
 
 docs: install
-	- mkdir install/docs
-	- (cd install && doxygen ../.doxygen)
-	- python generate_markdown_from_doxygen_xml.py
+	- (cd $(INSTALL_PREFIX)/include/dronecore \
```
hamishwillee: Shouldn't this be  $(CURRENT_DIR) not $(INSTALL_PREFIX) ?

example/fly_mission/CMakeLists.txt
hamishwillee: So the plan is to do that before this is merged? It doesn't "matter" except that the example docs will need to be changed to add "not supported" for windows.

example/fly_mission/CMakeLists.txt
hamishwillee: Thanks for the clarification.

example/fly_mission/CMakeLists.txt
hamishwillee: Ignore that, I see below that you're hopefully going to do that as part of this PR :-)

hamishwillee:
>> 2. If system wide is supported as default, is there still a case for local install? ie is this something we need to document?

> I think we can document that you can set the prefix where to install it. I'd leave it to devs to figure out how to link to it for that case.

Sounds good. We can provide an example in docs.

>> 3. How do other things that previously installed to local system work - e.g. building docs.

> `make docs` works if you run it with `sudo` which is a little awkward, or we could have a script that builds locally and generates the docs there.

OK. Up to you. IMO having script is "nice to have", but hardly a priority. Worth adding an issue so it can be addressed at some later point.

hamishwillee:
I'm good with this once Windows stuff addressed (by which I mean "fixed or explicitly not supported"), but leave @mrpollo to review.

julianoes:
@hamishwillee ok agreed.

PR #131, opened 2017-10-24
mission: add delay before camera action

julianoes:
This adds the possibility to add a delay before taking a picture or
starting a video. This is used to wait for a gimbal to reach its
orientation, or a vehicle to slow down.

The current implementation is not optimal because it is based on the
`LOITER_TIME` waypoint instead of the `NAV_DELAY` command which would be
more appropriate but is not supported (yet).

### DIFF
--- a/plugins/mission/mission_impl.cpp
+++ b/plugins/mission/mission_impl.cpp
@@ -306,6 +306,12 @@ void MissionImpl::assemble_mavlink_messages()
 {
     _mavlink_mission_item_messages.clear();
 
+    bool last_position_valid = false; // This flag is to protect us from using an invalid x/y.
+    MAV_FRAME last_frame;
+    int32_t last_x;
+    int32_t last_y;
+    float last_z;
+
     unsigned item_i = 0;
     for (auto item : _mission_items) {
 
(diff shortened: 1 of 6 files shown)


plugins/mission/mission_impl.cpp
```diff
@@ -326,6 +338,52 @@ void MissionImpl::assemble_mavlink_messages()
...
         }
 
+        // FIXME: It is a bit of a hack to set a LOITER_TIME waypoint to add a delay.
+        //        A better solution would be to properly use NAV_DELAY instead. This
+        //        would not require us to keep the last lat/lon.
+        if (std::isfinite(mission_item_impl.get_camera_action_delay_s())) {
+            if (!last_position_valid) {
+                // In the case where we get a delay without a previous position, we will have to
+                // ignore it.
+                LogErr() << "Can't set camera actoin delay without previous position set.";
```
hamishwillee: Typo "actoin"

hamishwillee:
LGTM. 
Generally I hate delays as mechanisms for getting things ready - they either have to be tuned for particular use cases or made conservatively over-long. Is it possible/reasonable to instead provide checks that speed is correct and gimbal is in position (or don't they provide that feedback)?

julianoes:
>  Generally I hate delays as mechanisms for getting things ready - they either have to be tuned for particular use cases or made conservatively over-long. Is it possible/reasonable to instead provide checks that speed is correct and gimbal is in position (or don't they provide that feedback)?

You're right. The problem is that we just send a command to the gimbal what it's setpoint is but we really don't know how long it will take it. Even if the gimbal would only acknowledge the command when it has reached the setpoint, the firmware might not even wait for a command to be completed before continuing its mission. What I'm saying is that this would require a bunch of changes to avoid a delay like this.

hamishwillee:
Fair enough. I guess also that even if one camera gimbal provided position feedback, that wouldn't be true of all of them.
@mrpollo Want to review this?

mrpollo:
@julianoes please rebase and merge!

PR #134, opened 2017-10-25
offboard: automatically send setpoins at set rate

julianoes:
This changes the offboard API so that it is no longer required to send
setpoints at least every 0.5 seconds. Instead, this enables to set a
setpoint once and only set it again when it changes. The implementation
will automatically send it at 20 Hz internally.

If the setpoints need to be sent at a higher rate, e.g. for very smooth
control, that is still possible by just setting the setpoints at high
rate. Whenever a setpoint changes, it is immediately sent to the
vehicle.

Tested in SITL using the integration tests as well as the offboard example.

### DIFF
--- a/example/offboard_velocity/offboard_velocity.cpp
+++ b/example/offboard_velocity/offboard_velocity.cpp
@@ -56,7 +56,7 @@ inline void connection_error_exit(DroneCore::ConnectionResult result, const std:
 // Logs during Offboard control
 inline void offboard_log(const std::string &offb_mode, const std::string msg)
 {
-    std::cout << "[" << offb_mode << "]" << msg << std::endl;
+    std::cout << "[" << offb_mode << "] " << msg << std::endl;
 }
 
 /**
@@ -69,59 +69,45 @@ bool offb_ctrl_ned(Device &device)
     const std::string offb_mode = "NED";
     // Send it once before starting offboard, otherwise it will be rejected.
     device.offboard().set_velocity_ned({0.0f, 0.0f, 0.0f, 0.0f});
-    offboard_log(offb_mode, " Sent Null velocity command once before starting OFFBOARD");
 
     Offboard::Result offboard_result = device.offboard().start();
     offboard_error_exit(offboard_result, "Offboard start failed");
-    offboard_log(offb_mode, " OFFBOARD started");
-    sleep_for(seconds(1));
+    offboard_log(offb_mode, "Offboard started");
 
-    // Let yaw settle.
-    offboard_log(offb_mode, " Let yaw settle...");
+    offboard_log(offb_mode,  "Turn to face East");
     device.offboard().set_velocity_ned({0.0f, 0.0f, 0.0f, 90.0f});
-    sleep_for(seconds(1));
-    offboard_log(offb_mode, " Done...");
+    sleep_for(seconds(1)); // Let yaw settle.
 
     {
         const float step_size = 0.01f;
         const float one_cycle = 2.0f * (float)M_PI;
-        const unsigned steps = (unsigned)(one_cycle / step_size);
+        const unsigned steps = 2 * unsigned(one_cycle / step_size);
 
-        offboard_log(offb_mode,  " Go right & oscillate");
+        offboard_log(offb_mode,  "Go North and back South");
         for (unsigned i = 0; i < steps; ++i) {
             float vx = 5.0f * sinf(i * step_size);
             device.offboard().set_velocity_ned({vx, 0.0f, 0.0f, 90.0f});
             sleep_for(milliseconds(10));
         }
     }
-    offboard_log(offb_mode, " Done...");
-    // NOTE: Use sleep_for() after each velocity-ned command to closely monitor their behaviour.
 
-    offboard_log(offb_mode,  " Turn clock-wise 270 deg");
+    offboard_log(offb_mode,  "Turn to face West");
     device.offboard().set_velocity_ned({0.0f, 0.0f, 0.0f, 270.0f});
     sleep_for(seconds(2));
 
-    offboard_log(offb_mode, " Done");
 
-    offboard_log(offb_mode, " Go UP 2 m/s, Turn clock-wise 180 deg");
+    offboard_log(offb_mode, "Go up 2 m/s, turn to face South");
     device.offboard().set_velocity_ned({0.0f, 0.0f, -2.0f, 180.0f});
     sleep_for(seconds(4));
-    offboard_log(offb_mode, " Done...");
 
-    offboard_log(offb_mode,  " Turn clock-wise 90 deg");
-    device.offboard().set_velocity_ned({0.0f, 0.0f, 0.0f, 90.0f});
-    sleep_for(seconds(4));
-    offboard_log(offb_mode, " Done...");
-
-    offboard_log(offb_mode,  " Go DOWN 1.0 m/s");
+    offboard_log(offb_mode,  "Go down 1 m/s, turn to face North");
     device.offboard().set_velocity_ned({0.0f, 0.0f, 1.0f, 0.0f});
     sleep_for(seconds(4));
-    offboard_log(offb_mode, " Done...");
 
     // Now, stop offboard mode.
     offboard_result = device.offboard().stop();
     offboard_error_exit(offboard_result, "Offboard stop failed: ");
-    offboard_log(offb_mode, " OFFBOARD stopped");
+    offboard_log(offb_mode, "Offboard stopped");
 
     return true;
 }
@@ -135,53 +121,44 @@ bool offb_ctrl_body(Device &device)
 {
     const std::string offb_mode = "BODY";
 
+    // Send it once before starting offboard, otherwise it will be rejected.
     device.offboard().set_velocity_body({0.0f, 0.0f, 0.0f, 0.0f});
-    offboard_log(offb_mode,  " Sent once before starting OFFBOARD");
 
     Offboard::Result offboard_result = device.offboard().start();
     offboard_error_exit(offboard_result, "Offboard start failed: ");
-    offboard_log(offb_mode, " OFFBOARD started");
+    offboard_log(offb_mode, "Offboard started");
 
-    // Turn around yaw and climb
-    offboard_log(offb_mode, " Turn around yaw & climb");
+    offboard_log(offb_mode, "Turn clock-wise and climb");
     device.offboard().set_velocity_body({0.0f, 0.0f, -1.0f, 60.0f});
-    sleep_for(seconds(2));
+    sleep_for(seconds(5));
 
-    // Turn back
-    offboard_log(offb_mode, " Turn back");
+    offboard_log(offb_mode, "Turn back anti-clockwise");
     device.offboard().set_velocity_body({0.0f, 0.0f, 0.0f, -60.0f});
-    sleep_for(seconds(2));
+    sleep_for(seconds(5));
 
-    // Wait for a bit
-    offboard_log(offb_mode, " Wait for a bit");
+    offboard_log(offb_mode, "Wait for a bit");
     device.offboard().set_velocity_body({0.0f, 0.0f, 0.0f, 0.0f});
     sleep_for(seconds(2));
-    // NOTE: Use sleep_for() after each velocity-ned command to closely monitor their behaviour.
 
-    // Fly a circle
-    offboard_log(offb_mode, " Fly a circle");
-    device.offboard().set_velocity_body({5.0f, 0.0f, 0.0f, 60.0f});
-    sleep_for(seconds(5));
+    offboard_log(offb_mode, "Fly a circle");
+    device.offboard().set_velocity_body({5.0f, 0.0f, 0.0f, 30.0f});
+    sleep_for(seconds(15));
 
-    // Wait for a bit
-    offboard_log(offb_mode, " Wait for a bit");
+    offboard_log(offb_mode, "Wait for a bit");
     device.offboard().set_velocity_body({0.0f, 0.0f, 0.0f, 0.0f});
     sleep_for(seconds(5));
 
-    // Fly a circle sideways
-    offboard_log(offb_mode, " Fly a circle sideways...");
-    device.offboard().set_velocity_body({0.0f, -5.0f, 0.0f, 60.0f});
-    sleep_for(seconds(5));
+    offboard_log(offb_mode, "Fly a circle sideways");
+    device.offboard().set_velocity_body({0.0f, -5.0f, 0.0f, 30.0f});
+    sleep_for(seconds(15));
 
-    // Wait for a bit
-    offboard_log(offb_mode, " Wait for a bit");
+    offboard_log(offb_mode, "Wait for a bit");
     device.offboard().set_velocity_body({0.0f, 0.0f, 0.0f, 0.0f});
-    sleep_for(seconds(5));
+    sleep_for(seconds(8));
 
-    // Now, stop offboard mode.
     offboard_result = device.offboard().stop();
     offboard_error_exit(offboard_result, "Offboard stop failed: ");
-    offboard_log(offb_mode, " OFFBOARD stopped");
+    offboard_log(offb_mode, "Offboard stopped");
 
     return true;
 }
@@ -190,7 +167,6 @@ int main(int, char **)
 {
     DroneCore dc;
 
-    // add udp connection
     DroneCore::ConnectionResult conn_result = dc.add_udp_connection();
     connection_error_exit(conn_result, "Connection failed");
 
@@ -200,7 +176,7 @@ int main(int, char **)
         sleep_for(seconds(1));
     }
 
-    // Device got connected...
+    // Device got discovered.
     Device &device = dc.device();
 
     while (!device.telemetry().health_all_ok()) {
@@ -209,25 +185,22 @@ int main(int, char **)
     }
     std::cout << "Device is ready" << std::endl;
 
-    // Arm
     Action::Result arm_result = device.action().arm();
     action_error_exit(arm_result, "Arming failed");
     std::cout << "Armed" << std::endl;
 
-    // Takeoff
     Action::Result takeoff_result = device.action().takeoff();
     action_error_exit(takeoff_result, "Takeoff failed");
     std::cout << "In Air..." << std::endl;
     sleep_for(seconds(5));
 
-    //  using LOCAL NED co-ordinates
+    //  using local NED co-ordinates
     bool ret = offb_ctrl_ned(device);
     if (ret == false) {
         return EXIT_FAILURE;
     }
-    std::cout << "---------------------------" << std::endl;
 
-    //  using BODY NED co-ordinates
+    //  using body co-ordinates
     ret = offb_ctrl_body(device);
     if (ret == false) {
         return EXIT_FAILURE;
--- a/example/offboard_velocity/offboard_velocity.cpp
+++ b/example/offboard_velocity/offboard_velocity.cpp
@@ -1,7 +1,9 @@
 /**
 * @file offboard_velocity.cpp
 * @brief Example that demonstrates offboard velocity control in local NED and body coordinates
-* @author Author: Julian Oes <julian@oes.ch>, Shakthi Prashanth <shakthi.prashanth.m@intel.com>
+*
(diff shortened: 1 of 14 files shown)


plugins/offboard/offboard_impl.cpp
```diff
@@ -72,6 +83,16 @@ Offboard::Result OffboardImpl::stop() const
...
+    {
+        //std::lock_guard<std::mutex> lock(_mutex);
+
+        if (_mode == Mode::NOT_ACTIVE) {
+            if (callback) {
+                callback(Offboard::Result::NO_SETPOINT_SET);
+            }
+        }
+    }
+
```
hamishwillee: Shouldn't this return here so that the following code isn't executed?

hamishwillee:
In the example
```
/**
* @name connected device
* @brief Does Offboard control using BODY co-ordinates
* returns true if everything went well in Offboard control, exits with a log otherwise.
*/
```
Seems odd to have @ docs for examples. Even so, should be correct - what is @name ? should it be @returns ?

hamishwillee:
Looks awesome. Some thoughts:
1. Is it possible for the vehicle to decide to switch mode due to RC or whatever while the script thinks it is in RC mode? In this case, what should happen / how should the script respond?
2. We should add a paragraph up in the class description explaining a) how message resending should be used b)need to call setpoint before calling mode. 
E.g. update along the lines of:
```
/**
 * @brief This class is used to control a drone with velocity commands.
 *
 * The module is called offboard because the velocity commands can be sent from external sources
 * as opposed to onboard control right inside the autopilot "board".
 *
 * Client code must specify a setpoint before starting offboard mode.
 * DroneCore automatically resends setpoints at 20Hz (PX4 Offboard mode requires that setpoints are
 * minimally resent at 2Hz). If more precise control is required, clients can call the
 * setpoint methods at whatever rate is required.
 *
 * **Attention:** this is work in progress, use with caution!
 */
```
3. Is there ever a case where someone might want to resend at lower than 20Hz?
4. I still think it would be nice to automatically put into the mode rather than requiring "start" semantics. This is however fairly easy to understand.

Will leave for @shakthi-prashanth-m to do technical C++ review.

hamishwillee:
@julianoes As part of this can I please also get a small docs update. For all the members in `VelocityNEDYaw` and `VelocityBodyYawspeed` add "@brief" (easiest way to fix output). So

```
float north_m_s; /**< Velocity North in metres/second. */
```
to
```
float north_m_s; /**< @brief Velocity North in metres/second. */
```
etc.

example/offboard_velocity/offboard_velocity.cpp
```diff
@@ -76,12 +78,9 @@ bool offb_ctrl_ned(Device &device)
```
hamishwillee: @julianoes In line above, why sleep for one second after starting offboard? API should not require artificial timing breaks.

example/offboard_velocity/offboard_velocity.cpp
```diff
@@ -76,12 +78,9 @@ bool offb_ctrl_ned(Device &device)
 
     // Let yaw settle.
```
hamishwillee: Don't think we need this note - since we show it as log in next line.

example/offboard_velocity/offboard_velocity.cpp
```diff
@@ -76,12 +78,9 @@ bool offb_ctrl_ned(Device &device)
...
     offboard_log(offb_mode, " Let yaw settle...");
```
hamishwillee: This log is confusing - because at this point we haven't told the user we're yawing.  We also report done, which I think is redundant since we're not actually checking that we're done and we then say the next step.

Suggest replace with following text - states that we're yawing and gives user 2 seconds to watch that behaviour. We don't need to tell the watcher that yaw is settling, but we add a comment so programmers know why we added that sleep.
```
    offboard_log(offb_mode, " Turn (yaw) 90% clockwise...");
    device.offboard().set_velocity_ned({0.0f, 0.0f, 0.0f, 90.0f});
    sleep_for(seconds(2));  // Let yaw settle
```
Would also be cool if we could control the yaw rate - is that possible?

example/offboard_velocity/offboard_velocity.cpp
```diff
@@ -99,31 +98,24 @@ bool offb_ctrl_ned(Device &device)
     // NOTE: Use sleep_for() after each velocity-ned command to closely monitor their behaviour.
```
hamishwillee: Section above is confusing to watch and the code is a bit complicated:
- The vehicle does not "go right". It turns to face east unless it is already facing east.
- It technically does "oscillate", but that could mean almost anything.
- Propose change text to "Oscillate in a North-South path while facing East (turning if necessary)"
- Propose make this do two cycles to be easier to observe when running the example (ie. ` const unsigned steps = (unsigned)(one_cycle / step_size);` to `const unsigned steps = 2 * (unsigned)(one_cycle / step_size);`
- Maybe "Done" is OK here because the action is quite extended.

example/offboard_velocity/offboard_velocity.cpp
```diff
@@ -99,31 +98,24 @@ bool offb_ctrl_ned(Device &device)
...
 
     offboard_log(offb_mode,  " Turn clock-wise 270 deg");
```
hamishwillee: "Turn clockwise to face West (270 deg)."

example/offboard_velocity/offboard_velocity.cpp
```diff
@@ -99,31 +98,24 @@ bool offb_ctrl_ned(Device &device)
...
-    for (unsigned i = 0; i < 400; ++i) {
-        device.offboard().set_velocity_ned({0.0f, 0.0f, 0.0f, 270.0f});
-        sleep_for(milliseconds(10));
-    }
+    device.offboard().set_velocity_ned({0.0f, 0.0f, 0.0f, 270.0f});
+    sleep_for(seconds(2));
+
     offboard_log(offb_mode, " Done");
 
     offboard_log(offb_mode, " Go UP 2 m/s, Turn clock-wise 180 deg");
```
hamishwillee: "Go UP 2 m/s. Turn clockwise to face South (180 deg.)"

hamishwillee: @julianoes When this command executes the vehicle should turn clockwise from 270 through to 0 and back round to 180 - because the yaw value is positive. What I'm seeing in SITL is that it goes anticlockwise back to 180 irrespective of the sign on the yaw value. So either a bug in our docs, or dronecore or PX4.

example/offboard_velocity/offboard_velocity.cpp
```diff
@@ -99,31 +98,24 @@ bool offb_ctrl_ned(Device &device)
...
-    for (unsigned i = 0; i < 400; ++i) {
-        device.offboard().set_velocity_ned({0.0f, 0.0f, -2.0f, 180.0f});
-        sleep_for(milliseconds(10));
-    }
+    device.offboard().set_velocity_ned({0.0f, 0.0f, -2.0f, 180.0f});
+    sleep_for(seconds(4));
     offboard_log(offb_mode, " Done...");
 
     offboard_log(offb_mode,  " Turn clock-wise 90 deg");
```
hamishwillee: Turn clockwise to face East (90 deg.)"

example/offboard_velocity/offboard_velocity.cpp
```diff
@@ -99,31 +98,24 @@ bool offb_ctrl_ned(Device &device)
...
-    for (unsigned i = 0; i < 400; ++i) {
-        device.offboard().set_velocity_ned({0.0f, 0.0f, 0.0f, 90.0f});
-        sleep_for(milliseconds(10));
-    }
+    device.offboard().set_velocity_ned({0.0f, 0.0f, 0.0f, 90.0f});
+    sleep_for(seconds(4));
     offboard_log(offb_mode, " Done...");
 
     offboard_log(offb_mode,  " Go DOWN 1.0 m/s");
```
hamishwillee: Go DOWN 1.0 m. Turn clockwise to face North (0 deg.)

example/offboard_velocity/offboard_velocity.cpp
hamishwillee: It would also be good if at least one of these cases turned anticlockwise to demonstrate that working. Certainly we need that in the test code.

example/offboard_velocity/offboard_velocity.cpp
```diff
@@ -152,53 +144,39 @@ bool offb_ctrl_body(Device &device)
 
     // Turn around yaw and climb
     offboard_log(offb_mode, " Turn around yaw & climb");
```
hamishwillee: Text should be: " Go UP 1 m/s. Turn clockwise at 60 degrees/second."

example/offboard_velocity/offboard_velocity.cpp
```diff
@@ -152,53 +144,39 @@ bool offb_ctrl_body(Device &device)
...
-    for (unsigned i = 0; i < 200; ++i) {
-        device.offboard().set_velocity_body({0.0f, 0.0f, -1.0f, 60.0f});
-        sleep_for(milliseconds(10));
-    }
+    device.offboard().set_velocity_body({0.0f, 0.0f, -1.0f, 60.0f});
+    sleep_for(seconds(2));
```
hamishwillee: Shouldn't all these cases print "Done" for consistency? Either all should do or all should not do.

example/offboard_velocity/offboard_velocity.cpp
```diff
@@ -152,53 +144,39 @@ bool offb_ctrl_body(Device &device)
...
 
     // Turn back
     offboard_log(offb_mode, " Turn back");
```
hamishwillee: "  Turn the other direction (anticlockwise) at 60 degrees/second."

example/offboard_velocity/offboard_velocity.cpp
```diff
@@ -152,53 +144,39 @@ bool offb_ctrl_body(Device &device)
 
     // Turn around yaw and climb
     offboard_log(offb_mode, " Turn around yaw & climb");
-    for (unsigned i = 0; i < 200; ++i) {
-        device.offboard().set_velocity_body({0.0f, 0.0f, -1.0f, 60.0f});
-        sleep_for(milliseconds(10));
-    }
+    device.offboard().set_velocity_body({0.0f, 0.0f, -1.0f, 60.0f});
+    sleep_for(seconds(2));
```
hamishwillee: Maybe sleep 5 seconds - so this is easy to observe for someone viewing the example.

example/offboard_velocity/offboard_velocity.cpp
```diff
@@ -152,53 +144,39 @@ bool offb_ctrl_body(Device &device)
...
 
     // Turn back
     offboard_log(offb_mode, " Turn back");
-    for (unsigned i = 0; i < 200; ++i) {
-        device.offboard().set_velocity_body({0.0f, 0.0f, 0.0f, -60.0f});
-        sleep_for(milliseconds(10));
-    }
+    device.offboard().set_velocity_body({0.0f, 0.0f, 0.0f, -60.0f});
+    sleep_for(seconds(2));
```
hamishwillee: Maybe sleep 5 seconds - so this is easy to observe for someone viewing the example.

example/offboard_velocity/offboard_velocity.cpp
```diff
@@ -152,53 +144,39 @@ bool offb_ctrl_body(Device &device)
...
 
     // Wait for a bit
     offboard_log(offb_mode, " Wait for a bit");
-    for (unsigned i = 0; i < 200; ++i) {
-        device.offboard().set_velocity_body({0.0f, 0.0f, 0.0f, 0.0f});
-        sleep_for(milliseconds(10));
-    }
+    device.offboard().set_velocity_body({0.0f, 0.0f, 0.0f, 0.0f});
+    sleep_for(seconds(2));
     // NOTE: Use sleep_for() after each velocity-ned command to closely monitor their behaviour.
```
hamishwillee: Duplicate Note. If this is to stay, make sure it changes to "after velocity-body".

example/offboard_velocity/offboard_velocity.cpp
```diff
@@ -152,53 +144,39 @@ bool offb_ctrl_body(Device &device)
...
 
     // Fly a circle
     offboard_log(offb_mode, " Fly a circle");
-    for (unsigned i = 0; i < 500; ++i) {
-        device.offboard().set_velocity_body({5.0f, 0.0f, 0.0f, 60.0f});
-        sleep_for(milliseconds(10));
-    }
+    device.offboard().set_velocity_body({5.0f, 0.0f, 0.0f, 60.0f});
```
hamishwillee: Propose rotate at 30.0f (bigger circle) and wait for 15 seconds.

example/offboard_velocity/offboard_velocity.cpp
```diff
@@ -152,53 +144,39 @@ bool offb_ctrl_body(Device &device)
...
+    sleep_for(seconds(5));
 
     // Wait for a bit
     offboard_log(offb_mode, " Wait for a bit");
-    for (unsigned i = 0; i < 500; ++i) {
-        device.offboard().set_velocity_body({0.0f, 0.0f, 0.0f, 0.0f});
-        sleep_for(milliseconds(10));
-    }
+    device.offboard().set_velocity_body({0.0f, 0.0f, 0.0f, 0.0f});
+    sleep_for(seconds(5));
```
hamishwillee: Change to at least 8 seconds.

plugins/offboard/offboard_impl.cpp
julianoes: Very good catch, thank you!

julianoes:
> Seems odd to have @ docs for examples. Even so, should be correct - what is @name ? should it be @returns ?

Agreed, I would remove the whole comment.

julianoes:
> @julianoes As part of this can I please also get a small docs update. For all the members in VelocityNEDYaw and VelocityBodyYawspeed add "@brief" (easiest way to fix output). So

Sure, done.

example/offboard_velocity/offboard_velocity.cpp
julianoes: Removed.

example/offboard_velocity/offboard_velocity.cpp
julianoes: Removed.

example/offboard_velocity/offboard_velocity.cpp
julianoes: Thanks, changed it along these lines.

example/offboard_velocity/offboard_velocity.cpp
julianoes: It just goes the quickest way to the new setpoint. I'll get rid of the "(anti-)clock-wise" indications.

example/offboard_velocity/offboard_velocity.cpp
julianoes: Ok, I'll add that.

example/offboard_velocity/offboard_velocity.cpp
julianoes: Ok.

example/offboard_velocity/offboard_velocity.cpp
julianoes: I've removed all but the one after the loop.

example/offboard_velocity/offboard_velocity.cpp
julianoes: Ok.

example/offboard_velocity/offboard_velocity.cpp
julianoes: Removed.

example/offboard_velocity/offboard_velocity.cpp
julianoes: Sure.

julianoes:
Note to self:

- add proper checks to integration test
- make example nicer, don't fly too far away.

julianoes:
@hamishwillee thanks a lot for the thorough review.

hamishwillee:
All my concerns addressed thanks :-). @shakthi-prashanth-m Are you planning on reviewing this too? Would be great to get this in!

hamishwillee:
@julianoes One more doc fix please. For stop and stop_async in API it would be good to note that when offboard mode stops we put it the vehicle into Hold mode.

So something like:
```
    /**
     * @brief Stop offboard control (synchronous).
     *
     * The vehicle will be put into Hold mode: https://docs.px4.io/en/flight_modes/hold.html
     *
     * @return Result of request.
     */
    Offboard::Result stop() const;

```

hamishwillee:
If I move into Hold mode without calling stop(), will DroneCore realise that it has to stop sending setpoints?

plugins/offboard/offboard_impl.cpp
```diff
@@ -151,6 +180,60 @@ void OffboardImpl::report_offboard_result(const Offboard::result_callback_t &cal
...
+        SEND_INTERVAL_S,
+        &_call_every_cookie);
+
+        _mode = Mode::VELOCITY_NED;
+    } else {
+        // We're already sending these kind of setpoints. Since the setpoint change, let's
+        // reschedule the next call, so we don't send setpoints too often.
+        _parent->reset_call_every(_call_every_cookie);
+    }
+    _velocity_ned_yaw = velocity_ned_yaw;
```
shakthi-prashanth-m: shouldn't we assign this before entering calling `send_velocity_ned()` at line:191.

plugins/offboard/offboard_impl.cpp
```diff
@@ -151,6 +180,60 @@ void OffboardImpl::report_offboard_result(const Offboard::result_callback_t &cal
...
+        SEND_INTERVAL_S,
+        &_call_every_cookie);
+
+        _mode = Mode::VELOCITY_BODY;
+    } else {
+        // We're already sending these kind of setpoints. Since the setpoint change, let's
+        // reschedule the next call, so we don't send setpoints too often.
+        _parent->reset_call_every(_call_every_cookie);
+    }
+    _velocity_body_yawspeed = velocity_body_yawspeed;
```
shakthi-prashanth-m: shouldn't this happen before Line 218 ?

plugins/offboard/offboard_impl.cpp
```diff
@@ -164,17 +247,19 @@ void OffboardImpl::set_velocity_ned(Offboard::VelocityNEDYaw velocity_ned_yaw)
     //const static uint16_t IGNORE_YAW = (1 << 10);
     const static uint16_t IGNORE_YAW_RATE = (1 << 11);
 
-    const float yaw = to_rad_from_deg(velocity_ned_yaw.yaw_deg);
+    _mutex.lock();
```
shakthi-prashanth-m: Does this lock makes sure `_velocity_body_ned_yaw` has updated value?

example/offboard_velocity/offboard_velocity.cpp
shakthi-prashanth-m: makes sense

plugins/offboard/offboard_impl.cpp
julianoes: Good spotting! Call every will only schedule it and not run it immediately, however, it is still a race, so you're right!
Changed in 2c0a887.

plugins/offboard/offboard_impl.cpp
julianoes: Yep, same thing.

plugins/offboard/offboard_impl.cpp
julianoes: Right, I guess the lock prevents the race from happening. Still changed it makes more sense.

julianoes:
> @julianoes One more doc fix please. For stop and stop_async in API it would be good to note that when offboard mode stops we put it the vehicle into Hold mode.

Added, thx.

> If I move into Hold mode without calling stop(), will DroneCore realise that it has to stop sending setpoints?

Note yet, but it makes sense to add a check for it and stop sending setpoints... done in a9c83ec1a8c5c8514a5fed362587fad9522de912.

julianoes:
We could merge this for now. I'll figure out the CI issues in a separate pull request. The problem is that timing sensitive tests fail on slow or busted CI instances.

julianoes:
CI troubles addressed in #147.

PR #139, opened 2017-10-26
core: add a lock around mavlink communication

julianoes:
This adds a lock to receive and send mavlink messages. This can be used
by a plugin to lockdown DroneCore.

The lock is disabled by default, so this change should not have any
effects apart from a few more CPU cycles.

### DIFF
--- a/core/device_impl.cpp
+++ b/core/device_impl.cpp
@@ -98,6 +98,10 @@ void DeviceImpl::unregister_timeout_handler(const void *cookie)
 
 void DeviceImpl::process_mavlink_message(const mavlink_message_t &message)
 {
+    if (_communication_locked) {
+        return;
+    }
+
     std::lock_guard<std::mutex> lock(_mavlink_handler_table_mutex);
 
     for (auto it = _mavlink_handler_table.begin(); it != _mavlink_handler_table.end(); ++it) {
@@ -267,6 +271,10 @@ void DeviceImpl::send_heartbeat(DeviceImpl *self)
 
 bool DeviceImpl::send_message(const mavlink_message_t &message)
 {
+    if (_communication_locked) {
+        return false;
+    }
+
     return _parent->send_message(message);
 }
 
@@ -512,6 +520,16 @@ void DeviceImpl::set_msg_rate_async(uint16_t message_id, double rate_hz,
             MavlinkCommands::Params {float(message_id), interval_us, NAN, NAN, NAN, NAN, NAN},
             callback);
     }
+}
+
+void DeviceImpl::lock_communication()
+{
+    _communication_locked = true;
+}
+
+void DeviceImpl::unlock_communication()
+{
+    _communication_locked = false;
 }
 
 
--- a/core/device_impl.h
+++ b/core/device_impl.h
@@ -95,6 +95,10 @@ public:
 
     Time &get_time() { return _time; };
 
+    // This allows a plugin to lock and unlock all mavlink communication.
+    void lock_communication();
+    void unlock_communication();
+
     // Non-copyable
     DeviceImpl(const DeviceImpl &) = delete;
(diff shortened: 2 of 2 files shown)


hamishwillee:
@julianoes Any examples/guidance on how/when this should used? Ie when should you do this? When is it a risk not to do this/to do this. What do you need to #include?
Basically what would you need to know if you were a brand new developer coming to the project?

julianoes:
> @julianoes Any examples/guidance on how/when this should used? Ie when should you do this? When is it a risk not to do this/to do this. What do you need to #include?

You should use that if you want to restrict access to the SDK in any way. You don't need to `#include` anything, you just call `_parent->lock_communication()` in a plugin to use it.

> Basically what would you need to know if you were a brand new developer coming to the project?

I don't think you need to worry about it as a new developer, you can just ignore it as an unused functionality. Does that make sense, or should I document that?

hamishwillee:
>> @julianoes Any examples/guidance on how/when this should used? Ie when should you do this? When is it a risk not to do this/to do this. What do you need to #include?

> You should use that if you want to restrict access to the SDK in any way. You don't need to #include anything, you just call _parent->lock_communication() in a plugin to use it.

>> Basically what would you need to know if you were a brand new developer coming to the project?

> I don't think you need to worry about it as a new developer, you can just ignore it as an unused functionality. Does that make sense, or should I document that?

If this is not used by anything then it shouldn't exist :-). I suspect that is not the case. I would document this in the "as yet to be written" [Plugin guide](https://github.com/dronecore/docs/issues/49). However I don't understand what you mean by "restrict access to the SDK". We will need to provide a couple of real word examples.

julianoes:
@hamishwillee: it is not used in this repository's tree but it can be used by plugins added externally.

We need this functionality to lockdown any mavlink communication until the SDK is enabled/verified.

hamishwillee:
> @hamishwillee: it is not used in this repository's tree but it can be used by plugins added externally.
> We need this functionality to lockdown any mavlink communication until the SDK is enabled/verified.

Nope, still means nothing to me. What is this  SDK you are talking about? How can you lock down mavlink comms using this - they can just choose not to use it.???

julianoes:
> Nope, still means nothing to me. What is this SDK you are talking about? How can you lock down mavlink comms using this - they can just choose not to use it.???

The SDK I am talking about is the Yuneec SDK which is built on top of DroneCore. If (as by default) this API is not used, everything works as expected and mavlink comms are enabled.

If however, the Yuneec SDK decides it does not want that and instead wants to lockdown the mavlink communication (for whatever reason) it can do it using this API.

hamishwillee:
@julianoes Thanks for the clarification. So why/when do Yuneec use this? FMI only, would help me understand one real use case.

julianoes:
@hamishwillee The Yuneec SDK needs to be activated by the developer using an app key.

hamishwillee:
Thanks for clarification. 

> I don't think you need to worry about it as a new developer, you can just ignore it as an unused functionality. Does that make sense, or should I document that?

IMO this has been created to address a use case - locking down API access for commercial SDKs. Other vendors are likely to want to do the same thing (right?). So we should document it. 

I might change the docs to say.
> // These methods lock and unlock all MAVLink communication.
> // The methods are not used internally. They are provided to allow vendors to provide controlled access to commercial SDKs.

Ideally we would have a specific doc in the DroneCore guide called SDK Developer Guide that covers this sort of thing.

hamishwillee:
@Sushma7785 @mrpollo  Do you guys plan on reviewing this? It requires approval.

Sushma7785:
already reviewed and approved.

julianoes:
Rebased...

mrpollo:
@julianoes please rebase and merge!

PR #142, opened 2017-10-27
Adding FollowMe plugin

shakthi-prashanth-m:
**Highlights of the FollowMe PR:**
- `FollowMe::start()` changes Flight mode to _FollowMe_
- For testing, we can set Follow info ` FollowMe::set_follow_info(const FollowMe::FollowInfo &info)` which involves initial position of Ground control station, etc.
- _FollowMe_ behavior can be configured using ` FollowMe::set_config(const FollowMe::Config &config)` as per [PX4 FollowMe configuration](https://docs.px4.io/en/advanced_config/parameter_reference.html#follow-target).
- FollowMe plugin periodically sends GCS location updates to Vehicle
- `FollowMe::stop()` leaves the device in _Loiter_


example/follow_me/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
+cmake_minimum_required(VERSION 2.8.12)
+
+project(follow_target)
```
julianoes: CMakeLists.txt for examples get a lot easier once this is in: https://github.com/dronecore/DroneCore/pull/130/files#diff-b33c729b1001aff667a2a60dcf6be327R6.

example/follow_me/follow_target.cpp
```diff
@@ -0,0 +1,100 @@
+/**
+* @file follow_target.cpp
+* @brief Example that switches Vehicle mode to FollowMe.
+* @author Author: Shakthi Prashanth M <shakthi.prashanth.m@intel.com>
```
julianoes: In general, I recommend doing integration tests when adding a new features. The reasons are:

- We need integration tests for all the functionality anyway, so we might as well do it while implementing it.
- The example should only cover a simple part of the API to quickly get a developer up to speed.
- It is easier to compile and run it because you don't always need to install the library, and then build the example and link it.
- You can use the internal `LogDebug()` statements.

I propose you move this over to an integration test and add an example at the very end when everything is working.

Oh, I see now you also added an integration test. I would suggest you keep developing there and create a small example at the end.

integration_tests/CMakeLists.txt
```diff
@@ -22,10 +22,11 @@ list(APPEND integration_tests
     offboard_velocity
     logging
     info
-	mission
     mission_change_speed
 	mission_survey
-    gimbal
+	mission
+    followme_target
```
julianoes: indenting.

integration_tests/followme_target.cpp
```diff
@@ -0,0 +1,41 @@
...
+#include <cmath>
+#include <thread>
+#include <chrono>
+#include "integration_test_helper.h"
+#include "global_include.h"
+#include "dronecore.h"
+
+using namespace dronecore;
+
+TEST_F(SitlTest, FollowMeTargetTest)
```
julianoes: No `Test` in name needed, just `FollowMe` is fine.

integration_tests/followme_target.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+    Action::Result action_ret = device.action().arm();
+    ASSERT_EQ(Action::Result::SUCCESS, action_ret);
+
+    action_ret = device.action().takeoff();
+    ASSERT_EQ(Action::Result::SUCCESS, action_ret);
+
+    std::this_thread::sleep_for(std::chrono::seconds(5));
+
+    FollowMe::Result followme_result = device.followme().start();
```
julianoes: Why is it `followme` but the files are called `follow_me`? Can you make it consistent?

plugins/follow_me/follow_me.h
```diff
@@ -0,0 +1,102 @@
...
+        BUSY, /**< @brief Vehicle busy. */
+        COMMAND_DENIED, /**< @brief Command denied. */
+        TIMEOUT, /**< @brief %Request timeout. */
+        UNKNOWN /**< @brief Unknown error. */
+    };
+
+    /**
+     * @brief Altitude flag for Follow Target command.
+     */
+    enum class ESTCapabilities {
```
julianoes: This should be internal and not part of the public API.

plugins/follow_me/follow_me.h
```diff
@@ -0,0 +1,102 @@
...
+        ACCEL = 2,
+        ATT_RATES = 3
+    };
+
+    /**
+     * @brief Motion report info which will be emitted peridically to the Vehicle
+     */
+    struct MotionReport {
+        // set to home position used by PX4.
+        int32_t lat_int = static_cast<uint32_t>(47.3977419 *
```
julianoes: 1. A motion report should be as user-friendly as possible and therefore use `double` for lat/lon.
2. Also, it should not have any default values since the user needs to set them.
3. All variables need to be clearly named, no abbreviations, and always with units in the variable name.
4. The docs need to have `@brief` and end with `.`.

plugins/follow_me/follow_me.h
```diff
@@ -0,0 +1,102 @@
...
+        // velocity
+        float vx = cos(0.13) * 5.0;               //	X velocity in NED frame in meter / s
+        float vy = sin(0.13) * 5.0;              //	Y velocity in NED frame in meter / s
+        float vz = 0;               //	Z velocity in NED frame in meter / s
+        // acceleration
+        float afx;              //	X acceleration in NED frame in meter / s^2 or N
+        float afy;              //	Y acceleration in NED frame in meter / s^2 or N
+        float afz;              //	Z acceleration in NED frame in meter / s^2 or N
+
+        float pos_std_dev[3] = { 0.8f, 0.8f, 0.f };   // -1 for unknown
```
julianoes: You can use `NAN` for unknown.

plugins/follow_me/follow_me_impl.cpp
```diff
@@ -0,0 +1,213 @@
...
+        MAVLINK_MSG_ID_HEARTBEAT,
+        std::bind(&FollowMeImpl::process_heartbeat, this, _1), static_cast<void *>(this));
+}
+
+void FollowMeImpl::deinit()
+{
+    _parent->unregister_all_mavlink_message_handlers(this);
+}
+
+void FollowMeImpl::timeout_occurred()
```
julianoes: Here, please use `call_every` as soon as it's merged.

plugins/follow_me/follow_me_impl.cpp
```diff
@@ -0,0 +1,213 @@
...
+{
+    // FIXME: This is HARD-coded as of now.
+    // Ideally GPS info should come from a platform-specific framwork.
```
julianoes: Obviously, this will go away. You just need the API to set a `MotionReport` and you can then do that in the integration test and example.

plugins/follow_me/follow_me_impl.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+    // update current location coordinates
+    _motion_report.lat_int += 10;
```
julianoes: It would be nice to have an example that does a box or a circle.

plugins/follow_me/follow_me_impl.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+/**
+ * @brief FollowMeImpl::send_gcs_motion_report
+ */
+void FollowMeImpl::send_gcs_motion_report()
+{
+    using namespace std::chrono;
+
+    auto end = system_clock::now();
+    auto elapsed_seconds = system_clock::to_time_t(end) - system_clock::to_time_t(_start);
```
julianoes: For everything with time please re-use the time functionality from `global_include.h`. This way we can also mock time for testing, and keep it consistent.

plugins/follow_me/follow_me_impl.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+    float vel[3] = { _motion_report.vx, _motion_report.vy, 0.f };
```
julianoes: I would probably make these `const float vel[3] ...` since you're not going to change them.

plugins/follow_me/follow_me_impl.cpp
```diff
@@ -0,0 +1,213 @@
...
+                                   _motion_report.alt,
+                                   vel,
+                                   accel_unknown,
+                                   attitude_q_unknown,
+                                   rates_unknown,
+                                   _motion_report.pos_std_dev,
+                                   custom_state);
+
+    std::cout << __func__ <<
+              ((_parent->send_message(msg)) ? ("--------------------> sent follow_target")
```
julianoes: Please use `LogDebug()` etc.

plugins/follow_me/follow_me_impl_test.cpp
```diff
@@ -0,0 +1,13 @@
+#include "follow_me_impl.h"
+#include "dronecore.h"
+#include "global_include.h"
+#include <gtest/gtest.h>
+
+namespace dronecore {
+
+TEST(FollowMeImpl, NoTest)
+{
+    ASSERT_TRUE(true);
```
julianoes: This would be a unit test but if there are none, you can probably leave the file out.

LorenzMeier:
Firmware fix merged.

plugins/follow_me/follow_me.h
```diff
@@ -0,0 +1,102 @@
...
+     */
+    enum class ESTCapabilities {
+        POS = 0,
+        VEL = 1,
+        ACCEL = 2,
+        ATT_RATES = 3
+    };
+
+    /**
+     * @brief Motion report info which will be emitted peridically to the Vehicle
```
hamishwillee: peridically => periodically

plugins/follow_me/follow_me.h
hamishwillee: if this is home position, name it appropriately => home_latitude. No need to put "type" in name. If this is public API then have a doc string output with @brief tag.

plugins/follow_me/follow_me.h
hamishwillee: So what is the purpose of this? You're sending current motion of vehicle? Why does a user of the API need to know this/set this?

plugins/follow_me/follow_me_impl.cpp
shakthi-prashanth-m: Yes. I am thinking to export a simple `Poistion {lat, long, alt}` data structure to API users & construct `MotionReport` myself.

plugins/follow_me/follow_me.h
shakthi-prashanth-m: @hamishwillee You're right. This should be auto-populated. But for testing purpose, we need to export Position (lat, long, alt) to API users. And, this structure will be part of impl and is auto-populated.

hamishwillee: @shakthi-prashanth-m (& @julianoes ) Just to be clear, are you saying that this will move to be internal?, and you'll test it via the internal API?
For me the API should look something like `start()`, `stop()`, `setFollowInfo(FollowMe::FollowInfo info)`, `getFollowInfo()`
Where `FollowInfo` would be info to specify the relative location that the follower should adopt. This might be a co-ordinate or "standard position + height" (front, back, front left, back left, height) or behaviour (e.g. circle me). 

I don't think anything else is needed, though possibly a string for printing output if we use standard positions (and hence have an enum)

plugins/follow_me/follow_me_impl.cpp
hamishwillee: Why does that need to be exported? Please comment below: https://github.com/dronecore/DroneCore/pull/142#discussion_r147604903

plugins/follow_me/follow_me.h
shakthi-prashanth-m: Agree to points 1, 3 & 5.
But for the point 2, I prefer to have them start with default values initialized.
In real scenario, I believe  user don't provide position at all. I see that QGroundControl doesn't prompt for position in _FollowMe_ mode.
Aren't we exporting `set_position` API to for testing purpose ?
As per [PX4 Flight modes](https://docs.px4.io/en/flight_modes/) documentation, In Follow Me mode, the vehicle autonomously follows a user with an Android phone/tablet running QGroundControl.

plugins/follow_me/follow_me.h
julianoes: I agree with @hamishwillee. We don't need to internally check GPS. We just expose it via an API.

shakthi-prashanth-m:
```
[==========] Running 1 test from 1 test case.
[ RUN      ] SitlTest.FollowMe
[07:39:23|Info ] New device on: 127.0.0.1:14557 (udp_connection.cpp:210)
FlightMode: Takeoff
FlightMode: Takeoff
FlightMode: Takeoff
FlightMode: Takeoff
FlightMode: Takeoff
FlightMode: FollowMe
FlightMode: FollowMe
FlightMode: FollowMe
FlightMode: FollowMe
FlightMode: FollowMe
FlightMode: FollowMe
FlightMode: FollowMe
FlightMode: FollowMe
FlightMode: FollowMe
FlightMode: FollowMe
FlightMode: Hold
FlightMode: Hold
[       OK ] SitlTest.FollowMe (23078 ms)
[----------] 1 test from SitlTest (23078 ms total)

[==========] 1 test from 1 test case ran. (23078 ms total)
[  PASSED  ] 1 test.
```

example/follow_me/follow_target.cpp
shakthi-prashanth-m: Agree. Done.

integration_tests/CMakeLists.txt
shakthi-prashanth-m: Done.

integration_tests/followme_target.cpp
shakthi-prashanth-m: Done

integration_tests/followme_target.cpp
shakthi-prashanth-m: I think, it is taking this name from classname. It won't be proper if I name plugin class name as `Follow_Me`. So, I felt `follow_me` is intuitive than followme.

plugins/follow_me/follow_me.h
shakthi-prashanth-m: Done

plugins/follow_me/follow_me_impl.cpp
shakthi-prashanth-m: Only `set_position` API is exported for testing purpose.

plugins/follow_me/follow_me_impl.cpp
shakthi-prashanth-m: I will figure out the coordinates for bounding box and then add.

plugins/follow_me/follow_me_impl.cpp
shakthi-prashanth-m: Done.

plugins/follow_me/follow_me_impl.cpp
shakthi-prashanth-m: Done.

plugins/follow_me/follow_me_impl.cpp
shakthi-prashanth-m: Done

plugins/follow_me/follow_me_impl_test.cpp
shakthi-prashanth-m: Removed, as it is redundant

plugins/follow_me/follow_me.h
shakthi-prashanth-m: Done

plugins/follow_me/follow_me.h
shakthi-prashanth-m: It is not public API now.

plugins/follow_me/follow_me.h
shakthi-prashanth-m: As per https://docs.px4.io/en/flight_modes/follow_me.html QGroundControl doesn't prompt user for position. It captures current position & other GPS related details using platform-specific framework and sends to device. So, in our case we're setting position only for testing purpose.

hamishwillee: @shakthi-prashanth-m @julianoes 

> So, in our case we're setting position only for testing purpose.

No no no no. That is horrible. Nothing to do with testing should be in a public API file. Making it conditional does not make it any less ugly.  If you need to have some sort of test stuff, then have it in an implementation file and test that file. 

So I had another look at the mode docs. Updating my comment above in line with what PX4 does:

The API should be `start()`, `stop()`, `setFollowInfo(FollowMe::FollowInfo info)`, `getFollowInfo()`
Where `FollowInfo` would be and object that captures all the config params [here](https://docs.px4.io/en/flight_modes/follow_me.html#configuration).  You would then set/get the params as needed to drive the mode. You would have an enum and a string converter for `NAV_FT_FS`

shakthi-prashanth-m: I agree about FollowInfo parameters. But somehow need to mock GPS position of the GCS and send periodic updates to device. It could be done internally, without a user API. As per FollowMe mode description [here](https://docs.px4.io/en/flight_modes/follow_me.html#configuration), GPS is needed. It's mentioned in the top,

**By using GPS and other positioning information a multicopter is able to automatically yaw to face and follow a user at a specified position and distance. While in this mode no user input is required.
This mode requires GPS.**

As discussed with @hamishwillee in slack, we can do as below:
1. Cleanup public header file `follow_me.h` and handle mocking position stuff in `follow_me_impl.h`.
2. Export public APIs for set/get FollowInfo
3. To use `CallEveryHandler` instead of `TimeoutHandler`, as per @julianoes suggestion.

hamishwillee: > No, why would you mock this position. Just feed it to DroneCore via the public API.

@julianoes Perhaps I'm dim, but what public API does DroneCore have for setting its **own**position? If such an API exists great - I wasn't aware we had platform-independent methods of getting position of dronecore in place.

shakthi-prashanth-m:
Kindly don't review now. WIP. Thanks.

hamishwillee:
You'd be surprised how easy I find it not to review things :-)

plugins/follow_me/follow_me.h
```diff
@@ -0,0 +1,206 @@
...
+
+    /**
+     * @brief FollowMe configuration.
+     * Reference: https://docs.px4.io/en/advanced_config/parameter_reference.html#follow-target
+     */
+    class Configuration
+    {
+    public:
+        /**
+         * @brief Side to follow target from
```
hamishwillee: Looks good, but all sentences should terminate in a full stop, everywhere.

plugins/follow_me/follow_me.h
```diff
@@ -0,0 +1,206 @@
...
+            FRONT_LEFT = 3 /**< @brief Follow from front left */
+        };
+
+        // Default configuration
+        ///////////////////////////////////////////////////////////
+        static constexpr const float DEF_HEIGHT_ABOVE_HOME =
+            8.0f; /**< @brief Default follow target altitude */
+        static constexpr const float MIN_HEIGHT_ABOVE_HOME = 8.0f; /**< @brief Min follow target altitude */
+
+        static constexpr const float DEF_DIST_WRT_FT = 8.0f; /**< @brief Distance to follow target from */
```
hamishwillee: Default follow-distance in metres.

plugins/follow_me/follow_me.h
```diff
@@ -0,0 +1,206 @@
...
+        static constexpr const float MIN_DIST_WRT_FT =
+            1.0f; /**< @brief Min distance to follow target from */
+
+        static constexpr const FollowDirection DEF_FOLLOW_DIR =
+            FollowDirection::BEHIND; /**< @brief Side to follow target from */
+        static constexpr const float DEF_DYN_FLT_ALG_RSP =
+            0.5f; /**< @brief Default Dynamic filtering algorithm responsiveness to target movement */
+        static constexpr const float MIN_DYN_FLT_ALG_RSP =
+            0.0f; /**< @brief Min Dynamic filtering algorithm responsiveness to target movement */
```
hamishwillee: "Dynamic" does not need to be capitalised. Add full stop.

plugins/follow_me/follow_me.h
```diff
@@ -0,0 +1,206 @@
...
+        static constexpr const float MAX_DYN_FLT_ALG_RSP =
+            1.0f; /**< @brief Max Dynamic filtering algorithm responsiveness to target movement */
+        ///////////////////////////////////////////////////////////
+
+        // FollowMe Configuration Accessors
+        /**
+         * @brief min_height_m
```
hamishwillee: All these accessors need proper docs. For example, this one is "Minimum follow height, in metres. "

plugins/follow_me/follow_me.h
```diff
@@ -0,0 +1,206 @@
...
+         * @return
```
hamishwillee: @ return Minimum follow height.

And yes, this is repetitive :-0

plugins/follow_me/follow_me.h
shakthi-prashanth-m: @hamishwillee  Thanks. I've a small doubt about best practice of configuration class. Generally Configuration or Options classes are not nested in another class. For example `class FollowMeConfiguration`, can be external to `class FollowMe`.  I am not sure what is the right way here.

hamishwillee: Hi @shakthi-prashanth-m 

Having classes in classes makes perfect sense where the definition is intimately associated with the parent and doesn't need to be shared around outside of the parent. That is IMO the case here. 

But @julianoes owns the broad API definition, so lets see what he says.

plugins/follow_me/follow_me.h
```diff
@@ -0,0 +1,206 @@
...
+         */
+        float min_height_m() const
```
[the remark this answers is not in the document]
shakthi-prashanth-m: True, I shall keep it consistent with other plugins.

plugins/follow_me/follow_me_impl_test.cpp
hamishwillee: Still looks to be here to me :-)

hamishwillee:
@shakthi-prashanth-m Definitely getting there - thanks for tidying all the docs! CMakeLists.txt will still need update when this goes into develop.

The example is a bit uninteresting, because of course it can't simulate actual movement of the computer. 

@shakthi-prashanth-m @julianoes Possibly dumb idea, but what do you think of the idea of moving MotionReport into the public API so that you can specify the position of the GCS from the PUBLIC api? 
Obviously this would allow us to make a nicer example, but that is not sufficient justification. I was wondering if there is a better justification for doing so - perhaps "giving users the ability to provide flexible off-centre follow me behaviour?"

shakthi-prashanth-m:
@hamishwillee example was removed. It is here by accident. I shall remove it. Same is for `follow_me_impl_test.cpp`

shakthi-prashanth-m:
@hamishwillee Motion Report details (`lat, lon, alt, eph, epv,` etc) are user friendly info. I couldn't get to see FollowMe featured apps providing such option, including QGroundControl. I feel it's not intuitive for user to provide it. Any thoughts ?

shakthi-prashanth-m:
@hamishwillee Motion Report details (`lat, lon, alt, eph, epv,` etc) are user friendly info. I couldn't get to see FollowMe featured apps providing such option, including QGroundControl. I feel it's not intuitive for user to provide it. Any thoughts ?

core/global_include.cpp
```diff
@@ -32,6 +32,11 @@ double Time::elapsed_since_s(const dl_time_t &since)
            static_cast<double>(steady_clock::period::den);
 }
 
+double Time::elapsed_since_ms(const dl_time_t &since)
```
julianoes: I don't think we should supply this function. You only us it once in the code you added, and it is weird to have a `double` for milliseconds.

core/mavlink_parameters.cpp
```diff
@@ -394,7 +394,7 @@ void MavlinkParameters::receive_timeout()
                 ParamValue empty_value;
                 // Notify about timeout
                 LogErr() << "Error: get param busy timeout: " << work.param_name;
-                // LogDebug() << "Got it after: " << elapsed_since_s(_last_request_time);
+                // LogErr() << "Got it after: " << _parent->get_time().elapsed_since_s(_last_request_time);
```
julianoes: Thanks, these are good changes.

core/mavlink_parameters.h
```diff
@@ -192,7 +192,7 @@ class MavlinkParameters
 
     void *_timeout_cookie = nullptr;
 
-    // dl_time_t _last_request_time = {};
+    dl_time_t _last_request_time = {};
```
julianoes: Is this change needed?

integration_tests/follow_me.cpp
```diff
@@ -0,0 +1,148 @@
...
+    sleep_for(seconds(5));
+
+    ////////////////////////////////////////
+    /// Uses Default PX4 FollowMe config////
+    // Min height: 8mts
+    // Distance to device: 8mts
+    // Responsiveness: 0.5
+    // Device follows from: Behind
+    ////////////////////////////////////////
+    auto defult_cfg = device.followme().get_config();
```
julianoes: `defult_cfg` -> `default_config`

integration_tests/follow_me.cpp
```diff
@@ -0,0 +1,148 @@
...
+    print_config(defult_cfg);
+
+    // start following with default configuration
+    FollowMe::Result followme_result = device.followme().start();
```
julianoes: Why is the API `followme` it should be `follow_me`, that's odd.

plugins/follow_me/follow_me.h
```diff
@@ -0,0 +1,194 @@
...
+     * Reference: https://docs.px4.io/en/advanced_config/parameter_reference.html#follow-target
+     */
+    class Configuration
+    {
+    public:
+        /**
+         * @brief Side to follow target from
+         */
+        enum class FollowDirection {
+            NONE = -1,
```
julianoes: I suggest to use enums without values like the Other APIs.

plugins/follow_me/follow_me.h
```diff
@@ -0,0 +1,194 @@
...
+            FRONT_RIGHT = 0, /**< @brief Follow from front right. */
+            BEHIND = 1, /**< @brief Follow from behind. */
+            FRONT = 2, /**< @brief Follow from front. */
+            FRONT_LEFT = 3 /**< @brief Follow from front left. */
+        };
+
+        // Default configuration
+        ///////////////////////////////////////////////////////////
+        /**< @brief Default follow target height. */
+        static constexpr const float DEF_HEIGHT_ABOVE_HOME = 8.0f;
```
julianoes: I don't particularly like these constants in the header file. If there is some way to prevent that would be good.
Also, they all need to have units!

If you just need to have a default config, why don't you add default values to the Config. I mean like this:

```
struct Config {
   float min_altitude_m = 5.0f;
   FollowDirection follow_direction = FollowDirection::BEHIND;
};
```

Then, the default config can be created with:
```
Config config {};
```

plugins/follow_me/follow_me.h
```diff
@@ -0,0 +1,194 @@
...
+         * @brief String form of %FollowDirection
+         * @param dir enum %FollowDirection
+         * @return std::string representation of enum %FollowDirection
+         * @sa follow_dir(), set_follow_dir(FollowDirection)
+         */
+        static std::string to_str(FollowDirection dir);
+
+        /**
+         * @brief Gets current dynamic filtering algorithm responsiveness value.
+         * @return Current responsiveness value
```
julianoes: What is the unit of this?

plugins/follow_me/follow_me.h
```diff
@@ -0,0 +1,194 @@
...
+         * @sa follow_dir(), to_str(FollowDirection)
+         */
+        void set_follow_dir(FollowDirection dir);
+
+        /**
+         * @brief Sets dynamic filtering algorithm responsiveness to target movement.
+         * Lower numbers increase the responsiveness to changing long lat but also ignore less noise.
+         * @param responsiveness
+         * @return true (success) if responsiveness is in range (0.0 to 1.0), false (failure) otherwise.
+         * @sa responsiveness()
```
julianoes: What is `@sa?`

plugins/follow_me/follow_me.h
```diff
@@ -0,0 +1,194 @@
...
+         */
+        bool set_responsiveness(float resp);
```
julianoes: `resp` -> `responsiveness`. Please error on the side of verbosity.

plugins/follow_me/follow_me.h
```diff
@@ -0,0 +1,194 @@
...
+
+    private:
+        float _min_height_m = 8.0f; // Default & Minimum follow target height in meters.
```
julianoes: This is a public header files, it should not have anything private.

plugins/follow_me/follow_me_impl.cpp
```diff
@@ -0,0 +1,294 @@
...
+}
+
+void FollowMeImpl::deinit()
+{
+    _parent->unregister_all_mavlink_message_handlers(this);
+}
+
+void FollowMeImpl::timeout_occurred()
+{
+#ifdef FOLLOW_ME_TESTING
```
julianoes: Testing should be done using the public API. This internal testing doesn't make sense in my opinion. The point of a unit test is to test the API, so it needs to access it from the outside.

plugins/follow_me/follow_me_impl.cpp
```diff
@@ -0,0 +1,294 @@
...
+                                               this, _1, dir));
+    if (cfg.responsiveness() != FollowMe::Configuration::DEF_DYN_FLT_ALG_RSP)
+        _parent->set_param_float_async("NAV_FT_RS", rsp,
+                                       std::bind(&FollowMeImpl::receive_param_responsiveness,
+                                                 this, _1, rsp));
+}
+
+/**
+ * @brief FollowMeImpl::start
+ * @return
```
julianoes: No doxygen comments needed here.

plugins/follow_me/follow_me_impl.cpp
```diff
@@ -0,0 +1,294 @@
...
+ */
+FollowMe::Result FollowMeImpl::start()
+{
+    _parent->add_call_every(std::bind(&FollowMeImpl::timeout_occurred, this), 0.6f, &_timeout_cookie);
```
julianoes: I see, this is a call every. Then I would rename the `timeout_occured` method.

plugins/follow_me/follow_me_impl.h
```diff
@@ -0,0 +1,116 @@
...
+    void init() override;
+    void deinit() override;
+
+    FollowMe::Configuration get_config() const;
+    void set_config(const FollowMe::Configuration &cfg);
+
+    FollowMe::Result start();
+    FollowMe::Result stop();
+private:
+    /* private methods */
```
julianoes: No need for this comment. There is already the keyword `private`.

plugins/follow_me/follow_me_impl.h
```diff
@@ -0,0 +1,116 @@
...
+    /*****************************************************/
+    void process_heartbeat(const mavlink_message_t &message);
+    FollowMe::Result to_follow_me_result(MavlinkCommands::Result result) const;
+    void timeout_occurred();
+    void send_gcs_motion_report();
```
julianoes: What does `gcs` mean? I would leave it away, spell it out, or at least add a comment somewhere. It makes code harder to read and understand.

plugins/follow_me/follow_me_impl.h
```diff
@@ -0,0 +1,116 @@
...
+        // GCS Position
+        int32_t lat_int;
+        int32_t lon_int;
+        float alt;
+        // Velocity in NED frame in meter / s
+        float vx;
+        float vy;
+        float vz;
+        // Acceleration in NED frame in meter/s^2 or NAN
+        float afx;
```
julianoes: Please add units to variable names. And don't use abbreviations. Someone from the outside does not understand this.

plugins/follow_me/follow_me_impl.h
```diff
@@ -0,0 +1,116 @@
...
+        float pos_std_dev[3];
+
+        /**
+         * @brief MotionReport
+         * @param _lat
+         * @param _lon
+         * @param _alt
+         */
+#ifdef FOLLOW_ME_TESTING
+        // Initial GCS position is set to PX4 Home
```
julianoes: As said, testing should be done using the outside API, not from the inside.

plugins/telemetry/telemetry.cpp
```diff
@@ -238,6 +238,7 @@ void Telemetry::flight_mode_async(flight_mode_callback_t callback)
 
 const char *Telemetry::flight_mode_str(FlightMode flight_mode)
 {
+    LogDebug() << "Flight mode val " << flight_mode;
```
julianoes: This debug line should probably eventually be removed, what do you think?

plugins/telemetry/telemetry.cpp
```diff
@@ -253,6 +254,8 @@ const char *Telemetry::flight_mode_str(FlightMode flight_mode)
             return "Land";
         case FlightMode::OFFBOARD:
             return "Offboard";
+        case FlightMode::FOLLOW_ME:
+            return "FollowMe";
```
julianoes: Nice addition!

plugins/follow_me/follow_me_impl.cpp
hamishwillee: That is a bit of a challenge currently - somehow going to have to test the public API while mocking movement of the GCS.

hamishwillee:
>> @shakthi-prashanth-m @julianoes Possibly dumb idea, but what do you think of the idea of moving MotionReport into the public API so that you can specify the position of the GCS from the PUBLIC api?
Obviously this would allow us to make a nicer example, but that is not sufficient justification. I was wondering if there is a better justification for doing so - perhaps "giving users the ability to provide flexible off-centre follow me behaviour?"

> @hamishwillee Motion Report details (lat, lon, alt, eph, epv, etc) are user friendly info. I couldn't get to see FollowMe featured apps providing such option, including QGroundControl. I feel it's not intuitive for user to provide it. Any thoughts ?

Is this regarding my comment up top? 

Possibly not. If you're going to test the public API though you are going to need some way to fake the GPS position of the machine Dronecore is running on.

integration_tests/follow_me.cpp
shakthi-prashanth-m: `followme()` is generated from `classname` which is set in [CMakeLists.txt](https://github.com/dronecore/DroneCore/pull/142/files#diff-1f0169d95c956b412a642e528a116f0fR2). How do I change it ?

core/mavlink_parameters.h
shakthi-prashanth-m: Oh yeah I forgot to remove.

core/global_include.cpp
shakthi-prashanth-m: ok, I'll remove it.

plugins/follow_me/follow_me.h
shakthi-prashanth-m: ok.

plugins/follow_me/follow_me.h
shakthi-prashanth-m: No unit for this: Check for `NAV_FT_RS ` in the table in https://docs.px4.io/en/advanced_config/parameter_reference.html#follow-target

plugins/follow_me/follow_me.h
shakthi-prashanth-m: Its see also tag in doxygen.

plugins/follow_me/follow_me_impl.cpp
shakthi-prashanth-m: ok

plugins/follow_me/follow_me_impl.h
shakthi-prashanth-m: ok

plugins/telemetry/telemetry.cpp
shakthi-prashanth-m: We can remove it. Caller may print this. We should make this `std::string`

plugins/telemetry/telemetry.cpp
shakthi-prashanth-m: Thanks

integration_tests/follow_me.cpp
julianoes: I see 😄 . I guess I need to look into the cmake magic to fix this.

plugins/follow_me/follow_me.h
julianoes: Ok, then describe what the range and meaning is, I guess.

plugins/follow_me/follow_me.h
julianoes: Ah thanks.

hamishwillee: OK. That may break docs build, but I won't know until @julianoes looks at https://github.com/dronecore/DroneCore/issues/159 :-)

plugins/follow_me/follow_me_impl.h
shakthi-prashanth-m: added.

plugins/follow_me/follow_me_impl.h
shakthi-prashanth-m: Its Ground Control Station. Now I simplified to `send_follow_info()`

plugins/follow_me/follow_me_impl.cpp
shakthi-prashanth-m: As of now we can test from PUBLIC API `void FollowMe::set_follow_info(const FollowInfo &info)`

plugins/follow_me/follow_me.h
shakthi-prashanth-m: Thanks. Changed

plugins/follow_me/follow_me.h
shakthi-prashanth-m: I had member with the same name. Windows build complained it. Anyways, now all changed :)

plugins/follow_me/follow_me.h
shakthi-prashanth-m: Constants are required for validating against values set by application. As it is header file, lets define them in `.cpp`

integration_tests/followme_target.cpp
julianoes: Yer, I need to fix it.

plugins/follow_me/follow_me.h
julianoes: Oh, that's funny.

plugins/follow_me/follow_me_impl.h
julianoes: Cool!

plugins/telemetry/telemetry.cpp
julianoes: Hm, good point, we might as well.

integration_tests/follow_me.cpp
```diff
@@ -37,22 +37,22 @@ TEST_F(SitlTest, FollowMe)
...
-    ////////////////////////////////////////
-    /// Uses Default PX4 FollowMe config////
-    // Min height: 8mts
-    // Distance to device: 8mts
-    // Responsiveness: 0.5
-    // Device follows from: Behind
-    ////////////////////////////////////////
-    auto defult_cfg = device.followme().get_config();
-    print_config(defult_cfg);
+    auto defult_config = device.followme().get_config();
```
hamishwillee: default_config ?

shakthi-prashanth-m: if no config is set, it is default PX4 follow me config.

hamishwillee: Sorry, I mean that this is miss-spelled.

plugins/follow_me/follow_me.h
```diff
@@ -0,0 +1,170 @@
...
+#include <iostream>
+#include <cmath>
+#include <functional>
+
+namespace dronecore {
+
+class FollowMeImpl;
+
+/**
+ * @brief FollowMe plugin is used to make your drone to follow your Ground Control Station (Phone/Tab). <br>
```
hamishwillee: "FollowMe plugin is used to make your drone to follow your Ground Control Station (Phone/Tab). <br>" TO
"This class is used to manage the vehicle FollowMe mode."

plugins/follow_me/follow_me.h
```diff
@@ -0,0 +1,170 @@
...
+    /**
+     * @brief Destructor (internal use only).
+     */
+    ~FollowMe();
+
+    /**
+     * @brief Results of FollowMe operations
+     */
+    enum class Result {
+        SUCCESS = 0, /**< @brief %Request succeeded. */
```
hamishwillee: Looks like a type " %Request" (the %)

plugins/follow_me/follow_me.h
```diff
@@ -0,0 +1,170 @@
...
+        NO_DEVICE, /**< @brief No device connected. */
+        CONNECTION_ERROR, /**< @brief %Connection error. */
+        BUSY, /**< @brief Vehicle busy. */
+        COMMAND_DENIED, /**< @brief Command denied. */
+        TIMEOUT, /**< @brief %Request timeout. */
+        NOT_ACTIVE, /**< @brief FollowMe is not activated yet */
```
hamishwillee: Remove "yet", add a fullstop.

plugins/follow_me/follow_me.h
```diff
@@ -0,0 +1,170 @@
...
+        UNKNOWN /**< @brief Unknown error. */
+    };
+
+    /**
+     * @brief FollowMe Configuration.
+     * Reference: https://docs.px4.io/en/advanced_config/parameter_reference.html#follow-target
+     */
+    struct Config {
+        /**
+         * @brief Side to follow target from.
```
hamishwillee: Perhaps "Relative position of following vehicle"

plugins/follow_me/follow_me.h
```diff
@@ -0,0 +1,170 @@
...
+        enum class FollowDirection {
+            FRONT_RIGHT, /**< @brief Follow from front right. */
+            BEHIND, /**< @brief Follow from behind. */
+            FRONT, /**< @brief Follow from front. */
+            FRONT_LEFT, /**< @brief Follow from front left. */
+            NONE
+        };
+        // Default Configuration
+        static const float DEF_HEIGHT_M; /**< @brief Default height above home: 8.0 mts */
+        static const float DEF_FOLLOW_DIST_M; /**< @brief Default Follow distance: 8.0 mts */
```
hamishwillee: lower case "f" in the  "DefaultFollow distance ..."

plugins/follow_me/follow_me.h
```diff
@@ -0,0 +1,170 @@
...
+        static const FollowDirection DEF_FOLLOW_DIR; /**< @brief Default follow direction: BEHIND */
+        static const float DEF_RESPONSIVENSS; /**< @brief Default responsiveness to target: 0.5 */
+
+        // Min & Max ranges as per PX4 FollowMe Configuration.
+        static const float MIN_HEIGHT_M; /**< @brief Min follow height, in mts. >= 8.0m */
+        static const float MIN_FOLLOW_DIST_M;  /**< @brief Min follow distance from, in mts. >= 1.0m */
+        static const float MIN_RESPONSIVENESS; /**< @brief Min responsiveness: 0.0 */
+        static const float MAX_RESPONSIVENESS; /**< @brief Max responsiveness: 1.0 */
+
+        float min_height_m = DEF_HEIGHT_M; /**< @brief Min follow height above home */
```
hamishwillee: Full stop.

plugins/follow_me/follow_me.h
```diff
@@ -0,0 +1,170 @@
...
+        float follow_dist_m =
+            DEF_FOLLOW_DIST_M; /**< @brief Follow horizontal distance b/w Vehicle & GCS. */
+        FollowDirection follow_dir = DEF_FOLLOW_DIR; /**< @brief Side to follow target from. */
```
hamishwillee: As above : https://github.com/dronecore/DroneCore/pull/142/files#r153111854

plugins/follow_me/follow_me.h
```diff
@@ -0,0 +1,170 @@
...
+        float responsiveness = DEF_RESPONSIVENSS; /**< @brief Responsiveness: Range (0.0 to 1.0)*/
+
+        /**
+         * @brief Human-readable string of enum `FollowDirection`.
```
hamishwillee: "string of enum" to "string for enum"

plugins/follow_me/follow_me.h
```diff
@@ -0,0 +1,170 @@
...
+         * @param dir follow direction
```
hamishwillee: Capital F.
Are you allowed to use shortened names like `dir` for parameters?

plugins/follow_me/follow_me.h
```diff
@@ -0,0 +1,170 @@
...
+         * @return std::string representing given direction
+         */
+        static std::string to_str(FollowDirection dir);
+    };
+
+    /**
+     * @brief Gets current FollowMe configuration.
+     * @return current FollowMe configuration.
```
hamishwillee: Current, not current

plugins/follow_me/follow_me.h
```diff
@@ -0,0 +1,170 @@
...
+     * @sa set_config(const Config &)
+     */
+    const Config &get_config() const;
+
+    /**
+     * @brief Applies FollowMe configuration by sending them to device.
```
hamishwillee: Sending "it" (not them)

plugins/follow_me/follow_me.h
```diff
@@ -0,0 +1,170 @@
...
+        double absolute_altitude_m; /**< @brief AMSL, in meters */
+
+        float velocity_x_mps; /**< @brief X-velocity in m/s2 */
+        float velocity_y_mps; /**< @brief Y-velocity in m/s2 */
+        float velocity_z_mps; /**< @brief Z-velocity in m/s2 */
+    };
+
+    /**
+     * @brief Sets current location of the moving target
+     * @note App can obtain location of the moving target from Location framework of the underlying platform. <br>
```
hamishwillee: Remove all the html br line breaks. If you really need a para break, add space between lines.

plugins/follow_me/follow_me.h
```diff
@@ -0,0 +1,170 @@
...
+     * Below is a hint to Apps to avail services available on their underlying platforms. <br>
```
hamishwillee: Change this to:

 The following links provide information about location services on different platforms:

plugins/follow_me/follow_me.h
```diff
@@ -0,0 +1,170 @@
...
+     * Android - https://developer.android.com/reference/android/location/Location.html <br>
+     * Apple - https://developer.apple.com/documentation/corelocation <br>
+     * Windows - https://docs.microsoft.com/en-us/uwp/api/Windows.Devices.Geolocation <br>
+     * @param pos current location of the target
```
hamishwillee: Capital "Current". Add full stop.

plugins/follow_me/follow_me.h
```diff
@@ -0,0 +1,170 @@
...
+     * @return Result::SUCCESS when location is set,
+     *         Result::NOT_STARTED_YET when you set the location before starting FollowMe.
+     */
+    Result set_curr_target_location(const TargetLocation &location);
+
+    /**
+     * @brief Return English string for FollowMe error codes.
+     *
+     * @param result FollowMe::Result code
```
hamishwillee: Add full stop.

plugins/follow_me/follow_me.h
```diff
@@ -0,0 +1,170 @@
...
+     * @return Returns std::string describing error code
+     */
+    static std::string result_str(Result result);
+
+    /**
+     * @brief Starts FollowMe mode
```
hamishwillee: Add full stop.

plugins/follow_me/follow_me.h
```diff
@@ -0,0 +1,170 @@
...
+     * @return Result::SUCCESS if succeeded, error otherwise. See FollowMe::Result for error codes.
+     * @sa stop()
+     */
+    FollowMe::Result start() const;
+
+    /**
+     * @brief Stops FollowMe mode
```
hamishwillee: Add full stop.

hamishwillee:
Looking pretty close now!

integration_tests/follow_me.cpp
```diff
@@ -0,0 +1,233 @@
...
+    config.min_height_m = 12.f; // increase min height
+    config.follow_dist_m = 20.f; // set distance b/w device and GCS during FollowMe mode
+    config.responsiveness = 0.2f; // set to higher responsiveness
+    config.follow_dir = FollowMe::Config::FollowDirection::FRONT; // Device follows you from FRONT side
+
+    // Apply configuration
+    bool configured = device.followme().set_config(config);
+    ASSERT_EQ(true, configured);
+
+    sleep_for(seconds(2)); // untill config is applied
```
julianoes: `untill` -> `until`

plugins/follow_me/follow_me.cpp
```diff
@@ -0,0 +1,103 @@
...
+FollowMe::Result FollowMe::set_curr_target_location(const TargetLocation &target_location)
+{
+    return _impl->set_curr_target_location(target_location);
+}
+
+std::string FollowMe::Config::to_str(FollowMe::Config::FollowDirection dir)
+{
+    switch (dir) {
+        case FollowMe::Config::FollowDirection::FRONT_RIGHT:
+            return "FRONT RIGHT";
```
julianoes: I would make the string `Front right`

plugins/follow_me/follow_me.cpp
```diff
@@ -0,0 +1,103 @@
...
+            return "Timeout";
+        case Result::NOT_ACTIVE:
+            return "FollowMe is not active";
+        case Result::UNKNOWN:
+        default:
+            return "Unknown";
+    }
+}
+
+const float FollowMe::Config::DEF_HEIGHT_M =
```
julianoes: Please spell `DEF` to `DEFAULT`.

plugins/follow_me/follow_me.h
```diff
@@ -0,0 +1,170 @@
...
+        static const float DEF_HEIGHT_M; /**< @brief Default height above home: 8.0 mts */
+        static const float DEF_FOLLOW_DIST_M; /**< @brief Default Follow distance: 8.0 mts */
+        static const FollowDirection DEF_FOLLOW_DIR; /**< @brief Default follow direction: BEHIND */
+        static const float DEF_RESPONSIVENSS; /**< @brief Default responsiveness to target: 0.5 */
+
+        // Min & Max ranges as per PX4 FollowMe Configuration.
+        static const float MIN_HEIGHT_M; /**< @brief Min follow height, in mts. >= 8.0m */
+        static const float MIN_FOLLOW_DIST_M;  /**< @brief Min follow distance from, in mts. >= 1.0m */
+        static const float MIN_RESPONSIVENESS; /**< @brief Min responsiveness: 0.0 */
+        static const float MAX_RESPONSIVENESS; /**< @brief Max responsiveness: 1.0 */
```
julianoes: You should be able to define the numbers right here using `static constexpr MAX_RESPONSIVENESS = 1.0f`.

plugins/follow_me/follow_me.h
```diff
@@ -0,0 +1,170 @@
...
+         */
+        enum class FollowDirection {
+            FRONT_RIGHT, /**< @brief Follow from front right. */
+            BEHIND, /**< @brief Follow from behind. */
+            FRONT, /**< @brief Follow from front. */
+            FRONT_LEFT, /**< @brief Follow from front left. */
+            NONE
+        };
+        // Default Configuration
+        static const float DEF_HEIGHT_M; /**< @brief Default height above home: 8.0 mts */
```
julianoes: For the default, can't you just make it the default initialization of the config struct?

plugins/follow_me/follow_me.h
julianoes: I was like "yes dir for directory is fair" and then realized it was direction not directory. I vote to spell it out.

plugins/follow_me/follow_me.h
```diff
@@ -0,0 +1,170 @@
...
+        static const float DEF_FOLLOW_DIST_M; /**< @brief Default Follow distance: 8.0 mts */
+        static const FollowDirection DEF_FOLLOW_DIR; /**< @brief Default follow direction: BEHIND */
```
julianoes: Just set the default right away:
```
enum class FollowDirection {
    FRONT_RIGHT, /**< @brief Follow from front right. */
...
};
...
FollowDirection follow_direction = FollowDirection::BEHIND;
```

plugins/follow_me/follow_me_impl.cpp
```diff
@@ -0,0 +1,362 @@
...
+                                   custom_state);
+
+    if (!_parent->send_message(msg)) {
+        LogErr() << "send_curr_target_location() failed..";
+    }
+}
+
+void FollowMeImpl::stop_sending_target_location()
+{
+    // We assume that mutex was acuired by the caller
```
julianoes: `acuired` -> `acquired`

plugins/follow_me/follow_me_impl.cpp
```diff
@@ -0,0 +1,362 @@
...
+        std::bind(&FollowMeImpl::process_heartbeat, this, _1), static_cast<void *>(this));
+    set_default_config();
+}
+
+void FollowMeImpl::deinit()
+{
+    _parent->unregister_all_mavlink_message_handlers(this);
+}
+
+bool operator!(FollowMe::TargetLocation &location)
```
julianoes: What is this used for? It seems strange to me.

plugins/follow_me/follow_me_impl.cpp
```diff
@@ -0,0 +1,362 @@
...
+{
+    // It is not possible for any of portion of location to be zero.
+    return location.lattitude_deg == 0.0;
```
julianoes: Checking for `==` with a double is usually not encouraged. I'd rather initialize with NAN and check for `std::isfinite`.

plugins/follow_me/follow_me.h
```diff
@@ -0,0 +1,170 @@
...
+     * @brief Checks whether FollowMe is active.
+     * @return true if FollowMe is active, false otherwise.
+     */
+    bool is_active() const;
+
+    /**
+     * @brief Represents geographic location of the target in motion
+     */
+    struct TargetLocation {
+        double lattitude_deg; /**< @brief Latitude */
```
julianoes: `lattitude_deg` -> `latitude_deg`

plugins/follow_me/follow_me_impl.h
```diff
@@ -0,0 +1,69 @@
...
+        NOT_ACTIVE,
+        ACTIVE
+    } _mode = Mode::NOT_ACTIVE;
+    void set_mode(Mode mode);
+
+    mutable std::mutex _mutex {};
+    FollowMe::TargetLocation _curr_target_location {}; // sent to vehicle
+    void *_curr_target_location_cookie = nullptr;
+
+    Time _time;
```
julianoes: You can initialize most variables here like this:
```
Time _time {};
```
In my opinion it's encouraged over having a big constructor.

plugins/follow_me/follow_me.h
shakthi-prashanth-m: Definitely I can. It will be messy in the public header file if line length becomes too long. This enables auto-warpping of the source code which decreases readability.

julianoes: What about having the comments before the field?

hamishwillee: > What about having the comments before the field?

What has been done here with comments is consistent with every other plugin so perhaps leave as is?

plugins/follow_me/follow_me_impl.cpp
shakthi-prashanth-m: :) I wanted to compare whether location was set or not by overloading `operator!`, I removed it now as it is not right way to compare real values. I am using `std::isfinite` as you suggested. Thank

plugins/follow_me/follow_me.h
shakthi-prashanth-m: Thanks @hamishwillee

plugins/follow_me/follow_me.h
shakthi-prashanth-m:
```
        /**< @brief Default height above home, in meters. */
        constexpr static const float DEFAULT_HEIGHT_M = 8.0;
        /**< @brief Default follow distance, in meters. */
        constexpr static const float DEFAULT_FOLLOW_DIST_M = 8.0;
        /**< @brief Default follow direction. */
        constexpr static const FollowDirection DEFAULT_FOLLOW_DIR = FollowDirection::BEHIND;
        /**< @brief Default responsiveness to target. */
        constexpr static const float DEFAULT_RESPONSIVENSS = 0.5f;

        /**< @brief Min follow height, in meters. */
        constexpr static const float MIN_HEIGHT_M = 8.0f;
        /**< @brief Min follow distance from, in meters. */
        constexpr static const float MIN_FOLLOW_DIST_M = 1.0f;
        /**< @brief Min responsiveness. */
        constexpr static const float MIN_RESPONSIVENESS = 0.f;
        /**< @brief Max responsiveness. */
        constexpr static const float MAX_RESPONSIVENESS = 1.0f;

        /**< @brief Min follow height above home, in meters. */
        float min_height_m = DEFAULT_HEIGHT_M;
        /**< @brief Horizontal distance b/w vehicle & target. */
        float follow_dist_m = DEFAULT_FOLLOW_DIST_M;
        /**< @brief Relative position of the following vehicle. */
        FollowDirection follow_dir = DEFAULT_FOLLOW_DIR;
        /**< @brief Responsiveness: Range (0.0 to 1.0)*/
        float responsiveness = DEFAULT_RESPONSIVENSS;
```

shakthi-prashanth-m: It would be like this.

shakthi-prashanth-m:
```
        constexpr static const float DEFAULT_HEIGHT_M =
            8.0; /**< @brief Default height above home, in meters. */
        constexpr static const float DEFAULT_FOLLOW_DIST_M =
            8.0; /**< @brief Default follow distance, in meters. */
        constexpr static const FollowDirection DEFAULT_FOLLOW_DIR =
            FollowDirection::BEHIND; /**< @brief Default follow direction. */
        constexpr static const float DEFAULT_RESPONSIVENSS =
            0.5f; /**< @brief Default responsiveness to target. */

        constexpr static const float MIN_HEIGHT_M = 8.0f; /**< @brief Min follow height, in meters. */
        constexpr static const float MIN_FOLLOW_DIST_M =
            1.0f; /**< @brief Min follow distance from, in meters. */
        constexpr static const float MIN_RESPONSIVENESS = 0.f; /**< @brief Min responsiveness. */
        constexpr static const float MAX_RESPONSIVENESS = 1.0f; /**< @brief Max responsiveness. */

        float min_height_m = DEFAULT_HEIGHT_M; /**< @brief Min follow height above home, in meters. */
        float follow_dist_m =
            DEFAULT_FOLLOW_DIST_M; /**< @brief Horizontal distance b/w vehicle & target. */
        FollowDirection follow_dir =
            DEFAULT_FOLLOW_DIR; /**< @brief Relative position of the following vehicle. */
        float responsiveness = DEFAULT_RESPONSIVENSS; /**< @brief Responsiveness: Range (0.0 to 1.0)*/
```
This is after in-line initialization.

julianoes:
@shakthi-prashanth-m it's fine if coverage drops, we can still merge it.

Let me know when I should review again.

shakthi-prashanth-m:
@julianoes  you can review now. Thanks.

integration_tests/follow_me.cpp
```diff
@@ -0,0 +1,263 @@
...
+
+    Action::Result action_ret = device.action().arm();
+    ASSERT_EQ(Action::Result::SUCCESS, action_ret);
+
+    device.telemetry().flight_mode_async(
+    std::bind([&](Telemetry::FlightMode flight_mode) {
+        FollowMe::TargetLocation last_location;
+        device.follow_me().get_last_location(last_location);
+
+        std::cout << "FlightMode: " << Telemetry::flight_mode_str(flight_mode)
```
julianoes: I think for integration tests we could re-use `LogInfo()`, etc. but we can do that in a later pull request because it's currently not consistent.

integration_tests/follow_me.cpp
```diff
@@ -0,0 +1,263 @@
...
+    action_ret = device.action().land();
+    ASSERT_EQ(Action::Result::SUCCESS, action_ret);
+    sleep_for(seconds(2)); // let it land
+}
+
+void print(const FollowMe::Config &config)
+{
+    std::cout << "Current FollowMe configuration of the device" << std::endl;
+    std::cout << "---------------------------" << std::endl;
+    std::cout << "Min Height: " << config.min_height_m << "m" << std::endl;
```
julianoes: Units! Yay :)

plugins/follow_me/follow_me.cpp
```diff
@@ -0,0 +1,90 @@
...
+        case FollowMe::Config::FollowDirection::BEHIND:
+            return "Behind";
+        case FollowMe::Config::FollowDirection::FRONT_LEFT:
+            return "Front left";
+        case FollowMe::Config::FollowDirection::FRONT:
+            return "Front";
+        case FollowMe::Config::FollowDirection::NONE:
+            return "None";
+    }
+    return nullptr; // to please compiler not to warn
```
julianoes: I think alternatively you could add the `default:`.

plugins/follow_me/follow_me.cpp
```diff
@@ -0,0 +1,90 @@
...
+            return "Connection error";
+        case Result::BUSY:
+            return "Busy";
+        case Result::COMMAND_DENIED:
+            return "Command denied";
+        case Result::TIMEOUT:
+            return "Timeout";
+        case Result::NOT_ACTIVE:
+            return "FollowMe is not active";
+        case Result::UNKNOWN:
```
julianoes: When you omit the `break` or `return` you need to add a comment `// FALLTHROUGH`, otherwise newer compilers will warn.

plugins/follow_me/follow_me.h
```diff
@@ -0,0 +1,178 @@
...
+        constexpr static const float DEFAULT_RESPONSIVENSS =
+            0.5f; /**< @brief Default responsiveness to target. */
+
+        constexpr static const float MIN_HEIGHT_M = 8.0f; /**< @brief Min follow height, in meters. */
+        constexpr static const float MIN_FOLLOW_DIST_M =
+            1.0f; /**< @brief Min follow distance from, in meters. */
+        constexpr static const float MIN_RESPONSIVENESS = 0.f; /**< @brief Min responsiveness. */
+        constexpr static const float MAX_RESPONSIVENESS = 1.0f; /**< @brief Max responsiveness. */
+
+        float min_height_m = DEFAULT_HEIGHT_M; /**< @brief Min follow height above home, in meters. */
```
julianoes: What about just setting the number here instead of having the #define? Do you think it is needed?

plugins/follow_me/follow_me_impl.cpp
```diff
@@ -0,0 +1,390 @@
...
+        default:
+            return FollowMe::Result::UNKNOWN;
+    }
+}
+
+bool FollowMeImpl::is_current_location_set() const
+{
+    // If the target's latitude is NAN, we assume that location is not set.
+    // We assume that mutex was acquired by the caller
+    return std::isfinite(_curr_target_location.latitude_deg);
```
julianoes: Yep, I think that's fine.

plugins/follow_me/follow_me_impl.cpp
```diff
@@ -0,0 +1,390 @@
...
+        px4_custom_mode.data = heartbeat.custom_mode;
+
+        if (px4_custom_mode.main_mode == px4::PX4_CUSTOM_MAIN_MODE_AUTO &&
+            px4_custom_mode.sub_mode == px4::PX4_CUSTOM_SUB_MODE_AUTO_FOLLOW_TARGET) {
+            follow_me_active = true; // we're in FollowMe mode
+        }
+    }
+
+    {
+        _mutex.lock();
```
julianoes: For the case where you exit early with a return you can also use the `std::lock_guard` which unlocks when it goes out of scope.

plugins/follow_me/follow_me.h
shakthi-prashanth-m: Yeah, it is not used anywhere. Initially I had used these constants for future use, if we need default values, etc. It seems we don't require them. I shall remove them and use literals.

shakthi-prashanth-m:
@julianoes  Thanks for the review. Next commit should resolve your comments + using new doxygen tags.

julianoes:
Clang is complaining :(
```
/Users/travis/build/dronecore/DroneCore/integration_tests/follow_me.cpp:59:75: error: implicit conversion increases floating-point precision: 'float' to 'double' [-Werror,-Wdouble-promotion]

    device.follow_me().set_curr_target_location({47.39768399, 8.54564155, 0.f, 0.f, 0.f, 0.f});
```

shakthi-prashanth-m:
@julianoes, Yeah corrected it. It done now.

shakthi-prashanth-m:
Thanks @julianoes and @hamishwillee for conducting detailed review of `this` pull request. :100:

julianoes:
@shakthi-prashanth-m thanks for your work, I'm just rebasing this and fixing the conflicts.

hamishwillee:
@shakthi-prashanth-m Awesome this is in. Now that it is all well understood, do you plan on creating an example?

shakthi-prashanth-m:
@hamishwillee FollowMe example should use either of Android/Apple/Windows platform. As of now, we've simulated in [integration_tests/follow_me.cpp::FollowMeMultiLocationWithConfig](https://github.com/dronecore/DroneCore/blob/develop/integration_tests/follow_me.cpp#L79) where it feeds GPS Location to FollowMe plugin. To build an example in one of the above platforms, we need Language bindings in-place.

PR #143, opened 2017-10-27
grpc : Dynamic plugin loading

Rjasuja:
Split proto files(one for each plugin), load plugins based on plugins.conf file


Rjasuja:
For now conf file is kept simple with plugin name in each new line, other option is to have xml file and have xml parser included.

julianoes:
> For now conf file is kept simple with plugin name in each new line, other option is to have xml file and have xml parser included.

We will have a xml parser included once the camera definition stuff is added that you could use.

grpc/python_client/CMakeLists.txt
```diff
@@ -16,4 +16,33 @@ if(${PYTHONINTERP_FOUND})
...
+    )
+
+    add_custom_command(TARGET pydronecore
+        COMMAND ${PYTHON_EXECUTABLE}
+            -m grpc_tools.protoc
+            -I ${proto_dir}
+            --python_out=.
+            --grpc_python_out=.
+            ${proto_dir}/telemetry.proto
+        DEPENDS ${proto_dir}/telemetry.proto
```
julianoes: With cmake you can list all files in a directory and loop over them.

grpc/server/CMakeLists.txt
```diff
@@ -27,10 +27,64 @@ add_custom_command(OUTPUT dronecore.pb.cc
...
 )
 
+add_custom_command(OUTPUT action.grpc.pb.cc
+    COMMAND ${Protobuf_PROTOC_EXECUTABLE}
+        -I ${proto_dir}
+        --grpc_out=.
+        --plugin=protoc-gen-grpc=${GRPC_CPP_PLUGIN}
+        --cpp_out=. ${proto_dir}/action.proto
+    DEPENDS ${proto_dir}/action.proto
+)
```
julianoes: Also, here I suggest to loop over the files automatically.

grpc/server/dronecore_server.cpp
```diff
@@ -21,16 +28,34 @@ using grpc::ServerReaderWriter;
...
+
 class DroneCoreRPCImpl final : public DroneCoreRPC::Service
 {
 
 public:
-    Status Arm(ServerContext *context, const Empty *request,
+
+};
+
+class ActionRPCImpl final : public ActionRPC::Service
```
julianoes: I think these classes should be in separate files and then inside the respective plugin folder.

grpc/proto/mission.proto
```diff
@@ -0,0 +1,51 @@
+syntax = "proto3";
```
julianoes: I think these proto files should be moved to the respective plugin folders.

grpc/server/dronecore_server.cpp
```diff
@@ -111,25 +140,59 @@ class DroneCoreRPCImpl final : public DroneCoreRPC::Service
...
+
 void RunServer()
 {
     std::string server_address("0.0.0.0:50051");
     DroneCoreRPCImpl service;
 
+    map_type map;
+    map["action"] = &createInstances<ActionRPCImpl>;
+    map["mission"] = &createInstances<MissionRPCImpl>;
+    map["telemetry"] = &createInstances<TelemetryRPCImpl>;
```
julianoes: This will need to be auto-generated in the same way as the plugins are currently added to device_impl.

avinash-palleti:
@julianoes Can you review the changes. If this is fine, we can go ahead for multiple device handling implementation through gRPC.

grpc/server/CMakeLists.txt
```diff
@@ -27,10 +29,39 @@ add_custom_command(OUTPUT dronecore.pb.cc
...
 add_executable(dronecore_server
     dronecore_server.cpp
     dronecore.pb.cc # generated
     dronecore.grpc.pb.cc # generated
+	action.pb.cc # generated
+    action.grpc.pb.cc # generated
+    mission.pb.cc # generated
+    mission.grpc.pb.cc # generated
+    telemetry.pb.cc # generated
+    telemetry.grpc.pb.cc # generated
```
julianoes: You could also generate that list using the same `foreach`. You just add it to all to a variable.

grpc/server/dronecore_server.cpp
julianoes: Do you agree?

grpc/server/plugins/action/action.proto
```diff
@@ -0,0 +1,30 @@
...
+
+package dronecorerpc;
+
+service ActionRPC {
+    rpc Arm(ActionEmpty) returns(ActionResult) {}
+    rpc TakeOff(ActionEmpty) returns(ActionResult) {}
+    rpc Land(ActionEmpty) returns(ActionResult) {}
+}
+
+// FIXME these different Empty types are ugly
```
julianoes: I think there is a protobuf empty that we should reuse: `import "google/protobuf/empty.proto";`.

grpc/server/plugins/telemetry/telemetry.proto
```diff
@@ -0,0 +1,16 @@
+syntax = "proto3";
+
+package dronecorerpc;
+
+service TelemetryRPC {
+    rpc TelemetryPositionSubscription(TelemetryEmpty) returns(stream TelemetryPosition) {}
+}
+
+message TelemetryEmpty {}
```
julianoes: Also re-use empty here.

grpc/server/dronecore_server.cpp
avinash-palleti: @julianoes createInstance takes class name as parameter, if not here somewhere there should be mapping between string to class name. Any other way we could do without hardcoded mapping.

Thanks for the review, we will address other comments in next submission.

mrpollo:
Same comments as @julianoes , thanks for the PR @Rjasuja

grpc/server/plugins/action/actionrpc_impl.h
```diff
@@ -1,22 +1,23 @@
 #include "action.h"
 #include "action.grpc.pb.h"
+#include <google/protobuf/empty.pb.h>
 
 using grpc::Status;
 using grpc::ServerContext;
 using dronecorerpc::ActionRPC;
-using dronecorerpc::ActionEmpty;
+using dronecorerpc::Empty;
```
julianoes: Is this line still needed?

Rjasuja:
Removed.

Rjasuja:
Removed.

Rjasuja:
@julianoes Added cmake file for automated generation of plugin objects. Please review.

julianoes:
@Rjasuja cool thanks, this looks better! Can you merge or rebase with develop and check the CI issues please?

julianoes:
Thanks @Rjasuja!

PR #145, opened 2017-10-30
Download mission capability

julianoes:
This changes the API from `send_mission_async()` to `upload_mission_async()` and adds `download_mission_async`.

This allows missions created with DroneCore to be downloaded from the vehicle. However, DroneCore only supports a subset of the mavlink mission protocol and will fail with the `Result::UNSUPPORTED` for anything it does not know.

### DIFF
--- a/plugins/mission/mission.h
+++ b/plugins/mission/mission.h
@@ -35,6 +35,8 @@ public:
         BUSY, /**< @brief %Vehicle busy. */
         TIMEOUT, /**< @brief Request timed out. */
         INVALID_ARGUMENT, /**< @brief Invalid argument. */
+        UNSUPPORTED, /**< @brief The mission downloaded from the device is not supported. */
+        NO_MISSION_AVAILABLE, /**< @brief No mission available on device. */
         UNKNOWN /**< @brief Unknown error. */
     };
 
@@ -59,6 +61,22 @@ public:
      */
     void upload_mission_async(const std::vector<std::shared_ptr<MissionItem>> &mission_items,
                               result_callback_t callback);
+
+    /**
+     * @brief Callback type for `download_mission_async` call to get mission items and result.
+     */
+    typedef std::function<void(Result, std::vector<std::shared_ptr<MissionItem>>)>
+    mission_items_and_result_callback_t;
+
+    /**
+     * @brief Downloads a vector of mission items from the device (asynchronous).
+     *
+     * The mission items are downloaded from a drone.
+     *
+     * @param mission_items reference to vector of mission items.
+     * @param callback callback to receive result of this request
+     */
+    void download_mission_async(mission_items_and_result_callback_t callback);
 
     /**
      * @brief Starts the mission (asynchronous).
--- a/plugins/mission/mission.h
+++ b/plugins/mission/mission.h
@@ -56,8 +56,8 @@ public:
      * The mission items are uploaded to a drone. Once uploaded the mission can be started and
      * executed even if a connection is lost.
      *
-     * @param mission_items reference to vector of mission items.
-     * @param callback callback to receive result of this request
+     * @param mission_items Reference to vector of mission items.
+     * @param callback Callback to receive result of this request.
      */
     void upload_mission_async(const std::vector<std::shared_ptr<MissionItem>> &mission_items,
                               result_callback_t callback);
@@ -73,8 +73,8 @@ public:
      *
      * The mission items are downloaded from a drone.
      *
-     * @param mission_items reference to vector of mission items.
-     * @param callback callback to receive result of this request
+     * @param mission_items Reference to vector of mission items.
+     * @param callback Callback to receive result of this request.
      */
     void download_mission_async(mission_items_and_result_callback_t callback);
 
@@ -84,7 +84,7 @@ public:
      * Note that the mission must be uplaoded to the vehicle using `upload_mission_async()` before
      * this method is called.
      *
-     * @param callback callback to receive result of this request
+     * @param callback callback to receive result of this request.
      */
     void start_mission_async(result_callback_t callback);
 
@@ -96,7 +96,7 @@ public:
      * A multicopter should just hover at the spot while a fixedwing vehicle should loiter
      * around the location where it paused.
      *
-     * @param callback callback to receive result of this request
+     * @param callback Callback to receive result of this request.
      */
     void pause_mission_async(result_callback_t callback);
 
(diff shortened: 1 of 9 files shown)


plugins/mission/mission.h
```diff
@@ -49,21 +51,37 @@ class Mission
...
+     */
+    typedef std::function<void(Result, std::vector<std::shared_ptr<MissionItem>>)>
+    mission_items_and_result_callback_t;
+
+    /**
+     * @brief Downloads a vector of mission items from the device (asynchronous).
+     *
+     * The mission items are downloaded from a drone.
+     *
+     * @param mission_items reference to vector of mission items.
```
hamishwillee: Capitalise  R in reference

plugins/mission/mission.h
```diff
@@ -49,21 +51,37 @@ class Mission
...
+     * @param callback callback to receive result of this request
```
hamishwillee: Capitisalise C in "Callback to ..."

plugins/mission/mission.h
```diff
@@ -49,21 +51,37 @@ class Mission
...
+    void download_mission_async(mission_items_and_result_callback_t callback);
 
     /**
      * @brief Starts the mission (asynchronous).
      *
-     * Note that the mission must be uplaoded to the vehicle using `send_mission_async()` before
+     * Note that the mission must be uplaoded to the vehicle using `upload_mission_async()` before
      * this method is called.
      *
      * @param callback callback to receive result of this request
```
hamishwillee: Capitalise sentence

plugins/mission/mission.h
```diff
@@ -49,21 +51,37 @@ class Mission
...
 
     /**
-     * @brief Sends a vector of mission items to the device (asynchronous).
+     * @brief Uploads a vector of mission items to the device (asynchronous).
      *
      * The mission items are uploaded to a drone. Once uploaded the mission can be started and
      * executed even if a connection is lost.
      *
      * @param mission_items reference to vector of mission items.
      * @param callback callback to receive result of this request
```
hamishwillee: Capitalize C in sentence. Add full stop.

hamishwillee:
Can we have 

1. an integration test attempts to download a complicated mission.
2. Example or example idea for a download, modify, upload example

hamishwillee:
So to be very clear, you can't download a "partial mission". It either fails and returns nothing or succeeds. Correct?

So is the assumption that a user will create mission in QGC (say) and that DroneCore will allow you to download it if it is valid? Are we expecting people to then manipulate the mission in DroneCore and reupload? Essentially, why is it people want this?

I still think that as a user I'd sometime like to bypass DC altogether and upload/download raw mavlink mission files. Then I could support any mission I liked. Not as "robust" but an easy way to get dronecore working with any vehicle.

This looks fine to me, but leave to @mrpollo to do proper C++ review.

julianoes:
> Can we have 
> 1. an integration test attempts to download a complicated mission.
> 2. Example or example idea for a download, modify, upload example

1. I have added an integration test that uploads and then downloads the mission and verifies it.
2. I don't think that is needed. It seems quite complicated and not a common use case. We don't need examples for all functionality in my opinion.

julianoes:
> So to be very clear, you can't download a "partial mission". It either fails and returns nothing or succeeds. Correct?

Correct. I don't want bugs or corner cases because something gets omitted or changed.

> So is the assumption that a user will create mission in QGC (say) and that DroneCore will allow you to download it if it is valid? Are we expecting people to then manipulate the mission in DroneCore and reupload? How useful is this?

No, that's not really the use case. This is to download DroneCore missions or super simple QGC missions if really needed.

> I have created linked issue to update the mission example and guide docs when this goes in.

Thanks.

> I still think that as a user I'd sometime like to bypass DC altogether and upload/download raw mavlink mission files. Then I could support any mission I liked. Not as "robust" but an easy way to get dronecore working with any vehicle.

The point of DroneCore is to give you a thoroughly tested API. Once more functionality is tested and implemented easy to use, then we can add that. If somebody wants to write its own mission, they can just write their own mission plugin or use something else altogether. I know this is very opinionated but I see people struggling with the mavlink API every day and if we just replicate that with DroneCore we are no better but only added another layer.

> This looks fine to me, but leave to @mrpollo to do proper C++ review.

Yes please.

hamishwillee:
@julianoes 

So yes to almost everything. The "no" is I still don't see why we are bothering. Consider

> > Example or example idea for a download, modify, upload example

> I have added an integration test that uploads and then downloads the mission and verifies it.
> I don't think that is needed. It seems quite complicated and not a common use case. We don't need examples for all functionality in my opinion.

>> So is the assumption that a user will create mission in QGC (say) and that DroneCore will allow you to download it if it is valid? Are we expecting people to then manipulate the mission in DroneCore and reupload? How useful is this?

> No, that's not really the use case. This is to download DroneCore missions or super simple QGC missions if really needed.

Yes, but if you're not going to manipulate them, why would you bother? If its a DroneCore mission you already created it, so why bother downloading it?

Perhaps doesn't matter for this PR, but for the docs it is nice to be able to explain why functionality exists.

julianoes:
> Yes, but if you're not going to manipulate them, why would you bother? If its a DroneCore mission you already created it in your app, so why bother downloading it from the vehicle? Just seems a bit pointless.

Good point. Because it is on my feature todo list 😄.

One reason could be that you plan a mission using a desktop application and upload it to the drone. Later when you view the telemetry using an app on your phone you can download the mission, so you can follow the progress.

mrpollo:
@hamishwillee here's the list of supported commands this PR introduces

```
MAV_CMD_NAV_WAYPOINT
MAV_CMD_DO_MOUNT_CONTROL
MAV_CMD_IMAGE_START_CAPTURE
	START_PHOTO_INTERVAL
	TAKE_PHOTO
MAV_CMD_IMAGE_STOP_CAPTURE
MAV_CMD_VIDEO_START_CAPTURE
MAV_CMD_VIDEO_STOP_CAPTURE
MAV_CMD_DO_CHANGE_SPEED
```

hamishwillee:
Thanks @mrpollo - https://github.com/dronecore/DroneCore/pull/145#issuecomment-342265399 not above very useful. I won't do update in my docs tree until this has been merged into develop.

julianoes:
> @julianoes What other ones are supported - I should add them all.

It's just these, I think.

hamishwillee:
Thanks. Any ETA on this going into develop?

julianoes:
@hamishwillee rebased, if CI goes through we can merge it.

hamishwillee:
Looks like some checks were not successful :-(

PR #147, opened 2017-11-01
Add fake time

### LINKED ISSUE #43: Intermittent CI fail
Sometimes the unit tests fail in CI, presumably because of timing issues.

```
/Users/travis/build/dronecore/DroneCore/core/global_include_test.cpp:27: Failure
Expected: (seconds_elapsed) < (0.11), actual: 0.110119 vs 0.11
[  FAILED  ] GlobalInclude.ElapsedSinceAboutRight (110 ms)
```
https://travis-ci.org/dronecore/DroneCore/jobs/269515690

julianoes:
In order to avoid sketchy unit test results on bursted or slow CI
machines, it makes sense to provide mocked time functionality.

This should fix the CI troubles introduced by #134.

Should fix #43 for good.


julianoes:
Note to self. It should probably called fake time.

Also, this doesn't actually work yet because the fake is selected at compile time, however, the library is not compiled with the unit test flags but only the unit test part is. Therefore, part of the time API is still using the real time. Something like dependency injection could fix this.

julianoes:
Note: I have local changes and an updated docker image for this that I could not push yet.

julianoes:
I'm merging this now because I need to rebase other PRs on top of it that are failing because of the flaky unit tests.

PR #150, opened 2017-11-03
Implement transition to fw as an action.

xgerrmann:
Implementation of the transition to Fixed Wing from VTOL.
One problem: It only builds if the action header files in the install directory contain the correct definitions. I did not include these files since the directory is placed in the .gitignore.

Checks to perform prior to acceptance:
- [ ] Test how commands behave if used in wrong modes or on wrong vehicles and confirm docs match that.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -110,14 +110,8 @@ if (IOS)
     include_directories(curl-android-ios/prebuilt-with-ssl/ios/include)
 endif()
 
-if (IOS OR ANDROID OR MSVC OR APPLE)
+if (IOS OR ANDROID OR MSVC)
     set(library_type "STATIC")
-
-    # We need tinyxml2 for the camera definition parsing.
-    # We use the submodule when linking statically.
-    add_subdirectory(libs/tinyxml2 EXCLUDE_FROM_ALL)
-    include_directories(SYSTEM libs/tinyxml2)
-
 else()
     set(library_type "SHARED")
 endif()
@@ -149,7 +143,7 @@ add_library(dronecore ${library_type}
 target_link_libraries(dronecore
     ${CMAKE_THREAD_LIBS_INIT}
     ${curl_lib}
-    tinyxml2
+#    tinyxml2
 )
 
 # cmake should check for C++11
@@ -160,9 +154,9 @@ set_target_properties(dronecore
     PROPERTIES COMPILE_FLAGS ${warnings}
 )
 
-# Use tinyxml2 from the host system for Linux.
-if(NOT IOS AND NOT ANDROID AND NOT MSVC AND NOT APPLE)
-endif()
+# We need tinyxml2 for the camera definition parsing.
+#add_subdirectory(libs/tinyxml2 EXCLUDE_FROM_ALL)
+#include_directories(SYSTEM libs/tinyxml2)
 
 # We support install in order to use the header and library files in
 # other applications.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -110,8 +110,14 @@ if (IOS)
     include_directories(curl-android-ios/prebuilt-with-ssl/ios/include)
 endif()
 
-if (IOS OR ANDROID OR MSVC)
+if (IOS OR ANDROID OR MSVC OR APPLE)
     set(library_type "STATIC")
+
+    # We need tinyxml2 for the camera definition parsing.
+    # We use the submodule when linking statically.
+    add_subdirectory(libs/tinyxml2 EXCLUDE_FROM_ALL)
+    include_directories(SYSTEM libs/tinyxml2)
+
 else()
     set(library_type "SHARED")
 endif()
@@ -143,7 +149,7 @@ add_library(dronecore ${library_type}
 target_link_libraries(dronecore
     ${CMAKE_THREAD_LIBS_INIT}
     ${curl_lib}
-#    tinyxml2
+    tinyxml2
 )
 
 # cmake should check for C++11
@@ -154,9 +160,9 @@ set_target_properties(dronecore
     PROPERTIES COMPILE_FLAGS ${warnings}
 )
 
-# We need tinyxml2 for the camera definition parsing.
-#add_subdirectory(libs/tinyxml2 EXCLUDE_FROM_ALL)
-#include_directories(SYSTEM libs/tinyxml2)
+# Use tinyxml2 from the host system for Linux.
+if(NOT IOS AND NOT ANDROID AND NOT MSVC AND NOT APPLE)
+endif()
 
 # We support install in order to use the header and library files in
 # other applications.
--- a/example/transition_vtol_fixed_wing/CMakeLists.txt
+++ b/example/transition_vtol_fixed_wing/CMakeLists.txt
@@ -1,0 +1,39 @@
+cmake_minimum_required(VERSION 2.8.12)
+
+project(transition_vtol_fixed_wing)
+
+if(NOT MSVC)
+    add_definitions("-std=c++11 -Wall -Wextra -Werror")
+else()
+    add_definitions("-std=c++11 -WX -W2")
+    set(platform_libs "Ws2_32.lib")
+endif()
+
+# Add DEBUG define for Debug target
+set(CMAKE_CXX_FLAGS_DEBUG "-DDEBUG")
+
+# This finds thread libs on Linux, Mac, and Windows.
+find_package(Threads REQUIRED)
+
+# Not needed if DroneCore installed system-wide
+include_directories(
+    ${CMAKE_SOURCE_DIR}/../../install/include
+)
+
+add_executable(transition_vtol_fixed_wing
+    transition_vtol_fixed_wing.cpp
+)
+
+# Not needed if DroneCore installed system-wide
+if(MSVC)
+    set(dronecore_lib "${CMAKE_SOURCE_DIR}/../../install/lib/dronecore.lib")
+else()
+    set(dronecore_lib "${CMAKE_SOURCE_DIR}/../../install/lib/libdronecore.so")
+endif()
+
+target_link_libraries(transition_vtol_fixed_wing
+    ${dronecore_lib}  # Remove/comment out this line if DroneCore used locally
+    # dronecore  # Uncomment/add this line if DroneCore installed system-wide
+    ${CMAKE_THREAD_LIBS_INIT}
+    ${platform_libs}
+)
--- a/example/transition_vtol_fixed_wing/transition_vtol_fixed_wing.cpp
+++ b/example/transition_vtol_fixed_wing/transition_vtol_fixed_wing.cpp
@@ -90,10 +90,10 @@ int main(int /*argc*/, char ** /*argv*/)
         return 1;
     }
 
-	// Wait
-	std::this_thread::sleep_for(std::chrono::seconds(10));
+    // Wait
+    std::this_thread::sleep_for(std::chrono::seconds(10));
 
-	// Change to FW
+    // Change to FW
     const Action::Result fw_result = device.action().transition_to_fixedwing();
 
     if (fw_result != Action::Result::SUCCESS) {
@@ -102,30 +102,30 @@ int main(int /*argc*/, char ** /*argv*/)
         //return 1;
     }
 
-	// Wait
-	std::this_thread::sleep_for(std::chrono::seconds(10));
+    // Wait
+    std::this_thread::sleep_for(std::chrono::seconds(10));
 
     // Change to VTOL
     const Action::Result mc_result = device.action().transition_to_multicopter();
     if (mc_result != Action::Result::SUCCESS) {
         std::cout << ERROR_CONSOLE_TEXT << "Transition to multi copter failed:" << Action::result_str(
                       mc_result) << NORMAL_CONSOLE_TEXT << std::endl;
-    //    return 1;
+        //    return 1;
     }
 
-	// Wait
-	std::this_thread::sleep_for(std::chrono::seconds(5));
+    // Wait
+    std::this_thread::sleep_for(std::chrono::seconds(5));
 
     // Return to launch
     const Action::Result rtl_result = device.action().return_to_launch();
     if (rtl_result != Action::Result::SUCCESS) {
         std::cout << ERROR_CONSOLE_TEXT << "Returning to launch failed:" << Action::result_str(
                       rtl_result) << NORMAL_CONSOLE_TEXT << std::endl;
-    //    return 1;
+        //    return 1;
     }
 
-	// Wait
-	std::this_thread::sleep_for(std::chrono::seconds(20));
+    // Wait
+    std::this_thread::sleep_for(std::chrono::seconds(20));
 
     // Land
     std::cout << "Landing..." << std::endl;
@@ -133,7 +133,7 @@ int main(int /*argc*/, char ** /*argv*/)
     if (land_result != Action::Result::SUCCESS) {
         std::cout << ERROR_CONSOLE_TEXT << "Land failed:" << Action::result_str(
                       land_result) << NORMAL_CONSOLE_TEXT << std::endl;
-    //    return 1;
+        //    return 1;
     }
 
     // We are relying on auto-disarming but let's keep watching the telemetry for a bit longer.
--- a/example/transition_vtol_fixed_wing/transition_vtol_fixed_wing.cpp
+++ b/example/transition_vtol_fixed_wing/transition_vtol_fixed_wing.cpp
@@ -1,0 +1,143 @@
+
+#include <dronecore/dronecore.h>
+#include <iostream>
+#include <thread>
+#include <chrono>
+#include <cstdint>
+
+using std::this_thread::sleep_for;
+using std::chrono::milliseconds;
+using namespace dronecore;
+
+#define ERROR_CONSOLE_TEXT "\033[31m" //Turn text on console red
+#define TELEMETRY_CONSOLE_TEXT "\033[34m" //Turn text on console blue
+#define NORMAL_CONSOLE_TEXT "\033[0m"  //Restore normal console colour
+
+int main(int /*argc*/, char ** /*argv*/)
+{
+    DroneCore dc;
+
+    bool discovered_device = false;
+
+    DroneCore::ConnectionResult connection_result = dc.add_udp_connection();
+
+    if (connection_result != DroneCore::ConnectionResult::SUCCESS) {
+        std::cout << ERROR_CONSOLE_TEXT << "Connection failed: "
+                  << DroneCore::connection_result_str(connection_result)
+                  << NORMAL_CONSOLE_TEXT << std::endl;
+        return 1;
+    }
+
+    std::cout << "Waiting to discover device..." << std::endl;
+    dc.register_on_discover([&discovered_device](uint64_t uuid) {
+        std::cout << "Discovered device with UUID: " << uuid << std::endl;
+        discovered_device = true;
+    });
+
+    // We usually receive heartbeats at 1Hz, therefore we should find a device after around 2 seconds.
+    std::this_thread::sleep_for(std::chrono::seconds(2));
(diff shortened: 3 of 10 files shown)


plugins/action/action.h
```diff
@@ -119,6 +119,13 @@ class Action
     Result return_to_launch() const;
 
     /**
+     * @brief Send command to *change flight mode* (FW).
```
julianoes: Please describe what this does for an outsider that doesn't know the concept of flight modes and neither the acronym FW.

plugins/action/action.cpp
```diff
@@ -42,6 +42,11 @@ Action::Result Action::return_to_launch() const
     return _impl->return_to_launch();
 }
 
+Action::Result Action::transition_to_fw() const
```
julianoes: I'm tempted to suggest to rename this to `transition_to_fixedwing`. The goal is to have an easy to use, and self explanatory API. I would error on the side of verbosity.

plugins/action/action.h
julianoes: And also comment that this is only applicable for hybrid/VTOL vehicles.

julianoes:
@xgerrmann for your build problems: 

It should build just fine if you remove what you have installed, and just do `make default`.

hamishwillee:
@xgerrmann @julianoes 
1.  How do you transition back to MC mode? IMO this should probably have a boolean to indicate the mode (although I guess you could do a matching transition to mc method.
2. There should be a matching method, probably in Telemetry  to get the current mode. 
3. There should be a corresponding async method to match the sync method.

xgerrmann:
@julianoes  @hamishwillee 
Thanks for your comments.
I've implemented the switch back to multicopter as transition_to_multicopter();
One problem I'm running into is that the device.action().transition_to_fixedwing() is giving met timeouts as result, thus Action::Result::Timeout. It seems there is no acknowledgement from the VTOL, any ideas on this?

julianoes:
> One problem I'm running into is that the device.action().transition_to_fixedwing() is giving met timeouts as result, thus Action::Result::Timeout. It seems there is no acknowledgement from the VTOL, any ideas on this?

Ok, does the transition happen when you try it in SITL?

julianoes:
@xgerrmann Can you please remove the install folder that you checked in from this PR?

xgerrmann:
I've added an example now for the transition between multicopter and fixedwing. In the mean time I rebased on the latest head of the develop branch. I removed the install directory from the git, since I now know how to correctly install with 'make default install'.

Timeouts still occur in the example, I switched off the 'return 1', so that the actions still are performed. So it seems like the requested behaviour is performed, but with timeout errors from MAVLINK.

mrpollo:
@xgerrmann once @julianoes is done with his review can you please do some cleanup with your commits? perhaps merging some of them with an interactive rebase, I want to encourage better commit titles and messages

xgerrmann:
@mrpollo Ill clean it up right now. Some changes were indeed not related to this branch and therefore i later removed them from here.

plugins/action/action.h
```diff
@@ -119,6 +119,26 @@ class Action
     Result return_to_launch() const;
 
     /**
+     * @brief Send command to transition the drone to fixedwing.
+     *
+     * Note that this is only for the vtol type.
```
hamishwillee: Suggest replace both of these points (and update similarly elsewhere) to:

> The associated action will only be executed for VTOL vehicles in multicopter mode. 
> On other vehicles/modes the command will fail with a `Result`.

I assume that the statement is correct - ie that the command will fail with an error?

hamishwillee:
I'm going to go so far as to suggest you abandon this PR and re-add to latest develop - a lazy man's rebase. There are just too many "accidental" changes in here for anyone to be sure it is safe to acccept. 

The method signatures look OK to me. I have added comment inline about document improvement. Can we also get current  VTOL mode added to telemetry? @julianoes Would be good to hear your ideas on best way to do this - as part of a VTOL info class or something more generic?

xgerrmann:
Ok guys, I squashed the commits. It's much cleaner now :)
Just try to run the example and you'll notice the timeout errors.

plugins/action/action.h
```diff
@@ -187,6 +207,26 @@ class Action
     void return_to_launch_async(result_callback_t callback);
 
     /**
+     * @brief Send command to transition the drone to fixedwing (asynchronous).
+     *
+     * Note that this is only for the vtol type.
```
hamishwillee: Argg, my comment from yesterday lost. Can you check your email for my suggested wording?

hamishwillee:
As per [this note](https://docs.dronecore.io/v/develop/en/contributing/plugins.html#example-code) it is a good idea to first write integration tests BEFORE writing example code. You'll need them anyway!

@julianoes I ran this, and the standard retries/timeout code is indeed being shown. What is interesting though is that the commands are being obeyed - ie the transition is happening. So either the ack is not being sent or it is being handled incorrectly.

Such a pity that logging isn't enabled so we could check all the traffic and see if "something" is really coming back. Any ideas on debugging?

plugins/action/action.h
xgerrmann: Found it :) 
Next commit will include your comment.Found it :) 
Next commit will include your comment.

xgerrmann:
In the latest commit, I've implemented an integrationtest.
The same problem with the acknowledgement, still occurs. 
I've been searching for a MAVLINK message that asks for the state of the vtol (multicopter or fixedwing), so that in the test script I can check wether the states are correct and put it in a EXPECT, but so far I have not been able to find such a state request message.
Any ideas on:
- How to request the multicopter or fixedwing state?
- How to solve the acknowledgement problem?

julianoes:
@xgerrmann thanks a lot. I will look into the issues. I might not get to it today I'm sorry but definitely tomorrow.

plugins/action/action.h
```diff
@@ -119,6 +119,26 @@ class Action
     Result return_to_launch() const;
 
     /**
+     * @brief Send command to transition the drone to fixedwing.
+     *
+     * The associated action will only be executed for VTOL vehicles in multicopter mode.
```
hamishwillee: Thanks! This will need to be checked of course (once we have got the timeout issue fix). It may be that PX4 would simply ignore the command if sent while already in the current mode (rather than fail) and similarly it might ignore the command for non vtol rather than failing it.

plugins/action/action.h
hamishwillee: Thanks. Looks good. We will need to test that the comment is actually true once the timeout is fixed :-)

example/transition_vtol_fixed_wing/transition_vtol_fixed_wing.cpp
```diff
@@ -0,0 +1,143 @@
...
+    if (fw_result != Action::Result::SUCCESS) {
+        std::cout << ERROR_CONSOLE_TEXT << "Transition to fixed wing failed: " << Action::result_str(
+                      fw_result) << NORMAL_CONSOLE_TEXT << std::endl;
+        //return 1;
+    }
+
+	// Wait
+	std::this_thread::sleep_for(std::chrono::seconds(10));
+
+    // Change to VTOL
```
hamishwillee: I'm not planning on fully reviewing this example just yet, but please add console output for vtol steps in the same way as we have for other (e.g. for taking off ... ` std::cout << "Taking off..." << std::endl;`). That makes the whole thing much more interesting when viewing operation.

When you have API to show vehicle state  you would probably use that to show transitions too.

hamishwillee:
> How to request the multicopter or fixedwing state?

Without having actually tried it, using [EXTENDED_SYS_STATE](https://mavlink.io/en/messages/common.html#EXTENDED_SYS_STATE)

julianoes:
Ok, I'm debugging the missing ack and writing down my steps as suggested by @hamishwillee:

I check out the PR:
```
cd DroneCore
make clean
git fetch origin pull/150/head:pr_150
git checkout pr_150
make
```

I look for the new integration test:
```
make && build/default/integration_tests_runner --gtest_list_tests
```

Ok found it let's run it:
```
make && build/default/integration_tests_runner --gtest_filter="SitlTest.ActionSimpleTransition"
```

And of course run px4 latest master:
```
cd Firmware
make clean
git checkout master
git fetch
git merge
git submodule update --init --recursive
make posix gazebo_standard_vtol
```

I can reproduce the missing ack:
```
[04:00:24|Info ] sending again, retries to do: 3  (3000). (mavlink_commands.cpp:184)
[04:00:24|Info ] sending again, retries to do: 2  (3000). (mavlink_commands.cpp:184)
[04:00:25|Info ] sending again, retries to do: 1  (3000). (mavlink_commands.cpp:184)
[04:00:25|Error] Retrying failed (3000) (mavlink_commands.cpp:202)
```

Command `3000` is indeed [MAV_CMD_DO_VTOL_TRANSITION](http://mavlink.org/messages/common#MAV_CMD_DO_VTOL_TRANSITION).

Let's grep where this command is handled in Firmware:
```
rg MAV_CMD_DO_VTOL_TRANSITION src
src/modules/mavlink/mavlink_mission.cpp
1437:		case MAV_CMD_DO_VTOL_TRANSITION:
```

This seems wrong because we're not interested in mission items.

Let's look for just `DO_VTOL_TRANSITION` instead:

```
rg DO_VTOL_TRANSITION src 
src/modules/vtol_att_control/vtol_att_control_main.cpp
484:	if (_vehicle_cmd.command == vehicle_command_s::VEHICLE_CMD_DO_VTOL_TRANSITION) {

src/modules/mavlink/mavlink_mission.cpp
1437:		case MAV_CMD_DO_VTOL_TRANSITION:
1511:		case NAV_CMD_DO_VTOL_TRANSITION:

src/modules/commander/commander.cpp
564:			.command = vehicle_command_s::VEHICLE_CMD_DO_VTOL_TRANSITION,
1185:	case vehicle_command_s::VEHICLE_CMD_DO_VTOL_TRANSITION:

src/modules/navigator/mission_feasibility_checker.cpp
292:		    missionitem.nav_cmd != NAV_CMD_DO_VTOL_TRANSITION) {

src/modules/navigator/mission_block.cpp
98:	case NAV_CMD_DO_VTOL_TRANSITION:
636:		vcmd.command = NAV_CMD_DO_VTOL_TRANSITION;

src/modules/navigator/rtl.cpp
204:			_mission_item.nav_cmd = NAV_CMD_DO_VTOL_TRANSITION;

src/modules/navigator/navigation.h
88:	NAV_CMD_DO_VTOL_TRANSITION = 3000,

src/modules/navigator/mission.cpp
641:			_mission_item.nav_cmd = NAV_CMD_DO_VTOL_TRANSITION;
695:			_mission_item.nav_cmd = NAV_CMD_DO_VTOL_TRANSITION;
753:		if (_mission_item.nav_cmd == NAV_CMD_DO_VTOL_TRANSITION
781:		if (_mission_item.nav_cmd == NAV_CMD_DO_VTOL_TRANSITION
984:	    || _mission_item.nav_cmd == NAV_CMD_DO_VTOL_TRANSITION
1021:		    && !(_mission_item.nav_cmd == NAV_CMD_DO_VTOL_TRANSITION || _navigator->get_vstatus()->in_transition_mode)) {
```

Let's look at the controller itself first:
https://github.com/PX4/Firmware/blob/master/src/modules/vtol_att_control/vtol_att_control_main.cpp#L484-L486

Looks like the command is just not acked at all. Let's change that:
https://github.com/PX4/Firmware/pull/8264

CMakeLists.txt
```diff
@@ -143,7 +143,7 @@ add_library(dronecore ${library_type}
 target_link_libraries(dronecore
     ${CMAKE_THREAD_LIBS_INIT}
     ${curl_lib}
-    tinyxml2
+#    tinyxml2
```
julianoes: Please don't change this stuff. Make sure to either merge or rebase on current develop and get rid of this diff.

example/transition_vtol_fixed_wing/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
+cmake_minimum_required(VERSION 2.8.12)
+
+project(transition_vtol_fixed_wing)
+
+if(NOT MSVC)
+    add_definitions("-std=c++11 -Wall -Wextra -Werror")
+else()
+    add_definitions("-std=c++11 -WX -W2")
+    set(platform_libs "Ws2_32.lib")
+endif()
```
julianoes: Once you merge or rebase you will see that this file gets a lot simpler.

example/transition_vtol_fixed_wing/transition_vtol_fixed_wing.cpp
julianoes: I agree.

integration_tests/transition_multicopter_fixedwing.cpp
```diff
@@ -0,0 +1,83 @@
...
+	DroneCore::ConnectionResult ret = dc.add_udp_connection();
+    ASSERT_EQ(ret, DroneCore::ConnectionResult::SUCCESS);
+
+    // Wait for device to connect via heartbeat.
+    std::this_thread::sleep_for(std::chrono::seconds(2));
+    ASSERT_TRUE(dc.is_connected());
+
+	Device &device = dc.device();
+
+	// Takeoff
```
julianoes: I would get rid of comments that are identical to a function name. They don't add any information. I suggest only to add comments if you need to explain something, add additional info for the reader, e.g. `We need to takeoff first, otherwise we can't actually transition` 😄

integration_tests/transition_multicopter_fixedwing.cpp
```diff
@@ -0,0 +1,83 @@
...
+    float altitude_m = 10.0f;
+    device.action().set_takeoff_altitude(altitude_m);
+
+    action_ret = device.action().takeoff();
+    EXPECT_EQ(action_ret, Action::Result::SUCCESS);
+    const int wait_time_s = 10;
+    std::this_thread::sleep_for(std::chrono::seconds(wait_time_s));
+
+
+    EXPECT_GT(device.telemetry().position().relative_altitude_m, altitude_m - 0.25f);
```
julianoes: Nice check, that will be useful once these integration tests make it into the PX4 Firmware testing.

plugins/action/action_impl.cpp
```diff
@@ -157,6 +157,46 @@ Action::Result ActionImpl::return_to_launch() const
                    MavlinkCommands::DEFAULT_COMPONENT_ID_AUTOPILOT));
 }
 
+Action::Result ActionImpl::transition_to_fixedwing() const
+{
+    return action_result_from_command_result(
+               _parent->send_command_with_ack(
+                   MAV_CMD_DO_VTOL_TRANSITION,
+                   MavlinkCommands::Params {float(MAV_VTOL_STATE_FW)},
+                   MavlinkCommands::DEFAULT_COMPONENT_ID_AUTOPILOT));
```
julianoes: Yeah, that's it. I like that you added both sync and async methods, thanks!

plugins/action/action.h
hamishwillee: @julianoes So based on your change https://github.com/PX4/Firmware/pull/8264 would I be correct in saying that
- on a vtol vehicle the result will either be "success" or timeout, where success is an ack of message received, not that the vehicle will act on it? In other words, if you're already in FW and you call this it will just succeed?
- What will happen on an MC or FW only vehicle - will your added code still get run? If so, then basically this will return success if received by vehicle, which will be a bit odd on MC. Should our code therefore check the type of vehicle we're talking to before even sending the message?

CMakeLists.txt
hamishwillee: I'd say rebase

example/transition_vtol_fixed_wing/CMakeLists.txt
hamishwillee: The required changes are being documented in https://github.com/dronecore/docs/pull/62 (in progress). The interesting bit is [here](https://github.com/hamishwillee/docs-1/blob/4dcd412abc72b448e9260f81f5922ea776e5262f/en/guide/toolchain.md#dronecore-installed-system-wide-dronecore_installed_system) - basically we're changing to use system-wide install by default AND dronecore itself is now linking to the thread library - so you no longer need to do most of this work.

integration_tests/transition_multicopter_fixedwing.cpp
xgerrmann: Thanks, but I have to give credit where due. I got it from the simple_hover example ;)

xgerrmann:
@julianoes I Really liked how you've written down your steps, this gives insight in your workflow and results in learning opportunities for people reading along (referring to myself here).

plugins/action/action.h
julianoes: > - on a vtol vehicle the result will either be "success" or timeout, where success is an ack of message received, not that the vehicle will act on it? In other words, if you're already in FW and you call this it will just succeed?

correct, it will just succeed.

> - What will happen on an MC or FW only vehicle - will your added code still get run? If so, then basically this will return success if received by vehicle, which will be a bit odd on MC. Should our code therefore check the type of vehicle we're talking to before even sending the message?

It will just time out because the VTOL controller is not running and therefore not responding.

A check to see if it is actually a VTOL and a specific error message for it would be nice!

integration_tests/transition_multicopter_fixedwing.cpp
```diff
@@ -30,28 +30,32 @@ void takeoff_and_transition_to_fixedwing()
...
-	// Takeoff
+	// We need to takeoff first, otherwise we can't actually transition
+    LogInfo() << "Taking off";
 	takeoff(device);
 
-	// Transition to fixedwing
+    LogInfo() << "Transitioning to fixedwing";
 	device.action().transition_to_fixedwing();
+
+	// Wait a little before the transition back to multicopter,
```
hamishwillee: And noting that we need to update to check transition once that is in place.

plugins/action/action.h
hamishwillee: @xgerrmann So
1. Can we check this is VTOL and have a specific error message for other types (rather than getting timeout).
2. So the text needs to change slightly as a  result of Julian's feedback:
    ```
   The associated action will fail with a Result if called for non-VTOL vehicles
   The action is ignored if called when the vehicle is in fixed-wing mode mode.
   ```

@julianoes What about if someone is using firmware pre your fix? Should we also have a PX4 version check?
The problem here is that even in an old version we always want to send the  message, resulting in the timeout. But in an old version we want to ignore the ack result. To me that sounds like we want to have a version of our send that if we don't get an ack on we can force to succeed anyway. THoughts?

mrpollo:
@hamishwillee I really liked @julianoes short write up, can we find it a home in our docs?

hamishwillee:
@mrpollo 

> @hamishwillee I really liked @julianoes short write up, can we find it a home in our docs?

I already created https://github.com/dronecore/docs/issues/63 to track this. It is not as easy as you might think to turn this into useful generic advice.

hamishwillee:
IMO there are a number of comments still not addressed.
- Example uses old CMakeLists: https://github.com/dronecore/DroneCore/pull/150/files#r150093535
- No API to get the current FW/MC state: https://github.com/dronecore/DroneCore/pull/150#issuecomment-342997474

example/transition_vtol_fixed_wing/transition_vtol_fixed_wing.cpp
```diff
@@ -0,0 +1,143 @@
...
+    if (mc_result != Action::Result::SUCCESS) {
+        std::cout << ERROR_CONSOLE_TEXT << "Transition to multi copter failed:" << Action::result_str(
+                      mc_result) << NORMAL_CONSOLE_TEXT << std::endl;
+        //    return 1;
+    }
+
+    // Wait
+    std::this_thread::sleep_for(std::chrono::seconds(5));
+
+    // Return to launch
```
hamishwillee: Actions in examples should be output to console rather than commented so that users have something to watch in the console. So "arming", "Taking off", "Landing", "Returning" etc.

PR #163, opened 2017-11-14
Follow up for VTOL transition additions

julianoes:
This follows up on https://github.com/dronecore/DroneCore/pull/150#issuecomment-344140095.

### DIFF
--- a/example/transition_vtol_fixed_wing/transition_vtol_fixed_wing.cpp
+++ b/example/transition_vtol_fixed_wing/transition_vtol_fixed_wing.cpp
@@ -93,7 +93,7 @@ int main(int /*argc*/, char ** /*argv*/)
     // Wait
     std::this_thread::sleep_for(std::chrono::seconds(10));
 
-    // Change to FW
+    std::cout << "Transition to fixedwing..." << std::endl;
     const Action::Result fw_result = device.action().transition_to_fixedwing();
 
     if (fw_result != Action::Result::SUCCESS) {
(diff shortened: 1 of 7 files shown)


example/transition_vtol_fixed_wing/transition_vtol_fixed_wing.cpp
```diff
@@ -106,6 +106,7 @@ int main(int /*argc*/, char ** /*argv*/)
     std::this_thread::sleep_for(std::chrono::seconds(10));
 
     // Change to VTOL
```
hamishwillee: Remove - incorrect and not needed.

julianoes: Thx will do.

julianoes:
> What will happen on PX4 prior to version with ACK added? I assume that this will still timeout. Should there be a test for version?

It will timeout but still work, so not really a problem in my opinion.

> Comments on methods not quite correct - below for the transition to fixed wing:

Thanks, fixed.

hamishwillee:
>> What will happen on PX4 prior to version with ACK added? I assume that this will still timeout. Should there be a test for version?

> It will timeout but still work, so not really a problem in my opinion.

I somewhat disagree, because you're getting a fail error even though the action will actually be executed (if the action was not being executed I'd feel differently).  The caller has to "magically" know that this fail is OK. 

Your call though. 

Otherwise all good.

julianoes:
> I somewhat disagree, because you're getting a fail error even though the action will actually be executed (if the action was not being executed I'd feel differently). The caller has to "magically" know that this fail is OK.

DroneCore does not "officially" support any kind of release yet but is only tested against master. Therefore, we don't need to worry about earlier releases/versions at this point.

hamishwillee:
Thanks!

PR #174, opened 2017-11-24
grpc:Update python client for dynamic plugin loading

Rjasuja:
Update python examples

### DIFF
--- a/grpc/python_client/CMakeLists.txt
+++ b/grpc/python_client/CMakeLists.txt
@@ -5,6 +5,7 @@ include(FindPythonInterp REQUIRED)
 add_custom_target(pydronecore ALL)
 
 set(plugins_dir ${CMAKE_SOURCE_DIR}/grpc/server/plugins)
+set(plugins_list action mission telemetry)
 
 if(${PYTHONINTERP_FOUND})
     set(proto_dir ${CMAKE_SOURCE_DIR}/grpc/proto)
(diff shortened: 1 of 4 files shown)


Rjasuja:
For async_client.py, thread for telemetry exits with Status Code as CANCELLED.

JonasVautherin:
How do you test this? When using this PR together with #173, and running both `dronecore_server` and either `async_client.py` or `sync_client.py`, I get:

> grpc._channel._Rendezvous: <_Rendezvous of RPC that terminated with (StatusCode.UNAVAILABLE, Endpoint read failed)>

grpc/python_client/CMakeLists.txt
```diff
@@ -5,6 +5,7 @@ include(FindPythonInterp REQUIRED)
 add_custom_target(pydronecore ALL)
 
 set(plugins_dir ${CMAKE_SOURCE_DIR}/grpc/server/plugins)
+set(plugins_list action mission telemetry)
```
JonasVautherin: I prefer the way it is done in #179: `file(STRINGS ../proto/plugins.conf plugins_list)`, so that it is not hard-coded.

julianoes: I agree, I will change it in #179.

julianoes:
Thanks @Rjasuja and sorry for not properly looking at your PR and duplicating the work. That was stupid from my side 😅.

PR #177, opened 2017-11-27
core: do blocking read on serial port

### LINKED ISSUE #176: Serial connection at 100% CPU
When using a Pixhawk connected via a serial connection (dronecore::DroneCore::add_serial_connection()), the CPU usage for one core goes to 100%. Looking at it with a profiler, I found that SerialConnection::receive() has a while loop that reads as fast as possible and the read() used is on a non-blocking file descriptor, leading to this behavior.

On embedded devices, having 100% CPU usage is not feasible. I tried solving the issue by removing the O_NONBLOCK option when opening the file descriptor, but that didn't work (i.e., it behaved the same way). Then I added small sleeps into the receive method which works, but feels more like a workaround.

```
void SerialConnection::receive(SerialConnection *parent)
{
    // Enough for MTU 1500 bytes.
    char buffer[2048];

    const std::chrono::milliseconds wait_time_after_read(5);

    while (!parent->_should_exit) {
        int recv_len = read(parent->_fd, buffer, sizeof(buffer));
        if (recv_len < -1) {
            LogErr() << "read failure: " << GET_ERROR(errno);
        }
        if (recv_len > static_cast<int>(sizeof(buffer)) || recv_len == 0) {
            std::this_thread::sleep_for(wait_time_after_read);
            continue;
        }
        parent->_mavlink_receiver->set_new_datagram(buffer, recv_len);
        // Parse all mavlink messages in one data packet. Once exhausted, we'll exit while.
        while (parent->_mavlink_receiver->parse_message()) {
            parent->receive_message(parent->_mavlink_receiver->get_last_message());
        }
        std::this_thread::sleep_for(wait_time_after_read);
    }
}  
```

The 5ms in the example are quite a lot, 1ms also reduces the CPU usage a lot already. So while this works, it seems like not a proper solution. I wonder if the unlimited loop speed is by design or if it's a bug and should/will be fixed.

julianoes:
We should be able to read in a blocking way on the serial port since it
happens on a separate thread.

Fixes #176.

### DIFF
--- a/core/serial_connection.cpp
+++ b/core/serial_connection.cpp
@@ -82,8 +82,8 @@ DroneCore::ConnectionResult SerialConnection::setup_port()
     tc.c_cflag &= ~(CSIZE | PARENB | CBAUD | CRTSCTS);
     tc.c_cflag |= CS8 | BOTHER;
 
-    tc.c_cc[VMIN] = 0;
-    tc.c_cc[VTIME] = 0;
+    tc.c_cc[VMIN] = 1; // We want at least 1 byte to be available.
+    tc.c_cc[VTIME] = 0; // We don't timeout but wait indefinitely.
     tc.c_ispeed = _baudrate;
     tc.c_ospeed = _baudrate;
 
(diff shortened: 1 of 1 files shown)


shakthi-prashanth-m:
@julianoes  is `O_NONBLOCK` is the culprit ?

julianoes:
Tested with a Pixhawk, will merge now because it's a bug.

PR #178, opened 2017-11-27
Fix Windows linking

### LINKED ISSUE #162: Building example locally on windows fails after #130
I buil DroneCore locally on windows and installed it locally using latest instructions on master. I then updated the takeoff_and_land CMakeLists.txt for local install and built it.

There are a bunch of errors: https://gist.github.com/hamishwillee/eea2b8d9699fda1ff5c09bc79e4554e5

julianoes:
This should fix the linking on Windows. We now properly link the static library itself to ws2_32.lib.

Also, the examples are now built in CI.

Closes #162.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -160,6 +160,11 @@ target_link_libraries(dronecore
     ${curl_lib}
     tinyxml2
 )
+if (MSVC)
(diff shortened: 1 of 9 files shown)


julianoes:
Yay!

PR #181, opened 2017-11-28
genrate_docs: Removed annoying prompts to clean build/install dir

shakthi-prashanth-m:
generate_doc.sh: Prompting repeatedly to clean build & install directory is annoying.
Fix is to let the script do it for us; we just need to re-run the script.

Added options `--clean` & `--help`
If the script is ran without options, it builds & installs the source and generate docs.
When `--clean` is used, it will perform clean build by deleting install folder.
When the script is ran with invalid/accidental args, help message is shown.

### DIFF
--- a/generate_docs.sh
+++ b/generate_docs.sh
@@ -25,8 +25,8 @@ if [ "$skip_checks" = false ]; then
     # We need to do a clean build, otherwise the INSTALL_PREFIX has no effect.
     # Check for leftover install artefacts.
     if [ -e $source_dir/build/ ] || [ -e $install_prefix ] ; then
-		printf "Cleaup your build & install directory using below command.\nmake clean && rm -rf $install_prefix\n"
-		exit 1
+	printf "Cleaup your build & install directory using below command.\nmake clean && rm -rf $install_prefix\n"
+	exit 1
     fi
 fi
 
--- a/generate_docs.sh
+++ b/generate_docs.sh
@@ -23,14 +23,11 @@ install_prefix="$source_dir/install"
 
 if [ "$skip_checks" = false ]; then
     # We need to do a clean build, otherwise the INSTALL_PREFIX has no effect.
-    if [ -e $source_dir/build/ ]; then
-        echo "Build directory ($install_prefix) already exists, you should do 'make clean' first."
-        exit 1
-    fi
-
     # Check for leftover install artefacts.
-    if [ -e $install_prefix ]; then
-        echo "Install directory ($install_prefix) already exists, you should delete it up first."
+    if [ -e $source_dir/build/ ] || [ -e $install_prefix ] ; then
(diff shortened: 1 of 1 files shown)


shakthi-prashanth-m:
OK.. It is interactive by default.
Generally interactive should be on demand. For eg: `rm -i` prompts before every removal.
How about below method ?
`./generate_docs.sh --interactive` => Prompt for cleaning/removing build & install dir.
`./generate_docs.sh` => behaves as `--skip-checks`

julianoes:
If you want to make it interactive, feel free to do so. Then you could add `--skip-checks` for people that know what they are doing and that do not need a clean build.

generate_docs.sh
```diff
@@ -13,22 +13,37 @@ set -e
...
-skip_checks=false
-if [ "$#" -eq 1 -a "$1" == "--skip-checks" ]; then
-    skip_checks=true
+help_msg="
+Usage: ./generate_docs.sh [OPTIONS]
+1. Builds and installs the source
+2. Runs doxygen and create HTML and XML docs
+3. Runs script to generate markdown from XML
+
+  --clean    for clean build and removing install dir.
```
julianoes: This basically changes the default to a non-clean build which means this script will fail for anyone that has not done a clean build beforehand. The reason for this is that it needs a clean build in order for the install directory to change to a local one as required by the docs script.

Therefore, a user that does not know this will run the script without `--clean` and it will fail and they will be unhappy. My vote is to error on the annoying side and have the option `--skip-checks` or `--skip-clean` for anyone that knows what they are doing.

shakthi-prashanth-m: Yes.. right!
We can do like this: 
Prompt user once to cleanup build & install directory (*with commands*) when ran without `--skip-checks`. Updated this in the latest commit.

generate_docs.sh
```diff
@@ -23,15 +23,10 @@ install_prefix="$source_dir/install"
...
-        echo "Build directory ($install_prefix) already exists, you should do 'make clean' first."
-        exit 1
-    fi
-
     # Check for leftover install artefacts.
-    if [ -e $install_prefix ]; then
-        echo "Install directory ($install_prefix) already exists, you should delete it up first."
-        exit 1
+    if [ -e $source_dir/build/ ] || [ -e $install_prefix ] ; then
+	printf "Cleaup your build & install directory using below command.\nmake clean && rm -rf $install_prefix\n"
```
julianoes: `Cleaup` -> `Cleanup`.

generate_docs.sh
```diff
@@ -23,15 +23,10 @@ install_prefix="$source_dir/install"
 
 if [ "$skip_checks" = false ]; then
     # We need to do a clean build, otherwise the INSTALL_PREFIX has no effect.
```
julianoes: Comments out of sync.

julianoes:
Thanks @shakthi-prashanth-m.

Issue #180, opened 2017-11-28
Few useful doxygen tags in the source leads to Doc build error


shakthi-prashanth-m:
Adding doxygen tags `@sa, @note, @para` leads to below error during `./generate_doc.sh`
```
  Generating: /home/intel/karma/fork/follow-me/fresh/DroneCore/install/docs/xml/classdronecore_1_1_follow_me.xml (class xml)
Unhanded kind in simplesect tag in markdown_any_tag:kind: see
Traceback (most recent call last):
  File "/home/intel/karma/fork/follow-me/fresh/DroneCore/generate_markdown_from_doxygen_xml.py", line 968, in <module>
    markdown_string=currentClass.markdown()
  File "/home/intel/karma/fork/follow-me/fresh/DroneCore/generate_markdown_from_doxygen_xml.py", line 871, in markdown
    outputstring+=the_function.markdown()
  File "/home/intel/karma/fork/follow-me/fresh/DroneCore/generate_markdown_from_doxygen_xml.py", line 586, in markdown
    functionstring+='\n\n%s' % markdown_any_tag(self.detaileddescription).strip()
  File "/home/intel/karma/fork/follow-me/fresh/DroneCore/generate_markdown_from_doxygen_xml.py", line 112, in markdown_any_tag
    child_text += markdown_any_tag(child,html=html,para=para) #.strip()
  File "/home/intel/karma/fork/follow-me/fresh/DroneCore/generate_markdown_from_doxygen_xml.py", line 100, in markdown_any_tag
    child_text+=str(ET.tostring(child),'utf-8')
```
@hamishwillee Does it need fix in  `generate_markdown_from_doxygen_xml.py` ?

hamishwillee:
Have you any actual test code I can run this against - ie showing that we need this? the follow-me current tree doesn't appear to use it?

I'm very tempted by @note, possibly by @sa. Why @para though? You get para support automatically just by having a new line.

shakthi-prashanth-m:
@hamishwillee  yes I had the code using tags @note, @sa in the code. As the build failed I removed them. I can upload the code which has these tags usage, if you want to test.

hamishwillee:
Thanks @shakthi-prashanth-m , but I wrote some test examples. You can try out the new version in #182 if you like.

hamishwillee:
This should be fixed now in develop. Give it a whirl :-)

PR #182, opened 2017-11-29
Add docs support for sa and note tags

hamishwillee:
This adds support see also tags (sa) and note tags. The sa tag groups all marked lines as bullet points under a heading **See Also** as requested by #180 

The notes render as a gitbook note (ie `> **Note** Some text`). The problem is that you can only have single line notes - multi-line notes render as separate notes. 

It may be better to render these JUST like see also (under their own heading). Thoughts?

### DIFF
--- a/generate_markdown_from_doxygen_xml.py
+++ b/generate_markdown_from_doxygen_xml.py
@@ -94,6 +94,10 @@ def markdown_any_tag(aTag, html=False,para=True,consume=False):
         if child.tag=='simplesect':
             if child.attrib['kind']=='return':
                 pass # simplesect contains return info for function. We parse it specifically in other code so here we make sure it this tag doesn't get processed.
+            elif child.attrib['kind']=='see':
+                pass #Prevent further handling of see here.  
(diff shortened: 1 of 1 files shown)


generate_markdown_from_doxygen_xml.py
```diff
@@ -116,7 +120,8 @@ def markdown_any_tag(aTag, html=False,para=True,consume=False):
             print('Unhanded tag in markdown_any_tag: %s' % child.tag)
             child_text+=str(ET.tostring(child),'utf-8')
 
-    # He we handle this tag.
+    # Here we handle this tag.
```
julianoes: I don't think it's clear what `this` means here.

hamishwillee: Fixed!

hamishwillee:
Modified so that notes render like see alsos - as bulleted list under a "Notes" heading. This is more robust in terms of the input it can handle. 
Complicated notes could still break this, because the bulleting is markdown, not html.

PR #183, opened 2017-11-29
core: SerialConnection Initialize _should_exit with false

bertbuchholz:
When _should_exit isn't initialized on an ARM platform (Udoo Dual), it defaults to true and the serial connection receive thread exits immediately. Explicitly initializing the variable solves the issue.


julianoes:
@bertbuchholz thanks for the fix. I don't think it "defaults" to anything though. It's just uninitialized and happens to be true.

PR #186, opened 2017-12-01
telemetry: fix missing param description

julianoes:
XCode was complaining.


julianoes:
That was quick, thanks @hamishwillee.

PR #187, opened 2017-12-04
Fix up issues with doc strings and doc string generator tool

hamishwillee:
This adds a few missing strings to follow_me.h and improves the rendering handling of the docs generator script.

### DIFF
--- a/generate_markdown_from_doxygen_xml.py
+++ b/generate_markdown_from_doxygen_xml.py
@@ -206,12 +206,13 @@ def markdown_any_tag(aTag, html=False,para=True,consume=False):
         note_text=''
         if note_tags:
             for item in note_tags:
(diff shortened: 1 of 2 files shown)


generate_markdown_from_doxygen_xml.py
```diff
@@ -239,7 +241,20 @@ def markdown_any_tag(aTag, html=False,para=True,consume=False):
     return tag_text
 
 
-
+def seealso_from_tag(aTag):
```
hamishwillee: FYI, the default position of seealso tags was not the optimal. Using this function we make them part of the class so we can render them wherever we need.

shakthi-prashanth-m:
@hamishwillee, Thanks for the doc-related changes in `follow_me.h`

PR #188, opened 2017-12-04
More flexible plugin architecture

julianoes:
This is a prototype for a more flexible plugin architecture.

The goals of the refactoring are:
- Enabling plugins to be added to a compiled dronecore "core" without having to recompile everything.
- Remove some of the autogenerated (magic) code which is hard to understand.

What changes:
- Instead of including the plugin folders and compiling them in the plugins are added using `add_subdirectory`. Each plugin contains then it's own `add_library` and creates a `dronecore_pluginname.so`.
- The API changes from:
  ```
  device.action().takeoff();
  ```
  to
  ```
  auto action = std::make_shared<Action>(&device);
  action.takeoff();
  ```
  (`action` is now specific to the `device` which is not entirely obvious).

**I would appreciate feedback, tips, opinions.**


CMakeLists.txt
```diff
@@ -195,12 +192,17 @@ install(TARGETS dronecore
 
 install(FILES
     include/dronecore.h
-    include/device.h
-    ${CMAKE_CURRENT_BINARY_DIR}/include/device_plugin_container.h
+    core/device.h
```
hamishwillee: I like this sort of cleanup.

CMakeLists.txt
```diff
@@ -195,12 +192,17 @@ install(TARGETS dronecore
...
     ${plugin_header_paths}
     DESTINATION "include/dronecore"
 )
 
+set(dronecore_install_include_dir "include/dronecore")
+set(dronecore_install_lib_dir ${lib_path})
+
+add_subdirectory(plugins/action)
```
hamishwillee: So this allows you to compile action separately, which is good. But has to already have been compiled before you build the core right?
So in other words if a plugin wasn't known at compile time for the core you can't add it later? WOuldn't it be better for the core not to know about dependency plugins?

hamishwillee:
Hi Julian

1. So is `add_subdirectory`  to add the plugins required to build the core and later use the plugins with it? Better for the core not to have to know what plugins exist. 
  - If not, what does this do for you?

2. So you're creating a new action and providing it a reference to the associated device? It automatically gets destroyed when the `action` goes out of scope? 
  - it is possible for an action to be called on a device that no longer exists?
  - Presumably if you have a swarm of devices to manage you'd have to create them each unique "action" items?
  - I don't really have much to say about this. It seems a little more complicated, but I suspect developers would get used to it as long as it is completely consistent.

CMakeLists.txt
```diff
@@ -146,12 +143,12 @@ add_library(dronecore ${library_type}
...
     core/tcp_connection.cpp
     core/udp_connection.cpp
+    core/plugin_base.cpp
     core/plugin_impl_base.cpp
     core/curl_wrapper.cpp
     core/http_loader.cpp
     core/timeout_handler.cpp
     core/call_every_handler.cpp
-    ${CMAKE_CURRENT_BINARY_DIR}/core/device_plugin_container.cpp
     ${plugin_source_files}
```
JonasVautherin: This can be removed, right?

CMakeLists.txt
```diff
@@ -195,12 +192,17 @@ install(TARGETS dronecore
...
+    core/device.h
     ${plugin_header_paths}
     DESTINATION "include/dronecore"
 )
 
+set(dronecore_install_include_dir "include/dronecore")
+set(dronecore_install_lib_dir ${lib_path})
+
+add_subdirectory(plugins/action)
+add_subdirectory(plugins/telemetry)
```
shakthi-prashanth-m: Are we adding these plugins to show as example or is it required ?

JonasVautherin:
@hamishwillee: Let me try to share my opinion, @julianoes can correct me if necessary :-).
 
> So is add_subdirectory to add the plugins required to build the core and later use the plugins with it? Better for the core not to have to know what plugins exist.

The core does not need the plugins to build, but the plugins need the core. I usually see it like this: you have the root CMakeLists.txt that sets the global environment stuff (some variables, some platform-specific things). Then you have your different components (executables, libraries) in their own folder, each with their own CMakeLists.txt. The root CMakeLists.txt `add_subdirectory` as "let's now build this component based on its own CMakeLists.txt and whatever I defined myself".

In our case, each plugin is becoming a component (a library, to be specific). And core is another library (it should probably have its own CMakeLists.txt, btw). And they are all linked by the `integration_tests_runner` executable (lines 272 and following). The gRPC server will also link them.

Now, what I think makes the whole thing a bit confusing is that the plugins are actually needing some files of the core (e.g. `action` uses includes from `include`, `core` and `libs/include`, that are included in line 96 of the root CMakeLists.txt).

CMakeLists.txt
JonasVautherin: I believe they are in the root CMakeLists.txt just so their libraries get built, and also because they are required by the integration tests. But somebody writing his own executable can choose which library to link (at least `libdronecore.so`, obviously).

avinash-palleti:
@julianoes 
everything looks good to me other than point which @hamishwillee brought up
>1.So is add_subdirectory to add the plugins required to build the core and later use the plugins with it? Better for the core not to have to know what plugins exist.
•If not, what does this do for you?

hamishwillee:
@JonasVautherin That would make perfect sense, except for the fact that the CMakeLists I referred to is the core, depending on the plugins. As per https://github.com/dronecore/DroneCore/pull/188#issuecomment-348895571

CMakeLists.txt
hamishwillee: That would make sense. I'm just making sure :-)

JonasVautherin:
@hamishwillee: I am not sure I understand your last point. There is no "core" CMakeLists.txt (`core/CMakeLists.txt`) right now. There is a "root" one (`CMakeLists.txt` at the root of the repo), and there are CMakeLists.txt for each plugin (`plugins/<plugin-name>/CMakeLists.txt`).

We should move some stuff from the root CMakeLists.txt into a newly-created core one, indeed. But it is perfectly normal to have the root CMakeLists.txt include (`add_subdirectory`) the different components we want to build in the project (here the core library and the plugins). This doesn't show a dependence between them. What does show a weird dependency, though, is the [`include_directories`](https://github.com/dronecore/DroneCore/blob/9d590ec348fc56830f05f8464aba49bdf8b68b44/CMakeLists.txt#L96). If we have files that are required by both core and the plugins, then they should probably go in their own library.

Does that make it clearer on your side, or am I off-topic?

avinash-palleti:
@JonasVautherin Agree. But the concern here is the top `CMakeLists.txt` should be aware of plugins that exists and their directory path should be added. 
In this case still there is dependency on plugins to build core.

JonasVautherin:
@avinash-palleti: Oh, you mean that some `add_subdirectory` lines are missing for some plugins? I thought the other plugins were not finished yet, but I guess they will be added later.

avinash-palleti:
@julianoes one more concern, Have you thought about gRPC ?
I think one should recompile gRPC_Server alone to add or remove plugins with this architecture unlike where user need to recompile everything in current architecture.

avinash-palleti:
> @avinash-palleti: Oh, you mean that some add_subdirectory lines are missing for some plugins? I thought the other plugins were not finished yet, but I guess they will be added later.

@JonasVautherin Had offline discussion with @shakthi-prashanth-m , it is clear now :) Thanks

CMakeLists.txt
shakthi-prashanth-m: OK. As I understand, we invoke CMake build for all the existing plugins in the directory [plugins/](https://github.com/dronecore/DroneCore/tree/develop/plugins). Those that are developed later will be built separately in their respective plugin directory.
Application which wants to use plugins need to link corresponding `dronecore_<plugin>.so` + `libdronecore.a`.

shakthi-prashanth-m:
Even earlier, `action` was specific to `device` isn't it ?
We get `action` from `device` itself => `device().action().takeoff()`. I am not quite clear about this.

CMakeLists.txt
julianoes: @shakthi-prashanth-m correct, but it would probably be all `.so`.

julianoes:
Thanks for the review/comments!

> @hamishwillee:
> 1. So is add_subdirectory to add the plugins required to build the core and later use the plugins with it? Better for the core not to have to know what plugins exist.

You're right. The core (libdronecore.so) library should not know about which plugins are there. However, the integration tests need these libraries and therefore we need to add/specify which plugins to include.

> 2. So you're creating a new action and providing it a reference to the associated device? It automatically gets destroyed when the action goes out of scope?

Correct.

> - it is possible for an action to be called on a device that no longer exists?

Until now, we never delete a device but it is a good point to keep in mind.

> - Presumably if you have a swarm of devices to manage you'd have to create them each unique "action" items?

Unfortunately, yes, that's how it would work. On the upside however, it will be closer to how the gRPC interface will look like.

> - I don't really have much to say about this. It seems a little more complicated, but I suspect developers would get used to it as long as it is completely consistent.

It removes a bunch of the cmake automagic, so it also gets less complicated, at least to add a plugin.

> @JonasVautherin:
> Now, what I think makes the whole thing a bit confusing is that the plugins are actually needing some files of the core (e.g. action uses includes from include, core and libs/include, that are included in line 96 of the root CMakeLists.txt).

Yes, that's ugly. We definitely need to move the core into its own folder and it needs to be its own library. The question is how to get the includes from that library in the plugins since they need it.

hamishwillee:
Thanks for explanations @julianoes . FWIW this is a lot clearer, particularly now you split out core. The client/user side code is not significantly more complicated.

It would be nice if you could still access the plugins from the device. It doesn't matter in our simple examples, but I can imagine an architecture where we have a "management" class for each vehicle in a  swarm, and each would have to manage multiple plugins as well as the device.

hamishwillee:
@julianoes Feels like I am getting a new coveralls scan every time someone adds a comment? Misconfiguration?

julianoes:
> @julianoes Feels like I am getting a new coveralls scan every time someone adds a comment? Misconfiguration?

I'm pretty sure it's every time CI runs. I guess we could disable the comments and just have the status at the bottom.

julianoes:
> It would be nice if you could still access the plugins from the device. It doesn't matter in our simple examples, but I can imagine an architecture where we have a "management" class for each vehicle in a swarm, and each would have to manage multiple plugins as well as the device. I'm probably just over-complicating the problem.

I don't know how we could do this without losing compile team safety.
Basically, we can't have:
```
dc.device().action().takeoff();
```
This is because when we compile the core which includes `Device` we don't know `Action`.

One way to "automatically" connect some of the device state such as `init()` `enable()` etc. to a device would be by using something like:

```
auto telemetry = std::make_shared<Telemetry>();
dc.device().register_plugin(reinterpret_cast<PluginBase *>(telemetry.get()));
```

We could maybe hide the ugly cast and make it:
```
dc.device().register_plugin(telemetry->get_base());`
```

Do you have other suggestions, ideas?

hamishwillee:
So how would you then access a registered Telemetry module? And what happens if someone accesses a module that is not registered?

My feeling is that this sort of approach could be added later if the issues I am envisaging proved to be real problems. 

> Do you have other suggestions, ideas?

Automatic run-time registration of plugins based on presence of config files (or just **.so**) in a specified relative location to the core?

JonasVautherin:
> It would be nice if you could still access the plugins from the device.

Thinking about it, I believe it is very fine (actually even better) the way it is proposed in this PR:

    auto action = std::make_shared<Action>(&device);

For multiple reasons:
* **Dependency inversion:** this way we can test `action` by passing a mock of `device`. Which would be a hell with `dc.device().action().takeoff();`.
* **Out-of-process plugins:** actually also because of dependency inversion :-). With this way of constructing the plugin, we could have a `device` implementation for in-process plugins (the way it is now), but we could also create another `device` implementation that actually sends the data over gRPC from the plugin to the core. Sounds like a nice way to abstract away the "in-process vs out-of-process" question.

Then for a swarm situation, I don't see a problem with this construct (as opposed to the previous one).

> Automatic run-time registration of plugins based on presence of config files (or just .so) in a specified relative location to the core?

Good idea, but I think this does not belong to the core. Could be done in `dronecore_server`, if we go for the "in-process" version. And for customers using the DroneCore libraries only (i.e. `libdronecore.so` and the corresponding `libplugins.so`), it's their problem.

julianoes:
> @JonasVautherin 
> Thinking about it, I believe it is very fine (actually even better) the way it is proposed in this PR:
>
>    `auto action = std::make_shared<Action>(&device);`

I agree and it might also be matching the gRPC interface when you use an `action` message and specifiy the `device` as an argument.

julianoes:
@shakthi-prashanth-m @anitha-suresh I got around to update the remaining integration tests. Hopefully I didn't do any duplicate work.

plugins/action/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@ add_library(dronecore_action SHARED
 
 include_directories(
     ${CMAKE_SOURCE_DIR}/core
-    ${CMAKE_SOURCE_DIR}/libs/include
+    SYSTEM ${CMAKE_SOURCE_DIR}/libs/include
```
JonasVautherin: Just wondering why `SYSTEM` comes between `core` and `libs/include`. Does the order matter here? It doesn't seem so from what I understand in the [documentation](https://cmake.org/cmake/help/v3.0/command/include_directories.html), in which case I would put it before the directories.

It is a detail, but the way it is now seems to me like if it was supposed to be applied only for `libs/include`. So it is just for the sake of double-checking (I have never used `SYSTEM`).

julianoes: I only want to use it for `libs` which I consider "external" libraries. I had to add it because the mavlink headers had build warnings, and I don't want to go in there right now and fix it.

core/CMakeLists.txt
```diff
@@ -0,0 +1,144 @@
...
+if (ANDROID)
+    include_directories(curl-android-ios/prebuilt-with-ssl/android/include)
+endif()
+
+if (IOS)
+    include_directories(curl-android-ios/prebuilt-with-ssl/ios/include)
+endif()
+
+if (IOS OR ANDROID OR MSVC OR APPLE)
+    set(library_type "STATIC")
```
JonasVautherin: @julianoes: I don't know about MSVC, but why not creating shared libraries for the other platforms, as we already do for Linux?

core/CMakeLists.txt
```diff
@@ -0,0 +1,144 @@
...
+
+    # We need tinyxml2 for the camera definition parsing.
+    # We use the submodule when linking statically.
+    add_subdirectory(libs/tinyxml2 EXCLUDE_FROM_ALL)
+    include_directories(SYSTEM libs/tinyxml2)
+else()
+    # Use tinyxml2 from the host system for Linux.
```
JonasVautherin: @julianoes This comment does not apply to the next line, which says that libdronecore will be a shared library, right?

core/CMakeLists.txt
```diff
@@ -0,0 +1,144 @@
...
+if (IOS)
+    include_directories(curl-android-ios/prebuilt-with-ssl/ios/include)
+endif()
+
+if (IOS OR ANDROID OR MSVC OR APPLE)
+    set(library_type "STATIC")
+
+    # We need tinyxml2 for the camera definition parsing.
+    # We use the submodule when linking statically.
+    add_subdirectory(libs/tinyxml2 EXCLUDE_FROM_ALL)
```
JonasVautherin: @julianoes: Why not having `libtinyxml2` statically linked for everybody, even Linux? That would simplify the build system, and I don't think it hurts (anyway it is statically linked for all the others).

core/CMakeLists.txt
julianoes: That's true. What's your suggestion?

core/CMakeLists.txt
julianoes: Right, why not.

core/CMakeLists.txt
JonasVautherin: Removing the comment xD.

hamishwillee:
@julianoes FYI, docs updates will be potentially quite significant. Hope to get onto that next week: https://github.com/dronecore/docs/issues/87

example/follow_me/follow_me.cpp
```diff
@@ -10,11 +10,15 @@
 * @date 2018-01-03
 */
 
+#include <chrono>
+#include <dronecore/action.h>
```
hamishwillee: @julianoes So are plugins now allowed to depend on other plugins?

julianoes: No, they are not. This is the example though which can depend on whatever plugins it requires.

hamishwillee: Thanks. Phew. Thought I was looking at a plugin.

PR #197, opened 2017-12-08
grpc: reorganize grpc

JonasVautherin:
* The protos are now in a separate repo (DroneCore-Proto)
* DroneCore-Proto is a submodule in grpc/server
* gRPC is built automatically from cmake
* python_client disappears from this repo and goes in DroneCore-Python

I tested and could compile/run the server on Ubuntu, Arch Linux and Alpine Linux.


.gitmodules
```diff
@@ -11,3 +11,6 @@
 [submodule "libs/tinyxml2"]
 	path = libs/tinyxml2
 	url = https://github.com/leethomason/tinyxml2.git
+[submodule "grpc/server/proto"]
```
shakthi-prashanth-m: @JonasVautherin So we have now created DroneCore-Proto repo for maintaining `.proto` files. I couldn't figure out why it is done so. We can as well have this in `grpc/server/proto/` isn't it ?

shakthi-prashanth-m: @JonasVautherin 
What I understand is that, we now have
- **DroneCore-Python**:  Repo that has Python-Client samples for DroneCore like `fly_mission.py`, `takeoff_land.py`, etc.

- **DroneCore-Proto**: Repo that contains `.proto` files alone to be used by all the language bindings.

We're going to have in future:
- **DroneCore-Android:** Repo that is going to have Android app samples for DroneCore like `FlyMission` app, etc.
Please feel free to correct me.

JonasVautherin: You're perfectly right! The idea is to split the clients from the backend. And since we need to share the proto files between them, I figured having a repo used as a submodule would be nice.

I am working on DroneCore-Python and DroneCore-Java that I will push very soon to get your feedback :-).

shakthi-prashanth-m: Okay makes sense to me.

grpc/server/cmake/helpers/build_target.cmake
```diff
@@ -0,0 +1,15 @@
+function(build_target SRC_DIR TARGET_DIR)
+    file(MAKE_DIRECTORY ${TARGET_DIR}/build)
+
+    execute_process(
+        COMMAND ${CMAKE_COMMAND}
```
julianoes: I see. So you run cmake out of cmake.

.gitmodules
```diff
@@ -11,3 +11,6 @@
...
+	path = grpc/server/proto
+	url = git@github.com:dronecore/DroneCore-Proto.git
```
julianoes: Did you choose the ssh link here on purpose? Do you know if this works for anyone without github ssh key?

grpc/server/cmake/external/grpc/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
+cmake_minimum_required(VERSION 2.8)
+
+project(external-grpc)
+
+include(ExternalProject)
```
julianoes: Where do external projects end up? Are they rmoeved when you do `make clean`? Do we need something like `make distclean` to remove everything, so we can have `make clean` to only remove dronecore itself?

grpc/server/src/CMakeLists.txt
```diff
@@ -0,0 +1,71 @@
...
+    dronecore_server
+    PUBLIC
+    ${CMAKE_BINARY_DIR}/grpc/server/src
+    ${PLUGINS_DIR}
+)
+
+target_link_libraries(
+    dronecore_server
+    dronecore
+    gRPC::grpc++
```
julianoes: I didn't know libraries could have `::`.

CMakeLists.txt
```diff
@@ -302,29 +301,7 @@ if(NOT IOS AND NOT ANDROID)
...
-            message(STATUS "PROTOBUF_DIR not set, checking automatically for Protobuf")
-            find_package(Protobuf REQUIRED)
-        endif()
-
-        add_subdirectory(grpc/server)
-        add_subdirectory(grpc/python_client)
-    else()
-        message(STATUS "gRPC not found, skipping gRPC build")
-    endif()
+    add_subdirectory(grpc/server)
```
julianoes: Should we add a cmake option, so you need to enable or can disable the gRPC build?

grpc/server/cmake/external/cares/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
+cmake_minimum_required(VERSION 2.8)
+
+project(external-cares)
+
+include(ExternalProject)
+
+ExternalProject_add(
+    cares
+    GIT_REPOSITORY https://github.com/c-ares/c-ares.git
```
julianoes: Why is this needed? Doesn't gRPC already come with c-ares?

.gitmodules
JonasVautherin: Good point :-). Gonna change this.

grpc/server/cmake/helpers/build_target.cmake
JonasVautherin: Yes

grpc/server/src/CMakeLists.txt
JonasVautherin: I learnt that as well :-). That is coming from the targets defined in gRPCConfig.cmake.

grpc/server/cmake/external/grpc/CMakeLists.txt
JonasVautherin: The target is defined a few lines below:

    -DCMAKE_INSTALL_PREFIX=$ENV{DRONECORESERVER_CARES_BINARY_DIR}

So it goes in `build/default/external` in the end. I am not sure what could be a good way to achieve what you suggest, though.

grpc/server/cmake/external/cares/CMakeLists.txt
JonasVautherin: Well, I had to add the options in `grpc/server/cmake/external/grpc/CMakeLists.txt` one by one because it was requesting them for some reason. But honestly I am not completely sure why it could not use the one from the grpc repo =/.

Actually, in the previous version (the current `develop` branch), it was required to install protobuf separately (and to set PROTOBUF_DIR). Not sure why exactly.

CMakeLists.txt
JonasVautherin: Checking this.

grpc/server/cmake/external/cares/CMakeLists.txt
JonasVautherin: Well, I am actually checking this, and I think I needed the "PACKAGE" lines in order to "install" grpc. But I cannot reproduce it correctly.

CMakeLists.txt
avinash-palleti: I agree with @julianoes on this.

grpc/server/CMakeLists.txt
```diff
@@ -1,74 +1,11 @@
-cmake_minimum_required(VERSION 2.8.12)
+cmake_minimum_required(VERSION 2.8)
```
avinash-palleti: I remember we faced issue while testing with older cmake version(not sure how old though), that's why it was kept specific to 2.8.12
Hope you have tested with 2.8.0 so that anything above 2.8 will work.

grpc/server/src/dronecore_server.cpp
```diff
@@ -50,13 +50,15 @@ void RunServer()
...
     std::fstream file;
-    file.open("grpc/proto/plugins.conf");
+    file.open("grpc/server/src/plugins/plugins.conf");
     if (!file) {
         LogErr() << "Error in reading conf file";
         return;
     }
     std::vector<::grpc::Service *> list;
-${PLUGIN_MAP_APPEND_STRING}
+    map["action"] = &createInstances<ActionRPCImpl>;
```
avinash-palleti: Any specific reason to again having this hardcoded plugin names ?

grpc/server/src/plugins/plugins.conf
```diff
@@ -1,3 +1,3 @@
-telemetry
 action
```
avinash-palleti: Is this conf file used somewhere in current implementation ?

grpc/server/CMakeLists.txt
JonasVautherin: I can put it back to 2.8.12.

grpc/server/src/dronecore_server.cpp
JonasVautherin: Right now, the includes are hardcoded anyway, so having both is not consistent to me. I kept the simpler one. But the auto-generation is still in the git history, so we could easily go back to it.

My feeling is that in the end, if we really want plugins, we cannot keep them in-process in dronecore_server. More specifically, if we require the customers to build their own version of dronecore_server in order to use their set of plugins, then it starts to look like a fork (i.e. they will provide their own SDK to their customers). And in this case their are not really "plugins".

So if we want real plugins, they should run in their own process and have a way to register to dronecore_server, other than being detected by cmake and compiled this way. That's probably something we should discuss during the next call :-).

Rjasuja: >Right now, the includes are hardcoded anyway, so having both is not consistent to me.
For includes also, we can have one more run-time variable like PLUGIN_MAP_APPEND_STRING and assign it in autogenerate_plugin_container.cmake.

JonasVautherin: Yes, definitely.

Still, I'd like to discuss that, as I am not completely sure this is what we want to do in the end. To be honest, I am not a big fan of auto-generating source files, as it adds some magic to the build system.

Again, I believe that either we go towards real plugins, in which case the plugins register themselves to the server, or we go towards some kind of drivers, in which case customers have to compile their own SDK. In case we go for the latter, I'd rather have them understand which library they link and how than having this auto-generated.

I would say we keep this discussion for the next WG call, if you don't mind :-).

CMakeLists.txt
JonasVautherin: I added a BUILD_DRONECORESERVER option (see last commit).

grpc/server/cmake/external/grpc/CMakeLists.txt
JonasVautherin: I tried to add `make distclean` and modify `make clean` to keep `build/default/external` (I kept this change in a commit that can be squashed later).

I am not sure where the `default` in `build/default` comes from, though. Moreover, I am not sure if my command is understandable and if it works on all platforms:

    clean:
        @find build/default/ -mindepth 1 -path build/default/external -prune -o -exec rm -rf {} +

grpc/server/cmake/external/cares/CMakeLists.txt
JonasVautherin: So, I figured c-ares has to be compiled as a package because of this message:

> gRPC_INSTALL will be forced to FALSE because gRPC_CARES_PROVIDER is
  "module"

Moreover, if I don't put protobuf as a package, somehow `gRPCTargets.cmake` is not generated.

I want grpc to get installed in `build/default/external/grpc`, so I need them as packages. Maybe there is another way using the in-tree versions and without installing grpc, but in this case I am not sure if the targets would be generated...

The way I see it is that either way, c-ares and grpc are dependencies. Here we have control on the protobuf version we use, and this setup works. If you feel like it is much cleaner to use the in-tree dependencies of grpc, I can try it. But I am not sure it is worth it (especially because the way I do it in this PR is another way, and not a hack).

grpc/server/cmake/external/grpc/CMakeLists.txt
julianoes: `default` was just for the "native" build, so not ios and not android.

grpc/server/src/dronecore_server.cpp
julianoes: I vote against compile time code generation as well. It has created quite a bit of confusion in the past and makes it hard to get into this project.

grpc/server/cmake/external/grpc/CMakeLists.txt
JonasVautherin: Ok so maybe it is fine for now, and we'll have to see how we adapt that when we compile `dronecore_server` for Android and iOS

grpc/server/src/dronecore_server.cpp
JonasVautherin: I could not agree more! To me, code generation is the same as `#ifdef`s. It is **always** worth spending time doing everything we can for not doing it :D.

PR #199, opened 2017-12-13
Improve Follow Me plug-in

shakthi-prashanth-m:
Improved Follow Me API signature as per [this](https://github.com/dronecore/DroneCore/issues/191) issue.

### DIFF
--- a/plugins/follow_me/follow_me.h
+++ b/plugins/follow_me/follow_me.h
@@ -1,8 +1,6 @@
 #pragma once
 
-#include <iostream>
-#include <cmath>
-#include <functional>
+#include <string>
 
 namespace dronecore {
 
--- a/plugins/follow_me/follow_me.h
+++ b/plugins/follow_me/follow_me.h
@@ -41,6 +41,7 @@ public:
         COMMAND_DENIED, /**< @brief Command denied. */
         TIMEOUT, /**< @brief Request timeout. */
         NOT_ACTIVE, /**< @brief FollowMe is not activated. */
+        SET_CONFIG_FAILED, /**< @brief Failed to set FollowMe configuration. */
         UNKNOWN /**< @brief Unknown error. */
     };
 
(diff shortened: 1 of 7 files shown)


plugins/follow_me/follow_me.h
```diff
@@ -107,7 +107,7 @@ class FollowMe
     bool is_active() const;
 
     /**
-     * @brief Geographical location of the target in motion.
+     * @brief Type for Geographical location of the target in motion.
```
hamishwillee: probably better as "Type for geographical location of the target being followed." - the target may not be in motion :-)

shakthi-prashanth-m:
@julianoes you're right.
It seems `auto` is used when the type of expression on right side

- is obvious OR
- need not be known 

In either case, its usage increases readability; if its not, we shouldn't use.
I removed such improper usages from the patch. References: [1](https://stackoverflow.com/questions/6434971/how-much-is-too-much-with-c11-auto-keyword), [2](https://stackoverflow.com/questions/6900459/the-new-keyword-auto-when-should-it-be-used-to-declare-a-variable-type).

hamishwillee:
@julianoes I didn't know anything about where they should be used so I checked  [examples here](https://stackoverflow.com/questions/6900459/the-new-keyword-auto-when-should-it-be-used-to-declare-a-variable-type). The advice seems to be as you would expect - "use if they make it easier to *read and understand the code* - reducing unnecessary verbosity. 

In practise that seems to be where type is obvious (e.g. on return from new), where it doesn't matter, or where it reduces unreadable garbage (expression templates etc.). 

Anyway, @shakthi-prashanth-m  reverted the ones that didn't meet that criteria.

hamishwillee:
Ha ha @shakthi-prashanth-m - we overlapped :-)

plugins/follow_me/follow_me.h
shakthi-prashanth-m: Thanks @hamishwillee . I'll correct.

Issue #193, opened 2017-12-06
Exposing Flight mode APIs in core/device_impl.h


shakthi-prashanth-m:
We have [methods](https://github.com/dronecore/DroneCore/blob/develop/core/device_impl.h#L86-L89) in `DeviceImpl` for PX4 param get/set; but we don't have methods to know/change flight mode in `DeviceImpl`. These methods are useful for plugins. 

Below are the plugins each of which maps to a **PX4 flight mode**:
- [Offboard plugin](https://github.com/dronecore/DroneCore/blob/develop/plugins/offboard/offboard_impl.cpp#L48)
- [FollowMe plugin](https://github.com/dronecore/DroneCore/blob/develop/core/device_impl.h#L86-L89)
- [Mission plugin](https://github.com/dronecore/DroneCore/blob/develop/plugins/mission/mission_impl.cpp#L682)

As of now (in above plugins), we're doing the hard way by using  `MAV_CMD_DO_SET_MODE`. I think it makes sense to expose a method for the same as part of `DeviceImpl` interface. 

At present we've a [public API to get flight mode](https://github.com/dronecore/DroneCore/blob/develop/plugins/telemetry/telemetry.h#L392) is part of `Telemetry `plugin. If other plugins want to know flight mode, they can't call `Telemetry` plugin API (which creates plugin dependency). So it is suitable and intuitive to move some of the useful flight mode APIs from `Telemetry` plugin to `Device` as well.

hamishwillee:
Having this in the `DeviceImpl` for plugins makes sense to me. I would keep the public API in telemetry -
 that is where I would expect to look for it as an API user.keeping the "public" API in Telemetry

julianoes:
I agree that this would be handy.

hamishwillee:
So the answer is Yes. Feel free to close  - and add the functionality/or change this to an issue and keep open.

julianoes:
The answer is yes but I won't get to it just yet.

shakthi-prashanth-m:
#206 addressed it.

PR #206, opened 2017-12-22
Set Flight Mode using new interface set_flight_mode in DeviceImpl

Rjasuja:
Modifies plugins to use `set_flight_mode` instead of using MAV_CMD_DO_SET_MODE
Issue : #193

### DIFF
--- a/core/device_impl.cpp
+++ b/core/device_impl.cpp
@@ -453,29 +453,30 @@ MavlinkCommands::Result DeviceImpl::set_flight_mode(FlightMode device_mode)
     uint8_t custom_sub_mode = 0;
 
     switch (device_mode) {
-        case FlightMode::HOLD :
+        case FlightMode::HOLD:
             custom_sub_mode = px4::PX4_CUSTOM_SUB_MODE_AUTO_LOITER;
             break;
-        case FlightMode::RETURN_TO_LAUNCH :
+        case FlightMode::RETURN_TO_LAUNCH:
             custom_sub_mode = px4::PX4_CUSTOM_SUB_MODE_AUTO_RTL;
             break;
-        case FlightMode::TAKEOFF :
+        case FlightMode::TAKEOFF:
             custom_sub_mode = px4::PX4_CUSTOM_SUB_MODE_AUTO_TAKEOFF;
             break;
-        case FlightMode::LAND :
+        case FlightMode::LAND:
             custom_sub_mode = px4::PX4_CUSTOM_SUB_MODE_AUTO_LAND;
             break;
-        case FlightMode::MISSION :
+        case FlightMode::MISSION:
             custom_sub_mode = px4::PX4_CUSTOM_SUB_MODE_AUTO_MISSION;
             break;
-        case FlightMode::FOLLOW_ME :
+        case FlightMode::FOLLOW_ME:
             custom_sub_mode = px4::PX4_CUSTOM_SUB_MODE_AUTO_FOLLOW_TARGET;
             break;
-        case FlightMode::OFFBOARD :
+        case FlightMode::OFFBOARD:
             custom_mode = px4::PX4_CUSTOM_MAIN_MODE_OFFBOARD;
             break;
         default :
-            return MavlinkCommands::Result::UNSUPPORTED_MODE;
+            LogErr() << "Unknown Flight mode.";
+            return MavlinkCommands::Result::ERROR;
 
     }
 
@@ -490,8 +491,7 @@ MavlinkCommands::Result DeviceImpl::set_flight_mode(FlightMode device_mode)
     return ret;
 }
 
-void DeviceImpl::set_flight_mode_async(FlightMode device_mode,
-                                       command_result_callback_t callback)
+void DeviceImpl::set_flight_mode_async(FlightMode device_mode, command_result_callback_t callback)
 {
     uint8_t flag_safety_armed = is_armed() ? MAV_MODE_FLAG_SAFETY_ARMED : 0;
 
@@ -503,40 +503,43 @@ void DeviceImpl::set_flight_mode_async(FlightMode device_mode,
     uint8_t custom_sub_mode = 0;
 
     switch (device_mode) {
-        case FlightMode::HOLD :
+        case FlightMode::HOLD:
             custom_sub_mode = px4::PX4_CUSTOM_SUB_MODE_AUTO_LOITER;
             break;
-        case FlightMode::RETURN_TO_LAUNCH :
+        case FlightMode::RETURN_TO_LAUNCH:
             custom_sub_mode = px4::PX4_CUSTOM_SUB_MODE_AUTO_RTL;
             break;
-        case FlightMode::TAKEOFF :
+        case FlightMode::TAKEOFF:
             custom_sub_mode = px4::PX4_CUSTOM_SUB_MODE_AUTO_TAKEOFF;
             break;
-        case FlightMode::LAND :
+        case FlightMode::LAND:
             custom_sub_mode = px4::PX4_CUSTOM_SUB_MODE_AUTO_LAND;
             break;
-        case FlightMode::MISSION :
+        case FlightMode::MISSION:
             custom_sub_mode = px4::PX4_CUSTOM_SUB_MODE_AUTO_MISSION;
             break;
-        case FlightMode::FOLLOW_ME :
+        case FlightMode::FOLLOW_ME:
             custom_sub_mode = px4::PX4_CUSTOM_SUB_MODE_AUTO_FOLLOW_TARGET;
             break;
-        case FlightMode::OFFBOARD :
+        case FlightMode::OFFBOARD:
             custom_mode = px4::PX4_CUSTOM_MAIN_MODE_OFFBOARD;
             break;
         default :
+            LogErr() << "Unknown flight mode.";
+            if (callback) {
+                callback(MavlinkCommands::Result::ERROR, NAN);
+            }
             return ;
     }
 
 
-    send_command_with_ack_async(
-        MAV_CMD_DO_SET_MODE,
-        MavlinkCommands::Params {float(mode),
-                                 float(custom_mode),
-                                 float(custom_sub_mode),
-                                 NAN, NAN, NAN, NAN},
-        callback,
-        MavlinkCommands::DEFAULT_COMPONENT_ID_AUTOPILOT);
+    send_command_with_ack_async(MAV_CMD_DO_SET_MODE,
+                                MavlinkCommands::Params {float(mode),
+                                                         float(custom_mode),
+                                                         float(custom_sub_mode),
+                                                         NAN, NAN, NAN, NAN},
+                                callback,
+                                MavlinkCommands::DEFAULT_COMPONENT_ID_AUTOPILOT);
 }
 
 void DeviceImpl::get_param_int_async(const std::string &name,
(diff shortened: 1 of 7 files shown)


core/device_impl.cpp
```diff
@@ -443,6 +444,40 @@ void DeviceImpl::get_param_float_async(const std::string &name,
                                             callback));
 }
 
+MavlinkCommands::Result DeviceImpl::set_flight_mode(uint8_t custom_sub_mode, uint8_t custom_mode)
```
shakthi-prashanth-m: Lets have `DeviceImpl::FlightMode` similar to `Telemetry::FlightMode`. Telemetry flight mode methods shall uses the services of those in `DeviceImpl`. We should have simple signature of `set_flight_mode` as below:
```
MavlinkCommands::Result DeviceImpl::set_flight_mode(DeviceImpl::FlightMode flight_mode);
```

core/device_impl.cpp
```diff
@@ -443,6 +444,40 @@ void DeviceImpl::get_param_float_async(const std::string &name,
...
+{
+    uint8_t flag_safety_armed = DeviceImpl::is_armed() ? MAV_MODE_FLAG_SAFETY_ARMED : 0;
```
shakthi-prashanth-m: `is_armed()` doesn't require explicit qualification, as it's a non-static method of `DeviceImpl`.

core/device_impl.h
```diff
@@ -18,10 +18,19 @@ namespace dronecore {
 
 class DroneCoreImpl;
 
-
 class DeviceImpl
 {
 public:
+    enum FLIGHT_MODE {
```
shakthi-prashanth-m: it should be
```
enum FlightMode {
        HOLD = 0,
        TAKEOFF,
        LAND,
        RETURN_TO_LAUNCH,
        MISSION,
        OFFBOARD,
        FOLLOW_ME,
}
```

core/device_impl.h
```diff
@@ -110,6 +119,8 @@ class DeviceImpl
     DeviceImpl(const DeviceImpl &) = delete;
     const DeviceImpl &operator=(const DeviceImpl &) = delete;
 
+
```
shakthi-prashanth-m: Please remove extra spaces.

plugins/action/action_impl.cpp
```diff
@@ -45,11 +45,8 @@ Action::Result ActionImpl::arm() const
     }
 
     // Go to LOITER mode first.
-    uint8_t custom_mode = px4::PX4_CUSTOM_MAIN_MODE_AUTO;
-    uint8_t custom_sub_mode = px4::PX4_CUSTOM_SUB_MODE_AUTO_LOITER;
-
     ret = action_result_from_command_result(
-              _parent->set_flight_mode(custom_sub_mode, custom_mode));
+              _parent->set_flight_mode(DeviceImpl::HOLD));
```
shakthi-prashanth-m: Flight modes are accessed as `DeviceImpl::FlightMode::HOLD`, etc.

plugins/action/action_impl.cpp
```diff
@@ -121,12 +114,9 @@ Action::Result ActionImpl::return_to_launch() const
 {
     // Note: the safety flag is not needed in future versions of the PX4 Firmware
     //       but want to be rather safe than sorry.
-    uint8_t custom_mode = px4::PX4_CUSTOM_MAIN_MODE_AUTO;
-    uint8_t custom_sub_mode = px4::PX4_CUSTOM_SUB_MODE_AUTO_RTL;
 
     return action_result_from_command_result(_parent->set_flight_mode(
-                                                 custom_sub_mode,
-                                                 custom_mode));
+                                                 DeviceImpl::RETURN_TO_LAUNCH));
```
shakthi-prashanth-m: Flight modes are accessed as `DeviceImpl::FlightMode::RETURN_TO_LAUNCH`, etc.

plugins/action/action_impl.cpp
```diff
@@ -391,22 +379,16 @@ void ActionImpl::process_extended_sys_state(const mavlink_message_t &message)
...
 }
 
 void ActionImpl::loiter_before_arm_async(const Action::result_callback_t &callback)
 {
-    uint8_t custom_mode = px4::PX4_CUSTOM_MAIN_MODE_AUTO;
-    uint8_t custom_sub_mode = px4::PX4_CUSTOM_SUB_MODE_AUTO_LOITER;
-
     _parent->set_flight_mode_async(
-        custom_sub_mode, custom_mode,
+        DeviceImpl::HOLD,
```
shakthi-prashanth-m: We can rename `HOLD` flight mode to `FlightMode::LOITER` to be consistent  with method name; or vice-versa.

plugins/mission/mission_impl.cpp
```diff
@@ -671,13 +671,11 @@ void MissionImpl::start_mission_async(const Mission::result_callback_t &callback
...
     // Note: the safety flag is not needed in future versions of the PX4 Firmware
     //       but want to be rather safe than sorry.
-    uint8_t custom_mode = PX4_CUSTOM_MAIN_MODE_AUTO;
-    uint8_t custom_sub_mode = PX4_CUSTOM_SUB_MODE_AUTO_MISSION;
 
     _activity = Activity::SEND_COMMAND;
 
     _parent->set_flight_mode_async(
-        custom_sub_mode, custom_mode,
+        DeviceImpl::MISSION,
```
shakthi-prashanth-m: `DeviceImpl::MISSION` -> `DeviceImpl::FlightMode::MISSION`

core/device_impl.cpp
```diff
@@ -444,12 +444,35 @@ void DeviceImpl::get_param_float_async(const std::string &name,
...
-    uint8_t flag_safety_armed = DeviceImpl::is_armed() ? MAV_MODE_FLAG_SAFETY_ARMED : 0;
+    uint8_t flag_safety_armed = is_armed() ? MAV_MODE_FLAG_SAFETY_ARMED : 0;
 
     uint8_t mode = MAV_MODE_FLAG_CUSTOM_MODE_ENABLED | flag_safety_armed;
 
+    uint8_t custom_mode = px4::PX4_CUSTOM_MAIN_MODE_AUTO;
+    uint8_t custom_sub_mode = 0;
+
+    switch (device_mode) {
+        case HOLD: custom_sub_mode = px4::PX4_CUSTOM_SUB_MODE_AUTO_LOITER;
```
shakthi-prashanth-m: Please add `custom_sub_mode = px4::PX4_CUSTOM_SUB_MODE_AUTO_LOITER;` next line for uniformity with other case values.

core/device_impl.cpp
```diff
@@ -461,13 +484,36 @@ MavlinkCommands::Result DeviceImpl::set_flight_mode(uint8_t custom_sub_mode, uin
...
+            break;
+        case LAND : custom_sub_mode = px4::PX4_CUSTOM_SUB_MODE_AUTO_LAND;
+            break;
+        case MISSION : custom_sub_mode = px4::PX4_CUSTOM_SUB_MODE_AUTO_MISSION;
+            break;
+        case FOLLOW_ME : custom_sub_mode = px4::PX4_CUSTOM_SUB_MODE_AUTO_FOLLOW_TARGET;
+            break;
+        case OFFBOARD : custom_mode = px4::PX4_CUSTOM_MAIN_MODE_OFFBOARD;
+            break;
+        default : return ;
```
shakthi-prashanth-m: We need to handle for unsupported mode, like it is handled [here](https://github.com/Rjasuja/DroneCore/blob/5e9b9703ece19bcce112a5be068c57442651baa8/core/device_impl.cpp#L582).

plugins/follow_me/follow_me_impl.cpp
```diff
@@ -117,14 +117,9 @@ FollowMe::Result FollowMeImpl::start()
 {
     // Note: the safety flag is not needed in future versions of the PX4 Firmware
```
shakthi-prashanth-m: These comments do not make sense now :) It will be in `set_flight_mode()`.

plugins/follow_me/follow_me_impl.cpp
```diff
@@ -149,13 +144,9 @@ FollowMe::Result FollowMeImpl::stop()
     }
     // Note: the safety flag is not needed in future versions of the PX4 Firmware
```
shakthi-prashanth-m: I think you can remove this comment everywhere it appeared other than in `set_flight_mode()`.

plugins/action/action_impl.cpp
hamishwillee: The public name of the mode in PX4 is "Hold". The names from DroneCore should match what PX4 uses, which tends to also match what QGC uses in selection lists.

core/device_impl.cpp
hamishwillee: Names like loiter I believe may be used by the internal state machine, but are not the public facing names.

shakthi-prashanth-m: OK..

core/device_impl.h
julianoes: And I'll add: it should be:
```
enum class FlightMode {
    ...
}
```

core/device_impl.cpp
```diff
@@ -443,6 +444,86 @@ void DeviceImpl::get_param_float_async(const std::string &name,
                                             callback));
 }
 
+MavlinkCommands::Result DeviceImpl::set_flight_mode(FLIGHT_MODE device_mode)
```
julianoes: I'm not sure if `MavlinkCommands::Result` is the correct result for a flight mode change. I vote that this has its own return type, and then map the `CommandResult` to a `SetFlightMode` result.

shakthi-prashanth-m: @julianoes As flight mode change is merely a MAVLink command operation, its result remains same; except for checking unsupported mode. Plugins should only use supported modes; but we are handling anyway for unsupported mode before sending command further. I think it should be fine. What do you think ?

julianoes: I would vote to just return `MavlinkCommand::ERROR` for the case of an unsupported mode, and add a `LogErr() << "Unsupported mode".

The reason is that any plugin using any kind of command, now needs to catch the command error `UNSUPPORTED_MODE` even for commands that have nothing to do with mode, and that is not intuitive.

julianoes:
@Rjasuja can you try to rebase on top of develop?

core/device_impl.cpp
```diff
@@ -440,6 +441,105 @@ void DeviceImpl::get_param_float_async(const std::string &name,
...
+        case FlightMode::MISSION:
+            custom_sub_mode = px4::PX4_CUSTOM_SUB_MODE_AUTO_MISSION;
+            break;
+        case FlightMode::FOLLOW_ME:
+            custom_sub_mode = px4::PX4_CUSTOM_SUB_MODE_AUTO_FOLLOW_TARGET;
+            break;
+        case FlightMode::OFFBOARD:
+            custom_mode = px4::PX4_CUSTOM_MAIN_MODE_OFFBOARD;
+            break;
+        default :
```
julianoes: In this case it would be helpful to add a `LogErr() << "Unknown Flight mode chosen.";`

julianoes:
@Rjasuja thanks, I only have one minor comment, then please do a rebase. You can do this as follows (assuming origin is dronecore/DroneCore):

```
git checkout upstream3
git fetch origin
git rebase origin/develop
git push --force
```

Rjasuja:
Updated and Rebased

julianoes:
Thanks @Rjasuja!

PR #217, opened 2018-01-05
Add Follow me example

shakthi-prashanth-m:
This adds [Follow Me example](https://github.com/dronecore/DroneCore/issues/189).

- Adds `FakeLocationProvider` class which mocks platform-specific location provider. It uses Boost Asio for sending periodic report of fake location update. [_Note_: Boost is chosen for portability across major platforms]

- Example registers with `FakeLocationProvider` for location updates and sends them to Follow Me plugin.

- Fixes [naming issue](https://github.com/dronecore/DroneCore/issues/204) in Follow Me plug-in implementation.

- Addresses #221 

_Note: Algorithm that generates latitude, longitude to make drone revolve is rudimentary. As this just serves as an example, it should be fine. If there's a better algorithm, please feel free to replace it._

Thanks


example/follow_me/fake_location_provider.cpp
```diff
@@ -0,0 +1,60 @@
+
+#include "fake_location_provider.h"
+
+void FakeLocationProvider::request_location_updates(location_callback_t callback)
+{
+    location_callback_ = callback;
+    timer_.async_wait(boost::bind(&FakeLocationProvider::compute_next_location, this));
```
julianoes: Do you really really need boost for this? I tried to keep everything to plain C++11, if at all possible, I'd vote to do this without boost.

I mean do you actually need to do it async? Or could you just use a loop where you set the location?

julianoes: In my opinion the example doesn't need to show how you could/should build this with async. It should only show how to use the DroneCore API. The simpler the better.

hamishwillee: @julianoes A simple example in this case can show how the API is called, but now how it *should* be used.  I think the way it is now will save readers time and energy. Your call though.

shakthi-prashanth-m: Thanks for the review @julianoes @hamishwillee. As discussed earlier our example should help application developers including Android, Windows, etc; I simulated approximate layout for the name. I tried simulating [Android Location Provider](https://developer.android.com/training/location/receive-location-updates.html#updates): `requestLocationUpdates`.
Even [Apple](https://developer.apple.com/documentation/corelocation/cllocationmanager/1423750-startupdatinglocation)  and [Windows](https://docs.microsoft.com/en-us/uwp/api/Windows.Devices.Geolocation.Geolocator#Examples) use this style. As mentioned earlier, I thought using Boost is portable way. I will try to check whether it can be done without Boost.

shakthi-prashanth-m:
@hamishwillee Sure, I can. I think Min height is most obvious among the Follow Me configuration. What do you think ?

hamishwillee:
> @hamishwillee Sure, I can. I think Min height is most obvious among the Follow Me configuration. What do you think ?

Yes. Perhaps even better, both that and the FollowDirection?

shakthi-prashanth-m:
@hamishwillee I added in the example.

julianoes:
@shakthi-prashanth-m if you rebase this on origin/develop and then force push I can merge it.

shakthi-prashanth-m:
@julianoes Sorry `CMakeLists.txt` & `.gitmodules` changes were accidental. I removed it. Its built successfully and ready to be merged unless you have some comments on the latest commit.

PR #219, opened 2018-01-06
make: always create openssl headers

julianoes:
This uses the build_Android.sh script to call the openssl configure step
without actually re-building all of openssl and curl.

The configure step is needed in order to be able to access the openssl
headers which are symlinked to a `include` folder, and to get the
opensslconf.h headerfile.


julianoes:
Nevermind, found a better way in #220.

PR #228, opened 2018-01-16
Fix issue: https://github.com/dronecore/DroneCore/issues/227

shakthi-prashanth-m:
Fixes Autostart SITL error during launch of integration tests by correcting second argument to [Tools/sitl_run.sh](https://github.com/PX4/Firmware/blob/master/Tools/sitl_run.sh#L66) as part of [start_px4_sitl.sh](https://github.com/dronecore/DroneCore/blob/develop/start_px4_sitl.sh#L53).

### DIFF
--- a/start_px4_sitl.sh
+++ b/start_px4_sitl.sh
@@ -50,7 +50,7 @@ cd $px4_firmware_dir/build/posix_sitl_default/tmp
 
 # And run
 $px4_firmware_dir/Tools/sitl_run.sh \
-    $px4_firmware_dir/build/posix_sitl_default/src/firmware/posix/px4 \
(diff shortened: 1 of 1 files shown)


shakthi-prashanth-m:
How about moving all scripts to a dedicated **scripts** directory ?

julianoes:
Sure, I think this one might go away once we change the integration tests a bit though.

shakthi-prashanth-m:
There are no errors in [travis-ci logs](https://travis-ci.org/dronecore/DroneCore/jobs/329375562)!

julianoes:
> There are no errors in travis-ci error logs!

Right, it's because the Mac builds are queued up. Let's ignore it.

Issue #121, opened 2017-10-19
Example Fly mission uses hard-coding of mission points


shakthi-prashanth-m:
Fly mission example uses[ hard-coding](https://github.com/dronecore/DroneCore/blob/develop/example/fly_mission/fly_mission.cpp#L73) of waypoints. It is hard for someone to compute longitude, latitude, etc and use this example.
I prefer, we can make this example take QGC Mission plan as input; where one can plan missions using QGC and feed its path via command-line to this example.

![image](https://user-images.githubusercontent.com/26615772/31763673-972c5bb6-b4dc-11e7-8ff0-f8b39b6b88c3.png)

We can parse QGC mission plan and pass mission items to DroneCore Mission plugin.
Make sense ?

Regards,
Shakthi

julianoes:
Thanks for the issue. This has been requested several times actually and it would be a good thing to have.

The only catch is that not all mission items are currently supported in DroneCore, so we won't be fully compatible. We would have to just ignore unsupported items and warn the user about it.

hamishwillee:
@julianoes @shakthi-prashanth-m 

> Fly mission example uses hard-coding of waypoints. It is hard for someone to compute longitude, latitude, etc and use this example.

The solution to this problem is NED based mission points - as per #100.
I really think it makes sense to implement this ourselves, sooner rather than later.

> The only catch is that not all mission items are currently supported in DroneCore, so we won't be fully compatible. We would have to just ignore unsupported items and warn the user about it.

My take on this is that mission import should completely bypass whatever we do with the API. It would be something to add flexibility over the top of our existing API. So perhaps:
- Allow upload and download of files *or strings* in generic waypoint format. Users can then dynamically construct the messages if they want.
- Provide an API to send a generic waypoint message or custom message - so the user can send whatever message they like. 

Yes I know it removes some of the "clean-ness" of the API, but if we don't do this then I personally as a user would be forced to write my own plugin to do this.

shakthi-prashanth-m:
#235

hamishwillee:
@shakthi-prashanth-m So what does https://github.com/dronecore/DroneCore/issues/121#issuecomment-364714307 add to this discussion? 

I still think we should implement #100

shakthi-prashanth-m:
> @shakthi-prashanth-m So what does #121 (comment) add to this discussion?

@hamishwillee, sorry which comment ? The link you referred says that **you opened this on Oct 20, 2017**! :-)

[...]

PR #232, opened 2018-01-19
Improve Fly mission example

shakthi-prashanth-m:
This is to improve Fly mission example based on issue #121.
Highlights:

- Adds [json11](https://github.com/dropbox/json11) as a submodule for parsing JSON files

- Example now handles waypoint missions fed from QGC plan (via command line arguments)

**Open question**: I feel the application code has more job to do. Can we provide an interface at Mission plugin ?

### DIFF
--- a/example/fly_mission/fly_mission.cpp
+++ b/example/fly_mission/fly_mission.cpp
@@ -1,20 +1,43 @@
-//
-// Example that demonstrates how add waypoint missions using DroneCore
-//
-// Author: Julian Oes <julian@oes.ch>, Shakthi Prashanth <shakthi.prashanth.m@intel.com>
-//
+/**
+* @file fly_mission.cpp
+*
+* @brief Demonstrates how add & fly waypoint missions from QGC plan using DroneCore.
+* The example is summarised below:
+* 1. Extracts mission items from QGroundControl mission plan (passed via command-line).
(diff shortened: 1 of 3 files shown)


shakthi-prashanth-m:
I strongly feel we can have add public method in Mission plugin which will do everything for the application.

example/fly_mission/fly_mission.cpp
```diff
@@ -255,3 +253,99 @@ std::shared_ptr<MissionItem> add_mission_item(double latitude_deg,
...
+        auto command = mission_item["command"].int_value();
+
+        // Extract parameters
+        std::vector<double> params;
+        for (auto &p : mission_item["params"].array_items()) {
+            params.push_back(p.number_value());
+        }
+
+        // This looks bit complicated at app users.
+        // Should we handle them in plugin ?
```
julianoes: Yes, we should think of an import/export function inside the plugin.

hamishwillee:
So just FYI this has been superseded by #235 , which adds the import/export function from a QGC plan within the module. 
I very much like the addition of the new API, but as documented in that PR, the example no longer shows much of the useful functionality. In either case, propose we close this item and proceed on the other PR?

shakthi-prashanth-m:
@hamishwillee  yes. this PR mostly going to become obsolete when #235 is merged. However, until that PR is merged this one may be used by testing team (#233); hence this is kept open.

PR #235, opened 2018-01-23
Add QGC Mission import capability

shakthi-prashanth-m:
This patch
- [x] Adds new capability to Mission plugin.
Now, missions can be imported from [QGroundControl plan file](https://dev.qgroundcontrol.com/en/file_formats/plan.html). This enables users to create plans on QGroundControl and simply import them into DroneCore applications. We use [json11](https://github.com/dropbox/json11) library for parsing QGC plan files.
- [x] Adds unit test for import QGC mission feature which tests with a sample QGC plan.
- [x] Re-factors Fly mission example
- [x] Adds Fly QGC mission example to demonstrate popular QGC plan.
- [x] Addresses #249
- [X] Adds methods to know whether a mission item has position info set.
- [x] Appveyor fixes for Windows build.
- [X] Adds QGC Mission Importer which currently handles
  - [X] Takeoff, Land
  - [X] Navigation waypoints
  - [X] Start image capture
    - [X] Take a photo
    - [X] Take a photo at intervals
  - [X] Stop Image capture
  - [X] Start Video capture
  - [X] Stop Video capture
  - [X] Control Gimbal pitch & yaw (MAVLink Triggering)
  - [X] Change speed


shakthi-prashanth-m:
So many errors reported in **json11**! We better ignore them.

CMakeLists.txt
```diff
@@ -99,6 +99,7 @@ include(autogenerate_plugin_container.cmake)
 include_directories(
     include
     libs/include
+    libs/json11
```
julianoes: put `libs/json11` to the bottom of this list and put a `SYSTEM` in front of it. This way the warnings are ignored.

plugins/mission/mission.cpp
```diff
@@ -81,4 +95,101 @@ void Mission::subscribe_progress(progress_callback_t callback)
...
+                    lon_deg = params[5];
+                    rel_alt_deg = params[6];
+                    break;
+                case MAV_CMD_NAV_RETURN_TO_LAUNCH:
+                    break;
+                default: // Unsupported mission command
+                    LogErr() << "Mission: Unsupported mission command: " << command;
+                    return Mission::Result::UNSUPPORTED_MISSION_CMD;
+            }
+            std::cout << "Cmd: " << command << " Lat: " << lat_deg << " Lon: " << lon_deg << " Alt: " <<
```
julianoes: Can you use `LogDebug()` please?

plugins/mission/mission.cpp
```diff
@@ -81,4 +95,101 @@ void Mission::subscribe_progress(progress_callback_t callback)
...
+                      rel_alt_deg <<
+                      " Speed " << speed_m_s << " Is fly thru: " << is_fly_through <<
```
julianoes: Does this print ` Is fly through: 1` ? Or `true` and `false`?

You could do: `<< (is_fly_through ? "yes" : "no") <<`

plugins/mission/mission.cpp
```diff
@@ -81,4 +95,101 @@ void Mission::subscribe_progress(progress_callback_t callback)
...
+    file.close();
+
+    std::string err;
+    const auto parsed_plan = Json::parse(ss.str(), err); // parse QGC plan
+    if (!err.empty()) { // Parse error
+        return Mission::Result::FAILED_TO_PARSE_QGC_PLAN;
+    }
+
+    // Compose mission items
+    Mission::Result result = [&items](const Json & mission_json) -> Mission::Result {
```
julianoes: I feel this function is too big. It would be better to split it up into logic that reads the file and a parse function.

shakthi-prashanth-m: I felt same too. But we need to make them static again! Also we need to add them to `mission.h`. C-style is to make them `static` functions. But lets not use C-style. Isn't it ?

shakthi-prashanth-m: Or we can have whole implemention in `MissionImpl` class. And lets invoke them via `Mission` class. Does it make sense ?

plugins/mission/mission.cpp
shakthi-prashanth-m: Yeah, I will. Thanks.

plugins/mission/mission.cpp
shakthi-prashanth-m: Forgot! I will add.

plugins/mission/mission.cpp
julianoes: Yes, move it to `MissionImpl` and invoke in `Mission`.

plugins/mission/mission_item.h
```diff
@@ -87,6 +88,13 @@ class MissionItem
         NONE /**< @brief No action. */
     };
 
+    /**
+     * @brief Converts CamerAction to English strings.
```
julianoes: `CamerAction` -> `CameraAction`.

plugins/mission/mission_impl.cpp
```diff
@@ -882,4 +884,108 @@ void MissionImpl::process_timeout()
...
+            case MAV_CMD_DO_CHANGE_SPEED:
+                speed_m_s = params[1];
+                break;
+            case MAV_CMD_DO_MOUNT_CONTROL:
+                // Possible bug in QGroundcontrol. It stores -ve values for pitch in deg.
+                gimbal_pitch_deg = -params[0];
+                gimbal_yaw_deg = params[2];
+                break;
+            case MAV_CMD_NAV_WAYPOINT:
+                is_fly_through = (params[0] == 0.0) ? true : false;
```
julianoes: no break on purpose?

plugins/mission/mission_impl.cpp
```diff
@@ -882,4 +884,108 @@ void MissionImpl::process_timeout()
...
+                camera_action = MissionItem::CameraAction::START_VIDEO;
+                break;
+            case MAV_CMD_VIDEO_STOP_STREAMING:
+                camera_action = MissionItem::CameraAction::STOP_VIDEO;
+                break;
+            case MAV_CMD_DO_CHANGE_SPEED:
+                speed_m_s = params[1];
+                break;
+            case MAV_CMD_DO_MOUNT_CONTROL:
+                // Possible bug in QGroundcontrol. It stores -ve values for pitch in deg.
```
julianoes: pitch down is `-` because it's all in the NED frame.

plugins/mission/mission_impl.cpp
shakthi-prashanth-m: Yes. Because, param 5,6,7 of both [NAV_WAYPOINT](http://mavlink.org/messages/common#MAV_CMD_NAV_WAYPOINT) & [NAV_TAKEOFF](http://mavlink.org/messages/common#MAV_CMD_NAV_TAKEOFF) provide Lat, Long and Alt respectively.

plugins/mission/mission_impl.cpp
shakthi-prashanth-m: Oh! Thanks @julianoes for the clarification.

plugins/mission/mission_impl.cpp
```diff
@@ -882,4 +884,107 @@ void MissionImpl::process_timeout()
...
+                break;
+            case MAV_CMD_DO_CHANGE_SPEED:
+                speed_m_s = params[1];
+                break;
+            case MAV_CMD_DO_MOUNT_CONTROL:
+                gimbal_pitch_deg = -params[0]; // Pitch value is -ve as its in NED frame.
+                gimbal_yaw_deg = params[2];
+                break;
+            case MAV_CMD_NAV_WAYPOINT:
+                is_fly_through = (params[0] == 0.0) ? true : false;
```
julianoes: Missing break, right?

plugins/mission/mission_impl.cpp
```diff
@@ -882,4 +884,107 @@ void MissionImpl::process_timeout()
...
+        // Maybe removed
+        LogDebug() << "Mission: Cmd: " << command <<
+                   ", Lat: " << lat_deg << " deg, Lon: " << lon_deg << " deg, Rel alt: " << rel_alt_deg <<
+                   " meters, Speed: " << speed_m_s << " m/s, Is fly thru: " << (is_fly_through ? "true" : "false") <<
+                   " Gimbal pitch: " << gimbal_pitch_deg << " deg,  Gimbal yaw: " << gimbal_yaw_deg <<
+                   " deg, Camera action: " << MissionItem::to_str(camera_action) << "\n";
+
+        // Add mission item to the list
+        mission_items.push_back([&]() -> std::shared_ptr<MissionItem> const {
+            std::shared_ptr<MissionItem> new_item(new MissionItem());
```
julianoes: Instead of `std::shared_ptr<MissionItem> new_item(new MissionItem());` you can do `auto new_item = std::make_shared<MissionItem>();`.

plugins/mission/mission_impl.cpp
```diff
@@ -882,4 +884,107 @@ void MissionImpl::process_timeout()
...
+                lon_deg = params[5];
+                rel_alt_deg = params[6];
+                break;
+            case MAV_CMD_NAV_RETURN_TO_LAUNCH:
+                break;
+            default: // Unsupported mission command
+                LogWarn() << "Mission: Unsupported mission command: " << command << ". Ignoring.";
+                break;
+        }
+        // Maybe removed
```
julianoes: Maybe what? 😄

plugins/mission/mission_impl.cpp
shakthi-prashanth-m: :) Like I said [here](https://github.com/dronecore/DroneCore/pull/235#discussion_r163311380), its intentional. Param 5,6,7 of both [NAV_WAYPOINT](http://mavlink.org/messages/common#MAV_CMD_NAV_WAYPOINT) & [NAV_TAKEOFF](http://mavlink.org/messages/common#MAV_CMD_NAV_TAKEOFF) need Lat, Long and Alt respectively, so combined them.

plugins/mission/mission_impl.cpp
shakthi-prashanth-m: I meant that we can remove debug log if its annoying for larger mission. :)

plugins/mission/mission_impl.cpp
shakthi-prashanth-m: @julianoes I agree with you. How about adding Lets add it in [integration_tests/mission_qgc.cpp](https://github.com/dronecore/DroneCore/blob/develop/integration_tests/)

plugins/mission/mission_impl.cpp
julianoes: I understand but it's a funny comment.

plugins/mission/mission_impl.cpp
julianoes: It should be a unit test in my opinion.

shakthi-prashanth-m: OK, you mean as part of [integration_tests/mission.cpp](https://github.com/dronecore/DroneCore/blob/develop/integration_tests/mission.cpp) ?

julianoes: No, not an integration test, a unit test.
https://github.com/dronecore/DroneCore/blob/develop/CMakeLists.txt#L221-L230

shakthi-prashanth-m: Ok! Thanks. I'll add it.

CMakeLists.txt
```diff
@@ -104,6 +104,7 @@ include_directories(
     ${CMAKE_CURRENT_BINARY_DIR}/include
     ${CMAKE_CURRENT_BINARY_DIR}/core
     SYSTEM ${additional_includes}
+    SYSTEM libs/json11
```
JonasVautherin: When #188 gets merged, I guess json11 can become a dependency of the mission plugin only, right?

shakthi-prashanth-m: Yes, right.

shakthi-prashanth-m:
@hamishwillee Thanks for your review comments.
>This does need a unit test.

Yes, next commit will have it.

>This replaces the example showing normal use of the API with use of the QGC plan. The implication is that you don't want to show people how to use the API, you only expect them to use the QGC plan.
IMO you either show both cases or you have a separate example for the qgc plan import.

Good observation. But we've example usage of it in integration tests. So, we don't need to show it in example.
>Personally I would revert to the previous example (but with your excellent documentation header) and either have a standalone example, or possibly just point users to the unit test - that should be enough information.

Thanks. Sure, I would prefer to show **Fly mission** example with _QGC import mission_ feature as it is user-friendly.

>If this is part of the example code, I'd supply a dummy plan. This will make it easier for users to play with.

Plan for example code ? Please share.

hamishwillee:
@shakthi-prashanth-m 

> Good observation. But we've example usage of it in integration tests. So, we don't need to show it in example.

Well you're going to have examples of everything in the integration tests or unit tests, so you might argue why have any examples at all. See below ....

>>Personally I would revert to the previous example (but with your excellent documentation header) and either have a standalone example, or possibly just point users to the unit test - that should be enough information.

> Thanks. Sure, I would prefer to show **Fly mission** example with _QGC import mission_ feature as it is user-friendly.

Example code should be designed to show off the most commonly used parts of the API or the tricky bits of the API. Essentially I think I can very easily work out how to use the QGC import function by inspection, but using the main API is more challenging. 

If you really think that 90% of users will be working off a QGC plan then maybe, but I doubt that is the case. 

I don't mind if you have both :-) 

Maybe @julianoes  can tie break.

>>If this is part of the example code, I'd supply a dummy plan. This will make it easier for users to play with.

> Plan for example code ? Please share.

What I mean is that for this example the instructions would have to say "go create a plan in QGC, download it, then run our example code using this command line ...". I am proposing that you provide a pre-created QGC plan (ie not a "dummy plan", sorry for wording) so that the instructions can be something like "You can test this by using the following command line".

shakthi-prashanth-m:
OK, I understand your point @hamishwillee .
I would like to summarize to next steps for this PR as below:

1. Add unit test for import QGC mission feature which tests with a sample QGC plan.

1. Lets keep the existing [ Fly mission](https://github.com/dronecore/DroneCore/tree/develop/example/fly_mission) example as it demonstrates core Mission APIs (improving documentation header).

1. Add **Fly QGC mission** example to demonstrate popular QGC plan.

@julianoes @hamishwillee please share your opinion.

Thanks

shakthi-prashanth-m:
@julianoes @hamishwillee @JonasVautherin Kindly review. Thanks.

shakthi-prashanth-m:
```
[----------] 1 test from QGCMissionImport
[ RUN      ] QGCMissionImport.ValidateQGCMissonItems
[       OK ] QGCMissionImport.ValidateQGCMissonItems (1 ms)
[----------] 1 test from QGCMissionImport (1 ms total)
```

.gitmodules
```diff
@@ -13,3 +13,6 @@
 [submodule "grpc/server/proto"]
 	path = grpc/server/proto
 	url = https://github.com/dronecore/DroneCore-Proto.git
+[submodule "plugins/mission/libs/json11"]
```
JonasVautherin: Nice to have it inside the plugin :-).

How about putting it in `third_party` instead of `libs`? I started using `third_party` because that's what they use in gRPC and many of its dependencies (maybe that's a Google convention, Idk), and I find it clearer.

shakthi-prashanth-m: >Nice to have it inside the plugin :-)

Yes, I placed it inside mission as it is dependency of the plugin alone.
>How about putting it in third_party instead of libs? I started using third_party because that's what they use in gRPC and many of its dependencies (maybe that's a Google convention, Idk), and I find it clearer.

I see. Yeah AFAIK we have a choice: `external` or `third_party`. I'm not particular about it, but I see that `external` is popular. What do you think ?

shakthi-prashanth-m:
Travis CI says below error while performing Docker build for Ubuntu & Fedora. 
```
This job ran on our Trusty environment,
which has been updated on Tuesday, December 12th, 2017.
Read all about this on our blog and take note that you can
add group: deprecated-2017Q4 in your .travis.yml file to use the previous image version. 
```
Rest are successful. Can we ignore this ?

.gitmodules
hamishwillee: I don't have a strong opinion on this but we should be consistent if possible. If it changes, then let me know because this is used in the docs.

JonasVautherin: I like `third_party`, firstly because there is no possible confusion for me, and also because that's the one I see in grpc, protobuf, boringssl, that are our dependencies already.

JonasVautherin:
Travis fails on Ubuntu and Fedora because of a style error (see last lines of the log). Regarding the update warning, IMO it is not so important for us, so we can ignore it.

shakthi-prashanth-m:
@JonasVautherin  Yeah, its strange; when I ran this locally, there are no such style errors reported.

JonasVautherin:
Does the `./fix_style.sh` work locally? Maybe you have to run it manually :-).

shakthi-prashanth-m:
Yes. I ran `make fix_style` which in-turn invokes `fix_style.sh`. But none reported.

plugins/mission/mission_impl.cpp
```diff
@@ -932,4 +934,115 @@ void MissionImpl::process_timeout()
...
+        LogDebug() << "Mission: Cmd: " << command <<
+                   ", Lat: " << lat_deg << " deg, Lon: " << lon_deg << " deg, Rel alt: " << rel_alt_deg <<
+                   " meters, Speed: " << speed_m_s << " m/s, Is fly thru: " << (is_fly_through ? "true" : "false") <<
+                   " Gimbal pitch: " << gimbal_pitch_deg << " deg,  Gimbal yaw: " << gimbal_yaw_deg <<
+                   " deg, Photo interval: " << photo_interval << ", Camera action: " << MissionItem::to_str(
+                       camera_action) << "\n";
+#endif
+
+        // Add mission item to the list
+        mission_items.push_back([&]() -> std::shared_ptr<MissionItem> const {
```
julianoes: What bugs me a little bit is that the importer does not combine the mavlink mission items together into dronecore mission items.

So, for the optimal case, a location + speed command + gimbal command would optimally end up in the same dronecore mission item. Basically, all commands following a location should be cominbed with the location.

plugins/mission/mission_impl.cpp
```diff
@@ -932,4 +934,115 @@ void MissionImpl::process_timeout()
...
+                speed_m_s = params[1];
+                break;
+            case MAV_CMD_DO_MOUNT_CONTROL:
+                gimbal_pitch_deg = params[0]; // Pitch value is -ve as its in NED frame.
+                gimbal_yaw_deg = params[2];
+                break;
+            case MAV_CMD_NAV_WAYPOINT:
+                is_fly_through = (params[0] == 0.0) ? true : false;
+            // NO break here, because params 4, 5 & 6 of both
+            // of these commands are Lat, Lon & Alt respectively.
```
julianoes: Add `// FALLTHROUGH` for GCC 7.

shakthi-prashanth-m:
@julianoes Yes, it runs as part of `make run_unit_tests`

shakthi-prashanth-m:
@julianoes Need your help in debugging. As discussed, in grouping mission items, I was successful in getting expected number of DroneCore mission items. But However, when I run Fly QGC Mission example with sample QGC plan, I get below error:
```
Found 6 mission items in the given QGC plan.
Uploading mission...
[09:44:56|Debug] Send mission item 0 (mission_impl.cpp:781)
[09:44:56|Debug] Send mission item 1 (mission_impl.cpp:781)
[09:44:56|Debug] Send mission item 2 (mission_impl.cpp:781)
[09:44:56|Debug] Send mission item 3 (mission_impl.cpp:781)
[09:44:56|Debug] Send mission item 4 (mission_impl.cpp:781)
[09:44:56|Debug] Send mission item 5 (mission_impl.cpp:781)
[09:44:57|Debug] Send mission item 6 (mission_impl.cpp:781)
[09:44:57|Debug] Send mission item 7 (mission_impl.cpp:781)
[09:44:57|Debug] Send mission item 8 (mission_impl.cpp:781)
Mission uploaded.
Arming...
[09:44:57|Info ] Mission accepted (mission_impl.cpp:146)
Armed.
Starting mission.
[09:44:57|Debug] MAVLink: info: ARMED by arm/disarm component command (device.cpp:247)
**[09:44:57|Debug] MAVLink: critical: Distance between waypoints too far: 5335120 meter (device.cpp:247)**
[09:44:57|Debug] MAVLink: info: [logger] file: rootfs/fs/microsd/log/2018-02-06/1 (device.cpp:247)
[09:44:57|Warn ] command temporarily rejected (176). (mavlink_commands.cpp:143)
Started mission.
Mission start failed: Error
```

julianoes:
`Distance between waypoints too far: 5335120 meter (device.cpp:247)`

That's far 😄 Is mission item 0 something without position?

shakthi-prashanth-m:
@julianoes I think so.
I think its safe to make this assumption while importing QGC Mission.

- First mission item is `TAKEOFF` command  (When you select to add waypoints, `TAKEOFF` gets selected by default)

shakthi-prashanth-m:
AppVeyor Windows [build succeeded](https://ci.appveyor.com/project/julianoes/dronecore/build/754/job/tpj8n5fc8tt69730#L4952). But, at the end it did [exit with an error code](https://ci.appveyor.com/project/julianoes/dronecore/build/754/job/tpj8n5fc8tt69730#L4963). @julianoes it happened sometime ago too, isn't it ?

julianoes:
I don't understand what happens in AppVeyor :(. Need to try it on my Windows box at some point.

JonasVautherin:
Weird... seems like one of those fails: `cd build\core\Debug && unit_tests_runner.exe`?

shakthi-prashanth-m:
@JonasVautherin Like I mentioned [here](https://github.com/dronecore/DroneCore/pull/235/files#diff-5f4a237561b8048385b3ba09ef74fc96R107), is that correct ?

shakthi-prashanth-m:
@JonasVautherin  Some of the CurlTest tests are failing;

shakthi-prashanth-m:
Re-based with `develop`.

plugins/mission/mission.cpp
```diff
@@ -83,4 +92,10 @@ void Mission::subscribe_progress(progress_callback_t callback)
     _impl->subscribe_progress(callback);
 }
 
+Mission::Result Mission::import_QGC_mission(Mission::mission_items_t &mission_items,
```
shakthi-prashanth-m: I've little concern about this name: `import_QGC_mission()`. This has a mix of snake_case with upper case in it. I choose `QGC` over `qgc` for popularity & being catchy. But I am open to adapt the right way.
- `import_QGC_mission()` (Now)
- `import_qgc_mission()` ?
- `import_qgroundcontrol_mission()` ?
@hamishwillee @julianoes @JonasVautherin

julianoes: `import_qgroundcontrol_mission()` is good, or even `import_mission_from_qgroundcontrol()`

shakthi-prashanth-m: OK. Thanks. I would go with `import_qgroundcontrol_mission()` as it is simple enough.

core/CMakeLists.txt
```diff
@@ -138,7 +141,9 @@ if(NOT IOS AND NOT ANDROID)
         timeout_handler_test.cpp
         call_every_handler_test.cpp
         curl_test.cpp
+        # FIXME: Below line is NOT capturing plugin unit tests sources. So adding manually.
         ${plugin_unittest_source_files}
```
JonasVautherin: Isn't it supposed to be extended by the plugins? So in your case, in `plugins/mission/CMakeLists.txt`, you do something like `list(APPEND plugin_unittest_source_files ${CMAKE_CURRENT_SOURCE_DIR}/mission_import_qgc_test.cpp)`?

@julianoes is it how it is meant to be?

.gitmodules
```diff
@@ -17,3 +17,6 @@
 	path = third_party/zlib
 	url = https://github.com/madler/zlib
 	ignore = dirty
+[submodule "plugins/mission/libs/json11"]
+	path = plugins/mission/libs/json11
```
JonasVautherin: Let's have it in `plugins/mission/third_party/json11` instead, to be consistent with the rest of the project.

core/CMakeLists.txt
```diff
@@ -128,6 +128,9 @@ if(NOT IOS AND NOT ANDROID)
     # SYSTEM because we don't want warnings for gtest headers.
     include_directories(SYSTEM third_party/gtest/googletest/include)
     include_directories(SYSTEM third_party/gtest/googlemock/include)
+    include_directories(SYSTEM libs/gtest/googletest/include)
+    include_directories(SYSTEM libs/gtest/googlemock/include)
```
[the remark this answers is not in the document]
shakthi-prashanth-m: Sorry, I think I missed your latest changes. I will delete these lines.

core/CMakeLists.txt
shakthi-prashanth-m: As per my understanding, `${plugin_unittest_source_files}` should capture unit tests of plugins if we set it in [plugin's CMakeLists.txt](https://github.com/dronecore/DroneCore/pull/235/files#diff-4373271e3de1813cb660c9771ee9ae3cR39). But it was not captured! So I added it manually here.

shakthi-prashanth-m: I see that `set` command handles PARENT_SCOPE and [list handles only current scope](https://cmake.org/cmake/help/v3.1/command/list.html). Probably that is why old plugin template had CMake's `set` command usage, but removed now.
Sadly even `set` is not working in my case.

JonasVautherin: Oh wait, but actually this is now in `core`. I don't think that the unit tests in "core" should include unit tests from the plugins.

Question then is whether we want to split the tests per plugin, or if we want to move the `unit_tests_runner` up into the root CMakeLists and use `${UNIT_TEST_SOURCE_FILES}` instead of `${plugin_unittest_source_files}`?

@shakthi-prashanth-m @julianoes what's your opinion on that? For me it's fine to make one unit test executable in the root CMakeLists.txt, I think.

shakthi-prashanth-m: Yeah, it makes sense to keep a single unit test executable runner in root CMakeLists.txt. In that case, all plugins should append their unit test source path to `${UNIT_TEST_SOURCE_FILES}`. 
My Question: Why does setting a global variable (using `set` with PARENT_SCOPE) doesn't work in CMake. I've been setting mission plugin unit test and expecting it to reflect in root's another CMake. But not working.

JonasVautherin: If you set it in `plugins/.../CMakeLists.txt` with `PARENT_SCOPE`, it will be available in the parent scope, i.e. `plugins/CMakeLists.txt`. Not in the root CMakeLists, and not in `core/CMakeLists.txt`.

So you're right in that in order to do that, we need to propagate `${UNIT_TEST_SOURCE_FILES}` up with `PARENT_SCOPE` at each level, as described in the documentation [here](https://cmake.org/cmake/help/v3.0/command/list.html):

> The list subcommands APPEND, INSERT, REMOVE_AT, REMOVE_ITEM, REMOVE_DUPLICATES, REVERSE and SORT may create new values for the list within the current CMake variable scope. Similar to the SET command, the LIST command creates new variable values in the current scope, even if the list itself is actually defined in a parent scope. To propagate the results of these operations upwards, use SET with PARENT_SCOPE, SET with CACHE INTERNAL, or some other means of value propagation.

I made a toy example [here](https://ptpb.pw/l3nF).

plugins/mission/mission_impl.cpp
```diff
@@ -932,4 +934,135 @@ void MissionImpl::process_timeout()
...
+    const auto parsed_plan = Json::parse(ss.str(), err); // parse QGC plan
+    if (!err.empty()) { // Parse error
+        return Mission::Result::FAILED_TO_PARSE_QGC_PLAN;
+    }
+
+    // Compose mission items
+    return compose_mission_items(mission_items, parsed_plan);
+}
+
+Mission::Result MissionImpl::compose_mission_items(MAV_CMD cmd, std::vector<double> params,
```
julianoes: The name of this method is confusing. When I read the name I don't understand what it does and why it requires these arguments.

plugins/mission/mission_impl.h
```diff
@@ -74,6 +79,14 @@ class MissionImpl : public PluginImplBase
     void download_next_mission_item();
     void assemble_mission_items();
 
+    static Mission::Result
+    compose_mission_items(Mission::mission_items_t &mission_items,
+                          const Json &mission_json);
+    static Mission::Result
+    compose_mission_items(MAV_CMD cmd, std::vector<double> params,
+                          std::shared_ptr<MissionItem> &new_mission_item,
+                          Mission::mission_items_t &mission_items);
```
julianoes: Two methods with different arguments? That's again confusing, I think.

plugins/mission/CMakeLists.txt
```diff
@@ -5,8 +5,24 @@ add_library(dronecore_mission SHARED
...
+include_directories(
+    ${CMAKE_SOURCE_DIR}/core
+    SYSTEM ${CMAKE_SOURCE_DIR}/libs/include
+    SYSTEM libs/json11
+)
+
+# JSON parser library for parsing QGC plan for Mission
+add_subdirectory(libs/json11)
+
+if(ANDROID)
```
julianoes: Some of the things here like this Android hack are no longer required.

plugins/mission/mission_impl.cpp
```diff
@@ -932,4 +934,135 @@ void MissionImpl::process_timeout()
...
+            new_mission_item->set_camera_action(MissionItem::CameraAction::STOP_PHOTO_INTERVAL);
+
+        } else if (cmd == MAV_CMD_VIDEO_START_CAPTURE) {
+            new_mission_item->set_camera_action(MissionItem::CameraAction::START_VIDEO);
+
+        } else if (cmd == MAV_CMD_VIDEO_STOP_CAPTURE) {
+            new_mission_item->set_camera_action(MissionItem::CameraAction::STOP_VIDEO);
+
+        } else if (cmd == MAV_CMD_DO_CHANGE_SPEED) {
+            enum { AirSpeed, GroundSpeed };
```
julianoes: You are using `AirSpeed` and `GroundSpeed` as `int` but don't set the int values which seems dangerous if a compiler does not use the usual 0, 1, 2...

plugins/mission/mission_impl.cpp
shakthi-prashanth-m: First of all, its a static method; hence it doesn't deal with **this** and hence it doesn't modify the state of the mission object.
It is intended to compose mission items into an OUT parameter `mission_items` (last one). It builds a mission item out of the first two parameters it receives. 
So every invocation will build one mission item and add it to mission_items vector. 
Probably I can name this as `compose_each_mission_item()` ?

plugins/mission/mission_impl.cpp
shakthi-prashanth-m: Oh! Ok, Did you mean to do this: `enum { AirSpeed = 0, GroundSpeed = 1 };`
Is that possible ? I am not sure.

plugins/mission/mission_impl.h
shakthi-prashanth-m: Yes, I can change as:
```
static Mission::Result
compose_mission_items(Mission::mission_items_t &mission_items, const Json &mission_json);

static Mission::Result
compose_each_mission_item(Mission::mission_items_t &mission_items, MAV_CMD cmd, std::vector<double> params, std::shared_ptr<MissionItem> &new_mission_item);
```
I could do this in a single method, but it will too big.

plugins/mission/CMakeLists.txt
shakthi-prashanth-m: OK. I will delete them.

plugins/mission/mission_impl.cpp
julianoes: Yes, that's certainly possible.

shakthi-prashanth-m: OK! Thanks.

core/CMakeLists.txt
shakthi-prashanth-m: Thank you @JonasVautherin. Its useful.

shakthi-prashanth-m: @julianoes are we building both Integration & Unit tests in [root CMakeLists.txt](https://github.com/dronecore/DroneCore/blob/develop/CMakeLists.txt#L134) ?

plugins/mission/mission_impl.h
julianoes: No, I think the method name doesn't really explain what it does and the fact that there are two methods with the same name is confusing.

plugins/mission/mission_impl.cpp
julianoes: I think my choice with `compose` was bad because it doesn't reflect what is going on.

> It builds a mission item out of the first two parameters it receives.
> So every invocation will either build one mission item or adds actions to the last mission item and when finally done

So could you split it up into two methods named: `build_a_mission_item(cmd, params)` and `add_to_mission_items(mission_item, all_mission_items)`?

Something like that would be clearer.

core/CMakeLists.txt
julianoes: I think propagating the unit tests to the main one makes sense.

plugins/mission/mission_impl.cpp
shakthi-prashanth-m: I wanted to so. But, `build_a_mission_item()` also involves [adding to mission item list](https://github.com/dronecore/DroneCore/pull/235/files#diff-3a666f956b3fd53eaf1268ec05ea8d29R973). Its like `assemble_mavlink_mission_items()`. In fact, I did refactoring for my case.

shakthi-prashanth-m:
Why do "submodule" in [.gitmodules](https://github.com/dronecore/DroneCore/blob/develop/.gitmodules) still contain "libs" while we did replace with "third-party" ?
@jones

JonasVautherin:
Actually I think it's fine for json11, but it's right that, interestingly, we still have some modules identified with the "libs/..." name (but with the correct "thirdparty/..." path). Not sure why and if it matters. Is it the id of the module?

shakthi-prashanth-m:
Thanks @JonasVautherin for the json11 submodule fix.

julianoes:
@hamishwillee any more comments? Are you ok with this?

plugins/mission/mission_item.h
```diff
@@ -87,6 +89,13 @@ class MissionItem
         NONE /**< @brief No action. */
     };
 
+    /**
+     * @brief Converts #CameraAction to English strings.
```
hamishwillee: Why is it #CameraAction ?

hamishwillee:
@julianoes 

> @hamishwillee any more comments? Are you ok with this?

Lot of change there, but it looks good. 

@shakthi-prashanth-m It would be cool if you could add a QGroundControl screenshot of the planned mission, the executed mission (ie path vehicle took) and the console log. This will enable me to do documentation even if my VM is playing up.

.gitmodules
```diff
@@ -1,19 +1,22 @@
-[submodule "libs/include/mavlink/v2.0"]
+[submodule "c_library_v2"]
```
JonasVautherin: I did not know we could actually rename them here. But in this case, why not "mavlink_v2"?

.gitmodules
```diff
@@ -1,19 +1,22 @@
...
 	url = https://github.com/google/googletest
 [submodule "curl-android-ios"]
 	path = core/third_party/curl-android-ios
 	url = https://github.com/dronecore/curl-android-ios.git
-[submodule "libs/tinyxml2"]
+[submodule "tinyxml2"]
 	path = core/third_party/tinyxml2
 	url = https://github.com/leethomason/tinyxml2.git
-[submodule "grpc/server/proto"]
+[submodule "DronCore-Proto"]
```
JonasVautherin: `dronecore-proto`, to be consistend with the others that are lowercase.

shakthi-prashanth-m:
@hamishwillee Sure, I'll provide screenshot of QGC and console logs of Fly QGC example.

.gitmodules
shakthi-prashanth-m: Yes. Good one, I'll change it.

.gitmodules
shakthi-prashanth-m: OK!

shakthi-prashanth-m:
Screenshot of QGroundControl when Fly QGC Mission example:

flyqgcmission

shakthi-prashanth-m:
Fly QGC Mission example console logs:
```
$ ./fly_qgc_mission 
Usage: ./fly_qgc_mission <path of QGC Mission plan>
Importing mission from Default mission plan: ../qgroundcontrol_sample.plan
Waiting to discover device...
[02:25:09|Info ] New device on: 127.0.0.1:14557 (udp_connection.cpp:211)
[02:25:09|Debug] MAVLink: info: DISARMED by auto disarm on land (device.cpp:247)
[02:25:09|Debug] MAVLink: info: data link #1 lost (device.cpp:247)
[02:25:09|Debug] MAVLink: info: data link #0 regained (device.cpp:247)
[02:25:09|Debug] MAVLink: info: data link #1 regained (device.cpp:247)
[02:25:09|Debug] Discovered 4294967298 (dronecore_impl.cpp:219)
Discovered device with UUID: 4294967298
Waiting for device to be ready
Waiting for device to be ready
Device ready
Found 8 mission items in the given QGC plan.
Uploading mission...
[02:25:11|Debug] Send mission item 0 (mission_impl.cpp:781)
[02:25:11|Debug] Send mission item 1 (mission_impl.cpp:781)
[02:25:11|Debug] Send mission item 2 (mission_impl.cpp:781)
[02:25:11|Debug] Send mission item 3 (mission_impl.cpp:781)
[02:25:11|Debug] Send mission item 4 (mission_impl.cpp:781)
[02:25:11|Debug] Send mission item 5 (mission_impl.cpp:781)
[02:25:11|Debug] Send mission item 6 (mission_impl.cpp:781)
[02:25:11|Debug] Send mission item 7 (mission_impl.cpp:781)
[02:25:11|Debug] Send mission item 8 (mission_impl.cpp:781)
[02:25:11|Debug] Send mission item 9 (mission_impl.cpp:781)
[02:25:11|Debug] Send mission item 10 (mission_impl.cpp:781)
[02:25:11|Debug] Send mission item 11 (mission_impl.cpp:781)
[02:25:12|Debug] Send mission item 12 (mission_impl.cpp:781)
[02:25:12|Debug] Send mission item 13 (mission_impl.cpp:781)
[02:25:12|Debug] Send mission item 14 (mission_impl.cpp:781)
Mission uploaded.
Arming...
[02:25:12|Info ] Mission accepted (mission_impl.cpp:146)
Armed.
Starting mission.
[02:25:12|Debug] MAVLink: info: ARMED by arm/disarm component command (device.cpp:247)
[02:25:12|Debug] MAVLink: info: [logger] file: rootfs/fs/microsd/log/2018-02-15/0 (device.cpp:247)
Started mission.
[02:25:12|Debug] MAVLink: info: Executing mission. (device.cpp:247)
[02:25:12|Debug] MAVLink: info: Takeoff to 15.0 meters above home. (device.cpp:247)
[02:25:12|Debug] MAVLink: info: Takeoff detected (device.cpp:247)
Mission status update: 0 / 8
Mission status update: 1 / 8
[02:25:30|Debug] MAVLink: info: data link #0 lost (device.cpp:247)
Mission status update: 1 / 8
Mission status update: 1 / 8
Mission status update: 1 / 8
Mission status update: 2 / 8
[02:25:32|Debug] MAVLink: info: data link #0 regained (device.cpp:247)
Mission status update: 2 / 8
Mission status update: 2 / 8
Mission status update: 2 / 8
Mission status update: 3 / 8
Mission status update: 3 / 8
Mission status update: 3 / 8
Mission status update: 3 / 8
Mission status update: 4 / 8
Mission status update: 4 / 8
Mission status update: 4 / 8
Mission status update: 4 / 8
Mission status update: 5 / 8
Mission status update: 5 / 8
Mission status update: 5 / 8
Mission status update: 5 / 8
Mission status update: 6 / 8
Mission status update: 6 / 8
Mission status update: 6 / 8
Mission status update: 6 / 8
Mission status update: 7 / 8
Mission status update: 7 / 8
Mission status update: 7 / 8
[02:26:19|Debug] MAVLink: info: Mission finished, loitering. (device.cpp:247)
Mission status update: 8 / 8
Commanding RTL...
Commanded RTL.
```

shakthi-prashanth-m:
```
Build succeeded.
    0 Warning(s)
    0 Error(s)
Time Elapsed 00:00:04.38
cd ..\..\..
cd C:\dronecore
if "%configuration%"=="Debug" ( copy build\third_party\gtest\googlemock\gtest\Debug\gtestd.dll build\core\Debug\ && copy build\third_party\gtest\googlemock\gtest\Debug\gtest_maind.dll build\core\Debug\ && copy build\third_party\gtest\googlemock\Debug\gmockd.dll build\core\Debug\ && cd build\core\Debug && unit_tests_runner.exe ) else ( copy build\third_party\gtest\googlemock\gtest\Release\gtest.dll build\core\Release\ && copy build\third_party\gtest\googlemock\gtest\Release\gtest_main.dll build\core\Release\ && copy build\third_party\gtest\googlemock\Release\gmock.dll build\core\Release\ && cd build\core\Release && unit_tests_runner.exe )
        1 file(s) copied.
        1 file(s) copied.
        1 file(s) copied.
'unit_tests_runner.exe' is not recognized as an internal or external command,
operable program or batch file.
Command exited with code 9009
```
This indicates successful build; but it says:
```
'unit_tests_runner.exe' is not recognized as an internal or external command
```
This is mostly due to PATH settings in Windows. I didn't get this on My Windows machine. Should setting Windows PATH be done in `cmake`  ?

JonasVautherin:
I guess that since the unit tests moved back to the root CMakeLists, the output path ([here](https://github.com/dronecore/DroneCore/blob/develop/appveyor.yml#L121) is not true anymore. Can you check on your machine that this is right and correct it in `appveyor.yml`?

JonasVautherin:
@hamishwillee What's your current opinion on the PR? Are the changes good enough for you, or do you still require changes?

plugins/mission/mission_import_qgc_test.cpp
```diff
@@ -11,6 +11,13 @@
 #include "global_include.h"
 #include "log.h"
 
+// To locate QGroundControl plan file during Unit test.
+#if defined(WIN32) || defined(_WIN32) || defined(__WIN32)
```
julianoes: @shakthi-prashanth-m you can just use `#ifdef WINDOWS` it's what we already have in the build system.

example/fly_qgc_mission/fly_qgc_mission.cpp
```diff
@@ -48,9 +48,8 @@ inline void handle_connection_err_exit(DroneCore::ConnectionResult result,
 
 int main(int argc, char **argv)
 {
-    // Sample QGC Mission plan.
-    const std::string DEFAULT_QGC_MISSION_PLAN = "../qgroundcontrol_sample.plan";
-    std::string qgc_plan(DEFAULT_QGC_MISSION_PLAN);
+    // Locate path of QGC Sample plan
+    std::string qgc_plan = "../../../plugins/mission/qgroundcontrol_sample.plan";
```
julianoes: Eventually, I would duplicate the .plan file but for now that's ok.

shakthi-prashanth-m:
@julianoes now ready to be merged unless @hamishwillee requests for changes.

julianoes:
Thanks @shakthi-prashanth-m for the continuous fixing and improvements 😄.

hamishwillee:
Congratulations on getting this in @shakthi-prashanth-m ! I'll **try* to get associated docs updates in next week: https://github.com/dronecore/docs/issues/99

plugins/mission/mission_item.h
shakthi-prashanth-m: Doxygen tag/command for [linking other members](https://www.stack.nl/~dimitri/doxygen/manual/autolink.html#linkother).

shakthi-prashanth-m:
>Congratulations on getting this in @shakthi-prashanth-m ! I'll *try to get associated docs updates in next week: dronecore/docs#99

Thanks @hamishwillee .
I would like to thank @julianoes and @JonasVautherin for the review and valuable suggestions! :+1:

plugins/mission/mission_item.h
hamishwillee: That would actually link to the constructor according to your docs - which I doubt is what you want. You don't need to mark it up at all to link to CameraAction - Doxygen does that automatically. More often I want to stop it linking :-) (and use `%`)

PR #239, opened 2018-02-01
Enable external plugins again

julianoes:
After the last refactor, external plugins were not included any longer.
However, it is as easy as including the folder where there should be the
next CMakeLists.txt to use.

### DIFF
--- a/external_example/integration_tests/CMakeLists.txt
+++ b/external_example/integration_tests/CMakeLists.txt
@@ -1,25 +1,16 @@
-#set(_libs
-#    dronecore
-#    pthread
-#)
-
-list(APPEND integration_tests
-    hello_world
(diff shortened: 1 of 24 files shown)


external_example/integration_tests/CMakeLists.txt
```diff
@@ -1,25 +1,16 @@
-#set(_libs
-#    dronecore
-#    pthread
-#)
-
-list(APPEND integration_tests
-    hello_world
+add_executable(yuneec_integration_tests_runner
```
hamishwillee: should this be "yuneec" - very vendor specific!

julianoes: Oops, thx.

PR #242, opened 2018-02-05
Fix up plugin constructor docs

hamishwillee:
Constructors are no longer "for internal use". This change also adds support for verbatim tag in the source, which allows us to specify arbitrary markdown (used in this doc change).

### DIFF
--- a/plugins/action/action.h
+++ b/plugins/action/action.h
@@ -25,7 +25,7 @@ class Action : public PluginBase
 {
 public:
     /**
-     * @brief Constructor. Creates the plugin and associates it with a specified Device.
+     * @brief Constructor. Creates the plugin for a specific Device.
      *
      * The plugin is typically created as shown below:
      *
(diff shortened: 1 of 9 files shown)


plugins/action/action.h
```diff
@@ -25,9 +25,17 @@ class Action : public PluginBase
...
+     * @brief Constructor. Creates the plugin and associates it with a specified Device.
      *
-     * @param impl Private internal implementation.
+     * The plugin is typically created as shown below:
+     *
+     *     ```cpp
+     *     auto action = std::make_shared<Action>(&device);
+     *     ```
+     *
+     * The plugin is owned by the device (and will be destroyed along with it).
```
julianoes: I would say: `Each plugin is specific to the device with which it got instantiated.`

I don't think the plugin is owned by the device because the plugin only gets destroyed when all shared pointers are deleted / out of scope.

hamishwillee: Thank you. That is a valid point - I have clearly not spent enough time with shared pointers.

hamishwillee: @julianoes I have updated the docs to remove the "owned/destroyed" stuff and simplify description for all plugins. I think this is enough.

hamishwillee:
@julianoes Updated, please check.

PR #246, opened 2018-02-06
Fix offboard plugin creation in example

hamishwillee:
Pass in offboard plugin instance rather than creating multiple shared pointers.

### DIFF
--- a/example/offboard_velocity/offboard_velocity.cpp
+++ b/example/offboard_velocity/offboard_velocity.cpp
@@ -67,7 +67,7 @@ inline void offboard_log(const std::string &offb_mode, const std::string msg)
  *
  * returns true if everything went well in Offboard control, exits with a log otherwise.
  */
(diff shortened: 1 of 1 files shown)


example/offboard_velocity/offboard_velocity.cpp
```diff
@@ -67,10 +67,8 @@ inline void offboard_log(const std::string &offb_mode, const std::string msg)
  *
  * returns true if everything went well in Offboard control, exits with a log otherwise.
  */
-bool offb_ctrl_ned(Device &device)
+bool offb_ctrl_ned(Offboard* offboard)
```
julianoes: You're passing the smart pointer `std::shared_ptr<Offboard> offboard` here.

hamishwillee: Yoiks.

hamishwillee:
@julianoes Thanks, fixed. Should be good to merge now.

PR #252, opened 2018-02-07
cmake: install tinyxml2 library for Android

julianoes:
This hack is needed because otherwise the tinyxml2 library gets
installed in a library directory that does not respect the ANDROID_ABI.


JonasVautherin:
So that's a temporary fix until we find a way to generate the debug lib without the `*d` at the end?

julianoes:
It's also a temporary fix to respect the ABI for tinyxml2.

PR #253, opened 2018-02-07
cmake: add tinyxml2 install hack for Mac

julianoes:
This is another hack like #252 but for iOS.

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -11,10 +11,8 @@ else()
     set(library_type "SHARED")
 endif()
 
-# We need tinyxml2 for the camera definition parsing.
-add_subdirectory(third_party/tinyxml2 EXCLUDE_FROM_ALL)
-include_directories(SYSTEM third_party/tinyxml2)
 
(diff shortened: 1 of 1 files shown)


core/CMakeLists.txt
```diff
@@ -11,7 +11,24 @@ else()
     set(library_type "SHARED")
 endif()
 
-if(IOS OR MSVC OR APPLE)
+if(IOS OR APPLE)
+    # We need tinyxml2 for the camera definition parsing.
+    add_subdirectory(third_party/tinyxml2 EXCLUDE_FROM_ALL)
```
JonasVautherin: Seems like this is done for all configs, so I'd move it in from of this `if` block. And if you know why we don't `include_directories(SYSTEM third_party/tinyxml2)` for Android, I'd be interested to know :-).

julianoes: You're right, and omitting the `include_directories` for Android is an oversight, thx. I'll fix it.

JonasVautherin:
Oh sorry, I was out of sync, I had not seen your last commit ^^. Let's see if it passes this time!

PR #255, opened 2018-02-08
Fix docker builds

julianoes:
Thanks @hamishwillee for trying. Both docker builds were broken in the meantime.

### DIFF
--- a/docker/Dockerfile-Ubuntu-16.04
+++ b/docker/Dockerfile-Ubuntu-16.04
@@ -24,6 +24,8 @@ RUN apt-get update \
         autoconf \
         automake \
         python \
+        autotools-dev \
+        libltdl-dev \
         python-pip \
         libtinyxml2-dev \
         ruby-dev \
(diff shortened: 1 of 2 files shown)


docker/Dockerfile-Ubuntu-16.04
```diff
@@ -36,35 +38,36 @@ RUN wget http://archive.ubuntu.com/ubuntu/pool/universe/a/astyle/astyle_3.0.1-1u
...
-    && make clean \
-    && ./configure --prefix=/opt/protobuf \
-    && make -j8 \
-    && make install \
-    && cd ../../.. \
-    && rm -rf grpc
-
-ENV GRPC_DIR=/opt/grpc
-ENV PROTOBUF_DIR=/opt/protobuf
+# Everything gRPC below is commented out because we'll build gRPC as part of the main build.
```
JonasVautherin: I'd say we remove the comments :-). IMO the next best solution after building grpc in our build system is to use a package manager.

hamishwillee: Tend to agree.  Change comment above to some unspecific comment about gRPC build being handled separately.

julianoes:
Thx @JonasVautherin

PR #257, opened 2018-02-08
[WIP] initial Jenkinsfile using docker

dagar:
http://ci.px4.io:8080/blue/organizations/jenkins/DroneCore/detail/pr-jenkins/1/pipeline

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,133 @@
+pipeline {
+  agent any
+  stages {
+    stage('Build') {
+      parallel {
+        stage('Ubuntu 16.04 Debug') {
+          agent {
+            docker {
+              image 'dronecore/dronecore-ubuntu-16.04'
+            }
+          }
+          steps {
+            sh 'git submodule deinit -f .'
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,31 @@
+pipeline {
+  agent any
```
JonasVautherin: So we only have linux agents, right? No MacOS from here?

dagar: OSX machines are labelled mac. https://github.com/mavlink/qgroundcontrol/blob/master/Jenkinsfile#L55

Docker slaves are labelled docker.

dagar:
TODO
 - create and archive packages
 - Examples: install package and build

dagar:
Merge and keep going with this?

hamishwillee:
When this is done it would be good if we (you guys) could provide an overview of the CI system for the docs. At the moment we have a little material on testing https://docs.dronecore.io/v/develop/en/contributing/test.html and building in docker: https://docs.dronecore.io/v/develop/en/contributing/build.html#building-in-docker

PR #258, opened 2018-02-09
Add Doxygen tag "\code" to include code block in plugins

shakthi-prashanth-m:
Currently plugin docs use [```cpp](https://github.com/dronecore/DroneCore/blob/develop/plugins/action/action.h#L32) to include code blocks that show plugin class usage.
But Doxygen tag [\code](http://www.stack.nl/~dimitri/doxygen/manual/commands.html#cmdcode) is meant for the same purpose. So, replacing with it.

### DIFF
--- a/generate_markdown_from_doxygen_xml.py
+++ b/generate_markdown_from_doxygen_xml.py
@@ -113,7 +113,7 @@ def markdown_any_tag(aTag, html=False,para=True,consume=False):
             if aTag.tag=='computeroutput':
                 ignore_current_tag=True #Ignore computer output markup in links
             child_text += markdown_any_tag(child,html=html,para=para)
-        elif child.tag=='computeroutput' or child.tag=='para' or child.tag=='listitem' or child.tag=='ulink' or child.tag=='bold' or child.tag=='emphasis' or child.tag=='verbatim':
+        elif child.tag=='computeroutput' or child.tag=='para' or child.tag=='listitem' or child.tag=='ulink' or child.tag=='bold' or child.tag=='emphasis' or child.tag=='verbatim' or child.tag=='programlisting' or child.tag=='codeline' or child.tag=='highlight' or child.tag=='sp':
             child_text += markdown_any_tag(child,html=html,para=para) #.strip()
         elif child.tag=='itemizedlist':
             child_text += markdown_any_tag(child,html=True,para=para)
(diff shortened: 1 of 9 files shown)


hamishwillee:
@shakthi-prashanth-m I've added an update to the script that handles this tag by default. 

That said, I do not "completely" agree with this change. In theory it is a good idea because the `code` tags should result in XML that can be output to any format - ie not tied to markdown in the source.  The problem though is that the output XML is **not generic**. The output of your code might look like this:
```
<para>
<programlisting><codeline><highlight class="keyword">auto</highlight><highlight class="normal"><sp/>follow_me<sp/>=<sp/>std::make_shared&lt;FollowMe&gt;(&amp;device);</highlight></codeline>
</programlisting>
</para>
```
As you can see it contains rendering information and loses the programming language information (cpp), which means that I have to hard code for programming language in the output. Also it is quite possible that a new code block will include addition tags that will break my parser ... because I can't know how it is going to try and render every block.

Given our output format is markdown and that is known, I think we should just use the previous format  and reject this PR.

Left open for your discussion.

shakthi-prashanth-m:
@hamishwillee when I built with docs using `./generate_docs.sh`, `html` rendered it like this, hope you've seen this.
image


After adding `\code` and it resulted as below:
image

**without any changes in doxygen md script!**. Latter looked better to me.
My worry was that it was throwing below error after my changes. 
```
Unhanded tag in markdown_any_tag: programlisting
Traceback (most recent call last):
  File "/home/intel/karma/fork/DroneCore/generate_markdown_from_doxygen_xml.py", line 1074, in <module>
    markdown_string=currentClass.markdown()
```
So I reported as issue to resolve it. Sorry that I am ignorant about XML output significance which you're pointing.

hamishwillee:
Yes, I'm aware of the issues. We still need to make a decision:
1. Use code tags, which work in html and markdown "for now", being aware that the support is possibly a bit flaky and that the language has to be hard coded (and that this cannot be fixed)
2. Use verbatim with markdown in the source, being aware that the HTML will render in markdown syntax.

I leaned towards 2, because HTML is not our output format. If you want to go with 1 then that is fine. We just need to make a decision.

shakthi-prashanth-m:
Ok then i am fine with (2). I shall reject my PR.

PR #260, opened 2018-02-10
Set position (originally was "set altitude")

xgerrmann:
This is a request for a functionality that sets the altitude through offboard control.
It requires a change in the PX4 software.
Add this in commander.cpp around line 4050 at the end of the case vehicle_status_s::NAVIGATION_STATE_OFFBOARD:
```
		control_mode.flag_control_altitude_enabled = true;
		control_mode.flag_control_attitude_enabled = true;
		control_mode.flag_control_rates_enabled = true;
		control_mode.flag_control_velocity_enabled = true;
		control_mode.flag_control_climb_rate_enabled = true;
		control_mode.flag_control_velocity_enabled = true;
```

I would like to get some feedback on this implementation because I do not know which flags I exactly need to set.

### DIFF
--- a/plugins/offboard/offboard.h
+++ b/plugins/offboard/offboard.h
@@ -94,7 +94,15 @@ public:
     };
 
     /**
-     * @brief Start offboard control (synchronous).
+     * @brief Type for position commands in NED coordinates (north, east, down).
+     */
+    struct PositionLocalNED {
+        float x_m; /**< @brief X Position in NED frame in meters.*/
+        float y_m; /**< @brief Y Position in NED frame in meters.*/
+        float z_m; /**< @brief Z Position in NED frame in meters.*/
+    };
+
+    /* @brief Start offboard control (synchronous).
      *
      * **Attention:** this is work in progress, use with caution!
      *
@@ -152,6 +160,13 @@ public:
      * @param velocity_body_yawspeed Velocity and yaw angular rate `struct`.
      */
     void set_velocity_body(VelocityBodyYawspeed velocity_body_yawspeed);
+
+    /**
+     * @brief Set the position in the local NED frame.
+     *
+     * @param position_local_ned Position 'struct'.
+     */
+    void set_position_local_ned(PositionLocalNED position_local_ned);
 
     /**
      * @brief Copy constructor (object is not copyable).
--- a/plugins/offboard/offboard_impl.cpp
+++ b/plugins/offboard/offboard_impl.cpp
@@ -2,6 +2,7 @@
 #include "offboard_impl.h"
 #include "dronecore_impl.h"
 #include "px4_custom_mode.h"
+#include <cmath>
 
 namespace dronecore {
 
@@ -159,6 +160,32 @@ void OffboardImpl::set_velocity_body(Offboard::VelocityBodyYawspeed velocity_bod
     // also send it right now to reduce latency
     send_velocity_body();
 }
+void OffboardImpl::set_position_local_ned(Offboard::PositionLocalNED position_local_ned)
+{
+    _mutex.lock();
+    _position_local_ned = position_local_ned;
+
+    if (_mode != Mode::POSITION_LOCAL_NED) {
+        if (_call_every_cookie) {
+            // If we're already sending other setpoints, stop that now.
+            _parent->remove_call_every(_call_every_cookie);
+            _call_every_cookie = nullptr;
+        }
+        // We automatically send altitude setpoint from now on.
+        _parent->add_call_every(
+            [this]() { send_position_local_ned(); }, SEND_INTERVAL_S, &_call_every_cookie);
+
+        _mode = Mode::POSITION_LOCAL_NED;
+    } else {
+        // We're already sending this kind of setpoints. Since the setpoint change, let's
+        // reschedule the next call, so we don't send setpoints too often.
+        _parent->reset_call_every(_call_every_cookie);
+    }
+    _mutex.unlock();
+
+    // also send it right now to reduce latency
+    send_position_local_ned();
+}
 
 void OffboardImpl::send_velocity_ned()
 {
@@ -252,6 +279,74 @@ void OffboardImpl::send_velocity_body()
         _parent->get_autopilot_id(),
         MAV_FRAME_BODY_NED,
         IGNORE_X | IGNORE_Y | IGNORE_Z | IGNORE_AX | IGNORE_AY | IGNORE_AZ | IGNORE_YAW,
+        x,
+        y,
+        z,
+        vx,
+        vy,
+        vz,
+        afx,
+        afy,
+        afz,
+        yaw,
+        yaw_rate);
+    _parent->send_message(message);
+}
+
+void OffboardImpl::send_position_local_ned()
+{
+    _mutex.lock();
+    const float yaw = 0.0f;
+    const float yaw_rate = 0.0f;
+    float x = (float)_position_local_ned.x_m;
+    float y = (float)_position_local_ned.y_m;
+    float z = (float)_position_local_ned.z_m;
+    const float vx = 0.0f;
+    const float vy = 0.0f;
+    const float vz = 0.0f;
+    const float afx = 0.0f;
+    const float afy = 0.0f;
+    const float afz = 0.0f;
+    _mutex.unlock();
+
+    const uint16_t IGNORE_X = std::isnan(x) << 0;
+    const uint16_t IGNORE_Y = std::isnan(y) << 1;
+    const uint16_t IGNORE_Z = std::isnan(z) << 2;
+    const static uint16_t IGNORE_VX = 1 << 3;
+    const static uint16_t IGNORE_VY = 1 << 4;
+    const static uint16_t IGNORE_VZ = 1 << 5;
+    const static uint16_t IGNORE_AX = 1 << 6;
+    const static uint16_t IGNORE_AY = 1 << 7;
+    const static uint16_t IGNORE_AZ = 1 << 8;
+    const static uint16_t IS_FORCE = 0 << 9; // Setpoint is NOT a force setpoint
+    const static uint16_t IGNORE_YAW = 1 << 10;
+    const static uint16_t IGNORE_YAW_RATE = 1 << 11;
+
+    // Set values to finite values if they are not supplied since PX4 check if the values are finite
+    if (std::isnan(x)) {
+        x = 0.0f;
+    }
+    if (std::isnan(y)) {
+        y = 0.0f;
+    }
+    if (std::isnan(z)) {
+        z = 0.0f;
+    }
+
+    const static uint16_t IGNORE_MASK = IGNORE_X | IGNORE_Y | IGNORE_Z | IGNORE_VX | IGNORE_VY |
+                                        IGNORE_VZ | IGNORE_AX | IGNORE_AY | IGNORE_AZ | IS_FORCE |
+                                        IGNORE_YAW | IGNORE_YAW_RATE;
+
+    mavlink_message_t message;
+    mavlink_msg_set_position_target_local_ned_pack(
+        GCSClient::system_id,
+        GCSClient::component_id,
+        &message,
+        static_cast<uint32_t>(_parent->get_time().elapsed_s() * 1e3),
+        _parent->get_system_id(),
+        _parent->get_autopilot_id(),
+        MAV_FRAME_LOCAL_NED,
+        IGNORE_MASK,
         x,
         y,
         z,
--- a/plugins/offboard/offboard_impl.h
+++ b/plugins/offboard/offboard_impl.h
@@ -29,10 +29,12 @@ public:
 
     void set_velocity_ned(Offboard::VelocityNEDYaw velocity_ned_yaw);
     void set_velocity_body(Offboard::VelocityBodyYawspeed velocity_body_yawspeed);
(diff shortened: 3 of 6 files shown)


integration_tests/offboard_altitude.cpp
```diff
@@ -0,0 +1,72 @@
...
+    float altitude_m = 20.0f;
+    offboard->set_altitude(-altitude_m);
+
+    Offboard::Result offboard_result = offboard->start();
+
+    EXPECT_EQ(offboard_result, Offboard::Result::SUCCESS);
+
+    offboard->set_altitude(-altitude_m);
+    std::this_thread::sleep_for(std::chrono::seconds(5));
+    EXPECT_GT(telemetry->position().relative_altitude_m, altitude_m - 0.25f);
```
julianoes: I would make a `const` or define for `0.25f` so it can easily be changed.

plugins/offboard/offboard.h
```diff
@@ -91,7 +91,12 @@ class Offboard : public PluginBase
         float yawspeed_deg_s; /**< @brief Yaw angular rate in degrees/second (positive for clock-wise
                                    looking from above). */
     };
-
+    
+    /**
+     * @brief Type for altitude commands.
+     */
+    typedef float Altitude;
```
julianoes: Please specify if this is absolute or relative (to home) altitude and the unit.

plugins/offboard/offboard_impl.cpp
```diff
@@ -257,6 +284,55 @@ void OffboardImpl::send_velocity_body()
     _parent->send_message(message);
 }
 
+void OffboardImpl::send_altitude()
+{
+    const static uint16_t IGNORE_X			= true	<< 0;
+    const static uint16_t IGNORE_Y			= true	<< 1;
+    const static uint16_t IGNORE_Z			= false	<< 2; // No ignore for altitude
```
julianoes: I find it weird to shift a boolean. It probably works but `1` is probably more common.

plugins/offboard/offboard.h
```diff
@@ -152,6 +157,13 @@ class Offboard : public PluginBase
      */
     void set_velocity_body(VelocityBodyYawspeed velocity_body_yawspeed);
 
+    /**
+     * @brief Set the altitude.
+     *
+     * @param altitude Altitude 'float'.
+     */
+    void set_altitude(Altitude atlitude);
```
julianoes: I wonder if there shouldn't be just an interface for position which then includes altitude, instead of altitude alone.

plugins/offboard/offboard.h
shakthi-prashanth-m: I don't think this typedef is needed to be consistent with the codebase.
```
float rel_altitude_m;
```
is fine I guess.

plugins/offboard/offboard.h
xgerrmann: I chose it to be separate since sometimes you'd want to only specify an altitude and ignore x and y position. In our specific case, we want to specify an attitude together with an altitude.

plugins/offboard/offboard.h
xgerrmann: Considering these comments I suggest that I remove de type definition in offboard.h; in offboard_impl.h (line 58) I will simply use float _rel_altitude_m. The *.cpp files will then set _rel_altitude_m accordingly.

plugins/offboard/offboard.h
julianoes: Ok, do you want to specify an attitude or a velocity? Because an attitude leads to different velocities due to wind or airframe control imbalances.

xgerrmann: I need to specify an attitude, no velocity. I am aware that this will result in different velocities for different circumstances. I will most likely do this through the set_attitude MAVLINK command.

plugins/offboard/offboard.h
shakthi-prashanth-m: I observed there's no unit in set method, but in get method. For eg [set and get](https://github.com/dronecore/DroneCore/blob/develop/plugins/action/action.h#L250-L271) examples.

plugins/offboard/offboard.h
julianoes: May I ask why you need attitude commands and not velocity commands? DroneCore aims to be high-level and based on autonomous capabilites of drones, away from rate or attitude control.

hamishwillee: >> I wonder if there shouldn't be just an interface for position which then includes altitude, instead of altitude alone.

>   @xgerrmann I chose it to be separate since sometimes you'd want to only specify an altitude and ignore x and y position. In our specific case, we want to specify an attitude together with an altitude.

FWIW I think that we do want a generic position controller - someone is going to want that, and then we'd need to change the API. There is nothing to stop us accepting NAN (or similar) for parameters that should default to current values.

julianoes: @hamishwillee yes I agree. But this here is essentially "altitude control".

xgerrmann: I understand that also a position control is required for dronecore, e.g. offboard->set_position(x,y,z). The set_altitude functionality should not be a replacement for this. I see offboard control a way to do the more 'advanced' things, like the set_altitude while ignoring position x,y. In the MAVLINK specification it is possible to set ignore flags for specific control targets and I think this functionality should be covered.
As I understood from what hamishwillee suggested, we can also make a set_position(x,y,z) where either x,y,z or a combination of those can be set to null (null=ignore). To me this gives the functionality I need without being too specific. Does this sound like a decent solution to you guys @julianoes @hamishwillee ?

julianoes: Yes, I think that's a solution, however, we would use `NaN` instead of null because you might want go to position `0` 😄.

I need to also look into the changes that this requires on the Firmware side to see the implications.

xgerrmann: How is the MAVLINK message SET_POSITION_TARGET_LOCAL_NED interpreted and how is it different from SET_POSITION_TARGET_GLOBAL_INT (which is not yet implemented on the px4 side btw) ? The LOCAL NED, is this always with respect to home, of always with respect to the current position?

julianoes: I don't actually know how and if this is all implemented. I suggest feel free to dig into the offboard control code and find what is happening and what is missing. That would certainly be a great!

hamishwillee: > `set_position(x,y,z) where either x,y,z or a combination of these can be set to null (null=ignore). To me this gives the functionality I need without being too specific. Does this sound like a decent solution to you guys @julianoes @hamishwillee ?

Yes, my suggestion is  `set_position(x,y,z)` with a `NaN` for ignored values (ie "use current"). 

> How is the MAVLINK message SET_POSITION_TARGET_LOCAL_NED interpreted and how is it different from SET_POSITION_TARGET_GLOBAL_INT (which is not yet implemented on the px4 side btw) ? The LOCAL NED, is this always with respect to home, of always with respect to the current position?

My recollection is that SET_POSITION_TARGET_GLOBAL_INT is global GPS co-ordinates, [SET_POSITION_TARGET_LOCAL_NED](http://mavlink.org/messages/common#SET_POSITION_TARGET_LOCAL_NED) is local co-cordinates where you specify the frame- e.g.  MAV_FRAME_LOCAL_NED is relative to home location. There are a number of different frames in the standard but I don't know which ones PX4 supports.

hamishwillee: A further complication is that the global position altitude can be relative to MSL or to home.  I'd make the API relative to home. 
Don't know what the NED API should look like yet - depends a bit on what frames PX4 supports.

xgerrmann: > Yes, my suggestion is set_position(x,y,z) with a NaN for ignored values (ie "use current").
So this is a crucial point. Here I would really ignore x and y and set the corresponding ignore_position flag, since the user does not specify them and apparently does not care.

julianoes: @xgerrmann yes if we set `NaN` in the DroneCore API, then it should set the ignore flags to true for these fields. 

Can you make a PR for the change needed on the Firmware side? If you haven't already.

xgerrmann: Update: I will continue on this PR soon (1-2 days).

plugins/offboard/offboard_impl.cpp
```diff
@@ -257,6 +284,61 @@ void OffboardImpl::send_velocity_body()
...
+    const        uint16_t IGNORE_Z			= std::isnan(z) << 2;
+    const static uint16_t IGNORE_VX			= 1 << 3;
+    const static uint16_t IGNORE_VY			= 1 << 4;
+    const static uint16_t IGNORE_VZ			= 1 << 5;
+    const static uint16_t IGNORE_AX			= 1 << 6;
+    const static uint16_t IGNORE_AY			= 1 << 7;
+    const static uint16_t IGNORE_AZ			= 1 << 8;
+    const static uint16_t IS_FORCE			= 0 << 9; // Setpoint is NOT a force setpoint
+    const static uint16_t IGNORE_YAW		= 1 << 10;
+    const static uint16_t IGNORE_YAW_RATE	= 1 << 11;
```
julianoes: This is all very clean, thank you! A nitpick is the tabs, I vote to change it to spaces.

plugins/offboard/offboard_impl.h
```diff
@@ -5,6 +5,7 @@
 #include "device.h"
 #include "offboard.h"
 #include <mutex>
+#include <cmath>
```
julianoes: Please include this one in the cpp file since it's not required in the header.

xgerrmann:
I have now created a PR on the PX4 side: https://github.com/PX4/Firmware/pull/8916
Also, I have implemented the discussed set_position functionality. @julianoes, @hamishwillee  can you please check this out and let me know if it needs any modifications.

julianoes:
@xgerrmann thanks for the Firmware change, I'll review it when I get to it.
Please make sure to rebase this PR at some point.

xgerrmann:
Is this rebase now correct? I rebased on my fork xgerrmann/develop, which I updated by merging DroneCore/develop into my fork.

julianoes:
@xgerrmann not sure, you'd have to rebase on top of upstream/develop. You can also just do a merge but then we have to squash it when merging the PR.

plugins/offboard/offboard.h
```diff
@@ -152,6 +161,13 @@ class Offboard : public PluginBase
      */
     void set_velocity_body(VelocityBodyYawspeed velocity_body_yawspeed);
 
+    /**
+     * @brief Set the position in the local NED frame.
```
hamishwillee: Perhaps "Set the target position in the local NED frame."

plugins/offboard/offboard.h
```diff
@@ -152,6 +161,13 @@ class Offboard : public PluginBase
...
+     *
+     * @param position_local_ned Position 'struct'.
```
hamishwillee: Perhaps "Target position.". The fact that is a struct is kind of irrelevant.

hamishwillee:
Hard to review with all the changes.

Added minor comments to set_position_local_ned() docs - looks good. Do we need/should we have set_position_global()?

hamishwillee:
Also, did I miss an example usage? (I did see a pretty thorough integration test).

julianoes:
I think the rebase was ok but now it's required yet again. :(

hamishwillee:
@xgerrmann Did my comments make sense? Let me know if clarification required on anything!

julianoes:
@xgerrmann I would like to get this in but I think we also need to get https://github.com/PX4/Firmware/pull/8916 in, right?

julianoes:
@xgerrmann what is the status here?

xgerrmann:
I Need to inspect why the offboard attitude control test fails.  I cannot say when I will be able to look into this PR, but I can say that eventually we do need it. Please let the PR dangle for now, I will inspect it at a later time.

xgerrmann:
For position control, https://github.com/PX4/Firmware/pull/8916 is not required. Only when you want to use altitude control. This means that right now the position integration test succeeds and the Altitude test fails.

xgerrmann:
Ok, I meassed up. let me fix this.

plugins/offboard/offboard.h
```diff
@@ -94,7 +94,15 @@ class Offboard : public PluginBase {
...
 
     /**
-     * @brief Start offboard control (synchronous).
+     * @brief Type for position commands in NED coordinates (north, east, down).
+     */
+    struct PositionLocalNED {
+        float x_m; /**< @brief X Position in NED frame in meters.*/
+        float y_m; /**< @brief Y Position in NED frame in meters.*/
+        float z_m; /**< @brief Z Position in NED frame in meters.*/
+    };
```
Stifael: I suggest to reuse the convention naming for NED (which is already used by VelocityLocalNED)
north_m, east_m, down_m

Stifael:
I am working on something similar, but should not conflict with this PR.
However, I will name the individual setpoint component north_m, east_m and down_m. Same for velocity

Stifael:
similar implementation [here](https://github.com/dronecore/DroneCore/pull/434) but with additional `yaw` and a few other setpoint combination

xgerrmann:
The related pull request for PX4 has been closed by me, since that did not seem to be a proper implementation after all. Therefore I will close this PR as well. I am looking forward to seeing improved offboard support.

julianoes:
@xgerrmann we might still take some parts of this PR I think. I'll leave that to @Stifael. Thanks for following up.

PR #261, opened 2018-02-11
Cross compile dronecore_server for Android and iOS

### LINKED ISSUE #212: Cross-compile the backend for Android
We will need to run the full backend on Android, so it should be cross-compiled.

### LINKED ISSUE #213: Cross-compile the backend for iOS
We will need the full backend libraries compiled for iOS.

JonasVautherin:
Compile `dronecore_server` and its dependencies on Android and iOS (when `BUILD_DRONECORESERVER` is set).

It has been tested to run on Android `armeabi-v7a`, though it had some warnings in Debug mode. There was a segmentation fault with `armeabi`, which we may or may not want to address depending on the devices we target right now.

Some work now needs to be done on `dronecore_server` itself before it can be tested on an iOS app (I will work on that from tomorrow on).

Fixes #212, fixes #213


grpc/server/cmake/helpers/build_target.cmake
```diff
@@ -5,8 +5,17 @@ function(build_target SRC_DIR TARGET_DIR)
         COMMAND ${CMAKE_COMMAND}
             -G${CMAKE_GENERATOR}
             -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
+            "-DCMAKE_FIND_ROOT_PATH=${CMAKE_FIND_ROOT_PATH}"
```
julianoes: The `"` are a bit ugly but I guess that's required?

grpc/server/src/CMakeLists.txt
```diff
@@ -2,35 +2,42 @@ cmake_minimum_required(VERSION 2.8)
 
 file(STRINGS ${PLUGINS_DIR}/plugins.conf PLUGINS_LIST)
 
+set(PROTOC_BINARY ${CMAKE_BINARY_DIR}/../default/third_party/protobuf/bin/protoc)
+set(GRPC_CPP_PLUGIN_BINARY ${CMAKE_BINARY_DIR}/../default/third_party/grpc/bin/grpc_cpp_plugin)
+
+if(NOT EXISTS ${PROTOC_BINARY} OR NOT EXISTS ${GRPC_CPP_PLUGIN_BINARY})
+    message(FATAL_ERROR "Could not find 'protoc' or 'grpc_cpp_plugin' in the 'default' build folder. Please build for your host first (`make BUILD_DRONECORESERVER=YES default`).")
```
julianoes: Good idea to warn like that.

plugins/mission/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-add_library(dronecore_mission SHARED
+add_library(dronecore_mission ${PLUGIN_LIBRARY_TYPE}
```
julianoes: The other plugins will need that too, not just the three we have right now.

JonasVautherin: Good point, I'll change that.

grpc/server/cmake/helpers/build_target.cmake
JonasVautherin: Yes, unfortunately it is, for lists :-). I just tried to rebuild without the quotes and it fails.

plugins/mission/CMakeLists.txt
JonasVautherin: Actually I'll merge like this and will change that when adding those plugins to `dronecore_server`.

julianoes: Ok, sure!

Issue #238, opened 2018-01-26
Update dronecore_server to use newly-defined action protocol


JonasVautherin:
Update the proto submodule to commit `2af63b4b2e867d80d5054a40723487518e757414` and update dronecore_server.cpp accordingly.

PR #266, opened 2018-02-15
Refactor dronecore server

JonasVautherin:
* some cleanup on the proto generation (build_system)
* use references and smart pointers instead of old-style pointers
* make a library out of `dronecore_server` (necessary for running from iOS)
* rename `dronecore_server` into `backend`

A PR actually updating the proto files will come just after this one is merged.

Connects to #238.

### DIFF
--- a/core/device.cpp
+++ b/core/device.cpp
@@ -384,7 +384,7 @@ void Device::set_disconnected()
         _parent->notify_on_timeout(_target_uuid);
 
         // Let's reset the flag hope again for the next time we see this target.
-        _target_uuid_initialized = 0;
+        _target_uuid_initialized = false;
     }
 
     {
--- a/grpc/server/src/CMakeLists.txt
+++ b/grpc/server/src/CMakeLists.txt
@@ -1,67 +1,33 @@
 cmake_minimum_required(VERSION 2.8)
 
-file(STRINGS ${PLUGINS_DIR}/plugins.conf PLUGINS_LIST)
+set(COMPONENTS_LIST core action mission telemetry)
 
-set(PROTOC_BINARY ${CMAKE_BINARY_DIR}/../default/third_party/protobuf/bin/protoc)
-set(GRPC_CPP_PLUGIN_BINARY ${CMAKE_BINARY_DIR}/../default/third_party/grpc/bin/grpc_cpp_plugin)
+include(cmake/compile_proto.cmake)
 
-if(NOT EXISTS ${PROTOC_BINARY} OR NOT EXISTS ${GRPC_CPP_PLUGIN_BINARY})
-    message(FATAL_ERROR "Could not find 'protoc' or 'grpc_cpp_plugin' in the 'default' build folder. Please build for your host first (`make BUILD_DRONECORESERVER=YES default`).")
-endif()
+foreach(COMPONENT_NAME ${COMPONENTS_LIST})
+    compile_proto_pb(${COMPONENT_NAME} PB_COMPILED_SOURCE)
+    list(APPEND PB_COMPILED_SOURCES ${PB_COMPILED_SOURCE})
 
-add_custom_command(OUTPUT core/core.grpc.pb.cc
-    COMMAND ${PROTOC_BINARY}
-        -I ${PROTO_DIR}
-        --grpc_out=.
-        --plugin=protoc-gen-grpc=${GRPC_CPP_PLUGIN_BINARY}
-        ${PROTO_DIR}/core/core.proto;
-)
-
-add_custom_command(OUTPUT core/core.pb.cc
-    COMMAND ${PROTOC_BINARY}
-        -I ${PROTO_DIR}
-        --cpp_out=.
-        ${PROTO_DIR}/core/core.proto
-)
-
-foreach(PLUGIN ${PLUGINS_LIST})
-    add_custom_command(OUTPUT ${PLUGIN}/${PLUGIN}.grpc.pb.cc
-        COMMAND ${PROTOC_BINARY}
-            -I ${PROTO_DIR}
-            --grpc_out=.
-            --plugin=protoc-gen-grpc=${GRPC_CPP_PLUGIN_BINARY}
-            --cpp_out=.
-            ${PROTO_DIR}/${PLUGIN}/${PLUGIN}.proto
-    )
+    compile_proto_grpc(${COMPONENT_NAME} GRPC_COMPILED_SOURCE)
+    list(APPEND GRPC_COMPILED_SOURCES ${GRPC_COMPILED_SOURCE})
 endforeach()
 
-foreach(PLUGIN ${PLUGINS_LIST})
-    add_custom_command(OUTPUT ${PLUGIN}/${PLUGIN}.pb.cc
-        COMMAND ${PROTOC_BINARY}
-            -I ${PROTO_DIR}
-            --cpp_out=.
-            ${PROTO_DIR}/${PLUGIN}/${PLUGIN}.proto
-    )
-endforeach()
-
-set(PLUGINS_SRC dronecore_server.cpp core/core.pb.cc core/core.grpc.pb.cc)
-foreach(PLUGIN ${PLUGINS_LIST})
-    list(APPEND PLUGINS_SRC ${PLUGIN}/${PLUGIN}.pb.cc)
-    list(APPEND PLUGINS_SRC ${PLUGIN}/${PLUGIN}.grpc.pb.cc)
-endforeach()
-
-add_executable(dronecore_server ${PLUGINS_SRC})
-
-target_compile_options(
-    dronecore_server
-    PUBLIC
-    -Wno-unused-parameter
-    -Wno-shadow
+add_library(backend
+    backend.cpp
+    ${GRPC_COMPILED_SOURCES}
+    ${PB_COMPILED_SOURCES}
 )
 
-target_include_directories(
-    dronecore_server
-    PUBLIC
+target_link_libraries(backend
+    dronecore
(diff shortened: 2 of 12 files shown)


core/device.cpp
```diff
@@ -384,7 +384,7 @@ void Device::set_disconnected()
         _parent->notify_on_timeout(_target_uuid);
 
         // Let's reset the flag hope again for the next time we see this target.
-        _target_uuid_initialized = 0;
+        _target_uuid_initialized = false;
```
julianoes: Better.

grpc/server/src/backend.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+void DroneCoreBackend::waitForDiscovery()
+{
+    LogInfo() << "Waiting to discover device...";
+    auto discoveryFuture = wrapped_register_on_discover();
+    discoveryFuture.wait();
+    LogInfo() << "Device discovered [UUID: " << discoveryFuture.get() << "]";
+}
+
+std::future<uint64_t> DroneCoreBackend::wrapped_register_on_discover()
```
julianoes: I think we should either go for `snake_case` or `camelCase`, or is the mix intentional?

grpc/server/src/backend.cpp
```diff
@@ -0,0 +1,104 @@
...
+    dc.register_on_timeout([](uint64_t uuid) {
+        LogInfo() << "Device timed out [UUID: " << uuid << "]";
+    });
+}
+
+void DroneCoreBackend::waitForDiscovery()
+{
+    LogInfo() << "Waiting to discover device...";
+    auto discoveryFuture = wrapped_register_on_discover();
+    discoveryFuture.wait();
```
julianoes: The problem here is that `std::promise` in C++11 does not support triggering the callback multiple times. Therefore, this will crash when the vehicle is discovered a second time.

grpc/server/src/backend.cpp
JonasVautherin: No it's not, sorry. Good catch.

julianoes: Same, I don't care either one as long as it's consistent. The rest of the API is `snake_case`, so presumably continue with that.

grpc/server/src/backend.cpp
JonasVautherin: That's one of those cases where I wish we had streams :D.

I thought it was not called the second time, as the `device` stays the same. I'll see what I can do here. But it sounds like a mutex and it's more complicated than what I would like to have :).

julianoes: `boost::fibers` would support something like that as well.

JonasVautherin: But since we don't depend on boost yet, I'm happy to try to avoid it unless necessary :D. I'm not sure I want to build it for Android/iOS...

grpc/server/src/CMakeLists.txt
```diff
@@ -70,12 +36,16 @@ target_include_directories(
     ${PLUGINS_DIR}
 )
 
-target_link_libraries(
-    dronecore_server
+add_executable(backend_bin
```
shakthi-prashanth-m: How about renaming it as `backend_server` ?

grpc/server/src/backend.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+#include "action/actionrpc_impl.h"
+#include "core/corerpc_impl.h"
+#include "log.h"
+#include "mission/missionrpc_impl.h"
+#include "telemetry/telemetryrpc_impl.h"
+
+namespace dronecore::backend {
+
+int DroneCoreBackend::run(const int mavlink_listen_port)
```
shakthi-prashanth-m: `const` doesn't serve purpose here, right ? Its call by value.

grpc/server/src/backend.cpp
```diff
@@ -0,0 +1,104 @@
...
+}
+
+int DroneCoreBackend::connect_to_vehicle(const int port)
+{
+    if (!add_udp_connection(port)) {
+        return 1;
+    }
+
+    log_uuid_on_timeout();
+    wait_for_discovery();
```
shakthi-prashanth-m: return value omitted.

grpc/server/src/backend.cpp
```diff
@@ -0,0 +1,104 @@
...
+    builder.RegisterService(&telemetryService);
+
+    server = builder.BuildAndStart();
+    LogInfo() << "Server started";
+    server->Wait();
+
+    return 0;
+}
+
+int DroneCoreBackend::connect_to_vehicle(const int port)
```
shakthi-prashanth-m: Why not boolean as return value ?

grpc/server/src/plugins/action/actionrpc_impl.h
```diff
@@ -36,13 +33,12 @@ class ActionServiceImpl final : public rpc::action::ActionService::Service
...
+        const Action::Result action_result = action.land();
         response->set_result(static_cast<rpc::action::ActionResult::Result>(action_result));
         response->set_result_str(Action::result_str(action_result));
         return Status::OK;
     }
 
 private:
-    DroneCore *dc;
-    std::shared_ptr<Action> action;
+    const Action &action;
```
shakthi-prashanth-m: If its `const`, then how does `action.arm()` can work ? IMO, `const` is used when you do **read-only**. Operations that change the state (such as arm, takeoff, etc) can't be `const`. Please correct me if I am wrong.

grpc/server/src/plugins/mission/missionrpc_impl.h
```diff
@@ -9,11 +11,8 @@ using namespace dronecore;
...
 public:
-    MissionServiceImpl(DroneCore *dc_obj)
-    {
-        dc = dc_obj;
-        mission = std::make_shared<Mission>(&dc->device());
-    }
+    MissionServiceImpl(Mission &mission)
+        : mission(mission) {}
 
     Status SendMission(ServerContext *context, const rpc::mission::SendMissionRequest *request,
```
shakthi-prashanth-m: Can we name it as `UploadMission()` (or just `Upload()` as its mission upload indeed, because we call as `mission.upload(...)`) ? Makes sense ?

grpc/server/src/CMakeLists.txt
JonasVautherin: But then should the library be called `libdronecore_backend_server`? I assumed that "backend" was enough, as a "grpc backend" is necessarily a server somehow :D.

What do you think?

grpc/server/src/backend.cpp
JonasVautherin: Doesn't hurt either :-). It says that `mavlink_listen_port` is not modified in `run`. Not much, but it's also about the principle: I try to make `const` everything I can :-).

grpc/server/src/backend.cpp
JonasVautherin: Good catch

JonasVautherin: Right, I forgot to log the UUID on discovery. Thanks!

JonasVautherin: Actually I'm not sure I understand this comment... what do you mean?

JonasVautherin: Good catch :-).

shakthi-prashanth-m: Ok! I said that you have omitted returning an `int` value from this function.

grpc/server/src/backend.cpp
JonasVautherin: Yeah, you're right.

grpc/server/src/plugins/action/actionrpc_impl.h
JonasVautherin: The `Action` object doesn't have state. It just passes the message further. By having it `const`, it makes it clear. `Mission` has some internal state and, therefore, isn't const.

grpc/server/src/plugins/mission/missionrpc_impl.h
JonasVautherin: I don't have a strong opinion on it, but note that this implies changing the proto files, and very soon we won't be able to change the proto files anymore as it will break the API for all the modules.

This said, I had put `SendMission` instead of `UploadMission` because I remember having had weird discussion with users who were assuming that "upload" is over the internet, and hence you "upload the mission to some website" :D. But maybe that's not relevant here.

@julianoes do you have an opinion on the naming?

grpc/server/src/plugins/action/actionrpc_impl.h
shakthi-prashanth-m: I see.

grpc/server/src/CMakeLists.txt
shakthi-prashanth-m: If "backend" is the name used in practice, I am fine with it.

grpc/server/src/plugins/mission/missionrpc_impl.h
julianoes: Hm, when it is `SendMission` what's the other direction? `ReceiveMission`, `RequestMission`?

JonasVautherin: Actually, I am completely wrong. It is already [UploadMission](https://github.com/dronecore/DroneCore-Proto/blob/master/mission/mission.proto#L9). It's just not in the scope of this PR :-). Sorry about that :-).

JonasVautherin:
Yes sorry, I got stuck on the iOS build. I'm working on this right now.

JonasVautherin:
Coming back with some unit tests soon.

PR #271, opened 2018-02-16
Update mission docs

hamishwillee:
Improves documentation for mission plan import.

### DIFF
--- a/plugins/mission/mission_item.h
+++ b/plugins/mission/mission_item.h
@@ -69,10 +69,11 @@ public:
     /**
      * @brief Set loiter time in seconds.
      *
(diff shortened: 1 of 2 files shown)


plugins/mission/mission_item.h
```diff
@@ -69,10 +69,11 @@ class MissionItem
     /**
      * @brief Set loiter time in seconds.
      *
-     * This can be used to wait for vehicle to slow down or a gimbal to arrive at the set
+     * This specifies the delay at a waypoint before executing next mission item.
```
hamishwillee: This in particular needs to be checked. I assume the delay is at a waypoint.

PR #276, opened 2018-02-16
build_system: build statically again for Mac

### LINKED ISSUE #233: Flight testing DroneCore examples on Intel Aero
@dronecore/testflights can you please help us verify the following documentation examples run on the Intel Aero:

* [Takeoff and Land](https://github.com/dronecore/DroneCore/tree/develop/example/takeoff_land)
* Fly QGC mission. You can use [dronecore/DroneCore/#235](#235) which flies QGroundControl Mission plan
* [Offboard velocity control](https://github.com/dronecore/DroneCore/tree/develop/example/offboard_velocity)
* [GPS Follow Me](https://github.com/dronecore/DroneCore/tree/develop/example/follow_me)

Please try to follow the steps in our docs to build and run the examples
* [Building DroneCore from source](https://docs.dronecore.io/en/contributing/build.html)
* [Running examples](https://docs.dronecore.io/en/examples/)

We are looking for all levels of feedback, from "steps are wrong, we can't build" to "vehicle flew in the wrong direction".

@shakthi-prashanth-m @anitha-suresh @julianoes @JonasVautherin how would you guys like the feedback from the test team? logs? command line output? screenshots?

julianoes:
This is just an interim fix in the meantime.

Should fix #233.

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -13,21 +13,21 @@ else()
 endif()
 
 # We need tinyxml2 for the camera definition parsing.
-if(IOS OR APPLE)
+if(APPLE)
     # We install the tinyxml2 library manually for macOS and iOS.
     # The install is commented out as long as we build statically.
-    #if (CMAKE_BUILD_TYPE STREQUAL "Debug")
-    #    # Need to remove that d again.
-    #    install(FILES ${CMAKE_CURRENT_BINARY_DIR}/third_party/tinyxml2/libtinyxml2d.5.0.1.dylib
-    #        DESTINATION ${lib_path}
-    #        RENAME libtinyxml2.dylib
-    #    )
-    #else()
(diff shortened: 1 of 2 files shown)


core/CMakeLists.txt
```diff
@@ -13,21 +13,21 @@ else()
 endif()
 
 # We need tinyxml2 for the camera definition parsing.
-if(IOS OR APPLE)
+if(APPLE)
     # We install the tinyxml2 library manually for macOS and iOS.
     # The install is commented out as long as we build statically.
```
JonasVautherin: Actually this comment is obsolete now, I guess it should be removed.

Also note that we never built tinyxml2 as a static library, I believe (and we'll need that for iOS). It requires `BUILD_STATIC_LIBS` to be set, which I guess we don't want to set in the whole scope of `core/CMakeLists.txt`. I am planning to work a bit more on the build system, but after I get the prototype running on iOS :-).

julianoes: Thanks, removed the comment.

PR #277, opened 2018-02-17
Add functionality for setting the actuator control targets directly.

xgerrmann:
This is an offboard functionality which allows for setting the actuator control targets directly for a specific control group in offboard mode. 

- This pull request goes accompanied with a pull request on the PX4 side https://github.com/PX4/Firmware/pull/8908.
- It includes an integration test, but I would like your opinion on which ASSERTS to add. 
- Also the integration test works for a specific model type (standard_vtol), since the goal is to set its ailerons directly. Any Ideas on how to generalize?

### DIFF
--- a/plugins/offboard/offboard.h
+++ b/plugins/offboard/offboard.h
@@ -93,6 +93,12 @@ public:
     };
 
     /**
+     * @brief Types required for actuator control commands.
+     */
+    typedef float ActuatorControlTarget[8];
+    typedef uint8_t ActuatorControlGroup;
+
+    /**
      * @brief Start offboard control (synchronous).
      *
      * **Attention:** this is work in progress, use with caution!
@@ -151,6 +157,13 @@ public:
      * @param velocity_body_yawspeed Velocity and yaw angular rate `struct`.
      */
     void set_velocity_body(VelocityBodyYawspeed velocity_body_yawspeed);
+
+    /**
+     * @brief Set the actuator control targets for a specific control group.
+     *
+     * @param actuator_control_target float[8] and control_group uint8_t.
+     */
+    void set_actuator_control_target(ActuatorControlTarget actuator_control_target, ActuatorControlGroup control_group);
 
     /**
      * @brief Copy constructor (object is not copyable).
--- a/integration_tests/CMakeLists.txt
+++ b/integration_tests/CMakeLists.txt
@@ -25,6 +25,7 @@ add_executable(integration_tests_runner
     async_hover.cpp
     takeoff_and_kill.cpp
     offboard_velocity.cpp
+    offboard_actuator_control_target.cpp
     logging.cpp
     info.cpp
     mission.cpp
--- a/integration_tests/offboard_actuator_control_target.cpp
+++ b/integration_tests/offboard_actuator_control_target.cpp
@@ -1,0 +1,67 @@
+#include <iostream>
+#include <cmath>
+#include "integration_test_helper.h"
+#include "dronecore.h"
+#include "plugins/action/action.h"
+#include "plugins/telemetry/telemetry.h"
+#include "plugins/offboard/offboard.h"
+#include "plugins/mission/mission.h"
+
+using namespace dronecore;
+
+
+TEST_F(SitlTest, OffboardActuatorControlTarget)
+{
+    DroneCore dc;
+
+    DroneCore::ConnectionResult ret = dc.add_udp_connection();
+    ASSERT_EQ(DroneCore::ConnectionResult::SUCCESS, ret);
+
+    // Wait for device to connect via heartbeat.
+    std::this_thread::sleep_for(std::chrono::seconds(2));
+
+    Device &device	= dc.device();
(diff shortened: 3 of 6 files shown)


plugins/offboard/offboard.h
```diff
@@ -152,6 +158,13 @@ class Offboard : public PluginBase
      */
     void set_velocity_body(VelocityBodyYawspeed velocity_body_yawspeed);
 
+    /**
+     * @brief Set the actuator control targets for a specific control group.
+     *
+     * @param actuator_control_target float[8] and control_group uint8_t.
```
hamishwillee: Docs is subset of doxygen standard so you will need the following (doxgen will pull in types). 
```
@param actuator_control_target Set the actuator control target.
@ control_group uint8_t Set the actuator control group.
```
```

File: plugins/offboard/offboard.h [hunk 2]
```diff
@@ -92,6 +92,12 @@ class Offboard : public PluginBase
                                    looking from above). */
     };
 
+    /**
+     * @brief Types required for actuator control commands.
+     */
+    typedef float ActuatorControlTarget[8];
```
hamishwillee: No, you'll need doxygen comments for each of these.

plugins/offboard/offboard.h
```diff
@@ -152,6 +158,13 @@ class Offboard : public PluginBase
      */
     void set_velocity_body(VelocityBodyYawspeed velocity_body_yawspeed);
 
+    /**
+     * @brief Set the actuator control targets for a specific control group.
+     *
+     * @param actuator_control_target float[8] and control_group uint8_t.
+     */
+    void set_actuator_control_target(ActuatorControlTarget actuator_control_target, ActuatorControlGroup control_group);
```
hamishwillee: The name `set_actuator_control_target()` sounds to me like you're setting what target you want to affect, and there would be another control that sets the value of that target. However I'm fairly sure that this actually sets the value of the target - correct? Perhaps `set_actuator_control`  would actually be more clear?

plugins/offboard/offboard.h
```diff
@@ -152,6 +158,13 @@ class Offboard : public PluginBase
      */
     void set_velocity_body(VelocityBodyYawspeed velocity_body_yawspeed);
 
+    /**
+     * @brief Set the actuator control targets for a specific control group.
+     *
```
hamishwillee: I think this is perhaps this is an advanced concept that we cannot assume users will know about. Perhaps provide a very brief overview and link to appropriate docs: https://dev.px4.io/en/concept/mixing.html#control-groups

plugins/offboard/offboard.h
julianoes: I vote to use different naming from mavlink for the API here and name it what it actually does.

Also, I don't think it's a good idea to have a `float[8]` but rather individual fields that all say what they are.

Basically, you have to imagine an app developer that doesn't know how the drone works and wants to command it without knowing what an actuator or an actuator group is.

hamishwillee: @julianoes I very much agree with you in principle, but I don't know whether this is feasible/reasonable.

So what has been proposed is a generic API for setting any target value for any [control group](https://dev.px4.io/en/concept/mixing.html#control-groups). There are 6 groups and they represent different uses of a physical bus of 8 outputs - e.g. flight control, payload, gimbal, manual passthrough. 

Choices are either to:
* Stick with current API - most generic approach
* Create (up to  6) separate methods that abstract the control groups. Then we can have specific names and types for each case
* Decide this API is too low level for our audience/core API and suggest this be provided as an external plugin by anyone who wants it. 

My problem here is that I don't really understand what, how and when each of the control groups would really be used - which makes it hard to assess whether we "really" need all or any of these. For example, I'm guessing group 0 is for any vehicle flight control and 1 is for a VTOL in MC mode, but I have no idea if that is correct, or what the system would do if you sent an invalid control group message for the current case, or why you would want to.

@julianoes Do you understand how the control groups would actually work?

@xgerrmann What is the use case you'd actually be trying to solve if you had the power of this API? Is it (say) to just drop a payload? If so, then it may be that we could just expose the payload control group.

julianoes: In my opinion if we just replicate the hard to understand somewhat generic mavlink interface, then DroneCore ends up being nothing else than what we already have with mavlink. It will expose a bunch of fields and no one knows what is supported and with what values/limits.

Therefore, we really need to break it down into what we're trying to achieve so that it is very clear to a future user of the API.

Therefore, I suggest to either change this one to something that is understandable and properly supported, or not include it for now but only in a fork for now and then separate plugin later.

> Do you understand how the control groups would actually work?

I'd have to loop up if/how it all works.

hamishwillee:
I agree with @julianoes on this: https://github.com/dronecore/DroneCore/pull/277#discussion_r170510803

We *shouldn't* merge something unless we understand the use case we're trying to meet and can provide a clear API to meet it. We *cannot* merge something without understanding the implications.

@xgerrmann - I'm going to close this now, but feel free to reopen if you wish to continue the discussion.

PR #280, opened 2018-02-20
core: send HITL bit with DO_SET_MODE command

julianoes:
Commander gets confused if this bit is not set correctly, so we need to
add it.
Also, the _armed and _hitl_enabled flags should both be atomic in order
to be thread-safe, also we should only update the flags if a heartbeat
originates from the autopilot but ignore other components such as the
camera or gimbal.

### DIFF
--- a/core/device.cpp
+++ b/core/device.cpp
@@ -154,7 +154,10 @@ void Device::process_heartbeat(const mavlink_message_t &message)
     mavlink_heartbeat_t heartbeat;
     mavlink_msg_heartbeat_decode(&message, &heartbeat);
 
-    _armed = ((heartbeat.base_mode & MAV_MODE_FLAG_SAFETY_ARMED) ? true : false);
+    if (message.compid == MavlinkCommands::DEFAULT_COMPONENT_ID_AUTOPILOT) {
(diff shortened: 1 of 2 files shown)


core/device.cpp
```diff
@@ -453,9 +456,12 @@ void Device::get_param_float_async(const std::string &name,
 
 MavlinkCommands::Result Device::set_flight_mode(FlightMode device_mode)
 {
-    uint8_t flag_safety_armed = is_armed() ? MAV_MODE_FLAG_SAFETY_ARMED : 0;
+    const uint8_t flag_safety_armed = is_armed() ? MAV_MODE_FLAG_SAFETY_ARMED : 0;
```
shakthi-prashanth-m: Don't we need `Device::is_hitl_enabled()` for plugins in the future ?
If you think its useful, then its good time to add it.

julianoes: I would add it once we need it.

julianoes:
> How did you notice it in firmware ?

I ran into this: https://github.com/PX4/Firmware/blob/a35abf2453e884987246e4e8d7565a94eddad705/src/modules/commander/state_machine_helper.cpp#L534-L538

PR #281, opened 2018-02-21
appveyor: try to run on more than one core

julianoes:
Appveyor CI should provide 2 cores, so we can make use of that.

This is also in the docs:
https://github.com/dronecore/docs/pull/98

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -41,6 +41,7 @@ before_build:
 build: on
 
 build_script:
+  - set CL=/MP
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -41,6 +41,7 @@ before_build:
 build: on
 
 build_script:
+  - set CL=/MP
```
shakthi-prashanth-m: So, Shall I remove `set cores=%NUMBER_OF_PROCESSORS%` ?

julianoes: I don't know. If it's the same, sure.

PR #282, opened 2018-02-22
telemetry: use absolute yaw for camera attitude

julianoes:
We now have an interface for the absolute yaw angle of a camera
(respectively a gimbal).

FYI @sanderux.

### DIFF
--- a/plugins/telemetry/telemetry.h
+++ b/plugins/telemetry/telemetry.h
@@ -364,12 +364,18 @@ public:
     /**
      * @brief Get the camera's attitude in quaternions (synchronous).
      *
+     * Note this attitude is relative to North, so in the absolute frame
+     * and not relative to the vehicle's attitude.
+     *
      * @return Camera's attitude as quaternion.
      */
     Quaternion camera_attitude_quaternion() const;
 
     /**
      * @brief Get the camera's attitude in Euler angles (synchronous).
+     *
+     * Note this attitude is relative to North, so in the absolute frame.
+     * and not relative to the vehicle's attitude.
      *
      * @return Camera's attitude as Euler angle.
      */
--- a/plugins/telemetry/telemetry.h
+++ b/plugins/telemetry/telemetry.h
@@ -364,8 +364,7 @@ public:
     /**
      * @brief Get the camera's attitude in quaternions (synchronous).
      *
-     * Note this attitude is relative to North, so in the absolute frame
-     * and not relative to the vehicle's attitude.
+     * Note that the yaw component of attitude is relative to North (absolute frame).
      *
      * @return Camera's attitude as quaternion.
      */
@@ -374,8 +373,7 @@ public:
     /**
      * @brief Get the camera's attitude in Euler angles (synchronous).
      *
-     * Note this attitude is relative to North, so in the absolute frame.
-     * and not relative to the vehicle's attitude.
+     * Note that the yaw component of attitude is relative to North (absolute frame).
      *
      * @return Camera's attitude as Euler angle.
      */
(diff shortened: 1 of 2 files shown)


plugins/telemetry/telemetry.h
```diff
@@ -364,13 +364,19 @@ class Telemetry : public PluginBase
     /**
      * @brief Get the camera's attitude in quaternions (synchronous).
      *
+     * Note this attitude is relative to North, so in the absolute frame
```
hamishwillee: This is OK, but possibly more than needs to be said. I'd suggest for both changes:

> Note that the yaw component of attitude is relative to North (absolute frame).

hamishwillee:
You also need to update the yaw value here: https://docs.dronecore.io/v/develop/en/api_reference/structdronecore_1_1_telemetry_1_1_euler_angle.html

I'd suggest updating the https://docs.dronecore.io/v/develop/en/api_reference/structdronecore_1_1_telemetry_1_1_quaternion.html but I don't really understand it :-0

Would it be useful for You also need to update the yaw value here: https://docs.dronecore.io/v/develop/en/api_reference/structdronecore_1_1_telemetry_1_1_euler_angle.html to include a vehicle-relative value too, that we calculate? I kind of think it would.

julianoes:
@hamishwillee the fields you're pointing to are just the data types and should not specify the coordinate frame as such. So, these structs can be used for relative or absolute angles.

plugins/telemetry/telemetry_impl.cpp
```diff
@@ -391,7 +391,7 @@ void TelemetryImpl::process_mount_orientation(const mavlink_message_t &message)
     Telemetry::EulerAngle euler_angle {
         mount_orientation.roll,
         mount_orientation.pitch,
-        mount_orientation.yaw
+        mount_orientation.yaw_absolute
```
shakthi-prashanth-m: This (http://mavlink.org/messages/common#MOUNT_ORIENTATION) has both `yaw` & `yaw_absolute`. Does the link hold message definitions of MAVLink2 ?
Also, if you change member here, we need to change its usage; isn't it ? :)

julianoes: Where do we need to change its usage? I'm confused.

hamishwillee:
> @hamishwillee the fields you're pointing to are just the data types and should not specify the coordinate frame as such. So, these structs can be used for relative or absolute angles.

I don't really understand. If you name something with "absolute" in it's title then you are forcing it to use an absolute frame - `mount_orientation.yaw_absolute`. I guess we aren't specifying which frame of course, but it isn't going to be relative to the vehicle.

plugins/telemetry/telemetry_impl.cpp
shakthi-prashanth-m: Ah OK! So, its just that [yaw_deg](https://github.com/dronecore/DroneCore/blob/742537418b5892bd30df3d58708826760c3dbef0/plugins/telemetry/telemetry.h#L74) of `EulerAngle` will be absolute now ?

julianoes:
@hamishwillee 
The data type used for an Euler angle or a quaternion in DroneCore just represent an attitude without a frame connected to it. They can be used for absolute or relative, depending on what telemetry information is sent/used.

However, the fields `yaw` and `yaw_absolute` sent in mavlink specify the frame.

JonasVautherin:
So from now on, the yaw of the camera is absolute, and we don't provide an interface to the camera yaw relative to the vehicle. Is that right? :-)

julianoes:
@JonasVautherin that's right. What's your opinion, would you also want the relative yaw?

JonasVautherin:
Fine for me for now. We can always add the relative yaw later if needed, right?

julianoes:
Yes we can.

PR #288, opened 2018-02-26
core: Move implementation into the pImpl.

JonasVautherin:
Fix the DroneCore pImpl (some of the implementation had leaked into the base class).

* Pass strings by reference
* Make const arguments const
* Move some defaults value to the consumer instead of the generic classes
* Remove unused members in udp_connection
* Reorder some headers

### DIFF
--- a/core/dronecore.cpp
+++ b/core/dronecore.cpp
@@ -1,13 +1,7 @@
 #include "dronecore.h"
+
 #include "dronecore_impl.h"
 #include "global_include.h"
-#include "connection.h"
-#include "udp_connection.h"
-#include "tcp_connection.h"
-#ifndef WINDOWS
-#include "serial_connection.h"
-#endif
-
 
 namespace dronecore {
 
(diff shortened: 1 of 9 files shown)


core/dronecore.cpp
```diff
@@ -112,22 +54,22 @@ bool DroneCore::is_connected() const
...
 }
 
-void DroneCore::register_on_timeout(event_callback_t callback)
+void DroneCore::register_on_timeout(const event_callback_t callback)
 {
     _impl->register_on_timeout(callback);
 }
 
-const char *DroneCore::connection_result_str(ConnectionResult result)
+const char *DroneCore::connection_result_str(const ConnectionResult &result)
```
julianoes: The result is only one value. Are you sure that it should be a reference?

core/dronecore.h
```diff
@@ -59,60 +56,35 @@ class DroneCore
...
      *
-     * @param dev_path COM or UART dev node name/path.
-     * @param baudrate Baudrate of the serial port.
+     * @param dev_path COM or UART dev node name/path (defaults to "/dev/ttyS0").
+     * @param baudrate Baudrate of the serial port (defaults to 9600).
      * @return The result of adding the connection.
      */
-    ConnectionResult add_serial_connection(std::string dev_path, int baudrate);
+    ConnectionResult add_serial_connection(const std::string &dev_path = "/dev/ttyS0",
+                                           int baudrate = 9600);
```
julianoes: The default is super low. I'd use 57600 or 115200, I think.

core/udp_connection.h
```diff
@@ -30,9 +30,6 @@ class UdpConnection : public Connection
 
     static void receive(UdpConnection *parent);
 
-    // This port is shared with mavros, so either one, this SDK or mavros can be used.
```
julianoes: I was going to complain that we lose this comment but I see you moved it.

core/dronecore.h
JonasVautherin: Ok. I have no clue, I just copied it from `SerialConnection` :D.

core/udp_connection.h
JonasVautherin: Yeah, I tend to be aggressive with comments. Good that you keep an eye on them xD.

core/dronecore.h
```diff
@@ -77,60 +81,35 @@ class DroneCore
...
      *
-     * @param dev_path COM or UART dev node name/path.
-     * @param baudrate Baudrate of the serial port.
+     * @param dev_path COM or UART dev node name/path (defaults to "/dev/ttyS0").
+     * @param baudrate Baudrate of the serial port (defaults to 57600).
      * @return The result of adding the connection.
      */
-    ConnectionResult add_serial_connection(std::string dev_path, int baudrate);
+    ConnectionResult add_serial_connection(const std::string &dev_path = DEFAULT_SERIAL_DEV_PATH,
+                                           int baudrate = DEFAULT_SERIAL_BAUDRATE);
```
julianoes: I think that's nice like that.

core/dronecore.h
```diff
@@ -22,6 +20,12 @@ class Device;
 class DroneCore
 {
 public:
+    static constexpr auto DEFAULT_UDP_CONNECTION_URL = "udp://:14540";
+    static constexpr int DEFAULT_UDP_PORT = 14540;
+    static constexpr auto DEFAULT_TCP_REMOTE_IP = "127.0.0.1";
+    static constexpr int DEFAULT_TCP_REMOTE_PORT = 5760;
+    static constexpr auto DEFAULT_SERIAL_DEV_PATH = "/dev/ttyS0";
+    static constexpr int DEFAULT_SERIAL_BAUDRATE = 57600;
```
julianoes: Yer, that's good, a nice overview.

PR #289, opened 2018-02-27
Fix initializers for attribute prototypes

hamishwillee:
Makes default values link better. 

For example, in previous version over overview prototype only `add_any_connection` would link - there would be no link to the definition of default value `DEFAULT_UDP_CONNECTION_URL`

```
add_any_connection (const std::string & connection_url=DEFAULT_UDP_CONNECTION_URL)
````
Furthermore, the definition for DEFAULT_UDP_CONNECTION_URL did not include the default value, making it pretty useless. This fixes a bug that was stopping the default value being displayed.

### DIFF
--- a/generate_markdown_from_doxygen_xml.py
+++ b/generate_markdown_from_doxygen_xml.py
@@ -321,19 +321,30 @@ class cppAttribute:
         output_string=''
         output_string+='\n\n### %s {#%s}\n' % (self.name,self.id)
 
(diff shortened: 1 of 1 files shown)


hamishwillee:
@julianoes FYI I forced this merge. I know it "works", and I don't want it to block changes I need to do for https://github.com/dronecore/DroneCore/pull/290

PR #290, opened 2018-02-27
core: move ConnectionResult out of DroneCore class

JonasVautherin:
Because it is used in many classes, `ConnectionResult` should not be nested into `DroneCore`, IMO. Moreover, having it separate makes the code a bit nicer (`DroneCore::ConnectionResult` becomes `ConnectionResult`).

Finally, I am mocking `DroneCore` for some unit tests and, in my situation, I need to have `ConnectionResult` separate from `DroneCore`.


core/connection_result.h
```diff
@@ -0,0 +1,67 @@
+#ifndef DRONECORE_CONNECTION_RESULT_H
+#define DRONECORE_CONNECTION_RESULT_H
```
julianoes: Usually, I used `#pragma once`.

JonasVautherin: Oh, right. I did not know `#pragma` once actually, and I had seen some guidelines on [isocpp.github.io](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rs-guards), saying:

> Some implementations offer vendor extensions like `#pragma once` as alternative to include guards. It is not standard and it is not portable.

Do you have an opinion on that? I really don't know, I just read it. I can change it for a `pragma` if you prefer.

julianoes: I prefer it because it seems to work and it prevents errors when you copy and paste files without changing the define correctly.

JonasVautherin: ok

hamishwillee:
To me it feels a little wrong that the ConnectionResult is not part of DroneCore, since that is the one user facing place that manages connections. I'll take your work for it that this is a good idea though :-)

I assume the header is exported for docs generation purposes?

JonasVautherin:
Oh, I did not think about the docs changes =/. One reason is that I want to be able to mock DroneCore from the backend. So I want to have `DroneCore` in production, and `MockDroneCore` in testing. And both are independant classes that need `ConnectionResult`.

And also in terms of code, it's not like if `DroneCore` was hiding any of `ConnectionResult` given that it is used in the 40 files I had to edit :D.

PR #295, opened 2018-02-28
238 refactor dronecore server

JonasVautherin:
Refactor dronecore_server in order to have it build as a library (for iOS) and to have the connection unit tested (saved me from a few bugs like having the discover callback called multiple times or me calling `register_on_discover` multiple times.

Connects to #238.

### DIFF
--- a/grpc/server/src/CMakeLists.txt
+++ b/grpc/server/src/CMakeLists.txt
@@ -1,67 +1,33 @@
-cmake_minimum_required(VERSION 2.8)
+cmake_minimum_required(VERSION 3.1)
 
-file(STRINGS ${PLUGINS_DIR}/plugins.conf PLUGINS_LIST)
+set(COMPONENTS_LIST core action mission telemetry)
 
-set(PROTOC_BINARY ${CMAKE_BINARY_DIR}/../default/third_party/protobuf/bin/protoc)
-set(GRPC_CPP_PLUGIN_BINARY ${CMAKE_BINARY_DIR}/../default/third_party/grpc/bin/grpc_cpp_plugin)
+include(cmake/compile_proto.cmake)
 
-if(NOT EXISTS ${PROTOC_BINARY} OR NOT EXISTS ${GRPC_CPP_PLUGIN_BINARY})
-    message(FATAL_ERROR "Could not find 'protoc' or 'grpc_cpp_plugin' in the 'default' build folder. Please build for your host first (`make BUILD_DRONECORESERVER=YES default`).")
-endif()
+foreach(COMPONENT_NAME ${COMPONENTS_LIST})
+    compile_proto_pb(${COMPONENT_NAME} PB_COMPILED_SOURCE)
+    list(APPEND PB_COMPILED_SOURCES ${PB_COMPILED_SOURCE})
 
-add_custom_command(OUTPUT core/core.grpc.pb.cc
-    COMMAND ${PROTOC_BINARY}
-        -I ${PROTO_DIR}
-        --grpc_out=.
-        --plugin=protoc-gen-grpc=${GRPC_CPP_PLUGIN_BINARY}
-        ${PROTO_DIR}/core/core.proto;
-)
-
-add_custom_command(OUTPUT core/core.pb.cc
-    COMMAND ${PROTOC_BINARY}
-        -I ${PROTO_DIR}
-        --cpp_out=.
-        ${PROTO_DIR}/core/core.proto
-)
-
-foreach(PLUGIN ${PLUGINS_LIST})
-    add_custom_command(OUTPUT ${PLUGIN}/${PLUGIN}.grpc.pb.cc
-        COMMAND ${PROTOC_BINARY}
-            -I ${PROTO_DIR}
-            --grpc_out=.
-            --plugin=protoc-gen-grpc=${GRPC_CPP_PLUGIN_BINARY}
-            --cpp_out=.
-            ${PROTO_DIR}/${PLUGIN}/${PLUGIN}.proto
-    )
+    compile_proto_grpc(${COMPONENT_NAME} GRPC_COMPILED_SOURCE)
+    list(APPEND GRPC_COMPILED_SOURCES ${GRPC_COMPILED_SOURCE})
 endforeach()
 
-foreach(PLUGIN ${PLUGINS_LIST})
-    add_custom_command(OUTPUT ${PLUGIN}/${PLUGIN}.pb.cc
-        COMMAND ${PROTOC_BINARY}
-            -I ${PROTO_DIR}
-            --cpp_out=.
-            ${PROTO_DIR}/${PLUGIN}/${PLUGIN}.proto
-    )
-endforeach()
-
-set(PLUGINS_SRC dronecore_server.cpp core/core.pb.cc core/core.grpc.pb.cc)
-foreach(PLUGIN ${PLUGINS_LIST})
-    list(APPEND PLUGINS_SRC ${PLUGIN}/${PLUGIN}.pb.cc)
-    list(APPEND PLUGINS_SRC ${PLUGIN}/${PLUGIN}.grpc.pb.cc)
-endforeach()
-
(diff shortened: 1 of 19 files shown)


grpc/server/test/connection_initiator_test.cpp
```diff
@@ -0,0 +1,84 @@
...
+    initiator.start(dc, ARBITRARY_PORT);
+}
+
+TEST(ConnectionInitiator, connectHangsUntilDeviceDiscovered)
+{
+    ConnectionInitiator initiator;
+    MockDroneCore dc;
+    event_callback_t discover_callback;
+    EXPECT_CALL(dc, register_on_discover(_))
+    .WillOnce(SaveCallback(&discover_callback));
```
JonasVautherin: This should be indented, but I can't seem to find how to do this with astyle -_-.

grpc/server/src/backend.h
```diff
@@ -0,0 +1,29 @@
...
+
+    bool run(const int mavlink_listen_port = 14540);
+
+private:
+    bool run_server();
+    void setup_port(grpc::ServerBuilder &builder);
+
+    dronecore::DroneCore _dc;
+    dronecore::backend::ConnectionInitiator<dronecore::DroneCore> _connection_initiator;
+    std::unique_ptr<grpc::Server> server;
```
julianoes: Should it be `_server`?

grpc/server/src/backend.h
```diff
@@ -0,0 +1,29 @@
...
+};
+
+} // namespace backend
```
julianoes: I like the namespacing.

grpc/server/test/connection_initiator_test.cpp
julianoes: Yer that's weird about astyle. I guess at some point we should try to out the clang tooling to enforce style.

grpc/server/test/connection_initiator_test.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+    auto async_future = std::async(std::launch::async, [&initiator, &dc]() {
+        initiator.start(dc, ARBITRARY_PORT);
+        initiator.wait();
+    });
+
+    EXPECT_TRUE(async_future.wait_for(std::chrono::milliseconds(100)) == std::future_status::timeout) <<
```
julianoes: I usually use fake time for unit tests. Have you seen that?

grpc/server/src/connection_initiator.h
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    std::future<uint64_t> wrapped_register_on_discover(DroneCore &dc)
+    {
+        auto future = _discovery_promise->get_future();
+
+        LogInfo() << "Waiting to discover device...";
+
+        dc.register_on_discover([this](uint64_t uuid) {
+            std::call_once(_discovery_flag, [this, uuid]() {
```
julianoes: Does this mean that it is only called once? So it is not called anymore after a timeout and then discovery.

grpc/server/test/connection_initiator_test.cpp
JonasVautherin: Oh, I've seen a `FAKE_TIME`, but immediately forgot about it. I'll have a look!

grpc/server/src/connection_initiator.h
JonasVautherin: I think it doesn't matter here, as I only want to initiate the connection? And the `dronecore` plugin (or telemetry? I don't remember) is the one sending the state to the client.

JonasVautherin: And actually as soon as the first connection is established, the `dronecore` plugin will probably overwrite the `register_on_discover` callback since, for now, there is only one callback saved at a time.

grpc/server/test/connection_initiator_test.cpp
JonasVautherin: I am not sure how I should use it here, actually. In `call_every_handler_test.cpp`, I see the benefit of the fake time because the object under test is using the fake time as well. But here, I don't know how I can make my `initiator.wait()` use the fake time?

I am a bit confused, maybe that's because I am not used to `FakeTime` ^^. How would you do it?

grpc/server/src/CMakeLists.txt
```diff
@@ -1,67 +1,33 @@
-cmake_minimum_required(VERSION 2.8)
+cmake_minimum_required(VERSION 3.1)
 
-file(STRINGS ${PLUGINS_DIR}/plugins.conf PLUGINS_LIST)
+set(COMPONENTS_LIST core action mission telemetry)
```
shakthi-prashanth-m: Oh! No plugin list ?

grpc/server/src/backend.cpp
```diff
@@ -0,0 +1,56 @@
+#include "backend.h"
+
+#include <grpc/grpc.h>
+#include <grpc++/server.h>
+#include <grpc++/server_builder.h>
+#include <grpc++/server_context.h>
+#include <grpc++/security/server_credentials.h>
+
+#include "action/actionrpc_impl.h"
```
shakthi-prashanth-m: So ideally, backend shall include all plugin header files as it doesn't know what a user wants. Is that right ?

grpc/server/src/connection_initiator.h
```diff
@@ -0,0 +1,82 @@
...
+namespace backend {
+
+template <typename DroneCore>
+class ConnectionInitiator
+{
+public:
+    ConnectionInitiator() {}
+    ~ConnectionInitiator() {}
+
+    bool start(DroneCore &dc, const int port)
```
shakthi-prashanth-m: What about connection URL ?

grpc/server/test/connection_initiator_test.cpp
```diff
@@ -0,0 +1,84 @@
...
+}
+
+TEST(ConnectionInitiator, registerDiscoverIsCalledExactlyOnce)
+{
+    ConnectionInitiator initiator;
+    MockDroneCore dc;
+    EXPECT_CALL(dc, register_on_discover(_))
+    .Times(1);
+
+    initiator.start(dc, ARBITRARY_PORT);
```
shakthi-prashanth-m: Can we also add one that initiates via connection URL ?

grpc/server/src/plugins/action/actionrpc_impl.h
```diff
@@ -36,13 +33,12 @@ class ActionServiceImpl final : public rpc::action::ActionService::Service
...
+        const Action::Result action_result = action.land();
         response->set_result(static_cast<rpc::action::ActionResult::Result>(action_result));
         response->set_result_str(Action::result_str(action_result));
         return Status::OK;
     }
 
 private:
-    DroneCore *dc;
-    std::shared_ptr<Action> action;
+    const Action &action;
```
shakthi-prashanth-m: Its good that you switched to references. :)

grpc/server/src/CMakeLists.txt
JonasVautherin: For now, it is hard-coded. We need to deliver a first version soon. So our first version will not really support plugins (i.e. the backend has all our plugins). Later, we'll upgrade for a more dynamic architecture.

Still, the frontend now should look like the plugins are loaded dynamically, so that it is transparent to the user when we update the backend :-).

grpc/server/src/backend.cpp
JonasVautherin: Ideally, plugins are somehow loaded dynamically. But the first version is hard-coded because it is faster to do xD.

grpc/server/src/connection_initiator.h
JonasVautherin: Some further enhancement as well :-). Shouldn't be difficult to add, but not part of the "minimum viable product".

grpc/server/src/CMakeLists.txt
shakthi-prashanth-m: OK :-)

PR #297, opened 2018-03-01
Add missing docs for ConnectionResult and connection_result_str

hamishwillee:
Add missing

### DIFF
--- a/core/connection_result.h
+++ b/core/connection_result.h
@@ -1,5 +1,9 @@
 #pragma once
 
+
+/**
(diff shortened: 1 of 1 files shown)


core/connection_result.h
```diff
@@ -1,5 +1,9 @@
 #pragma once
 
+
+/**
+ * @brief Namespace for all dronecore types.
```
hamishwillee: @JonasVautherin FYI, namespace documentation is mandatory for doxygen to be able to generate the contained global objects.

PR #298, opened 2018-03-01
Enable markdown generation for global functions and enums

hamishwillee:
@JonasVautherin changes in #290 introduced global enums and functions - before all were scoped within classes. Doxygen generates these in a new file namespacedronecore.xml. This change parses the xml file and creates namespacedronecore.md.

As an aside, while there is a fair bit of successful code reuse for this change, I do need to start thinking about whether this can be simplified/broken up to ease future maintenance.

### DIFF
--- a/generate_markdown_from_doxygen_xml.py
+++ b/generate_markdown_from_doxygen_xml.py
@@ -722,7 +722,10 @@ class cppFunction:
 
     def markdown_overview(self):
         # Print function in overview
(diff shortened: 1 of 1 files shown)


JonasVautherin:
I cannot really judge this, but from looking at the documentation generated by these changes, I'd say it's all good :-).

hamishwillee:
> I cannot really judge this, but from looking at the documentation generated by these changes, I'd say it's all good :-).

Thanks @JonasVautherin . I'll update with the base branch and merge then.

PR #299, opened 2018-03-02
294 ios backend framework

### LINKED ISSUE #294: Compile backend to a Framework on iOS
When building gRPC for iOS, it should directly create a Framework that can be used in an iOS app.

JonasVautherin:
When building for iOS, the backend is now automatically built as an iOS Framework that can be directly added in an Xcode project and used (it only exposes the one `runBackend(port)` function).

There are two small glitches regarding dependencies:
* Json11 has to be built as a static library. I still need to figure what's a good way to do that, but maybe for now I will just fork it.
* Tinyxml2 has to be built as a static library. Here, there is a way to do it, but with a not-so-generic variable (`-DBUILD_STATIC_LIBS`) that we may not want to set for the whole project. So I'd like to build tinyxml2 separately (like grpc and others). For now, I don't link it in iOS.

Fixes #294.

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -13,25 +13,27 @@ else()
 endif()
 
 # We need tinyxml2 for the camera definition parsing.
-if(APPLE AND NOT IOS)
-    # We install the tinyxml2 library manually for macOS and iOS.
-    if (CMAKE_BUILD_TYPE STREQUAL "Debug")
-        # Need to remove that d again.
-        install(FILES ${CMAKE_CURRENT_BINARY_DIR}/third_party/tinyxml2/libtinyxml2d.5.0.1.dylib
-            DESTINATION ${lib_path}
-            #RENAME libtinyxml2.dylib
-        )
-    else()
-        install(FILES ${CMAKE_CURRENT_BINARY_DIR}/third_party/tinyxml2/libtinyxml2.5.0.1.dylib
-            DESTINATION ${lib_path}
-            #RENAME libtinyxml2.dylib
-        )
-    endif()
+if(APPLE)
+    if(NOT IOS)
+        # We install the tinyxml2 library manually for macOS and iOS.
+        if (CMAKE_BUILD_TYPE STREQUAL "Debug")
+            # Need to remove that d again.
+            install(FILES ${CMAKE_CURRENT_BINARY_DIR}/third_party/tinyxml2/libtinyxml2d.5.0.1.dylib
+                DESTINATION ${lib_path}
+                #RENAME libtinyxml2.dylib
+            )
+        else()
+            install(FILES ${CMAKE_CURRENT_BINARY_DIR}/third_party/tinyxml2/libtinyxml2.5.0.1.dylib
+                DESTINATION ${lib_path}
+                #RENAME libtinyxml2.dylib
+            )
+        endif()
 
-    add_subdirectory(third_party/tinyxml2 EXCLUDE_FROM_ALL)
-    include_directories(SYSTEM third_party/tinyxml2)
-    link_directories(third_party/tinyxml2)
-    set(TINYXML2_LIBRARY tinyxml2)
+        add_subdirectory(third_party/tinyxml2 EXCLUDE_FROM_ALL)
+        include_directories(SYSTEM third_party/tinyxml2)
+        link_directories(third_party/tinyxml2)
+        set(TINYXML2_LIBRARY tinyxml2)
+    endif()
 elseif(ANDROID)
     # We install the tinyxml2 library manually for Android.
     if (CMAKE_BUILD_TYPE STREQUAL "Debug")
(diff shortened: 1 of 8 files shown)


core/CMakeLists.txt
```diff
@@ -14,23 +14,26 @@ endif()
...
-            DESTINATION ${lib_path}
-            #RENAME libtinyxml2.dylib
-        )
-    else()
-        install(FILES ${CMAKE_CURRENT_BINARY_DIR}/third_party/tinyxml2/libtinyxml2.5.0.1.dylib
-            DESTINATION ${lib_path}
-            #RENAME libtinyxml2.dylib
-        )
-    endif()
+    if(NOT IOS)
```
julianoes: You could just write `if (APPLE AND NOT IOS)`, I think.

grpc/server/src/CMakeLists.txt
```diff
@@ -36,16 +51,29 @@ target_include_directories(backend
...
-add_executable(backend_bin
-    dronecore_server.cpp
-)
+if(IOS)
+    set_target_properties(backend PROPERTIES
+        FRAMEWORK TRUE
+        BUILD_WITH_INSTALL_RPATH TRUE
+        INSTALL_NAME_DIR @rpath
+        MACOSX_FRAMEWORK_IDENTIFIER io.dronecore.backend
+        VERSION 0.0.1
```
julianoes: We can get the version later from the latest git tag.

grpc/server/src/backend.cpp
```diff
@@ -5,52 +5,74 @@
...
+    void setup_port(grpc::ServerBuilder &builder)
+    {
+        std::string server_address("0.0.0.0:50051");
+        builder.AddListeningPort(server_address, grpc::InsecureServerCredentials());
+        LogInfo() << "Server set to listen on " << server_address;
+    }
+
+    dronecore::DroneCore _dc;
+    dronecore::backend::ConnectionInitiator<dronecore::DroneCore> _connection_initiator;
+    std::unique_ptr<grpc::Server> server;
```
julianoes: Not `_server`?

grpc/server/src/backend.h
```diff
@@ -1,28 +1,22 @@
...
 
 class DroneCoreBackend
 {
 public:
-    DroneCoreBackend() {}
-    ~DroneCoreBackend() {}
+    DroneCoreBackend();
+    ~DroneCoreBackend();
+
+    DroneCoreBackend(DroneCoreBackend &&op) noexcept;
```
julianoes: Why is the `noexcept` needed? Shouldn't everything be compiled without exception anyway?

core/CMakeLists.txt
JonasVautherin: Nope, because there is an `else()` at the end of the tinyxml2 block, that would then take linux and iOS :)

grpc/server/src/backend.cpp
JonasVautherin: argh. I lost it in the rebase. And I actually verified that it was still `_server` xD. Will change, sorry.

grpc/server/src/backend.h
JonasVautherin: Not, sure, I copied a pImpl example to not mess it up. Should I remove it?

core/CMakeLists.txt
julianoes: Ah, you're right.

grpc/server/src/backend.h
julianoes: I guess so, not sure.

grpc/server/src/backend.cpp
JonasVautherin: Oh, I actually messed up, because it fails the test. Just the CI is not compiling grpc for now :D.

grpc/server/src/backend.cpp
```diff
@@ -5,52 +5,74 @@
...
+    }
+
+    dronecore::DroneCore _dc;
+    dronecore::backend::ConnectionInitiator<dronecore::DroneCore> _connection_initiator;
+    std::unique_ptr<grpc::Server> _server;
+};
+
+DroneCoreBackend::DroneCoreBackend() : _impl(new Impl()) {}
+DroneCoreBackend::~DroneCoreBackend() = default;
+DroneCoreBackend::DroneCoreBackend(DroneCoreBackend &&) = default;
```
julianoes: What does `default` mean for these? What about using `delete` ?

JonasVautherin: This time, I think it is explained in the article showing the pImpl coming from Scott Meyers' book ([article here](http://oliora.github.io/2015/12/29/pimpl-and-rule-of-zero.html). Specifically, it says:

> // Tell the compiler to generate default special members which utilize
// the power of std::unique_ptr.
// We can do it here because the implementation class is defined at this point thus
// std::unique_ptr can properly handle the implementation pointer.

JonasVautherin: I can still change it later if you want, but as far as I understand, `default` is not wrong. So I'll already merge that :-).

JonasVautherin: Let's continue this discussion [here](https://github.com/dronecore/DroneCore/pull/300).

PR #300, opened 2018-03-02
backend: clean pImpl

JonasVautherin:
_Continuing the discussion we had on the other PR, but this one is not blocking, so we can take more time :-)._

It seems like the destructor must be defined because keeping the default doesn't compile (something like it creates a default destructor, but doesn't know `Impl` at that time.

The other two were apparently there to define them as `noexcept`, but they can be removed since we don't need it. I can also delete them if I do it in `backend.h`, but I'm not sure we actually want that.


julianoes:
I think you misunderstood me. What I meant with delete is `= delete` so that copying and assigning is not prevented.

JonasVautherin:
@julianoes Is it fine for you with the `delete`s now?

PR #303, opened 2018-03-05
backend: change backend dir

JonasVautherin:
* backend is now in /backend instead of /grpc/server
* update readme
* change BUILD_DRONECORESERVER to BUILD_BACKEND in Makefile


backend/README.md
```diff
@@ -0,0 +1,15 @@
...
+```
+make BUILD_BACKEND=YES
+```
+
+### Run the backend
+
+The backend is built as an executable:
+
+```
+./build/default/backend/src/backend_bin
```
shakthi-prashanth-m: Are you sure you're placing backend executable in `src` ?

backend/src/CMakeLists.txt
```diff
@@ -45,7 +45,7 @@ target_include_directories(backend
     PRIVATE
     ${CMAKE_SOURCE_DIR}/core
     ${CMAKE_SOURCE_DIR}/plugins
-    ${CMAKE_BINARY_DIR}/grpc/server/src
+    ${CMAKE_BINARY_DIR}/backend/src
```
shakthi-prashanth-m: I just noticed. Are you sure you're placing backend executable in `src` ?

backend/README.md
shakthi-prashanth-m: I am not sure the reason behind having it in `src`. As you know, for DroneCore unit tests, we've in  `build/default/unit_tests_runner`.  I think `./build/default/backend/bin` is also fine.

JonasVautherin: Right. I can put them in `build/default`, of course.

JonasVautherin: Actually, I am not sure what a good practice would be for that (and also maybe `install` should be set for the backend). I don't have much time for this right now, I just wanted to move `grpc/server` to `backend` now that not so many people are working on it.

Maybe changing the build output/install targets can be done later?

shakthi-prashanth-m: Yeah its perfectly fine.

Issue #304, opened 2018-03-06
DoneCore handles only one component


shakthi-prashanth-m:
While working on Camera plugin, I realized that DroneCore core library can communicate with one component of a vehicle only.
For.eg: A Drone has components such as Autopilot, Camera, Gimbal, etc. Each component has a unique Component ID, but System ID / UUID is same for all.

As per the MAVLink protocol design, all the components send Heartbeat indicating they're ready for communication. DroneCore core library maintains a map of devices that are discovered in the process.
When autopilot component of a vehicle is discovered, it is added to the map. However when a another component (say, camera) of the same system/vehicle is discovered, [it won't add it to the map](https://github.com/dronecore/DroneCore/blob/f308f020205c08d930e1acaa849b332e365d4965/core/dronecore_impl.cpp#L327).

Due to this model, DroneCore can only interact with one component of a system only. Suppose I would like to interact with a camera component and autopilot as well, I can only think of **creating another instance of DroneCore. Is that recommended** ?

If we design core to handle multiple components,  camera plugin may be used as below:
```cpp
DroneCore dc;
dc.add_udp_connection(14540); // same port can be used for both Autopilot & Camera as well

uint64_t uuid;
uint8_t drone_id, camera_id;
dc.register_on_discover(std::bind(&on_drone_discover, _1, _2)); // UUID, Comp ID
dc.register_on_discover(std::bind(&on_camera_discover, _1, _2)); // UUID, Comp ID

while (!dc.is_autopilot_connected(drone_id) &&
       !dc.is_camera_connected(camera_id)) {
    sleep_for(seconds(1);
}

Device &camera = dc.get_device(uuid, camera_id);
camera->start_video_capture();

Device &drone = dc.get_device(uuid, drone_id);
auto offboard = std::make_shared<Offboard>(&drone);
offboard->set_velocity_ned(......);
....

camera->stop_video_capture();
```

Right now, camera plugin should handle all the discover, connect logic in itself. 

As already discussed that DroneCore is also meant for dealing with any component, how about adding these APIs ?

```cpp
bool Device::is_autopilot(uint64_t uuid, uint8_t comp_id);
bool Device::is_camera(uint64_t uuid, uint8_t comp_id);
std::vector<unit8_t> DroneCore::get_component_list(uint64_t uuid);

```

So,  we can maintain many manage multiple camera devices of a drone as well.

Also related to #292

julianoes:
[...]

PR #310, opened 2018-03-07
Multi component handling

shakthi-prashanth-m:
**This is the pull request that makes DroneCore to handle multiple components.**

image

New changes in **core**:

1. DroneCore will now maintain `std::map<uint8_t, std::vector<Device*>>`
2. All the components may share same UDP connection port (For eg: Both Autopilot and other components may send MAVLink messages on the same port)

1. As per existing model, `device_thread` continue to exist per component that deals with sending MAVLink commands over the same connection. This makes camera component to get connected.

New interface in DroneCore:
```cpp
// Return reference to autopilot device of the given system UUID
Device &autopilot(uint64_t uuid);

// Return reference to autopilot device of the first system
Device &autopilot();

// Returns vector of camera ids on the given system UUID
std::vector<uint8_t> camera_ids(uint64_t uuid);

// Returns vector of camera ids on the first system
std::vector<uint8_t> camera_ids();

// Returns reference to camera device #camera_id of the given system UUID
Device &camera(uint64_t uuid, uint8_t camera_id);

// Returns reference to first camera device on the given system UUID
Device &camera(uint64_t uuid);

// Returns reference to first camera device on the first connected system
Device &camera();

// Checks whether first camera on the first vehicle is connected.
bool is_camera_connected();

// Checks whether first vehicle is connected.
bool is_autopilot_connected();

```

Resulting camera application code would be:
```cpp
DroneCore connection;

connection.add_udp_connection(14550); // common port for all the components of the system

while (!connection.is_autopilot_connected() ||
       !connection.is_camera_connected())
    sleep_for(seconds(1));

auto offboard = std::make_shared<Offboard>(connection.autopilot());
auto camera = std::make_shared<Camera>(connection.camera());

...
```

### DIFF
--- a/example/fly_mission/fly_mission.cpp
+++ b/example/fly_mission/fly_mission.cpp
@@ -57,8 +57,9 @@ int main(int /*argc*/, char ** /*argv*/)
         auto future_result = prom->get_future();
 
         std::cout << "Waiting to discover device..." << std::endl;
-        dc.register_on_discover([prom](uint64_t uuid) {
-            std::cout << "Discovered device with UUID: " << uuid << std::endl;
+        dc.register_on_discover([prom](uint64_t uuid, uint8_t component_id) {
+            std::cout << "Discovered device with UUID: " << uuid
+                      << "Component ID: " << int(component_id) << std::endl;
             prom->set_value();
         });
 
@@ -68,8 +69,10 @@ int main(int /*argc*/, char ** /*argv*/)
         future_result.get();
     }
 
-    dc.register_on_timeout([](uint64_t uuid) {
-        std::cout << "Device with UUID timed out: " << uuid << std::endl;
+    dc.register_on_timeout([](uint64_t uuid, uint8_t component_id) {
+        std::cout << "Device with UUID: " << uuid
+                  << " and Component ID: " << int(component_id)
+                  << " timed out: " << std::endl;
         std::cout << "Exiting." << std::endl;
         exit(0);
     });
@@ -77,7 +80,7 @@ int main(int /*argc*/, char ** /*argv*/)
     // We don't need to specifiy the UUID if it's only one device anyway.
     // If there were multiple, we could specify it with:
     // dc.device(uint64_t uuid);
-    Device &device = dc.device();
+    Device &device = dc.autopilot();
     std::shared_ptr<Action> action = std::make_shared<Action>(&device);
     std::shared_ptr<Mission> mission = std::make_shared<Mission>(&device);
     std::shared_ptr<Telemetry> telemetry = std::make_shared<Telemetry>(&device);
--- a/example/fly_qgc_mission/fly_qgc_mission.cpp
+++ b/example/fly_qgc_mission/fly_qgc_mission.cpp
@@ -66,8 +66,9 @@ int main(int argc, char **argv)
         auto future_result = prom->get_future();
 
         std::cout << "Waiting to discover device..." << std::endl;
-        dc.register_on_discover([prom](uint64_t uuid) {
-            std::cout << "Discovered device with UUID: " << uuid << std::endl;
+        dc.register_on_discover([prom](uint64_t uuid, uint8_t component_id) {
+            std::cout << "Discovered device with UUID: " << uuid
+                      << "Component ID: " << int(component_id) << std::endl;
             prom->set_value();
         });
 
@@ -77,8 +78,10 @@ int main(int argc, char **argv)
         future_result.get();
     }
 
-    dc.register_on_timeout([](uint64_t uuid) {
-        std::cout << "Device with UUID timed out: " << uuid << std::endl;
+    dc.register_on_timeout([](uint64_t uuid, uint8_t component_id) {
+        std::cout << "Device with UUID: " << uuid
+                  << " and Component ID: " << int(component_id)
+                  << " timed out: " << std::endl;
         std::cout << "Exiting." << std::endl;
         exit(0);
     });
@@ -86,7 +89,7 @@ int main(int argc, char **argv)
     // We don't need to specify the UUID if it's only one device anyway.
     // If there were multiple, we could specify it with:
     // dc.device(uint64_t uuid);
-    Device &device = dc.device();
+    Device &device = dc.autopilot();
     std::shared_ptr<Action> action = std::make_shared<Action>(&device);
     std::shared_ptr<Mission> mission = std::make_shared<Mission>(&device);
     std::shared_ptr<Telemetry> telemetry = std::make_shared<Telemetry>(&device);
--- a/core/device.cpp
+++ b/core/device.cpp
@@ -166,7 +166,10 @@ void Device::process_heartbeat(const mavlink_message_t &message)
     // We do not call on_discovery here but wait with the notification until we know the UUID.
 
     /* If we don't know the UUID yet, we try to find out. */
-    if (_target_uuid == 0 && !_target_uuid_initialized) {
+    // Lets request autopilot version only from an Autopilot.
+    if (heartbeat.autopilot != MAV_AUTOPILOT_INVALID
+        && _target_uuid == 0
+        && !_target_uuid_initialized) {
         request_autopilot_version();
     }
 
@@ -357,7 +360,7 @@ void Device::set_connected()
 
         if (!_connected && _target_uuid_initialized) {
 
-            _parent->notify_on_discover(_target_uuid);
+            _parent->notify_on_discover(_target_uuid, _target_component_id);
             _connected = true;
 
             register_timeout_handler(std::bind(&Device::heartbeats_timed_out, this),
@@ -389,7 +392,7 @@ void Device::set_disconnected()
         //_heartbeat_timeout_cookie = nullptr;
 
         _connected = false;
-        _parent->notify_on_timeout(_target_uuid);
+        _parent->notify_on_timeout(_target_uuid, _target_component_id);
 
         // Let's reset the flag hope again for the next time we see this target.
         _target_uuid_initialized = false;
--- a/core/device.cpp
+++ b/core/device.cpp
@@ -15,8 +15,10 @@ namespace dronecore {
 using namespace std::placeholders; // for `_1`
 
 Device::Device(DroneCoreImpl *parent,
-               uint8_t target_system_id) :
+               uint8_t target_system_id,
+               uint8_t target_component_id) :
     _target_system_id(target_system_id),
+    _target_component_id(target_component_id),
     _parent(parent),
     _params(this),
     _commands(this),
@@ -29,9 +31,11 @@ Device::Device(DroneCoreImpl *parent,
         MAVLINK_MSG_ID_HEARTBEAT,
         std::bind(&Device::process_heartbeat, this, _1), this);
 
-    register_mavlink_message_handler(
-        MAVLINK_MSG_ID_AUTOPILOT_VERSION,
-        std::bind(&Device::process_autopilot_version, this, _1), this);
+    // Request for Autopilot version, only if its Autopilot
+    if (_target_component_id == MAV_COMP_ID_AUTOPILOT1)
+        register_mavlink_message_handler(
+            MAVLINK_MSG_ID_AUTOPILOT_VERSION,
+            std::bind(&Device::process_autopilot_version, this, _1), this);
 
     register_mavlink_message_handler(
         MAVLINK_MSG_ID_STATUSTEXT,
--- a/core/device.h
+++ b/core/device.h
@@ -33,7 +33,7 @@ public:
     };
 
     explicit Device(DroneCoreImpl *parent,
-                    uint8_t target_system_id);
+                    uint8_t target_system_id, uint8_t target_component_id);
     ~Device();
 
     void process_mavlink_message(const mavlink_message_t &message);
@@ -148,8 +148,7 @@ private:
 
     std::atomic<uint8_t> _target_system_id;
 
-    // The component ID is hardcoded for now.
-    uint8_t _target_component_id = MAV_COMP_ID_AUTOPILOT1;
+    uint8_t _target_component_id;
     uint64_t _target_uuid {0};
 
     int _target_uuid_retries = 0;
--- a/core/dronecore.cpp
+++ b/core/dronecore.cpp
@@ -56,22 +56,20 @@ Device &DroneCore::autopilot() const
     return _impl->get_autopilot();
 }
 
-#if 0
-Device &DroneCore::device(const uint64_t uuid) const
+Device &DroneCore::autopilot(uint64_t uuid) const
 {
-    return _impl->get_device(uuid);
+    return _impl->get_autopilot(uuid);
 }
 
-bool DroneCore::is_connected() const
+bool DroneCore::is_autopilot_connected() const
 {
-    return _impl->is_connected();
+    return _impl->is_autopilot_connected();
 }
 
-bool DroneCore::is_connected(const uint64_t uuid) const
+bool DroneCore::is_autopilot_connected(uint64_t uuid) const
 {
-    return _impl->is_connected(uuid);
+    return _impl->is_autopilot_connected(uuid);
 }
-#endif
 
 
 void DroneCore::register_on_discover(const event_callback_t callback)
--- a/core/dronecore.cpp
+++ b/core/dronecore.cpp
@@ -51,11 +51,12 @@ const std::vector<uint64_t> &DroneCore::device_uuids() const
     return _impl->get_device_uuids();
 }
 
-Device &DroneCore::device() const
+Device &DroneCore::autopilot() const
 {
-    return _impl->get_device();
+    return _impl->get_autopilot();
 }
 
+#if 0
 Device &DroneCore::device(const uint64_t uuid) const
 {
     return _impl->get_device(uuid);
@@ -70,6 +71,8 @@ bool DroneCore::is_connected(const uint64_t uuid) const
 {
     return _impl->is_connected(uuid);
 }
+#endif
+
 
 void DroneCore::register_on_discover(const event_callback_t callback)
 {
--- a/core/dronecore.h
+++ b/core/dronecore.h
@@ -97,16 +97,16 @@ public:
     const std::vector<uint64_t> &device_uuids() const;
 
     /**
-     * @brief Get the first discovered device.
+     * @brief Get the first discovered autopilot device.
      *
-     * This returns the first discovered device or a null device if no device has yet been found.
+     * This returns the first discovered autopilot device or a null device if no device has yet been found.
      *
-     * @return A reference to a device.
+     * @return A reference to an autopilot device.
      */
-    Device &device() const;
+    Device &autopilot() const;
 
     /**
-     * @brief Get the device with the specified UUID.
+     * @brief Get the autopilot device of the given system UUID.
      *
      * This returns a device for a given UUID if such a device has been discovered and a null
      * device otherwise.
@@ -114,27 +114,28 @@ public:
      * @param uuid UUID of device to get.
      * @return A reference to the specified device.
      */
-    Device &device(uint64_t uuid) const;
+    Device &autopilot(uint64_t uuid) const;
 
     /**
      * @brief Callback type for discover and timeout notifications.
      *
      * @param uuid UUID of device (or MAVLink system ID for devices that don't have a UUID).
      */
-    typedef std::function<void(uint64_t uuid)> event_callback_t;
+    typedef std::function<void(uint64_t uuid, uint8_t component_id)> event_callback_t;
 
     /**
-     * @brief Returns `true` if exactly one device is currently connected.
+     * @brief Returns `true` if exactly one autopilot device is currently connected.
      *
-     * Connected means we are receiving heartbeats from this device.
+     * Connected means we are receiving heartbeats from this autopilot device.
      * It means the same as "discovered" and "not timed out".
      *
-     * If multiple devices have connected, this will return `false`.
+     * If multiple autopilot devices have connected, this will return `false`.
      *
-     * @return `true` if exactly one device is connected.
+     * @return `true` if exactly one autopilot device is connected.
      */
-    bool is_connected() const;
+    bool is_autopilot_connected() const;
 
+#if 0
     /**
      * @brief Returns `true` if a device is currently connected.
      *
@@ -145,6 +146,7 @@ public:
      * @return `true` if device is connected.
      */
     bool is_connected(uint64_t uuid) const;
+#endif
 
     /**
      * @brief Register callback for device discovery.
--- a/core/dronecore.h
+++ b/core/dronecore.h
@@ -135,7 +135,6 @@ public:
      */
     bool is_autopilot_connected() const;
 
-#if 0
     /**
      * @brief Returns `true` if a device is currently connected.
      *
@@ -145,8 +144,7 @@ public:
      * @param uuid UUID of device to check.
      * @return `true` if device is connected.
      */
-    bool is_connected(uint64_t uuid) const;
-#endif
+    bool is_autopilot_connected(uint64_t uuid) const;
 
     /**
      * @brief Register callback for device discovery.
--- a/core/dronecore_impl.cpp
+++ b/core/dronecore_impl.cpp
@@ -275,29 +275,29 @@ Device &DroneCoreImpl::get_autopilot()
     return *_devices[system_id][component_id];
 }
 
-#if 0
-Device &DroneCoreImpl::get_device(const uint64_t uuid)
+Device &DroneCoreImpl::get_autopilot(const uint64_t uuid)
 {
     {
         std::lock_guard<std::recursive_mutex> lock(_devices_mutex);
         // TODO: make a cache map for this.
-        for (auto device : _devices) {
-            if (device.second->get_target_uuid() == uuid) {
-                return *device.second;
-            }
+        for (auto system_it = _devices.begin(); system_it != _devices.end(); ++system_it) {
+            for (auto component : system_it->second)
+                if (component->get_target_uuid() == uuid &&
+                    component->get_target_component_id() == MAV_COMP_ID_AUTOPILOT1) {
+                    return *component;
+                }
         }
+
+        // We have not found a device with this UUID.
+        // TODO: this is an error condition that we ought to handle properly.
+        LogErr() << "NO autopilot in UUID: " << uuid;
+
+        // Create a dummy
+        uint8_t system_id = 0, component_id = 0;
+        make_new_device(system_id, component_id);
+        return *_devices[system_id][component_id];
     }
-
-    // We have not found a device with this UUID.
-    // TODO: this is an error condition that we ought to handle properly.
-    LogErr() << "device with UUID: " << uuid << " not found";
-
-    // Create a dummy
-    uint8_t system_id = 0, component_id = 0;
-    create_device_if_not_existing(system_id, component_id);
-    return *_devices[system_id];
 }
-#endif
 
 bool DroneCoreImpl::is_autopilot_connected() const
 {
@@ -312,19 +312,19 @@ bool DroneCoreImpl::is_autopilot_connected() const
     return false;
 }
 
-#if 0
-bool DroneCoreImpl::is_connected(const uint64_t uuid) const
+bool DroneCoreImpl::is_autopilot_connected(uint64_t uuid) const
 {
     std::lock_guard<std::recursive_mutex> lock(_devices_mutex);
 
-    for (auto it = _devices.begin(); it != _devices.end(); ++it) {
-        if (it->second->get_target_uuid() == uuid) {
-            return it->second->is_connected();
-        }
+    if (_devices.size() == 1) {
+        for (auto component : _devices.begin()->second)
+            if (component->get_target_uuid() == uuid &&
+                component->get_target_component_id() == MAV_COMP_ID_AUTOPILOT1) {
+                return component->is_connected();
+            }
     }
     return false;
 }
-#endif
 
 bool DroneCoreImpl::make_new_device(uint8_t system_id, uint8_t comp_id)
 {
--- a/core/dronecore_impl.cpp
+++ b/core/dronecore_impl.cpp
@@ -28,8 +28,10 @@ DroneCoreImpl::~DroneCoreImpl()
         std::lock_guard<std::recursive_mutex> lock(_devices_mutex);
         _should_exit = true;
 
-        for (auto it = _devices.begin(); it != _devices.end(); ++it) {
-            delete it->second;
+        for (auto system_it = _devices.begin(); system_it != _devices.end(); ++system_it) {
+            for (auto component : system_it->second) {
+                delete component;
+            }
         }
         _devices.clear();
     }
@@ -68,13 +70,13 @@ void DroneCoreImpl::receive_message(const mavlink_message_t &message)
 
     // Change system id of null device
     if (_devices.find(0) != _devices.end()) {
-        auto null_device = _devices[0];
+        auto null_device = _devices.begin()->second[0];
         _devices.erase(0);
         null_device->set_target_system_id(message.sysid);
-        _devices.insert(std::pair<uint8_t, Device *>(message.sysid, null_device));
+        _devices.insert(std::pair<uint8_t, std::vector<Device *>>(message.sysid, {null_device}));
     }
 
-    create_device_if_not_existing(message.sysid, message.compid);
+    make_new_device(message.sysid, message.compid);
 
     if (_should_exit) {
         // Don't try to call at() if devices have already been destroyed
@@ -86,8 +88,12 @@ void DroneCoreImpl::receive_message(const mavlink_message_t &message)
         LogDebug() << "sysid: " << int(message.sysid);
     }
 
+    // Process the mavlink message of this component
     if (_devices.find(message.sysid) != _devices.end()) {
-        _devices.at(message.sysid)->process_mavlink_message(message);
+        for (auto component : _devices[message.sysid])
+            if (component->get_target_component_id() == message.compid) {
+                component->process_mavlink_message(message);
+            }
     }
 }
 
@@ -229,8 +235,9 @@ const std::vector<uint64_t> &DroneCoreImpl::get_device_uuids() const
     static std::vector<uint64_t> uuids = {};
     uuids.clear();
 
-    for (auto it = _devices.begin(); it != _devices.end(); ++it) {
-        uint64_t uuid = it->second->get_target_uuid();
+    for (auto system_it = _devices.begin(); system_it != _devices.end(); ++system_it) {
+        // All components will have same UUID, so get it from first component.
+        auto uuid = system_it->second[0]->get_target_uuid();
         if (uuid != 0) {
             uuids.push_back(uuid);
         }
@@ -239,36 +246,36 @@ const std::vector<uint64_t> &DroneCoreImpl::get_device_uuids() const
     return uuids;
 }
 
-Device &DroneCoreImpl::get_device()
+Device &DroneCoreImpl::get_autopilot()
 {
-    {
-        std::lock_guard<std::recursive_mutex> lock(_devices_mutex);
-        // In get_device withoiut uuid, we expect to have only
-        // one device conneted.
-        if (_devices.size() == 1) {
-            return *(_devices.at(_devices.begin()->first));
-        }
+    std::lock_guard<std::recursive_mutex> lock(_devices_mutex);
 
-        if (_devices.size() > 1) {
-            LogErr() << "Error: more than one device found:";
-
-            int i = 0;
-            for (auto it = _devices.begin(); it != _devices.end(); ++it) {
-                LogWarn() << "strange: " << i << ": " << int(it->first) << ", " << it->second;
-                ++i;
+    // If called without uuid, we expect to have only one autopilot conneted.
+    if (_devices.size() == 1) {
+        for (auto component : _devices.begin()->second) {
+            if (component->get_target_component_id() == MAV_COMP_ID_AUTOPILOT1) {
+                return *component;
             }
-
-            // Just return first device instead of failing.
-            return *_devices.begin()->second;
-        } else {
-            LogErr() << "Error: no device found.";
-            uint8_t system_id = 0, component_id = 0;
-            create_device_if_not_existing(system_id, component_id);
-            return *_devices[system_id];
         }
     }
+
+    if (_devices.size() > 1) {
+        LogErr() << "Error: more than one device found:";
+
+        // Just return first autopilot instead of failing.
+        for (auto component : _devices.begin()->second)
+            if (component->get_target_component_id() == MAV_COMP_ID_AUTOPILOT1) {
+                return *component;
+            }
+    }
+
+    LogErr() << "Error: no device found.";
+    uint8_t system_id = 0, component_id = 0;
+    make_new_device(system_id, component_id);
+    return *_devices[system_id][component_id];
 }
 
+#if 0
 Device &DroneCoreImpl::get_device(const uint64_t uuid)
 {
     {
@@ -290,17 +297,22 @@ Device &DroneCoreImpl::get_device(const uint64_t uuid)
     create_device_if_not_existing(system_id, component_id);
     return *_devices[system_id];
 }
+#endif
 
-bool DroneCoreImpl::is_connected() const
+bool DroneCoreImpl::is_autopilot_connected() const
 {
     std::lock_guard<std::recursive_mutex> lock(_devices_mutex);
 
     if (_devices.size() == 1) {
-        return _devices.begin()->second->is_connected();
+        for (auto component : _devices.begin()->second)
+            if (component->get_target_component_id() == MAV_COMP_ID_AUTOPILOT1) {
+                return component->is_connected();
+            }
     }
     return false;
 }
 
+#if 0
 bool DroneCoreImpl::is_connected(const uint64_t uuid) const
 {
     std::lock_guard<std::recursive_mutex> lock(_devices_mutex);
@@ -312,40 +324,51 @@ bool DroneCoreImpl::is_connected(const uint64_t uuid) const
     }
     return false;
 }
+#endif
 
-void DroneCoreImpl::create_device_if_not_existing(uint8_t system_id, uint8_t comp_id)
+bool DroneCoreImpl::make_new_device(uint8_t system_id, uint8_t comp_id)
 {
     std::lock_guard<std::recursive_mutex> lock(_devices_mutex);
 
     if (_should_exit) {
         // When the device got destroyed in the destructor, we have to give up.
-        return;
+        LogWarn() << "Device got destroyed";
+        return false;
     }
 
-    // existing already.
+    // If we already have a component from this system,
+    // add it to the list (ie, vector).
     if (_devices.find(system_id) != _devices.end()) {
-        //LogDebug() << "ID: " << int(system_id) << " exists already.";
-        return;
+        for (auto component : _devices[system_id]) {
+            // Don't add if the component already exists.
+            if (component->get_target_component_id() == comp_id) {
+                return false;
+            }
+        }
+        // Create a device
+        auto new_device = new Device(this, system_id, comp_id);
+        _devices[system_id].push_back(new_device);
+        return true;
     }
 
-    // Create both lists in parallel
-    Device *new_device = new Device(this, system_id, comp_id);
-    _devices.insert(std::pair<uint8_t, Device *>(system_id, new_device));
+    auto new_device = new Device(this, system_id, comp_id);
+    _devices.insert(std::pair<uint8_t, std::vector<Device *>>(system_id, {new_device}));
+    return true;
 }
 
-void DroneCoreImpl::notify_on_discover(const uint64_t uuid)
+void DroneCoreImpl::notify_on_discover(uint64_t uuid, uint8_t component_id)
 {
-    LogDebug() << "Discovered " << uuid;
+    LogDebug() << "Discovered component " << int(component_id) << " on vehicle " << uuid;
     if (_on_discover_callback != nullptr) {
-        _on_discover_callback(uuid);
+        _on_discover_callback(uuid, component_id);
     }
 }
 
-void DroneCoreImpl::notify_on_timeout(const uint64_t uuid)
+void DroneCoreImpl::notify_on_timeout(uint64_t uuid, uint8_t component_id)
 {
-    LogDebug() << "Lost " << uuid;
+    LogDebug() << "Lost component " << int(component_id) << " on vehicle " << uuid;
     if (_on_timeout_callback != nullptr) {
-        _on_timeout_callback(uuid);
+        _on_timeout_callback(uuid, component_id);
     }
 }
 
--- a/core/dronecore_impl.cpp
+++ b/core/dronecore_impl.cpp
@@ -74,7 +74,7 @@ void DroneCoreImpl::receive_message(const mavlink_message_t &message)
         _devices.insert(std::pair<uint8_t, Device *>(message.sysid, null_device));
     }
 
-    create_device_if_not_existing(message.sysid);
+    create_device_if_not_existing(message.sysid, message.compid);
 
     if (_should_exit) {
         // Don't try to call at() if devices have already been destroyed
@@ -262,8 +262,8 @@ Device &DroneCoreImpl::get_device()
             return *_devices.begin()->second;
         } else {
             LogErr() << "Error: no device found.";
-            uint8_t system_id = 0;
-            create_device_if_not_existing(system_id);
+            uint8_t system_id = 0, component_id = 0;
+            create_device_if_not_existing(system_id, component_id);
             return *_devices[system_id];
         }
     }
@@ -286,9 +286,8 @@ Device &DroneCoreImpl::get_device(const uint64_t uuid)
     LogErr() << "device with UUID: " << uuid << " not found";
 
     // Create a dummy
-    uint8_t system_id = 0;
-    create_device_if_not_existing(system_id);
-
+    uint8_t system_id = 0, component_id = 0;
+    create_device_if_not_existing(system_id, component_id);
     return *_devices[system_id];
 }
 
@@ -314,7 +313,7 @@ bool DroneCoreImpl::is_connected(const uint64_t uuid) const
     return false;
 }
 
-void DroneCoreImpl::create_device_if_not_existing(const uint8_t system_id)
+void DroneCoreImpl::create_device_if_not_existing(uint8_t system_id, uint8_t comp_id)
 {
     std::lock_guard<std::recursive_mutex> lock(_devices_mutex);
 
@@ -330,7 +329,7 @@ void DroneCoreImpl::create_device_if_not_existing(const uint8_t system_id)
     }
 
     // Create both lists in parallel
-    Device *new_device = new Device(this, system_id);
+    Device *new_device = new Device(this, system_id, comp_id);
     _devices.insert(std::pair<uint8_t, Device *>(system_id, new_device));
 }
 
--- a/core/dronecore_impl.h
+++ b/core/dronecore_impl.h
@@ -41,7 +41,7 @@ public:
     void notify_on_timeout(uint64_t uuid);
 
 private:
-    void create_device_if_not_existing(uint8_t system_id);
+    void create_device_if_not_existing(uint8_t system_id, uint8_t component_id);
 
     std::mutex _connections_mutex;
     std::vector<Connection *> _connections;
--- a/core/dronecore_impl.h
+++ b/core/dronecore_impl.h
@@ -32,9 +32,7 @@ public:
     Device &get_autopilot(uint64_t uuid);
 
     bool is_autopilot_connected() const;
-#if 0
-    bool is_connected(uint64_t uuid) const;
-#endif
+    bool is_autopilot_connected(uint64_t uuid) const;
 
     void register_on_discover(DroneCore::event_callback_t callback);
     void register_on_timeout(DroneCore::event_callback_t callback);
--- a/core/dronecore_impl.h
+++ b/core/dronecore_impl.h
@@ -28,26 +28,28 @@ public:
     ConnectionResult add_serial_connection(const std::string &dev_path, int baudrate);
 
     const std::vector<uint64_t> &get_device_uuids() const;
-    Device &get_device();
-    Device &get_device(uint64_t uuid);
+    Device &get_autopilot();
+    Device &get_autopilot(uint64_t uuid);
 
-    bool is_connected() const;
+    bool is_autopilot_connected() const;
+#if 0
     bool is_connected(uint64_t uuid) const;
+#endif
 
     void register_on_discover(DroneCore::event_callback_t callback);
     void register_on_timeout(DroneCore::event_callback_t callback);
 
-    void notify_on_discover(uint64_t uuid);
-    void notify_on_timeout(uint64_t uuid);
+    void notify_on_discover(uint64_t uuid, uint8_t component_id);
+    void notify_on_timeout(uint64_t uuid, uint8_t component_id);
 
 private:
-    void create_device_if_not_existing(uint8_t system_id, uint8_t component_id);
+    bool make_new_device(uint8_t ystem_id, uint8_t comp_id);
 
     std::mutex _connections_mutex;
     std::vector<Connection *> _connections;
 
     mutable std::recursive_mutex _devices_mutex;
-    std::map<uint8_t, Device *> _devices;
+    std::map<uint8_t, std::vector<Device *>> _devices;
 
     DroneCore::event_callback_t _on_discover_callback;
     DroneCore::event_callback_t _on_timeout_callback;
--- a/example/follow_me/follow_me.cpp
+++ b/example/follow_me/follow_me.cpp
@@ -43,13 +43,13 @@ int main(int, char **)
     connection_error_exit(conn_result, "Connection failed");
 
     // Wait for the device to connect via heartbeat
-    while (!dc.is_connected()) {
+    while (!dc.is_autopilot_connected()) {
         std::cout << "Wait for device to connect via heartbeat" << std::endl;
         sleep_for(seconds(1));
     }
 
     // Device got discovered.
-    Device &device = dc.device();
+    Device &device = dc.autopilot();
     std::shared_ptr<Action> action = std::make_shared<Action>(&device);
     std::shared_ptr<FollowMe> follow_me = std::make_shared<FollowMe>(&device);
     std::shared_ptr<Telemetry> telemetry = std::make_shared<Telemetry>(&device);
--- a/example/offboard_velocity/offboard_velocity.cpp
+++ b/example/offboard_velocity/offboard_velocity.cpp
@@ -175,13 +175,13 @@ int main(int, char **)
     connection_error_exit(conn_result, "Connection failed");
 
     // Wait for the device to connect via heartbeat
-    while (!dc.is_connected()) {
+    while (!dc.is_autopilot_connected()) {
         std::cout << "Wait for device to connect via heartbeat" << std::endl;
         sleep_for(seconds(1));
     }
 
     // Device got discovered.
-    Device &device = dc.device();
+    Device &device = dc.autopilot();
     std::shared_ptr<Action> action = std::make_shared<Action>(&device);
     std::shared_ptr<Offboard> offboard = std::make_shared<Offboard>(&device);
     std::shared_ptr<Telemetry> telemetry = std::make_shared<Telemetry>(&device);
--- a/example/takeoff_land/takeoff_and_land.cpp
+++ b/example/takeoff_land/takeoff_and_land.cpp
@@ -44,8 +44,9 @@ int main(int argc, char **argv)
     }
 
     std::cout << "Waiting to discover device..." << std::endl;
-    dc.register_on_discover([&discovered_device](uint64_t uuid) {
-        std::cout << "Discovered device with UUID: " << uuid << std::endl;
+    dc.register_on_discover([&discovered_device](uint64_t uuid, uint8_t component_id) {
+        std::cout << "Discovered device with UUID: " << uuid
+                  << " Component Id: " << int(component_id) << std::endl;
         discovered_device = true;
     });
 
@@ -60,10 +61,8 @@ int main(int argc, char **argv)
     // We don't need to specify the UUID if it's only one device anyway.
     // If there were multiple, we could specify it with:
     // dc.device(uint64_t uuid);
-    Device &device = dc.device();
-
-    auto telemetry = std::make_shared<Telemetry>(&device);
-    auto action = std::make_shared<Action>(&device);
+    auto telemetry = std::make_shared<Telemetry>(&dc.autopilot());
+    auto action = std::make_shared<Action>(&dc.autopilot());
 
     // We want to listen to the altitude of the drone at 1 Hz.
     const Telemetry::Result set_rate_result = telemetry->set_rate_position(1.0);
--- a/example/transition_vtol_fixed_wing/transition_vtol_fixed_wing.cpp
+++ b/example/transition_vtol_fixed_wing/transition_vtol_fixed_wing.cpp
@@ -31,8 +31,9 @@ int main(int /*argc*/, char ** /*argv*/)
     }
 
     std::cout << "Waiting to discover device..." << std::endl;
-    dc.register_on_discover([&discovered_device](uint64_t uuid) {
-        std::cout << "Discovered device with UUID: " << uuid << std::endl;
+    dc.register_on_discover([&discovered_device](uint64_t uuid, uint8_t component_id) {
+        std::cout << "Discovered device with UUID: " << uuid
+                  << "Component ID: " << int(component_id) << std::endl;
         discovered_device = true;
     });
 
@@ -47,7 +48,7 @@ int main(int /*argc*/, char ** /*argv*/)
     // We don't need to specify the UUID if it's only one device anyway.
     // If there were multiple, we could specify it with:
     // dc.device(uint64_t uuid);
-    Device &device = dc.device();
+    Device &device = dc.autopilot();
     std::shared_ptr<Telemetry> telemetry = std::make_shared<Telemetry>(&device);
 
     // We want to listen to the altitude of the drone at 1 Hz.
--- a/integration_tests/async_connect.cpp
+++ b/integration_tests/async_connect.cpp
@@ -9,9 +9,10 @@ using namespace std::placeholders; // for _1
 static bool _discovered_device = false;
 static bool _timeouted_device = false;
 static uint64_t _uuid = 0;
+static uint8_t _component_id = 0;
 
-void on_discover(uint64_t uuid);
-void on_timeout(uint64_t uuid);
+void on_discover(uint64_t uuid, uint8_t component_id);
+void on_timeout(uint64_t uuid, uint8_t component_id);
 
 TEST_F(SitlTest, AsyncConnect)
 {
@@ -19,8 +20,8 @@ TEST_F(SitlTest, AsyncConnect)
 
     ASSERT_EQ(dc.add_udp_connection(), ConnectionResult::SUCCESS);
 
-    dc.register_on_discover(std::bind(&on_discover, _1));
-    dc.register_on_timeout(std::bind(&on_timeout, _1));
+    dc.register_on_discover(std::bind(&on_discover, _1, _2));
+    dc.register_on_timeout(std::bind(&on_timeout, _1, _2));
 
     while (!_discovered_device) {
         std::cout << "waiting for device to appear..." << std::endl;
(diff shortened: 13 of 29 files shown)


JonasVautherin:
I am wondering one thing: why do we need dronecore to handle multiple devices? Isn't it enough to create one dronecore instance for each device and keep that complexity out of the core libraries?

More specifically, is it limiting for the user if, instead of doing something like:

```
DroneCore dc = new DroneCore();
dc.takeoff(UUID1);
dc.takeoff(UUID2);
```

the user has to do something more like:

```
DroneCore dc1 = new DroneCore(UUID1);
DroneCore dc2 = new DroneCore(UUID2);
dc1.takeoff();
dc2.takeoff();
```

shakthi-prashanth-m:
When you create multiple DroneCore instance, you will get to a situation in binding same port number twice. If vehicle (autopilot & other components) are sending heartbeat to 14540, then first DroneCore instance will bind onto it, not allowing other one to handle other component.

Also by design, DroneCore **Device** abstracts any MAVLink component, autopilot is a special case. We can very well handle just a standalone camera with a server support using current infrastructure.

shakthi-prashanth-m:
@JonasVautherin 
>I am wondering one thing: why do we need dronecore to handle multiple devices?

Its not multiple devices, but components. You may check #304 for issue details.

JonasVautherin:
Yeah sorry, bad wording on my side :-).

I'll have a look at the issue ASAP.

julianoes:
@shakthi-prashanth-m I think this API is too complex. I think all we want to connect to is a mavlink device and that device then happens to be e.g.:
- an autopilot
- an autopilot with camera connected
- an autopilot with gimbal and camera
- an autopilot with onboard computer
- an autopilot with multiple cameras connected
- a camera alone
- a camera with gimbal

If you now want to create an interface for all these combinations it will get messy. That's why I vote to just have a `device` and behind that device can be whatever is supported by plugins, etc.

hamishwillee:
@julianoes Agree that simpler is better, but I don't understand how https://github.com/dronecore/DroneCore/pull/310#issuecomment-371283195 would work in practice (as code).

Perhaps rambling but ...

This is a case where our old plugin model made more sense from the user perspective. In that case you would have a Device (lets call it a MAVLink "system") made up of components. DroneCore could then build up a picture of the system based on the SysIDs and component IDs and you would query the system for list of available cameras, autopilot, servos etc. 

In our current design though the plugins are created "manually" as needed. That means that the coder will have to know about all the camera configurations on all the vehicles in order to build up the picture of the vehicle(s) OR each component will have to be its own device, not associated with any other components (devices).

I "hate" it, but I think the simplest API is probably the later - every component is a separate `Device` and we lose the idea of complete systems. You can attach any plugin to any device, but of course a camera plugin will only work when attached to a Camera Device (will ignore the other mavlink messages), and similarly if you add an action plugin to a Camera device it will ignore those messages.

shakthi-prashanth-m:
How about this ?
Let's make a hash key out of uuid and component id combination and store one device per entry in the map. No vector. 
When a new component of the same system is discovered then we simply generate a hash key for that and store it.
We can return this hash key as a handle in discovery callback. 
So each device is identified by hash key which we can call DroneCore device identifier. 
Device could be any MAVLink device . Sounds good ?
So we provide methods in `Device` class to know what kind of device it is.
Let's keep device abstraction itself.  But let's change UUID to DevId. 
```
Device &device = dc.device(devid);
if (device.is_autopilot())
....
if (device.is_camera())
....
if (device.is_gimbal())
....
```

julianoes:
> I "hate" it, but I think the simplest API for both implementation and users is probably to treat every component as a separate Device (referenced by sysid and component id) and we lose the idea of complete systems. You can attach any plugin to any device, but of course a camera plugin will only work when attached to a Camera Device (will ignore the mavlink messages it doesn't understand), and similarly if you add an action plugin to a Camera device it will ignore those messages.

At some point you might throw everything about DroneCore out and go back to mavlink :smile: .

@shakthi-prashanth-m I think a device should contain both autopilot and camera as laid out before. Therefore, you could have:
```
DroneCore dc;
dc.add_udp_connection(bla);
dc.add_udp_connection(foo);
Device &device = dc.device(/* or uuid for specific device */);
if (device.has_camera()) {
    // do camera things
}
```

hamishwillee:
Yeah, lets go back to MAVLink :-)

OK, so in 
```
DroneCore dc;
dc.add_udp_connection(bla);
dc.add_udp_connection(foo);
Device &device = dc.device(/* or uuid for specific device */);
if (device.has_camera()) {
    // do camera things
}
```
Is your idea than the device itself creates camera objects using a camera plugin? Or does it just know that something is a camera, hand over a component id, and you create the plugin with that?

julianoes:
> Is your idea than the device itself creates camera objects using a camera plugin? Or does it just know that something is a camera, hand over a component id, and you create the plugin with that?

A camera plugin goes to the core and says hey let me talk to component ID of the camera.

shakthi-prashanth-m:
@julianoes we don't maintain components in device! We should ask to `DroneCore` class isn't it?

julianoes:
> we don't maintain components in device! 

We should though and that's where I agree with you.

shakthi-prashanth-m:
Can we safely assume System ID what we get in heartbeat is unique to identify a system ?
I know that UID is unique for autopilot hardware (if provided).
What about a stand alone MAVLink camera? I see in CSD, it fills system ID as 1. (Not sure need to check with Lalit)
How does a system choose it's system ID?

If we make System ID unique, the combination of System ID and component ID uniquely identifies a MAVLink device.
Makes sense?

hamishwillee:
> How does a system choose it's system ID?

Any system can use whatever id it likes, provided there are no conflicts (other devices on the system that use the same ID).  GCS tend to use 255 (true I believe for QGC, Mission Planner, Droid Planner). The default for PX4 is "1" (set in MAV_SYS_ID). DroneCore must have some standard id set too.

So in normal case if you were setting up a system of GCS and vehicle there would be no conflict. If you had multiple vehicles you'd set them up safely by giving consecutive numbers up from 1.

> Can we safely assume System ID what we get in heartbeat is unique to identify a system ?

In a properly set up MAVLink network that would be correct. However it is quite possible for people to change the system ID, or for there to be accidental clashes. 

So far we haven't assumed this - because we have a UUID in most cases that we know we can trust - we only use SYSID if we don't have UUID ... and then we trust it.

> What about a stand alone MAVLink camera? I see in CSD, it fills system ID as 1. (Not sure need to check with Lalit)

For CSD this is configurable in the [mavlink section of the config file](https://hamishwillee.gitbooks.io/camera-streaming-daemon/content/en/guide/configuration_file.html#mavlink). The default is actually 42, but must be over-ridden in each instance to match the associated system. 

> If we make System ID unique, the combination of System ID and component ID uniquely identifies a MAVLink device.
Makes sense?

Yes. Though the UUID is more unique :-), so combo of UUID and component ID perhaps :-)

shakthi-prashanth-m:
Thanks @hamishwillee.
@julianoes: If it's reasonably safe to use system ID, then we better use combo of system ID and comp ID to identify a MAVLink device.
Because UUID
- is that of autopilot hardware, can't be used for standalone MAVLink devices.
- comes in autopilot version MAVLink message, which we don't request if device is not autopilot. 

So,  how about having
```
Device &DroneCore::device(devid_t dev_id);
```
to get ALL MAVLink devices. `dev_id` is a unique device id (sysid, compid combo).
`dev_id` is notified to application instead of UUID. 

We could maintain `std::map<devid_t, Device *>` one entry per device. 
If an autopilot comes with a camera there'll be two entries. 
With that we can provide support to all MAVLink devices.

shakthi-prashanth-m:
We can still keep uuid discovery as well. And we can also query autopilot using uuid. 
```
Device &DroneCore::device(uint64t uuid);
```
To get access to autopilot of given UUID.

hamishwillee:
@lbegani
Actually I do see a problem. The component ID is specified in the [mavlink](https://hamishwillee.gitbooks.io/camera-streaming-daemon/content/en/guide/configuration_file.html#mavlink) section here. But note, *there is only one id field*. 
- How can a single system then declare multiple separate camera components? 
- There are 6 component IDs for cameras. What will CSD do for the component ID if a 7th camera is detected?

lbegani:
> How can a single system then declare multiple separate camera components?

component-id  field is not used. Component ID is assigned dynamically to the cameras as they get detected. Perhaps need to clean that up or add a note
> There are 6 component IDs for cameras. What will CSD do for the component ID if a 7th camera is detected?

CSD will print error message and ignore additional cameras once hit the limit. It will continue running with the components

julianoes:
> Can we safely assume System ID what we get in heartbeat is unique to identify a system ?

It's a mavlink limitation and therefore we need to assume it. It is up to the integrator of the whole system to make sure the system IDs don't clash and match up within a "system".

shakthi-prashanth-m:
@julianoes then it's good news.

shakthi-prashanth-m:
@julianoes @hamishwillee @JonasVautherin 

dronecore_multi_comp2

- A system may contain several components.
- Each system has a unique System ID and a unique Component ID for each of its components.
- Each component of a system is also termed as **Device**.
- Devices are discovered and grouped by system they belong to.
- To get access to any device of a system (For eg: autopilot, camera, gimbal, etc.), we expose DeviceID of  type `device_id_t` which has both System and Component IDs.
- Discovered Device IDs will look like **(1, 1), (2, 100), (3, 1), (3, 100), (3, 154)**, etc.
- Based on the device type, we can choose plugin to work on it. If the device is a camera, we instantiate camera plugin, etc. (See example below).

So, the application code looks like:
```cpp
if (discovered) {
    Device &device = dc.device(device_id);

    if (dc.is_autopilot(device_id))
        takeoff_land(deSo,vice);
    else if (dc.is_camera(device_id))
        control_camera(device);
    else
        cout << "No device found, exiting." << endl;
    
    return 0;
}
```
Full example that uses new interface can be found in Takeoff land example.

**Note:**
As said earlier, Using System ID is better over UUID, because:
1. Unlike System ID, UUID is specific to autopilot hardware.
2. We can't use it to control non-autopilot devices (for. eg MAVLink camera device)
3. It may not be provided by hardware vendor.
3. Its micro-controller (or NuttX) specific.
See here:
- https://github.com/PX4/Firmware/blob/ebe74ccb2efba5d47931e1d2327c61707e65775b/src/modules/mavlink/mavlink_main.cpp#L1305
- https://github.com/PX4/Firmware/blob/665ab03f5cfe2dbc4707c2b2fd3018b98ba8033f/src/drivers/boards/common/stm32/board_identity.c#L72-L74

### With very few changes below, we can support ALL MAVLink devices across systems:

- Use `device_id_t` instead of UUID (Method to access device remains same as earlier)
- Provide interface discover Device IDS and to check whether a device is an autopilot, camera, gimbal. Right now we can support only those whose plugins are available.
- To control multiple systems 1, 2, 3, 4 etc (see figure above), application may need system ids. So, we need to provide an interface to get System IDs as well.
- Methods in `Device` class to know about device type.
- Request autopilot version only if its autopilot.

Latest commit has these changes (**NOT updated integration and unit tests yet**)! Takeoff land example is built with this interface for sample.

Console logs for takeoff land example:
```
Waiting to discover device...
[06:07:42|Info ] New device on: 127.0.0.1:14557 (udp_connection.cpp:206)
[06:07:42|Debug] MAVLink: info: DISARMED by auto disarm on land (device.cpp:261)
[06:07:42|Debug] MAVLink: info: data link #0 lost (device.cpp:261)
[06:07:42|Debug] MAVLink: info: data link #0 regained (device.cpp:261)
[06:07:43|Debug] Discovered component 1 on system 1 (dronecore_impl.cpp:442)
Discovered component 1 on system 1
Handle autopilot
Vehicle is getting ready to arm
Altitude: 0.002 m
Arming...
Taking off...
[06:07:45|Debug] MAVLink: info: ARMED by arm/disarm component command (device.cpp:261)
[06:07:45|Debug] MAVLink: info: [logger] file: rootfs/fs/microsd/log/2018-03-12/1 (device.cpp:261)
[06:07:45|Debug] MAVLink: info: Using minimum takeoff altitude: 2.50 m (device.cpp:261)
[06:07:45|Debug] MAVLink: info: Takeoff detected (device.cpp:261)
[06:07:45|Debug] MAVLink: critical: Using minimum takeoff altitude: 2.50 m (device.cpp:261)
Altitude: 0 m
Altitude: 1.232 m
Altitude: 1.978 m
Altitude: 1.904 m
Altitude: 1.802 m
Altitude: 1.789 m
Altitude: 1.788 m
Altitude: 1.784 m
Altitude: 1.791 m
Altitude: 1.796 m
Landing...
[06:07:55|Debug] MAVLink: info: Landing at current position (device.cpp:261)
Altitude: 1.773 m
Altitude: 1.118 m
Altitude: 0.354 m
Altitude: -0.278 m
Altitude: -0.261 m
[06:07:59|Debug] MAVLink: info: Landing detected (device.cpp:261)
Finished...
[06:08:00|Warn ] Device got destroyed (dronecore_impl.cpp:414)
```
Kindly review and share your feedback.

shakthi-prashanth-m:
As the latest commit doesn't include Integration tests and Unit tests, CI and Appveyor build will fail. As this is a proposal commit, kindly review the design changes. I shall update them once review is approved.

julianoes:
@shakthi-prashanth-m why don't we use a `Device` for what you call a "system"?

shakthi-prashanth-m:
@julianoes I am fine with it. As you know

- Device is a uniquely identifiable component.
- System is a logical container of devices.

I thought the concept of "system" shall make app developers convenient to control multiple vehicles.
I think however, it can also be done without having another API for System ID discovery.

Suppose, I have two vehicles, each of them has autopilot, front & down camera.
How do I ask DroneCore

- first vehicle to do Offboard control with video streaming ?
- second vehicle to Follow Me ?

 ```cpp
auto device_ids = dc.device_ids(); // Gets 6 device ids: (1, 100), (1, 101), (2, 100) & (2, 101)
``` 
```cpp
for (auto device_id: device_ids) {
  Device &device = dc.device(device_id);
  
  switch (device_id.system_id) {
  case 1:
    if (dc.is_autopilot(device_id))
      offboard_control(device);
    else if (dc.is_camera(device_id))
      control_camera(device);
  case 2:
    if (dc.is_autopilot(device_id))
      follow_me(device);
    else if (dc.is_camera(device_id))
      control_camera(device);
  }
  ...
 }
}
```
Does it sound OK ?
```

**Comment by JonasVautherin** [comment]
Very interesting, good work @shakthi-prashanth-m!

I must say I am not yet really aware of how the mavlink stuff works in DroneCore, but I'll try to give some feedback anyway.

1. I am a bit confused by the "device" name. So is a device an instance of the `Device` (device.h) class? Or is it different in this context?
2. Does this assume that all the mavlink components of a "system" are emitting on the same port (e.g. 14540)?
3. This seems to assume that the components of a system advertise themselves using mavlink. What if a component is not mavlink-enabled? How do I deal with a PTP camera, for instance (e.g. a GoPro)? That could be a ptp-camera plugin that could only run onboard the drone, right?
4. It feels like `device.h` is discovering the different devices, and then you can create a `device.h` using one of them. Is that right?

More generally, I don't like the idea of having one device connecting to multiple systems/drones. On the frontend side, I was seeing `core` as being the entry point for __one drone__. So I would create my `DroneCore dc` object and, from it, I would get information about the components running on my drone.

Say my drone has an autopilot, a gimbal and a camera, I would do something like:

```
DroneCore dc = new DroneCore(); // This will discover and connect to the first drone it finds, ignoring the others

for (PluginInfo pluginInfo : dc.getPluginInfos()) {
    log.d(pluginInfo.getName() + ": " + pluginInfo.getId());
}

----- Output: ------
autopilot: 0
gimbal: 23
camera: 142
```

But maybe that becomes too complex when we have multiple components of the same type (e.g. multiple cameras)?

**Comment by julianoes** [comment]
@shakthi-prashanth-m sorry, I shouldn't be so blunt! Let me try again.

I really like your last diagram where you explain the possibilities, and it makes all sense. We are only disagreeing on terminology and how the API should look like.

I always envisioned a `Device` to be basically what you refer to as a system. That's why I keep saying, yes a `Device` should contain the components. I can see how that's confusing because the word `Device` is misleading and something like `Drone` would be better. However, `Drone` limits us to just aerial things, so an alternative would be `Robot`. I didn't choose `System` because a system can be everything or nothing. A `System` can include the ground station etc. I'm talking in general and not in the mavlink sense.

This means we need to chose a name for this thing that contains multiple components. Since it is already a system in the mavlink sense, we might as well use that.

Now, for the API, I think, we don't want to expose as much as you suggest. All that is required in my opinion is:
```
bool System::has_camera();
bool System::has_autopilot();
bool System::has_gimbal();
```

But the component IDs etc. should be hidden because that's mavlink specific and should not be leaked.

**Comment by hamishwillee** [comment]
@julianoes 
Re https://github.com/dronecore/DroneCore/pull/310#issuecomment-372340878

```
bool System::has_camera();
bool System::has_autopilot();
bool System::has_gimbal();
```

I agree component ID should be hidden from the API (particularly since for CSD we can't rely on same camera getting same ID each time vehicle restarted). I am happy enough with Device being called a System.

I don't think the API above works for cameras because there may be more than one - ie you can't just say "if has_camera() create the camera".

**Comment by JonasVautherin** [comment]
I agree with @julianoes in that `System` is not less confusing than `Device` to me. `Robot`, `Vehicle` or `Drone` work well, but may be limiting (even though one might call a handheld gimbal "robot", if that makes it clearer). Just my 2 cents, I don't have a strong opinion on that :-).

@shakthi-prashanth-m: could we say that `dronecore` keeps only one system id (e.g. the first one it detects, or the one passed as parameter) and ignores the other systems? I really believe that one `Device` should only maintain components of one system.

**Comment by shakthi-prashanth-m** [comment]
@julianoes  Ah! Until now I thought `Device` class represents one component!
I thought `core` represents _a discovered component_. So, as many are discovered, so many `Device` instances. `DroneCore` class maintains discovered devices in a map. So If a system has autopilot, camera and gimbal, then `DroneCore` class will have 1 entry in `_devices` map with system id as key and vector of 3 `Device` instances  as value (see below).
Key: `uint8_t`, Value: `std::vector<Device*>`
1: 1, 100, 154

It also made sense to me because, `Device` class has details that a component should have: system id, comp id, etc and `device_thread` handling MAVLink messages of **that** component only. 
I think its simple enough, isn't it ?

OK, Now when you said that, you designed `Device` class originally to represent an entire vehicle/robot, then I did re-look everything accordingly (i.e considering `core` is per vehicle/robot/system). So, now we are doing renaming `Device` to `System`.

But I have some questions about it.

- Why does `Device` class has members like `get_target_system_id()`, `get_target_component_id()`, because, its suppose to have represent entire system with several components inside it ? Whose component Id do we refer by that ?
- We don't have a class to represent a component, isn't it ? Do we just need a vector of component Ids ?
- When autopilot sends heartbeat, we just created `Device` instance with autopilot component id hard-coded. But ideally, we should maintain a vector of components, each of which should have its properties, isn't it?
- Same is for non-autopilot components. When other components of a vechicle/system send heartbeat, then we can simply append a new component to the existing vector. 
- We see that plugins `action`, `mission`, `follow_me`, `telemetry`, deal only with autopilot component; `gimbal` suppose to deal with gimbal component. `camera` plugin should deal with camera component (considering multiple cameras too). Each of the plugin holds a pointer to `Device` instance.  When a plugin sends a MAVLink message, it is intended for a specific component. So, plugins should make a decision whom the message is for (see below offboard plugin code)

```cpp
    mavlink_msg_set_position_target_local_ned_pack(_parent->get_own_system_id(),
                                                   _parent->get_own_component_id(),
                                                   &message,
                                                   static_cast<uint32_t>(_parent->get_time().elapsed_s() * 1e3),
                                                   _parent->get_target_system_id(),
                             /* here */ =>         _parent->get_target_component_id(),
                                                   MAV_FRAME_LOCAL_NED,
                                                   IGNORE_X | IGNORE_Y | IGNORE_Z |
                                                   IGNORE_AX | IGNORE_AY | IGNORE_AZ |
                                                   IGNORE_YAW_RATE,
                                                   x, y, z, vx, vy, vz, afx, afy, afz,
                                                   yaw, yaw_rate);
```
 
For multiple camera scenario, we need to let application choose it, because all the camera components handled by a single camera plugin. Isn't it ?

shakthi-prashanth-m:
@JonasVautherin 
>@shakthi-prashanth-m: could we say that dronecore keeps only one system id (e.g. the first one it detects, or the one passed as parameter) and ignores the other systems? I really believe that one Device should only maintain components of one system.

Yes, DroneCore does detect other systems. As said earlier, it doesn't detect non-first components of a system. On a lighter note,  _It assumes there's already an entry for this system, why add again ?_ :-)
So, the necessary changes to support multiple component would be to just add component list in the `Device` instance of **that** system id.

shakthi-prashanth-m:
Now I understand: the only thing we need to take care in handling multiple components of a system is **Target component ID**.
IMO, **each plugin impl** should take care of passing the right component id while composing MAVLink message to its component. Application interface need not worry about it.

However, in case of camera, there would be multiple cameras; so, **camera plugin impl** should maintain a list of camera component IDs (_can default to 1st camera_) and should also expose a public method resembling as below:
```cpp
std::vector<camera_id_t> Camera::camera_ids(uint8_t system_id);
```

shakthi-prashanth-m:
@julianoes , this is probably why you were suggesting API names `has_camera()`, `has_gimbal()`; and I was suggesting `is_camera()`, `is_autopilot()` etc :)

shakthi-prashanth-m:
@JonasVautherin 
>I am a bit confused by the "device" name. So is a device an instance of the Device (device.h) class? Or is it different in this context?

I misunderstood about `Device` class. Now I get that there'll be an instance of `Device` class for each  System discovered (_see System in my second diagram above_).

>Does this assume that all the mavlink components of a "system" are emitting on the same port (e.g. 14540)?

[Gazebo simulated camera](https://github.com/PX4/sitl_gazebo/blob/84e14e29497862d0cd04f4d42e43315a95b87bf9/src/gazebo_geotagged_images_plugin.cpp#L440) sends heartbeat to UDP port 14550. We may add both 14550 & 14540 UDP ports in DroneCore application.

>This seems to assume that the components of a system advertise themselves using mavlink. What if a component is not mavlink-enabled?

We only handle MAVLink based systems in DroneCore isn't it ? As I understand, `Device` is a MAVLink abstraction and obviously can't be used for non-MAVLink based systems. Did I miss something ?

>How do I deal with a PTP camera, for instance (e.g. a GoPro)? That could be a ptp-camera plugin that could only run onboard the drone, right?

In that case, we need to have a class that abstracts PTP based systems as part of DroneCore (I don't think we plan to do that, please correct me). 

>It feels like device.h is discovering the different devices, and then you can create a device.h using one of them. Is that right?

Right, discovered using MAVLink Heartbeat.

>More generally, I don't like the idea of having one device connecting to multiple systems/drones.

No, its not.
>On the frontend side, I was seeing core as being the entry point for one drone. So I would create my DroneCore dc object and, from it, I would get information about the components running on my drone.

Right.

shakthi-prashanth-m:
@hamishwillee 
>I don't think the API above works for cameras because there may be more than one - ie you can't just say "if has_camera() create the camera".

Does it make sense: https://github.com/dronecore/DroneCore/pull/310#issuecomment-372550632 ?

hamishwillee:
>> I don't think the API above works for cameras because there may be more than one - ie you can't just say "if has_camera() create the camera".

>v Does it make sense: #310 (comment) ?

No. Or to be more precise, I can see what you might be saying, and what @julianoes might be saying, but I am not sure they are the same thing. 

I'm going to wait and see what the end of the discussion is.

julianoes:
@shakthi-prashanth-m 
> - Why does Device class has members like get_target_system_id(), get_target_component_id(), because, its suppose to have represent entire system with several components inside it ? Whose component Id do we refer by that ?

You're exactly right. These don't really make sense and are very error prone. We've had quite a few cases where things would work sometimes, and sometimes not. The reason was that the target component ID was different depending on which heartbeat was received by a device first and therefore set and used.

> - We don't have a class to represent a component, isn't it ? Do we just need a vector of component Ids ?

I think just a vector should be enough. We'll see as we go.


> - When autopilot sends heartbeat, we just created Device instance with autopilot component id hard-coded. But ideally, we should maintain a vector of components, each of which should have its properties, isn't it?

I agree, we could do that more sophisticated.

> - Same is for non-autopilot components. When other components of a vechicle/system send heartbeat, then we can simply append a new component to the existing vector.

Right.

> - We see that plugins action, mission, follow_me, telemetry, deal only with autopilot component; gimbal suppose to deal with gimbal component. camera plugin should deal with camera component (considering multiple cameras too). Each of the plugin holds a pointer to Device instance. When a plugin sends a MAVLink message, it is intended for a specific component. So, plugins should make a decision whom the message is for (see below offboard plugin code)

Correct! And that's already how many of them work actually, see:
https://github.com/dronecore/DroneCore/blob/develop/plugins/action/action_impl.cpp#L107

@hamishwillee 
> I don't think the API above works for cameras because there may be more than one - ie you can't just say "if has_camera() create the camera".

Right, we might have to use something like:
```
has_camera() // for default
has_camera(0) // for default
has_camera(1) // for MAV_COMP_ID_CAMERA2
has_camera(2) // for MAV_COMP_ID_CAMERA3
```

shakthi-prashanth-m:
OK! Thanks @julianoes for answering. I shall get back soon.

JonasVautherin:
@julianoes:

> Correct! And that's already how many of them work actually, see:
https://github.com/dronecore/DroneCore/blob/develop/plugins/action/action_impl.cpp#L107

Should we initialize one plugin for each component, e.g. having two camera plugin instances if we have two cameras? Or do you think that a camera plugin should manage multiple compatible cameras?

julianoes:
Decisions from today's meeting:
- Rename `Device` to `System`
- Initiate a camera plugin per camera component. Do init given a camera ID (or default ?)

shakthi-prashanth-m:
**Note:** Assuming that `Device` => `System`.
 
There's another important point we need to consider about exposing APIs like `has_camera()`, etc: Class `System` _is NOT exposed to applications_. Applications have access to  `DroneCore` class.

So, we have to make them part of `DroneCore` instead:
```cpp
bool DroneCore::has_camera(uint64_t uuid); // check on the system with the given UUID
bool DroneCore::has_camera(); // check on the first discovered system
```
`DroneCore` will get it from `System`.

lbegani:
```
bool DroneCore::has_camera(uint64_t uuid); // check on the system with the given UUID
bool DroneCore::has_camera(); // check on the first discovered system
```
I would rather prefer an asynchronous flow, something like this - 
1. The app must be able to subscribe with dronecore for camera devices, by enabling some flag during dronecore init.
2. The app will receive a callback whenever there is a new camera device discovered

julianoes:
> There's another important point we need to consider about exposing APIs like `has_camera()`, etc: Class `System` is NOT exposed to applications. Applications have access to DroneCore class.

Why? Isn't it:
```
DroneCore dc = DroneCore();
dc.add_connection(...);
System &system dc.system();
if (system.has_camera()) {
   ...
}
```

shakthi-prashanth-m:
@julianoes Do we also allow applications to include `<system.h>` ? I think its not a good idea. As `system.h` is very MAVLink, its better to interface via `DroneCore` class. What do you think ?

shakthi-prashanth-m:
If we're exposing `system.h`, then we're indirectly letting applications to use its public methods, isn't it ?

shakthi-prashanth-m:
@julianoes if not, why do we need to export this method in `DroneCore`, if an app can call it via `System` class ?
https://github.com/dronecore/DroneCore/blob/a2b9db92f83e53bcce75c611f6e9b473f4341562/core/dronecore.h#L148

shakthi-prashanth-m:
@lbegani 
>The app will receive a callback whenever there is a new camera device discovered

Yes, we have
https://github.com/dronecore/DroneCore/blob/a2b9db92f83e53bcce75c611f6e9b473f4341562/core/dronecore.h#L161
for the same purpose, but it notifies only system id. I think, its a good point.

@julianoes 
How about this: As MAVLink supports only 6 cameras, `core` can subscribe to all of them internally. While processing heartbeat, it can make out if its a newly discovered component. If yes, it may notify application (about component type, NOT ID) if callback is set.
Application may use corresponding plugin to deal with a newly discovered component.

shakthi-prashanth-m:
```
[==========] Running 1 test from 1 test case.
[ RUN      ] SitlTest.MultiComponentDiscovery
We've nothing to do until we've some system to talk to. Lets sleep!
[07:34:28|Info ] New system on: 127.0.0.1:14556 (udp_connection.cpp:206)
[07:34:28|Debug] New: System ID: 1 Comp ID: 1 (dronecore_impl.cpp:312)
[07:34:28|Error] Remote IP unknown (udp_connection.cpp:135)
[07:34:28|Error] send fail (dronecore_impl.cpp:93)
[07:34:28|Debug] Component Autopilot added. (system.cpp:306)
[07:34:28|Info ] New system on: 127.0.0.1:14557 (udp_connection.cpp:206)
[07:34:28|Info ] System changed to: 127.0.0.1:36633 (udp_connection.cpp:217)
[07:34:28|Debug] Component Camera added. (system.cpp:306)
[07:34:28|Debug] Got HEARTBEAT from: Camera (system.cpp:154)
[07:34:28|Info ] System changed to: 127.0.0.1:14556 (udp_connection.cpp:217)
[07:34:29|Debug] Got HEARTBEAT from: Autopilot (system.cpp:154)
[07:34:29|Debug] Got HEARTBEAT from: Autopilot (system.cpp:154)
[07:34:29|Debug] # of components: 2 (system.cpp:406)
[07:34:29|Debug] Discovered 4294967298 (dronecore_impl.cpp:334)
[07:34:29|Info ] System changed to: 127.0.0.1:36633 (udp_connection.cpp:217)
[07:34:29|Debug] Got HEARTBEAT from: Camera (system.cpp:154)
[07:34:29|Info ] System changed to: 127.0.0.1:14556 (udp_connection.cpp:217)
[07:34:30|Debug] Got HEARTBEAT from: Autopilot (system.cpp:154)
[07:34:30|Debug] Got HEARTBEAT from: Autopilot (system.cpp:154)
[07:34:30|Info ] System changed to: 127.0.0.1:36633 (udp_connection.cpp:217)
[07:34:30|Debug] Got HEARTBEAT from: Camera (system.cpp:154)
[07:34:30|Info ] System changed to: 127.0.0.1:14556 (udp_connection.cpp:217)
[07:34:31|Debug] Got HEARTBEAT from: Autopilot (system.cpp:154)
[07:34:31|Debug] Got HEARTBEAT from: Autopilot (system.cpp:154)
Done with sleeping
We found a system with UUID: 4294967298
System 4294967298 has:  Autopilot + Camera
Done
[       OK ] SitlTest.MultiComponentDiscovery (7007 ms)
[----------] 1 test from SitlTest (7007 ms total)

[==========] 1 test from 1 test case ran. (7007 ms total)
[  PASSED  ] 1 test.
```

shakthi-prashanth-m:
@julianoes @JonasVautherin @hamishwillee  Please review. Thanks.

README.md
```diff
@@ -34,10 +34,10 @@ API consumers use the `DroneCore` class to discover and manage vehicles (`Device
```
hamishwillee: I can't mark it up, but docs above mention `Device`: 

>  API consumers use the `DroneCore` class to discover and manage vehicles (`Device` objects),

README.md
```diff
@@ -34,10 +34,10 @@ API consumers use the `DroneCore` class to discover and manage vehicles (`Device
 
 The links below take you to the respective header files:
 
-- [dronecore](include/dronecore.h): set up connection, discover devices
-- [device](include/device.h): a device providing access to modules below using ...
-- [device_plugin_container.h.in](include/device_plugin_container.h.in) which is auto-generated on build.
-- [info](plugins/info/info.h): general info about a device
+- [dronecore](include/dronecore.h): set up connection, discover systems
+- [system](include/system.h): a system providing access to modules below using ...
+- [system_plugin_container.h.in](include/system_plugin_container.h.in) which is auto-generated on build.
```
hamishwillee: Does this still exist?

README.md
```diff
@@ -34,10 +34,10 @@ API consumers use the `DroneCore` class to discover and manage vehicles (`Device
...
+- [info](plugins/info/info.h): general info about a system
```
hamishwillee: The old text wasn't much good either. Suggest change to:

> autopilot and vehicle product and version information.

core/system.h
```diff
@@ -19,7 +20,21 @@ namespace dronecore {
 class DroneCoreImpl;
 class PluginImplBase;
 
-class Device
+
+/**
```
hamishwillee: 1. This feels very odd sitting up here in the System class docs. 
2. Why is it called `Self`? What is implied by calling it Self? I hate this name - bit close to *self*.
3. What is it for? Why does a user need to know about this?

core/system.h
```diff
@@ -19,7 +20,21 @@ namespace dronecore {
...
+ * @brief This represents GCS DroneCore client.
+ */
+struct Self {
+    static constexpr uint8_t system_id = 0;
+    static constexpr uint8_t component_id = MAV_COMP_ID_SYSTEM_CONTROL;
+    static constexpr MAV_TYPE type = MAV_TYPE_GCS;
+};
+
+/**
```
hamishwillee: This needs work. How about:

> This class represents a MAVLink system, made up of one or more components (e.g. autopilot, cameras, servos, gimbals, etc). Commonly `System` objects are used to interact with UAVs (including their components) and standalone cameras.

Anything else we need to say?

core/system.h
```diff
@@ -74,13 +89,26 @@ class Device
...
+    void add_new_component(uint8_t component_id);
+    size_t total_components() const;
+
+    uint8_t get_autopilot_id() const;
+    std::vector<uint8_t> get_camera_ids() const;
+    uint8_t get_gimbal_id() const;
+
+    bool is_standalone() const;
+    bool has_autopilot() const;
+    bool has_camera() const;
```
hamishwillee: As discussed, there might be multiple cameras. This API not sufficient.

example/fly_mission/fly_mission.cpp
```diff
@@ -69,25 +69,25 @@ int main(int /*argc*/, char ** /*argv*/)
...
 
     dc.register_on_timeout([](uint64_t uuid) {
-        std::cout << "Device with UUID timed out: " << uuid << std::endl;
+        std::cout << "System with UUID timed out: " << uuid << std::endl;
         std::cout << "Exiting." << std::endl;
         exit(0);
     });
 
-    // We don't need to specifiy the UUID if it's only one device anyway.
+    // We don't need to specifiy the UUID if it's only one system anyway.
```
hamishwillee: typo "specifiy"

example/fly_qgc_mission/fly_qgc_mission.cpp
```diff
@@ -78,25 +78,25 @@ int main(int argc, char **argv)
...
+    // We don't need to specify the UUID if it's only one system anyway.
     // If there were multiple, we could specify it with:
-    // dc.device(uint64_t uuid);
-    Device &device = dc.device();
-    std::shared_ptr<Action> action = std::make_shared<Action>(device);
-    std::shared_ptr<Mission> mission = std::make_shared<Mission>(device);
-    std::shared_ptr<Telemetry> telemetry = std::make_shared<Telemetry>(device);
+    // dc.system(uint64_t uuid);
+    System &system = dc.system();
+    std::shared_ptr<Action> action = std::make_shared<Action>(system);
```
hamishwillee: If you plan on tidying this in future, might be good time to make these
```
auto action = std::make_shared<Action>(system);
```

core/system.cpp
```diff
@@ -282,7 +285,54 @@ void Device::device_thread(Device *self)
     }
 }
 
-void Device::send_heartbeat(Device &self)
+std::string System::name(uint8_t component_id)
```
hamishwillee: Add docs for ALL methods! (ie all the ones above and below here that don't have them!)

core/system.cpp
```diff
@@ -282,7 +285,54 @@ void Device::device_thread(Device *self)
...
+        LogDebug() << "Component " << name(component_id) << " added.";
+    }
+}
+
+size_t System::total_components() const
+{
+    return _components.size();
+}
+
+bool System::is_standalone() const
```
hamishwillee: HOw is this supposed to be used? i.e. what it for/what does it mean? 

I guess I'm just concerned that an autopilot on its own is technically also "standalone".

core/system.h
```diff
@@ -74,13 +89,26 @@ class Device
...
+    bool has_gimbal() const;
```
hamishwillee: Do we need has_servo and other options in https://mavlink.io/en/messages/common.html#MAV_COMPONENT ?

README.md
shakthi-prashanth-m: Sorry. Kindly ignore changes in README.md

core/system.h
shakthi-prashanth-m: >This feels very odd sitting up here in the System class docs.

Thanks for pointing this. I have created an for a discussion on this: https://github.com/dronecore/DroneCore/issues/325. I added docs because this interface is publicly accessible to apps.

>Why is it called Self? What is implied by calling it Self? I hate this name - bit close to self.
>What is it for? Why does a user need to know about this?

As mentioned in the header, this represents GCS DroneCore client. I mentioned in detail here: https://github.com/dronecore/DroneCore/issues/317 Its for plugins while sending MAVLink messages. We don't have class `SystemImpl` to add it in :)

core/system.h
shakthi-prashanth-m: Yeah, I just added whatever came to my mind. I think yours is better.

core/system.h
shakthi-prashanth-m: Ok! I shall change it. Thanks.

example/fly_qgc_mission/fly_qgc_mission.cpp
shakthi-prashanth-m: Sure I'll this time :) Thanks for reminding.

core/system.cpp
shakthi-prashanth-m: I think, if a system has autopilot, its subjected to move; isn't it ?

core/system.h
shakthi-prashanth-m: We can anytime. But I added only those whom we've plugins support.

core/system.h
shakthi-prashanth-m: @hamishwillee what name do you suggest after you read through #317  ?

shakthi-prashanth-m:
What can be done for these ?
```
Reading package lists...
Building dependency tree...
Reading state information...
E: Unable to locate package gcc-6
E: Unable to locate package g++-6
E: Couldn't find any package by regex 'g++-6'
```
Ignore ?

shakthi-prashanth-m:
OK I restarted Travis CI build.

core/system.h
hamishwillee: I'm a bit rushed, but I saw another you renamed it  `struct ControlSystem` and that is much less offensive :-). Why not `GCSClient`?
Also still feels odd sitting in system.h alongside System, the object for our System.

shakthi-prashanth-m: Yeah, `GCSClient` is better, I will change it. Its in [mavlink_system.h](https://github.com/dronecore/DroneCore/pull/310/files#diff-e3907820f42d7a290a4876d951b4ca80R25) not `system.h`.

core/mavlink_system.h
```diff
@@ -0,0 +1,223 @@
+#pragma once
+
+#include "global_include.h"
```
JonasVautherin: Just a detail, but following Google guidelines, I like putting the external includes `<...>` first, then the `"..."`. And each group in alphabetical order :-).

shakthi-prashanth-m: Sure. I thought fix_style takes care of it.

shakthi-prashanth-m:
Some open questions about handling multi-components:
dronecoreudpconnection

Consider above system with components Autopilot, Camera and some other component. We may receive heartbeat from any of the components first.

If we receive **first heartbeat** from **Autopilot**, then we can request for autopilot version and initialize UUID from after receiving autopilot version.

However, if we receive **first heartbeat** from **Camera** component, then we can't differentiate whether it is a _standalone camera_ OR _part of an Autopilot system_.

In case of a standalone system, we don't need to request for autopilot version; instead, use System ID as UUID.
If its not standalone, then do we need to wait for heartbeat from Autopilot ?

shakthi-prashanth-m:
>However, if we receive first heartbeat from Camera component,

When I ran CSD along with PX4 SITL, I **always**  get first heartbeat from Autopilot. The reason is that CSD reads heartbeat and collect System ID and then broadcast heartbeat. So, obviously there's a delay.

shakthi-prashanth-m:
OK, My conclusion regarding Multi component discovery:

- In case of a Multi component system, Camera server (CSD) would wait for Autopilot heartbeat for getting System ID, hence we **assume** that we always receive Autopilot's heartbeat first.
- In case of Standalone system, we don't request for Autopilot version and we consider System ID as UUID after we receive consecutive heartbeats from a non-autopilot component **twice**.

Please correct me if these assumptions are not acceptable.

shakthi-prashanth-m:
### Test scenario 1: Launch PX4 SITL alone.
```
Note: Google Test filter = SitlTest.MultiComponentDiscovery
[==========] Running 1 test from 1 test case.
[ RUN      ] SitlTest.MultiComponentDiscovery
We've nothing to do until we've some system to talk to. Lets sleep!
[07:35:05|Info ] New system on: 127.0.0.1:14557 (udp_connection.cpp:275)
[07:35:05|Debug] New: System ID: 1 Comp ID: 1 (dronecore_impl.cpp:311)
[07:35:05|Debug] Component Autopilot added. (mavlink_system.cpp:324)
[07:35:05|Debug] MAVLink: info: [logger] file: rootfs/fs/microsd/log/2018-03-23/1 (mavlink_system.cpp:261)
[07:35:05|Debug] MAVLink: critical: First waypoint too far away: 7094 meters, 900 max (mavlink_system.cpp:261)
[07:35:05|Debug] MAVLink: critical: Planned mission fails check. Please upload again. (mavlink_system.cpp:261)
[07:35:05|Debug] # of components: 1 (mavlink_system.cpp:434)
[07:35:05|Debug] Discovered 4294967298 (dronecore_impl.cpp:332)
Done with sleeping
//////////////////////////////
We found a System with UUID: 4294967298
We found an Autopilot alone.
//////////////////////////////
Done
[       OK ] SitlTest.MultiComponentDiscovery (9007 ms)
[----------] 1 test from SitlTest (9007 ms total)

[==========] 1 test from 1 test case ran. (9008 ms total)
[  PASSED  ] 1 test.

```

### Test scenario 2: Connect a camera and launch CSD. (No PX4 SITL).
```
Note: Google Test filter = SitlTest.MultiComponentDiscovery
[==========] Running 1 test from 1 test case.
[ RUN      ] SitlTest.MultiComponentDiscovery
We've nothing to do until we've some system to talk to. Lets sleep!
[07:32:29|Info ] New system on: 127.0.0.1:43672 (udp_connection.cpp:275)
[07:32:29|Debug] New: System ID: 1 Comp ID: 100 (dronecore_impl.cpp:311)
[07:32:29|Debug] Component Camera 1 added. (mavlink_system.cpp:324)
[07:32:30|Debug] # of components: 1 (mavlink_system.cpp:434)
[07:32:30|Debug] Discovered 1 (dronecore_impl.cpp:332)
Done with sleeping
//////////////////////////////
We found a System with UUID: 1
We found a Standalone camera.
//////////////////////////////
Done
[       OK ] SitlTest.MultiComponentDiscovery (9014 ms)
[----------] 1 test from SitlTest (9014 ms total)

[==========] 1 test from 1 test case ran. (9014 ms total)
[  PASSED  ] 1 test.
```

### Test scenario 3: Launch PX4 SITL + Connect a camera and launch CSD.
```
Note: Google Test filter = SitlTest.MultiComponentDiscovery
[==========] Running 1 test from 1 test case.
[ RUN      ] SitlTest.MultiComponentDiscovery
We've nothing to do until we've some system to talk to. Lets sleep!
[07:32:10|Info ] New system on: 127.0.0.1:14557 (udp_connection.cpp:275)
[07:32:10|Debug] New: System ID: 1 Comp ID: 1 (dronecore_impl.cpp:311)
[07:32:10|Debug] Component Autopilot added. (mavlink_system.cpp:324)
[07:32:10|Debug] MAVLink: info: data link #0 lost (mavlink_system.cpp:261)
[07:32:10|Debug] MAVLink: critical: Planned mission fails check. Please upload again. (mavlink_system.cpp:261)
[07:32:10|Debug] MAVLink: info: data link #0 regained (mavlink_system.cpp:261)
[07:32:10|Debug] # of components: 1 (mavlink_system.cpp:434)
[07:32:10|Debug] Discovered 4294967298 (dronecore_impl.cpp:332)
[07:32:11|Info ] New system on: 127.0.0.1:43672 (udp_connection.cpp:275)
[07:32:11|Debug] Component Camera 1 added. (mavlink_system.cpp:324)
Done with sleeping
We found a System with UUID: 4294967298
Its an Autopilot with a Camera.
Done
[       OK ] SitlTest.MultiComponentDiscovery (9005 ms)
[----------] 1 test from SitlTest (9005 ms total)

[==========] 1 test from 1 test case ran. (9006 ms total)

```

shakthi-prashanth-m:
@julianoes Can you please review ? I shall resolve the Appveyor issue in parallel.

hamishwillee:
> Re "In case of a Multi component system, Camera server (CSD) would wait for Autopilot heartbeat for getting System ID, hence we assume that we always receive Autopilot's heartbeat first."

CSD is only one possible implementation of the MAVLink Camera Protocol. I'd say you'd try avoid this assumption.

shakthi-prashanth-m:
>CSD is only one possible implementation of the MAVLink Camera Protocol. I'd say you'd try avoid this assumption.

Yeah I agree to your point. Even otherwise any implementation should get heartbeat info dynamically from Autopilot heartbeat. So there's definitely going to be a delay isn't it ?

hamishwillee:
>> CSD is only one possible implementation of the MAVLink Camera Protocol. I'd say you'd try avoid this assumption.

> Yeah I agree to your point. Even otherwise any implementation should get System ID info dynamically from Autopilot heartbeat. So there's definitely going to be a delay isn't it ?

CSD gets the ID it uses for camera system ID from a configuration file. I don't really understand the discussion re heartbeat.

shakthi-prashanth-m:
>I don't really understand the discussion re heartbeat.

System ID in configuration file is not dynamic. So CSD listens to Autopilot heartbeat to retrieve system ID dynamically and then it will compose its heartbeat for broadcasting.

hamishwillee:
>> I don't really understand the discussion re heartbeat.

> System ID in configuration file is not dynamic. So CSD listens to Autopilot heartbeat to retrieve system ID dynamically and then it will compose its heartbeat for broadcasting.

That seems odd. If true the CSD documentation is incorrect. 

According to docs CSD system ID comes from CSD. It is static - and that makes sense because you're not going have a camera having to guess that it is part of a particular system - that is something you'd configure.  

Even if you could guess it, listening for an autopilot heartbeat would not make sense, because  a MAVLink network could in theory have 255 autopilots - which one would you use?

shakthi-prashanth-m:
>According to docs CSD system ID comes from CSD. It is static - and that makes sense because you're not going have a camera having to guess that it is part of a particular system - that is something you'd configure.

Sorry there's a little change. As I understand CSD will wait for SystemID from Vehicle ONLY if the given system ID is invalid. Please correct me @lbegani .

shakthi-prashanth-m:
>In case of a Multi component system, Camera server (CSD) would wait for Autopilot heartbeat for getting System ID, hence we assume that we always receive Autopilot's heartbeat first.

I think this assumption is **no more valid** in the latest commit.

If we don't receive Autopilot heartbeat for 2 seconds, but received only Camera's heartbeat, then we consider System ID as the UUID and we don't request for Autopilot version which is obvious.

lbegani:
@hamishwillee @shakthi-prashanth-m  In CSD, there is an order of preference : system ID from vehicle > system ID from conf file. In the absence of autopilot heartbeat, static ID from conf file will get used.  However it seems the logic will fail if there exists multiple vehicles sending heartbeat. What can be done is to attach CSD with the first autopilot detected and use its system ID forever.

hamishwillee:
> @hamishwillee @shakthi-prashanth-m In CSD, there is an order of preference : system ID from vehicle > system ID from conf file. In the absence of autopilot heartbeat, static ID from conf file will get used. However it seems the logic will fail if there exists multiple vehicles sending heartbeat. 
> What can be done is to attach CSD with the first autopilot detected and use its system ID forever.

Seems a bit fragile. What if I want to use CSD as a stand alone camera on a mavlink network? 

What is the reason for every using system ID from network? Perhaps more sensible to have System ID in config have a possible value Network which means "fetch value"?

lbegani:
@hamishwillee @shakthi-prashanth-m 
This could be a strategy to follow for CSD - 
1. Read from the conf file (-1 : read from network, >=0 : use as it is )

    - If conf file is not there, use some default value (1) or read from nw ?

2. Read from the network (if system ID is set to -1 in conf file)

    - Wait for the autopilot heatbeat before sending component heartbeat
    - Wait forever or set a timeout and fallback to some default value (1) ?

> What if I want to use CSD as a stand alone camera on a mavlink network?

By stand-alone, if you mean component without autopilot, QGC will not accept the heartbeat message from camera component unless there is a corresponding vehicle with same system ID connected with QGC.

hamishwillee:
QGC is not the only possible GCS. Yes, I mean camera component without an autopilot. That is a valid MAVLink system AFAIK.

hamishwillee:
I think if the config is not there then use a default value (1). 

Why do we need to wait for the autopilot before sending component id?

lbegani:
@hamishwillee  

> Why do we need to wait for the autopilot before sending component id?

Fundamental premise was that camera component cannot exist standalone without autopilot. There has to be vehicle running. 
Secondly, any random system ID will not work. It has to match with the system id of autopilot running (either hardcode or fetch)
For CSD, the code changes are trivial, just need to finalize on handling.

hamishwillee:
@lbegani Can you please create issue for https://github.com/dronecore/DroneCore/pull/310#issuecomment-376081823 so I can track what is done and update docs?

lbegani:
> Can you please create issue for #310 (comment) so I can track what is done and update docs?

https://github.com/intel/camera-streaming-daemon/issues/149

integration_tests/CMakeLists.txt
```diff
@@ -2,11 +2,14 @@
 include_directories(
     SYSTEM ${CMAKE_SOURCE_DIR}/third_party/gtest/googletest/include
     SYSTEM ${CMAKE_SOURCE_DIR}/third_party/gtest/googlemock/include
+    SYSTEM ${CMAKE_SOURCE_DIR}/third_party/mavlink/include
 )
 
 # These are consumed in ../CMakeLists.txt
 set(integration_tests_src ${integration_tests_src} PARENT_SCOPE)
 
+file(GLOB_RECURSE integration_tests_src "*.cpp")
```
JonasVautherin: I think the good practice is to list the files, instead of doing a `GLOB_RECURSE`?!

plugins/action/action.cpp
```diff
@@ -113,4 +113,34 @@ float Action::get_max_speed_m_s() const
     return _impl->get_max_speed_m_s();
 }
 
+const char *Action::result_str(Result result)
```
JonasVautherin: I thought I had moved that to `plugins/action/action_result.h`?

plugins/action/action.h
```diff
@@ -27,23 +27,48 @@ class Action : public PluginBase
...
 
     /**
      * @brief Destructor (internal use only).
      */
     ~Action();
 
+    /**
+     * @brief Possible results returned for commanded actions.
+     */
+    enum class Result {
```
JonasVautherin: I thought I had moved that to `plugins/action/action_result.h`?

plugins/action/action_result.h
```diff
@@ -14,7 +14,7 @@ namespace dronecore {
 enum class ActionResult {
```
JonasVautherin: Yeah it is actually here. Why do you have another `Result` in `Action`, then?

integration_tests/CMakeLists.txt
julianoes: Yes that's what I read too.

plugins/action/action_result.h
julianoes: Probably re-appeared in a merge?

JonasVautherin: Yeah I did not formulate it in a good way. The other one should be removed xD.

integration_tests/CMakeLists.txt
shakthi-prashanth-m: I see, I could see that here: https://cmake.org/cmake/help/v3.9/command/file.html

Note: *We do not recommend using GLOB to collect a list of source files from your source tree.
If no CMakeLists.txt file changes when a source is added or removed
then the generated build system cannot know when to ask CMake to regenerate.*

plugins/action/action.cpp
shakthi-prashanth-m: Oops! Ok, I'll remove it.

core/mavlink_system.cpp
```diff
@@ -282,16 +295,82 @@ void Device::device_thread(Device *self)
...
+{
+    return _components.size();
+}
+
+bool MAVLinkSystem::is_standalone() const
+{
+    return !is_autopilot();
+}
+
+bool MAVLinkSystem::is_autopilot() const
```
julianoes: I would change this to `has_autopilot()` to make it consistent and because it's not clear if something **is** an autopilot if it also includes a camera.

core/mavlink_system.cpp
```diff
@@ -282,16 +295,82 @@ void Device::device_thread(Device *self)
...
+{
+    return get_autopilot_id() != uint8_t(0);
+}
+
+bool MAVLinkSystem::has_camera(uint8_t camera_id) const
+{
+    uint8_t camera_comp_id = MAV_COMP_ID_CAMERA + (camera_id - 1);
+
+    for (auto compid : _components) {
+        return compid == camera_comp_id;
```
julianoes: I think this is wrong and should be:
```
for (auto compid : _components) {
    if (compid == camera_comp_id) {
        return true;
    }
}
```

core/mavlink_system.cpp
```diff
@@ -544,92 +623,123 @@ void Device::set_flight_mode_async(FlightMode device_mode, command_result_callba
...
+            return compid;
+        }
+    // FIXME: Not sure what should be returned if autopilot is not found
+    return uint8_t(0);
+}
+
+std::vector<uint8_t> MAVLinkSystem::get_camera_ids() const
+{
+    std::vector<uint8_t> camera_ids;
+    camera_ids.clear();
```
julianoes: You can just initialize it wit:
```
std::vector<uint8_t> camera_ids {};
```

core/mavlink_system.cpp
```diff
@@ -544,92 +623,123 @@ void Device::set_flight_mode_async(FlightMode device_mode, command_result_callba
...
+            camera_ids.push_back(compid);
+        }
+    return camera_ids;
+}
+
+uint8_t MAVLinkSystem::get_gimbal_id() const
+{
+    for (auto compid : _components)
+        if (compid == MAV_COMP_ID_GIMBAL) {
+            return compid;
```
julianoes: Same error here, I think.

core/mavlink_system.h
```diff
@@ -19,7 +20,18 @@ namespace dronecore {
 class DroneCoreImpl;
 class PluginImplBase;
 
-class Device
+
+// This represents DroneCore client application which is a GCS.
+struct GCSClient {
```
julianoes: Please add in the comment or name what a `GCS` is.

core/system.h
```diff
@@ -0,0 +1,73 @@
...
+     *
+     * A System may have several cameras with ID starting from 1.
+     * When called without passing camera ID, it checks for the camera ID 1.
+     * @param camera_id ID of the camera starting from 1 onwards.
+     * @return true if camera with the given camera ID is found, false otherwise.
+     */
+    bool has_camera(uint8_t camera_id = 1) const;
+
+    /**
+     * @brief Checks whether the system has gimbal.
```
julianoes: `has a gimbal`

core/system.h
```diff
@@ -0,0 +1,73 @@
...
+    bool is_autopilot() const;
+
+    /**
+     * @brief Checks whether the system is a standalone (Non-autopilot).
+     * @return true if stand alone, false otherwise.
+     */
+    bool is_standalone() const;
+
+    /**
+     * @brief Checks whether the system has camera with the given camera ID.
```
julianoes: `has a camera`.

core/system.h
```diff
@@ -0,0 +1,73 @@
...
+ * are used to interact with UAVs (including their components) and standalone cameras.
+ */
+class System
+{
+public:
+    explicit System(DroneCoreImpl &parent, uint8_t system_id, uint8_t comp_id);
+    ~System();
+
+    /**
+     * @brief Checks whether the system is an autopilot.
```
julianoes: `has an autopilot`.

core/system.h
```diff
@@ -0,0 +1,73 @@
...
+     */
+    bool has_camera(uint8_t camera_id = 1) const;
+
+    /**
+     * @brief Checks whether the system has gimbal.
+     * @return true if the system has gimbal, false otherwise.
+     */
+    bool has_gimbal() const;
+
+    friend class DroneCoreImpl;
```
julianoes: Why are the friends needed? Please add a comment explaining why.

core/plugin_impl_base.h
```diff
@@ -56,7 +59,7 @@ class PluginImplBase
     const PluginImplBase &operator=(const PluginImplBase &) = delete;
 
 protected:
-    Device &_parent;
+    std::shared_ptr<MAVLinkSystem> _parent;
```
julianoes: Why didn't you leave this a reference?

core/system.h
JonasVautherin: Oh, I had not seen that. I'm interesting in why we need friends at all. I thought they were made for operators, but that they were not good practice in other uses...

core/mavlink_system.cpp
shakthi-prashanth-m: _(FYI: I am unable to use Slack temporarily, but I see your email notifications.)_

OK. I added `is_autopilot` thinking that we could say,
```cpp
...
if (system.is_autopilot() && system.has_camera())
    print "We've a drone that has a camera"
else if (system.is_autopilot())
    print "We've a drone"
...
```
I understand your point as well, so I am ok to change as you suggested.
I am also thinking to add `is_camera()` for standalone cameras. 
If something **is** a camera, it can't be anything else; because, it is drones that contain camera and other components. What do you think ?

core/mavlink_system.cpp
shakthi-prashanth-m: No its not. The result of the expression `compid == camera_comp_id` evaluates to `true` if they're equal and `false` otherwise.

shakthi-prashanth-m: Oh! ok sorry I missed the point. I'll correct.

core/mavlink_system.cpp
shakthi-prashanth-m: Here, I return gimbal `compid` if I find it.

core/mavlink_system.h
shakthi-prashanth-m: OK

core/plugin_impl_base.h
shakthi-prashanth-m: class `System` creates a `shared_ptr` to `MAVLinkSystem` during construction to finally share them to various plugins. I made it so, because each plugin holds a shared instance of the system. It made sense to me. Do you see any issue ?

core/system.h
shakthi-prashanth-m: OK

core/system.h
shakthi-prashanth-m: OK

core/system.h
shakthi-prashanth-m: OK

core/system.h
shakthi-prashanth-m: Yeah good point.

- `friend class PluginImplBase`:
While we create plugins, we pass a reference to `System` to plugin constructors. But eventually, plugins need access to `MAVLinkSystem` instance. So, we keep `MAVLinkSystem` as a private member of class `System`.
By making `DroneCoreImpl` class a friend of class `System`, we allow to `DroneCoreImpl` class to request access to instance of `MAVLinkSystem` class.
- `friend class DroneCoreImpl`:
class `DroneCoreImpl` wants to perform some operations on class `System`, but we don't want those to expose to public. Please suggest for alternative if any.

core/mavlink_system.cpp
julianoes: I don't think we need `is_camera()` because it will be confusing to have both. And by just having `has_...()` it is clear that a system can always include various components.

core/plugin_impl_base.h
julianoes: I think a reference was fine, and generally you wanted to favor references against pointers, right?

core/system.h
julianoes: Usually we use used the pimpl scheme to solve exactly this. I guess we can leave it for now but might want to optimize this later.

core/plugin_impl_base.h
shakthi-prashanth-m: Yeah, I favor reference over a normal (raw) pointers. I can change to references as well.

core/system.h
shakthi-prashanth-m: OK

integration_tests/multi_components.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+    for (auto uuid : uuids) {
+        std::cout << "We found a System with UUID: " << uuid << '\n';
+        System &system = dc.system(uuid);
+
+        auto is_autopilot = system.is_autopilot();
+
+        auto is_standalone = system.is_standalone();
+
+        auto has_camera = system.has_camera(); // by default checks for camera 1
```
shakthi-prashanth-m: I want to ask few suggestions. I'd done testing of this by attaching a V4L2 camera on Ubuntu.
But, when I tested it today with a new camera, its component ID turned out to be `101` (usually it was `100`).
The test queries **whether we discovered camera 1** ? This time we discovered camera 2. As the test didn't discover camera 1, it prints as *Its an Autopilot alone.* 

It should rather query **did we discover any camera ?** If yes, continue the rest of the test.

We need a method that returns `true` if the system has at least one camera. We have two options:
- Add a new method `has_any_camera()`.
- Overload `has_camera()`, when 0 is passed to know whether system has at least a camera.
What shall we do ?

core/plugin_impl_base.h
JonasVautherin: I guess the question of smart pointer vs reference is about ownership: say I have an object `A` that has an object `o`. If `o` only lives when `A` lives, then you can pass a reference to other objects using `o`. You want to have a pointer when ownership can change, i.e. when `o` may live longer than `A`. And there you have the question between `unique_ptr` and `shared_ptr`.

I am not completely sure of the design here, but it seems like a reference is what you want, indeed.

shakthi-prashanth-m: I create `MAVLinkSystem` instance in [System constructor](https://github.com/dronecore/DroneCore/pull/310/files#diff-de9116ca3f3e99f357eb697180be37f6R22). I've to use pointer here. So, as it is going to be shared by all the plugins, I made it as `shared_ptr`. Now, while giving access to plugins, it is better to give them as `shared_ptr` itself, rather than dereference and give reference. What do you think ?

integration_tests/multi_components.cpp
JonasVautherin: Interesting question, but I would not change it in this PR. What about opening an issue to discuss this point and make a new PR later with a fix for it? :-)

core/plugin_impl_base.h
julianoes: @shakthi-prashanth-m why do you have to create `MAVLinkSystem` on the heap in the system constructor?

integration_tests/multi_components.cpp
julianoes: Later is fine. I think `has_camera` without argument should be `true` if it has any camera.

core/plugin_impl_base.h
shakthi-prashanth-m: So that `System` can hide `MAVLinkSystem` class. Now we have pointer to it.

integration_tests/multi_components.cpp
shakthi-prashanth-m: @JonasVautherin I agree.
@julianoes That would break our original idea. We thought that `has_camera()`, by default should check for camera 1 (to avoid passing 1 every time). I think, `has_camera(0)` should be OK because, by concept: Comp ID 0 indicates all IDs. So it means check for any camera on the system.

shakthi-prashanth-m: @julianoes I think what you suggested makes sense.

shakthi-prashanth-m:
@julianoes @JonasVautherin 
I hope https://github.com/dronecore/DroneCore/pull/310/commits/f153bebe8a0d95d83d47bff33ed6052510a386bf addressed most of the review comments. If this is good to go now, lets merge soon after you have a re-look. We can work on the items later, that require optimization. Sounds good ?

plugins/telemetry/telemetry.cpp
```diff
@@ -6,9 +6,9 @@
 
 namespace dronecore {
 
-Telemetry::Telemetry(Device &device) :
+Telemetry::Telemetry(System &system) :
     PluginBase(),
-    _impl(new TelemetryImpl(device))
+    _impl { new TelemetryImpl(system) }
```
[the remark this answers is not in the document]
shakthi-prashanth-m: Its C++11 style to initialize class members: http://en.cppreference.com/w/cpp/language/initializer_list. See (2) in the item list.

backend/src/backend.cpp
```diff
@@ -38,16 +38,18 @@ class DroneCoreBackend::Impl
...
-        MissionServiceImpl<dronecore::Mission> missionService(mission);
+        Mission mission(_dc.system());
+        MissionServiceImpl missionService(mission);
+
         builder.RegisterService(&missionService);
 
-        Telemetry telemetry(_dc.device());
-        TelemetryServiceImpl<> telemetryService(telemetry);
+        Telemetry telemetry(_dc.system());
+        TelemetryServiceImpl<dronecore::Telemetry> telemetryService(telemetry);
```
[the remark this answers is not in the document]
shakthi-prashanth-m: Yeah, sorry I'll correct that.

core/mavlink_system.cpp
```diff
@@ -334,20 +334,42 @@ size_t MAVLinkSystem::total_components() const
...
+{
+    return (comp_id >= MAV_COMP_ID_CAMERA)
+           && (comp_id <= MAV_COMP_ID_CAMERA6);
+}
+
 bool MAVLinkSystem::has_camera(uint8_t camera_id) const
 {
-    uint8_t camera_comp_id = MAV_COMP_ID_CAMERA + (camera_id - 1);
+    uint8_t camera_comp_id = (camera_id == 0) ?
+                             camera_id : (MAV_COMP_ID_CAMERA + (camera_id - 1));
```
julianoes: Does that mean that the first camera has index 1? If yes, that works but should be documented that it is 1 based.

Otherwise, I'd make the default `int camera_id = -1` and check for that, and then use `0` input as the first component.

plugins/telemetry/telemetry.cpp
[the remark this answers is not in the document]
JonasVautherin: Is it necessary, or is it just a different style? If it's only the style, I'd rather keep it consistent (we usually use `()` and not `{}` in the code). Having it different here suggests that it requires a special treatment, making the code more difficult to read (IMO).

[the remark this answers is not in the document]
shakthi-prashanth-m: Actually, `{}` is the consistent style so far used in all the plugins and core.
```
plugins/gimbal/gimbal.cpp:    _impl { new GimbalImpl(system) }
plugins/offboard/offboard.cpp:    _impl { new OffboardImpl(system) }
plugins/info/info.cpp:    _impl { new InfoImpl(system) }
plugins/follow_me/follow_me.cpp:    _impl { new FollowMeImpl(system) }
plugins/mission/mission_item.cpp:    _impl { new MissionItemImpl() }
plugins/mission/mission.cpp:    _impl { new MissionImpl(system) }
plugins/logging/logging.cpp:    _impl { new LoggingImpl(system) }
plugins/action/action.cpp:    _impl { new ActionImpl(system) }
plugins/telemetry/telemetry.cpp:    _impl { new TelemetryImpl(system) }
core/dronecore.cpp:    _impl { new DroneCoreImpl() }
```
And it makes sense to follow C++11 style over the old as we've mentioned in `README.md` as well :)

core/mavlink_system.cpp
shakthi-prashanth-m: Yes, although it is documented, I think what you're saying is right. Users may feel index `0` as more intuitive.

julianoes:
Awesome, thanks @shakthi-prashanth-m!

plugins/telemetry/telemetry.cpp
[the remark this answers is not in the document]
JonasVautherin: Oh, my bad :)

PR #311, opened 2018-03-08
device: fix reconnect

julianoes:
This fix was merged earlier but somehow snuck back.

If we reset this flag but don't set the actual UUID to 0, we won't ever
request it again and therefore won't ever discover a device again that
we have discovered before.

This was already removed in #215 but somehow it came back.

@JonasVautherin maybe in a merge? Or was this on purpose?


JonasVautherin:
Not done on purpose, sorry about that.

Issue #314, opened 2018-03-12
Choosing reference over pointer


shakthi-prashanth-m:
It is recommended to use references when we can, and pointers when we have to.
(Reference: https://isocpp.org/wiki/faq/references#refs-vs-ptrs)

We can replace `Device *_parent` to `Device & _parent` as I see that, its possible. We initialize while activating plugin and never modify it. Also, we have a method (below) that actually gives, **a reference to a device**, not pointer. 
```cpp
Device &device(uint64_t uuid);
```

Consider, below piece of application code:
```cpp
Device &device = dc.device(); // we get reference to device
auto action = std::make_shared<Action>(&device); // why pass its address ?

// Why not pass it as-is?
auto action = std::make_shared<Action>(device);
```
I just noticed this.
@julianoes, was it done intentionally for some vital reason? If not, then we can find similar usages and replace them properly.

julianoes:
I tend to agree. Thanks for writing that up @shakthi-prashanth-m. Do you agree @JonasVautherin?

PR #320, opened 2018-03-14
fix backend following #314

JonasVautherin:
Very small thing, reminding me that I need to put the backend in the CI :)

### DIFF
--- a/backend/src/backend.cpp
+++ b/backend/src/backend.cpp
@@ -36,15 +36,15 @@ public:
         CoreServiceImpl core(_dc);
         builder.RegisterService(&core);
 
-        Action action(&_dc.device());
+        Action action(_dc.device());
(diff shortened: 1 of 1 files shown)


shakthi-prashanth-m:
@JonasVautherin , sorry I reviewed it. Am I supposed to wait when it says "in_progress" or until I see "waiting_for_review" ?

JonasVautherin:
Oh, `waiting for review` was created by waflle when I added a column for that, but then I figured it was not ideal.

IMO, a "pull request" is something that I am ready to merge. While I am working on a feature, I just make a branch. So I would say we should always review PRs, and `[WIP]` just means that it should not be merged yet.

What do you think?

shakthi-prashanth-m:
Yeah, I agree.

PR #324, opened 2018-03-15
Function parameter reference/pointer types fix

hamishwillee:
Addresses https://github.com/dronecore/DroneCore/issues/321. Basically a function parameter type for a reference or pointer will now render with the * or & immediately next to the type.
```
#old
Device * somedevice
Device & somedevice
#new
Device* somedevice
Device& somedevice
```
IFF the type is a pointer, then the "*" is escaped (so markdown doesn't think we want to change case to italic).

Note, I decided on aligning symbol to type rather than name, because this works better in more cases.


hamishwillee:
Thanks @shakthi-prashanth-m !

PR #327, opened 2018-03-17
267 update telemetry proto

### LINKED ISSUE #267: Update backend to new telemetry proto file

JonasVautherin:
This PR introduces testing for gRPC streams, which is a bit more tricky than unary RPC. In order to test streams, I am creating an in-process grpc server and use a mock to capture the `position_callback_t` callback called when the user subscribes to the position data.

Fixes #267.


backend/src/plugins/telemetry/telemetry_service_impl.h
```diff
@@ -0,0 +1,50 @@
...
+        : _telemetry(telemetry),
+          _stop_promise(std::promise<void>()),
+          _stop_future(_stop_promise.get_future()) {}
+
+    grpc::Status SubscribePosition(grpc::ServerContext * /* context */,
+                                   const dronecore::rpc::telemetry::SubscribePositionRequest * /* request */,
+                                   grpc::ServerWriter<rpc::telemetry::PositionResponse> *writer) override
+    {
+        _telemetry.position_async([&writer](dronecore::Telemetry::Position position) {
+            auto *rpc_position = new dronecore::rpc::telemetry::Position();
```
shakthi-prashanth-m: isn't it just `auto rpc_position` ?

backend/test/telemetry_service_impl_test.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+using testing::_;
+using testing::NiceMock;
+
+using MockTelemetry = NiceMock<dronecore::testing::MockTelemetry>;
+using TelemetryServiceImpl = dronecore::backend::TelemetryServiceImpl<MockTelemetry>;
+using TelemetryService = dronecore::rpc::telemetry::TelemetryService;
+
+using PositionResponse = dronecore::rpc::telemetry::PositionResponse;
+using Position = dronecore::Telemetry::Position;
```
shakthi-prashanth-m: These are nice!

backend/test/telemetry_service_impl_test.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+class TelemetryServiceImplTest : public ::testing::Test
+{
+protected:
+    virtual void SetUp()
+    {
+        _telemetry = std::unique_ptr<MockTelemetry>(new MockTelemetry());
+        _telemetry_service = std::unique_ptr<TelemetryServiceImpl>(new TelemetryServiceImpl(*_telemetry));
```
shakthi-prashanth-m:
```
_telemetry = new MockTelemetry();
_telemetry_service = new TelemetryServiceImpl(*_telemetry);
```
this would also do.
Ref: https://docs.microsoft.com/en-us/cpp/cpp/smart-pointers-modern-cpp#uses-for-smart-pointers

plugins/telemetry/telemetry.cpp
```diff
@@ -299,4 +299,19 @@ const char *Telemetry::result_str(Result result)
     }
 }
 
+bool operator==(const Telemetry::Position &lhs, const Telemetry::Position &rhs)
+{
+    return lhs.latitude_deg == rhs.latitude_deg
```
shakthi-prashanth-m: Can we compare two doubles ?

JonasVautherin: Good point. I am not very familiar with that. I am not sure if there is a generally accepted good practice for that. I know that we may have rounding issues when doing computations, but I don't think we do such computations with any of the fields of the Position.

What do you think? Is it good enough for now (to be modified if we ever do computations), or should I change it now? And in the latter case, what would be your suggestion? I've seen people using "epsilon"s, but I am not completely sure what it implies for e.g. latitudes and longitudes...

backend/test/telemetry_service_impl_test.cpp
JonasVautherin: Oh, I did not know that! I'll have a look, thanks :-)

JonasVautherin: Your proposition doesn't build for me... =/

plugins/telemetry/telemetry.cpp
JonasVautherin: After looking at it more closely, what you suggest is to implement a "close-enough" operator. Semantically, I've been told that `operator==` should actually be equal, and that I should use something else for "equal enough".

julianoes: I would suggest:
```
#include <limits>
return (abs(lhs.latitude_deg - rhs.latitude_deg) <= std::numeric_limits<T>::epsilon());
```

plugins/telemetry/telemetry.cpp
```diff
@@ -299,4 +299,19 @@ const char *Telemetry::result_str(Result result)
...
+           && lhs.longitude_deg == rhs.longitude_deg
+           && lhs.absolute_altitude_m == rhs.absolute_altitude_m
+           && lhs.relative_altitude_m == rhs.relative_altitude_m;
+}
+
+std::ostream &operator<<(std::ostream &str, Telemetry::Position const &position)
+{
+    return str << "[lat: " << position.latitude_deg << ", lng: " << position.longitude_deg
```
julianoes: we usually shorten longitude with `lon`.

backend/src/plugins/telemetry/telemetry_service_impl.h
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    void stop()
+    {
+        _stop_promise.set_value();
+    }
+
+private:
+    Telemetry &_telemetry;
+    std::promise<void> _stop_promise;
```
julianoes: Why is it called `stop`? That confuses me.

JonasVautherin: Because when this promise is satisfied, we stop the service. I'm happy to change the name :-). What do you suggest?

julianoes: I see, no that's fine then.

JonasVautherin:
@julianoes @shakthi-prashanth-m: I am not sure what I should change more in this PR? :-)

PR #328, opened 2018-03-17
action: move Action::Result out of action.h

JonasVautherin:
Just as for `ConnectionResult`, move `Action::Result` out of `Action` in order to make it available to others, like `MockAction`.

### DIFF
--- a/backend/src/plugins/action/action_service_impl.h
+++ b/backend/src/plugins/action/action_service_impl.h
@@ -20,7 +20,7 @@ public:
 
         auto *rpc_action_result = new dronecore::rpc::action::ActionResult();
         rpc_action_result->set_result(rpc_result);
-        rpc_action_result->set_result_str(dronecore::Action::result_str(action_result));
+        rpc_action_result->set_result_str(dronecore::action_result_str(action_result));
 
         response->set_allocated_action_result(rpc_action_result);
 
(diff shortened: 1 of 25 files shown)


JonasVautherin:
@hamishwillee Thanks for the review and  interesting comments :-). 

> Given that the result types and results strings for all plugins are similar ... does it make sense to have a common result class (or 2) rather than lots of little ones?

Similar, but not identical. Having a `DENIED_NOT_LANDED` `ActionResult` makes sense to me, but it would be weird in a `ConnectionResult`. I am in favor of sharing libraries between plugins when it makes sense, but in this case I think we will soon have some kind of `CameraResult` that will be quite different, etc. Finally, having one common result class will not simplify the code, I believe :-).

> Even if not, if we're going to end up pulling the other results classes out in future, let's do all in one go.

I guess it will come as I go in the backend implementation: I refactor them when I feel the need to do it from my code. I can try to make one plugin at a time, but I am not sure that doing everything at the same time is convenient for me (and that will make one huge PR). What is the problem with doing it independently for the different datastructs?

hamishwillee:
@JonasVautherin Sounds like you've thought this through so good for me.

> What is the problem with doing it independently for the different datastructs?

Nothing in particular. I'd just like to get it over and done with.

PR #338, opened 2018-03-23
Added camera plugin with basic functionality

### LINKED ISSUE #154: Add public API to control camera outside of missions
For example, as an Action. Possibly better as a Camera plugin, complementing the Gimbal plugin.

julianoes:
~Depends on https://github.com/mavlink/mavlink/pull/872.~

Todos:
- [x] take photos, videos
- [x] docstrings
- [x] change mode
- [x] integration tests
- [x] clean up timeouts
- [x] integration tests for camera status and modes
- [ ] test against SITL (can follow later)
- [ ] add example (can follow later)
- [x] camera definition and settings
- [x] test against Yuneec E90
- [x] test against Yuneec E50
- [x] test against Yuneec ET

camera definition and settings inspired by https://github.com/mavlink/qgroundcontrol/blob/master/src/Camera/QGCCameraControl.cc.

Fixes #154.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -148,6 +148,7 @@ test_script:
       copy build\third_party\gtest\googlemock\Debug\gmockd.dll build\Debug\ &&
       copy build\plugins\mission\Debug\dronecore_mission.dll build\Debug\ &&
       copy build\plugins\mission\third_party\json11\Debug\json11.dll build\Debug\ &&
+      copy build\plugins\camera\Debug\dronecore_camera.dll build\Debug\ &&
       build\Debug\unit_tests_runner.exe
     ) else (
       copy build\third_party\gtest\googlemock\gtest\Release\gtest.dll build\Release\ &&
@@ -155,5 +156,6 @@ test_script:
       copy build\third_party\gtest\googlemock\Release\gmock.dll build\Release\ &&
       copy build\plugins\mission\Release\dronecore_mission.dll build\Release\ &&
       copy build\plugins\mission\third_party\json11\Release\json11.dll build\Release\ &&
+      copy build\plugins\camera\Release\dronecore_camera.dll build\Release\ &&
       build\Release\unit_tests_runner.exe
     )
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -149,6 +149,7 @@ test_script:
       copy build\plugins\mission\Debug\dronecore_mission.dll build\Debug\ &&
       copy build\plugins\mission\third_party\json11\Debug\json11.dll build\Debug\ &&
       copy build\plugins\camera\Debug\dronecore_camera.dll build\Debug\ &&
+      copy build\third_party\tinyxml2\Debug\tinyxml2d.dll build\Debug\ &&
       build\Debug\unit_tests_runner.exe
     ) else (
       copy build\third_party\gtest\googlemock\gtest\Release\gtest.dll build\Release\ &&
@@ -157,5 +158,6 @@ test_script:
       copy build\plugins\mission\Release\dronecore_mission.dll build\Release\ &&
       copy build\plugins\mission\third_party\json11\Release\json11.dll build\Release\ &&
       copy build\plugins\camera\Release\dronecore_camera.dll build\Release\ &&
+      copy build\third_party\tinyxml2\Release\tinyxml2.dll build\Release\ &&
       build\Release\unit_tests_runner.exe
     )
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -33,7 +33,7 @@ clone_folder: C:\dronecore
 before_build:
 
   - cd C:\dronecore
-  - git submodule update --init --recursive --depth 20
+  - git submodule update --init --recursive --depth 100
   - cd C:\
   - appveyor DownloadFile https://curl.haxx.se/download/curl-7.56.1.zip
   - 7z x -y curl-7.56.1.zip
--- a/cmake/compiler_flags.cmake
+++ b/cmake/compiler_flags.cmake
@@ -35,11 +35,10 @@ else()
 endif()
 
 if(APPLE)
-    # We need a define if on APPLE
     add_definitions("-DAPPLE")
 endif()
 
-# Add DEBUG define for Debug target
+# Add DEBUG define for Debug target because that is not done automatically.
 set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -DDEBUG")
 
 set(CMAKE_CXX_FLAGS_COVERAGE "${CMAKE_CXX_FLAGS_COVERAGE} --coverage")
--- a/cmake/compiler_flags.cmake
+++ b/cmake/compiler_flags.cmake
@@ -29,6 +29,9 @@ else()
     elseif(CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang")
         set(warnings "${warnings} -Wno-missing-braces -Wno-unused-parameter")
     endif()
+
+    # Otherwise tinyxml2 complains.
+    set(warnings "${warnings} -Wno-old-style-cast")
 endif()
 
 if(APPLE)
@@ -42,6 +45,3 @@ set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -DDEBUG")
 set(CMAKE_CXX_FLAGS_COVERAGE "${CMAKE_CXX_FLAGS_COVERAGE} --coverage")
 set(CMAKE_EXE_LINKER_FLAGS_COVERAGE "${CMAKE_EXE_LINKER_FLAGS_COVERAGE} --coverage")
 set(CMAKE_LINKER_FLAGS_COVERAGE "${CMAKE_LINKER_FLAGS_COVERAGE} --coverage")
-
-# Otherwise tinyxml2 complains.
-set(warnings "${warnings} -Wno-old-style-cast")
--- a/core/call_every_handler.cpp
+++ b/core/call_every_handler.cpp
@@ -57,6 +57,7 @@ void CallEveryHandler::remove(const void *cookie)
     auto it = _entries.find(const_cast<void *>(cookie));
     if (it != _entries.end()) {
         _entries.erase(const_cast<void *>(cookie));
+        _iterator_invalidated = true;
     }
 }
 
@@ -80,6 +81,13 @@ void CallEveryHandler::run_once()
                 callback();
                 _entries_mutex.lock();
             }
+        }
+
+        // We leave the loop.
+        // FIXME: there should be a nicer way to do this.
+        if (_iterator_invalidated) {
+            _iterator_invalidated = false;
+            break;
         }
     }
     _entries_mutex.unlock();
--- a/integration_tests/camera_mode.cpp
+++ b/integration_tests/camera_mode.cpp
@@ -16,11 +16,11 @@ TEST(CameraTest, ChangeMode)
     ConnectionResult ret = dc.add_udp_connection();
     ASSERT_EQ(ret, ConnectionResult::SUCCESS);
 
-    // Wait for device to connect via heartbeat.
+    // Wait for system to connect via heartbeat.
     std::this_thread::sleep_for(std::chrono::seconds(2));
 
-    Device &device = dc.device();
-    auto camera = std::make_shared<Camera>(device);
+    System &system = dc.system();
+    auto camera = std::make_shared<Camera>(system);
 
     set_mode(camera, Camera::Mode::PHOTO);
     std::this_thread::sleep_for(std::chrono::seconds(2));
--- a/integration_tests/camera_mode.cpp
+++ b/integration_tests/camera_mode.cpp
@@ -3,6 +3,7 @@
 #include "integration_test_helper.h"
 #include "global_include.h"
 #include "dronecore.h"
+#include "system.h"
 #include "camera_test_helpers.h"
 
 
@@ -20,6 +21,7 @@ TEST(CameraTest, ChangeMode)
     std::this_thread::sleep_for(std::chrono::seconds(2));
 
     System &system = dc.system();
+    ASSERT_TRUE(system.has_camera());
     auto camera = std::make_shared<Camera>(system);
 
     set_mode(camera, Camera::Mode::PHOTO);
--- a/integration_tests/camera_settings.cpp
+++ b/integration_tests/camera_settings.cpp
@@ -12,6 +12,8 @@ using namespace std::placeholders; // for `_1`
 
 // To run specific tests for Yuneec cameras.
 const static bool is_e90 = false;
+const static bool is_e50 = false;
+const static bool is_et = false;
 
 TEST(CameraTest, ShowSettingsAndOptions)
 {
@@ -27,50 +29,65 @@ TEST(CameraTest, ShowSettingsAndOptions)
     ASSERT_TRUE(system.has_camera());
     auto camera = std::make_shared<Camera>(system);
 
-
     // Wait for download to happen.
     std::this_thread::sleep_for(std::chrono::seconds(2));
 
-    // Set to photo mode
-    set_mode(camera, Camera::Mode::PHOTO);
+    if (is_e90 || is_e50 || is_et) {
+        // Set to photo mode
+        set_mode(camera, Camera::Mode::PHOTO);
 
-    std::vector<std::string> settings;
-    EXPECT_TRUE(camera->get_possible_settings(settings));
+        std::vector<std::string> settings;
+        EXPECT_TRUE(camera->get_possible_settings(settings));
 
-    LogDebug() << "Possible settings in photo mode: ";
-    for (auto setting : settings) {
-        LogDebug() << "-" << setting;
-    }
+        LogDebug() << "Possible settings in photo mode: ";
+        for (auto setting : settings) {
+            LogDebug() << "-" << setting;
+        }
 
-    if (is_e90) {
-        EXPECT_EQ(settings.size(), 9);
-    }
+        if (is_e90) {
+            EXPECT_EQ(settings.size(), 9);
+        } else if (is_e50) {
+            EXPECT_EQ(settings.size(), 6);
+        } else if (is_et) {
+            EXPECT_EQ(settings.size(), 5);
+        }
 
-    set_mode(camera, Camera::Mode::VIDEO);
+        set_mode(camera, Camera::Mode::VIDEO);
 
-    std::this_thread::sleep_for(std::chrono::seconds(2));
+        std::this_thread::sleep_for(std::chrono::seconds(2));
 
-    EXPECT_TRUE(camera->get_possible_settings(settings));
+        EXPECT_TRUE(camera->get_possible_settings(settings));
 
-    LogDebug() << "Possible settings in video mode: ";
-    for (auto setting : settings) {
-        LogDebug() << "-" << setting;
-    }
+        LogDebug() << "Possible settings in video mode: ";
+        for (auto setting : settings) {
+            LogDebug() << "-" << setting;
+        }
 
-    if (is_e90) {
-        EXPECT_EQ(settings.size(), 6);
-    }
+        if (is_e90) {
+            EXPECT_EQ(settings.size(), 6);
+        } else if (is_e50) {
+            EXPECT_EQ(settings.size(), 5);
+        } else if (is_et) {
+            EXPECT_EQ(settings.size(), 5);
+        }
 
-    if (is_e90) {
         std::vector<std::string> options;
 
-        // Try something that is specific to the camera mode.
-        EXPECT_TRUE(camera->get_possible_options("CAM_VIDRES", options));
-        EXPECT_EQ(options.size(), 32);
+        if (is_e90) {
+            // Try something that is specific to the camera mode.
+            EXPECT_TRUE(camera->get_possible_options("CAM_VIDRES", options));
+            EXPECT_EQ(options.size(), 32);
+        } else if (is_e50) {
+            // Try something that is specific to the camera mode.
+            EXPECT_TRUE(camera->get_possible_options("CAM_VIDRES", options));
+            EXPECT_EQ(options.size(), 12);
+        }
 
-        // This param is not applicable, so we should get an empty vector back.
-        EXPECT_FALSE(camera->get_possible_options("CAM_PHOTOQUAL", options));
-        EXPECT_EQ(options.size(), 0);
+        if (is_e90) {
+            // This param is not applicable, so we should get an empty vector back.
+            EXPECT_FALSE(camera->get_possible_options("CAM_PHOTOQUAL", options));
+            EXPECT_EQ(options.size(), 0);
+        }
 
         // The same should happen with a param that does not exist at all.
         EXPECT_FALSE(camera->get_possible_options("CAM_BLABLA", options));
@@ -78,9 +95,11 @@ TEST(CameraTest, ShowSettingsAndOptions)
 
         set_mode(camera, Camera::Mode::PHOTO);
 
-        // Try something that is specific to the camera mode.
-        EXPECT_TRUE(camera->get_possible_options("CAM_PHOTOQUAL", options));
-        EXPECT_EQ(options.size(), 4);
+        if (is_e90) {
+            // Try something that is specific to the camera mode.
+            EXPECT_TRUE(camera->get_possible_options("CAM_PHOTOQUAL", options));
+            EXPECT_EQ(options.size(), 4);
+        }
 
         // This param is not applicable, so we should get an empty vector back.
         EXPECT_FALSE(camera->get_possible_options("CAM_VIDRES", options));
--- a/integration_tests/camera_settings.cpp
+++ b/integration_tests/camera_settings.cpp
@@ -107,7 +107,7 @@ TEST(CameraTest, SettingsAnySetting)
     auto ret = prom->get_future();
 
     // TODO: use a valid setting here
-    camera->set_option_key_async("FOO", "BAR",
+    camera->set_option_async("FOO", "BAR",
     [prom](Camera::Result result) {
         prom->set_value(result);
     });
--- a/integration_tests/camera_settings.cpp
+++ b/integration_tests/camera_settings.cpp
@@ -10,6 +10,9 @@
 using namespace dronecore;
 using namespace std::placeholders; // for `_1`
 
+// To run specific tests for Yuneec cameras.
+const static bool is_e90 = false;
+
 TEST(CameraTest, ShowSettingsAndOptions)
 {
     DroneCore dc;
@@ -38,7 +41,10 @@ TEST(CameraTest, ShowSettingsAndOptions)
     for (auto setting : settings) {
         LogDebug() << "-" << setting;
     }
-    EXPECT_EQ(settings.size(), 9);
+
+    if (is_e90) {
+        EXPECT_EQ(settings.size(), 9);
+    }
 
     set_mode(camera, Camera::Mode::VIDEO);
 
@@ -50,35 +56,40 @@ TEST(CameraTest, ShowSettingsAndOptions)
     for (auto setting : settings) {
         LogDebug() << "-" << setting;
     }
-    EXPECT_EQ(settings.size(), 6);
 
-    std::vector<std::string> options;
+    if (is_e90) {
+        EXPECT_EQ(settings.size(), 6);
+    }
 
-    // Try something that is specific to the camera mode.
-    EXPECT_TRUE(camera->get_possible_options("CAM_VIDRES", options));
-    EXPECT_EQ(options.size(), 32);
+    if (is_e90) {
+        std::vector<std::string> options;
 
-    // This param is not applicable, so we should get an empty vector back.
-    EXPECT_FALSE(camera->get_possible_options("CAM_PHOTOQUAL", options));
-    EXPECT_EQ(options.size(), 0);
+        // Try something that is specific to the camera mode.
+        EXPECT_TRUE(camera->get_possible_options("CAM_VIDRES", options));
+        EXPECT_EQ(options.size(), 32);
 
-    // The same should happen with a param that does not exist at all.
-    EXPECT_FALSE(camera->get_possible_options("CAM_BLABLA", options));
-    EXPECT_EQ(options.size(), 0);
+        // This param is not applicable, so we should get an empty vector back.
+        EXPECT_FALSE(camera->get_possible_options("CAM_PHOTOQUAL", options));
+        EXPECT_EQ(options.size(), 0);
 
-    set_mode(camera, Camera::Mode::PHOTO);
+        // The same should happen with a param that does not exist at all.
+        EXPECT_FALSE(camera->get_possible_options("CAM_BLABLA", options));
+        EXPECT_EQ(options.size(), 0);
 
-    // Try something that is specific to the camera mode.
-    EXPECT_TRUE(camera->get_possible_options("CAM_PHOTOQUAL", options));
-    EXPECT_EQ(options.size(), 4);
+        set_mode(camera, Camera::Mode::PHOTO);
 
-    // This param is not applicable, so we should get an empty vector back.
-    EXPECT_FALSE(camera->get_possible_options("CAM_VIDRES", options));
-    EXPECT_EQ(options.size(), 0);
+        // Try something that is specific to the camera mode.
+        EXPECT_TRUE(camera->get_possible_options("CAM_PHOTOQUAL", options));
+        EXPECT_EQ(options.size(), 4);
 
-    // The same should happen with a param that does not exist at all.
-    EXPECT_FALSE(camera->get_possible_options("CAM_BLABLA", options));
-    EXPECT_EQ(options.size(), 0);
+        // This param is not applicable, so we should get an empty vector back.
+        EXPECT_FALSE(camera->get_possible_options("CAM_VIDRES", options));
+        EXPECT_EQ(options.size(), 0);
+
+        // The same should happen with a param that does not exist at all.
+        EXPECT_FALSE(camera->get_possible_options("CAM_BLABLA", options));
+        EXPECT_EQ(options.size(), 0);
+    }
 }
 
 TEST(CameraTest, SetSettings)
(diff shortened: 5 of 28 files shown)


plugins/camera/camera_impl.cpp
```diff
@@ -0,0 +1,154 @@
...
+void CameraImpl::enable() {}
+
+void CameraImpl::disable() {}
+
+Camera::Result CameraImpl::take_photo()
+{
+    return camera_result_from_command_result(
+               _parent.send_command_with_ack(
+                   MAV_CMD_IMAGE_START_CAPTURE,
+                   MavlinkCommands::Params {0.0f, // all camera IDs
```
shakthi-prashanth-m: What is the difference between Camera ID and Camera Component ID ?

plugins/camera/camera_impl.cpp
```diff
@@ -0,0 +1,154 @@
...
+                                            0.0f, // only picture, no interval
+                                            1.0f, // take only one picture
+                                            float(_capture_sequence++),
+                                            NAN, NAN, NAN},
+                   MAV_COMP_ID_CAMERA));
```
shakthi-prashanth-m: As we discussed, there'll be one instance per camera; so, camera component ID would be a member. Isn't it ?

plugins/camera/camera_impl.cpp
julianoes: Camera ID should actually be a thing of the past. I removed the comments.

plugins/camera/camera_impl.cpp
julianoes: That's right. For now I'm just going to add the current single device plugin. Later we can extend it.

plugins/camera/camera.h
```diff
@@ -0,0 +1,294 @@
...
+#include <map>
+#include "plugin_base.h"
+
+namespace dronecore {
+
+class CameraImpl;
+class Device;
+
+/**
+ * @brief The Camera class allows to control generic mavlink cameras.
```
hamishwillee: Perhaps to
> The Camera class can be used to manage cameras that implement the MAVLink Camera Protocol: https://mavlink.io/en/protocol/camera.html

hamishwillee: This plugin is different in that multiple instances can be used with one device. We should capture that here too.

plugins/camera/camera.h
```diff
@@ -0,0 +1,294 @@
...
+ *
+ * Synchronous and asynchronous variants of the camera methods are supplied.
+ */
+class Camera : public PluginBase
+{
+public:
+    /**
+     * @brief Constructor. Creates the plugin for a specific Device.
```
hamishwillee: FYI, if that rename for `System` goes in then this will have to change.

hamishwillee:
This looks like it supports just a single camera per device (if not, then need to explain the "how"here). There is also no example.

integration_tests/camera_mode.cpp
```diff
@@ -0,0 +1,66 @@
...
+    std::this_thread::sleep_for(std::chrono::seconds(2));
+
+    set_mode(camera, Camera::Mode::PHOTO);
+    std::this_thread::sleep_for(std::chrono::seconds(2));
+    new_mode = get_mode(camera);
+    EXPECT_EQ(new_mode, Camera::Mode::PHOTO);
+    std::this_thread::sleep_for(std::chrono::seconds(2));
+
+    set_mode(camera, Camera::Mode::VIDEO);
+    std::this_thread::sleep_for(std::chrono::seconds(2));
```
shakthi-prashanth-m: I assume this is just like pausing and not mandatory for set mode activity ? Because, I see you're already waiting in `set_mode()`.

integration_tests/camera_settings.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+    // Set to photo mode
+    set_mode(camera, Camera::Mode::PHOTO);
+
+    std::map<std::string, std::string> settings;
+    EXPECT_TRUE(camera->get_possible_settings(settings));
+
+    LogDebug() << "Possible settings in photo mode: ";
+    for (auto setting : settings) {
+        LogDebug() << "- " << setting.second;
```
shakthi-prashanth-m: It looks if both name and value are printed. Isn't it ?

integration_tests/camera_mode.cpp
julianoes: Good catch, double is probably not needed.

integration_tests/camera_settings.cpp
julianoes: I think so, I need to work on this, it's far from finished.

plugins/camera/camera_impl.cpp
```diff
@@ -0,0 +1,739 @@
...
+
+    _parent.unregister_timeout_handler(_get_mode.timeout_cookie);
+}
+
+void CameraImpl::process_camera_information(const mavlink_message_t &message)
+{
+    mavlink_camera_information_t camera_information;
+    mavlink_msg_camera_information_decode(&message, &camera_information);
+
+    load_definition_file(camera_information.cam_definition_uri);
```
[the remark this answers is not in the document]
julianoes: Not yet, good point.

plugins/camera/camera_impl.cpp
```diff
@@ -0,0 +1,739 @@
...
+
+    _parent.send_command_with_ack_async(
+        MAV_CMD_REQUEST_CAMERA_CAPTURE_STATUS,
+        MavlinkCommands::Params {1.0f, // request it
+                                 NAN, NAN, NAN, NAN, NAN, NAN},
+        std::bind(&CameraImpl::receive_camera_capture_status_result, this, _1),
+        MAV_COMP_ID_CAMERA);
+
+    _parent.send_command_with_ack_async(
+        MAV_CMD_REQUEST_STORAGE_INFORMATION,
```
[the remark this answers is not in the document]
julianoes: Yer we should probably remove the WIP soon. It has been there for a long time.

plugins/camera/camera_definition_test.cpp
```diff
@@ -0,0 +1,89 @@
...
+#include <memory>
+#include <fstream>
+
+using namespace dronecore;
+
+static const char *e90_unit_test_file = "extensions/plugins/camera/e90_unit_test.xml";
+
+TEST(CameraDefinition, LoadE90InfoFile)
+{
+    CameraDefinition cd;
```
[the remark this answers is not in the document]
julianoes: :smile:

plugins/camera/camera.h
```diff
@@ -0,0 +1,298 @@
...
+
+    /**
+     * @brief Stop photo interval (synchronous).
+     *
+     * Stops a photo timelapse again.
+     *
+     * @return Result of request.
+     */
+    Result stop_photo_interval();
+
```
shakthi-prashanth-m: How about providing a static method that tells the number of cameras on the drone.
```
static size_t Camera::how_many_cameras() const;
```
OR
```
```
static size_t Camera::total_cameras() const;
```
whatever is more intuitive.

File: plugins/camera/camera.h [hunk 11]
```diff
@@ -0,0 +1,298 @@
...
+     * @brief Getter for camera mode (asynchronous).
+     *
+     * @param callback Function to call with result of request.
+     */
+    void get_mode_async(const mode_callback_t &callback);
+
+    /**
+     * @brief Information about a picture just captured.
+     */
+    struct CaptureInfo {
```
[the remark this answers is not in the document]
julianoes: I would say if it is standalone you actually need to look into it because you can't look at the autopilot's telemetry.

plugins/camera/camera.h
```diff
@@ -0,0 +1,298 @@
...
+     * @brief Get camera status.
+     *
+     * @param callback Function to call with camera status.
+     */
+    void get_status_async(get_status_callback_t callback);
+
+    bool get_possible_settings(std::map<std::string, std::string> &settings);
+    bool get_possible_options(const std::string &setting_name, std::vector<std::string> &options);
+
+    typedef std::function<void(Result, const std::string &)> get_option_callback_t;
```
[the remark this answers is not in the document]
julianoes: I tried to group together with methods that use the typedefs, so that you don't need to scroll around when looking at a method signature.

plugins/camera/camera.h
julianoes: As said, I'll work on multi camera support when single camera support is merged and properly working. I can't do all at once :smile:.

plugins/camera/camera.h
```diff
@@ -0,0 +1,298 @@
...
+     * @brief Callback type for capture info updates.
+     */
+    typedef std::function<void(CaptureInfo)> capture_info_callback_t;
+
+    /**
+     * @brief Subscribe to capture info updates (asynchronous).
+     *
+     * @param callback Function to call with updates.
+     */
+    void capture_info_async(capture_info_callback_t callback);
```
shakthi-prashanth-m: So, we've only async version for capture_info and not that of sync ?

plugins/camera/camera_impl.cpp
```diff
@@ -0,0 +1,739 @@
...
+            return Camera::Result::UNKNOWN;
+    }
+}
+
+void CameraImpl::set_mode_async(Camera::Mode mode, const Camera::mode_callback_t &callback)
+{
+    float mavlink_mode;
+    switch (mode) {
+        case Camera::Mode::PHOTO:
+            mavlink_mode = CAMERA_MODE_IMAGE;
```
shakthi-prashanth-m: I'm just curious: Didn't you get auto type conversion error from `uint8_t` to `float` ? (In MSVC at least :) )

plugins/camera/camera_impl.cpp
```diff
@@ -0,0 +1,739 @@
...
+            Camera::CaptureInfo capture_info = {};
+            capture_info.position.latitude_deg = image_captured.lat / 1e7;
+            capture_info.position.longitude_deg = image_captured.lon / 1e7;
+            capture_info.position.absolute_altitude_m = image_captured.alt / 1e3f;
+            capture_info.position.relative_altitude_m = image_captured.relative_alt / 1e3f;
+            capture_info.time_utc_us = image_captured.time_utc;
+            capture_info.quaternion.w = image_captured.q[0];
+            capture_info.quaternion.x = image_captured.q[1];
+            capture_info.quaternion.y = image_captured.q[2];
+            capture_info.quaternion.z = image_captured.q[3];
```
shakthi-prashanth-m: How about having friend method:
```cpp
friend const Camera::CaptureInfo &operator=(Camera::CaptureInfo& to, float quaternion[4]);
```
so that we can do:
```cpp
capture_info.quaternion = image_captured.q;
```
I mean, may be later.

plugins/camera/camera_impl.cpp
```diff
@@ -0,0 +1,739 @@
...
+bool CameraImpl::get_possible_options(const std::string &setting_name,
+                                      std::vector<std::string> &options)
+{
+    if (!_camera_definition) {
+        LogWarn() << "Error: no camera definition available yet";
+        return false;
+    }
+
+    CameraDefinition::parameter_map_t parameters;
+    _camera_definition->get_parameters(parameters, true);
```
shakthi-prashanth-m: Not checking return value ?

plugins/camera/camera_impl.cpp
```diff
@@ -0,0 +1,739 @@
...
+}
+
+void CameraImpl::get_option_key_async(const std::string &setting_key,
+                                      const Camera::get_option_callback_t &callback)
+{
+    if (_camera_definition) {
+        LogWarn() << "Error: no camera defnition available yet.";
+        if (callback) {
+            callback(Camera::Result::ERROR, "");
+        }
```
shakthi-prashanth-m: I am not sure. Don't we need to return from here ? It seems falling through.

plugins/camera/camera_definition.h
```diff
@@ -0,0 +1,64 @@
...
+    struct Parameter {
+        std::string description;
+        Type type;
+        option_list_t options;
+    };
+
+    void update_setting(const std::string &name, const ParameterValue &value);
+
+    typedef std::map<std::string, std::shared_ptr<Parameter>> parameter_map_t;
+    bool get_parameters(parameter_map_t &parameters, bool filter_possible);
```
shakthi-prashanth-m: Its a selector, isn't it ? If so, it should be
```cpp
bool get_parameters(parameter_map_t &parameters, bool filter_possible) const;
```

plugins/camera/camera_definition_test.cpp
```diff
@@ -0,0 +1,89 @@
...
+#include "log.h"
+#include <gtest/gtest.h>
+#include <vector>
+#include <map>
+#include <memory>
+#include <fstream>
+
+using namespace dronecore;
+
+static const char *e90_unit_test_file = "extensions/plugins/camera/e90_unit_test.xml";
```
shakthi-prashanth-m: `std::string` is better. :)

plugins/camera/camera_impl.cpp
```diff
@@ -0,0 +1,739 @@
...
+    return camera_result_from_command_result(
+               _parent.send_command_with_ack(
+                   MAV_CMD_IMAGE_STOP_CAPTURE,
+                   MavlinkCommands::Params {0.0f, // Reserved, set to 0
+                                            NAN, NAN, NAN, NAN, NAN, NAN},
+                   MAV_COMP_ID_CAMERA));
+}
+
+Camera::Result CameraImpl::start_video()
+{
```
shakthi-prashanth-m: Don't we need to check whether video capture in progress ? Also, we need to make sure we're in Video mode. Same applies to async versions.

plugins/camera/camera_impl.cpp
```diff
@@ -0,0 +1,739 @@
...
+
+void CameraImpl::enable() {}
+
+void CameraImpl::disable() {}
+
+Camera::Result CameraImpl::take_photo()
+{
+    std::lock_guard<std::mutex> lock(_capture.mutex);
+
+    return camera_result_from_command_result(
```
shakthi-prashanth-m: Don't we need to make sure we're in Photo mode? Same applies to async versions.

plugins/camera/camera.h
julianoes: I guess so, for now.

plugins/camera/camera_impl.cpp
```diff
@@ -0,0 +1,739 @@
...
+    }
+
+    _get_mode.callback(Camera::Result::SUCCESS, mode);
+    _get_mode.callback = nullptr;
+
+    _parent.unregister_timeout_handler(_get_mode.timeout_cookie);
+}
+
+void CameraImpl::process_camera_information(const mavlink_message_t &message)
+{
```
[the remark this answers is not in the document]
julianoes: This is all for one camera only, for now.

plugins/camera/camera_impl.cpp
```diff
@@ -0,0 +1,739 @@
...
+    std::lock_guard<std::mutex> lock(_get_mode.mutex);
+
+    if (_get_mode.callback) {
+        _get_mode.callback(Camera::Result::TIMEOUT, Camera::Mode::UNKNOWN);
+        _get_mode.callback = nullptr;
+    }
+}
+
+void CameraImpl::receive_int_param(const std::string &name, bool success, int value)
```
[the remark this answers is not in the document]
julianoes: I prefer `UNUSED(...)` because it makes it very explicit.

plugins/camera/camera_impl.cpp
julianoes: Hehe, don't know.

plugins/camera/camera_impl.cpp
julianoes: Nice, didn't know this.

plugins/camera/camera_impl.cpp
julianoes: Work in progress...

plugins/camera/camera_impl.cpp
julianoes: Very good catch, thank you!

plugins/camera/camera_impl.cpp
julianoes: I made a TODO.

plugins/camera/camera_impl.cpp
julianoes: Yap, thx.

plugins/camera/camera_impl.cpp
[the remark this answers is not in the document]
shakthi-prashanth-m: Yeah, got it :)

My point is that, when we support one camera, it should work well with one camera regardless of how many cameras the system has. But due to the existence of other camera, this implementation should not get affected.

In this case, we get bombarded with camera information of several cameras, so we overwrite first camera info by the other. So **we should have a mechanism to restrict information from other cameras**, and receive only information from **one camera**.

plugins/camera/camera_impl.cpp
shakthi-prashanth-m: OK.

plugins/camera/camera_impl.h
```diff
@@ -0,0 +1,122 @@
...
+
+    void check_status();
+
+    void status_timeout_happened();
+
+    void load_definition_file(const std::string &uri);
+    void receive_int_param(const std::string &name, bool success, int value);
+    void receive_float_param(const std::string &name, bool success, float value);
+
+    CameraDefinition *_camera_definition = nullptr;
```
shakthi-prashanth-m: It could be a unique pointer, I guess.

julianoes: Good idea.

CMakeLists.txt
```diff
@@ -29,6 +29,41 @@ endif()
 set(dronecore_install_include_dir "include/dronecore")
 set(dronecore_install_lib_dir ${lib_path})
 
+# We need tinyxml2 for the camera definition parsing.
```
JonasVautherin: Just to comment on comments (meta-commenting :D): this comment will get outdated as soon as we use tinyxml2 for another purpose than the camera definition parsing. So I will probably never trust it.

If at some point I question the need for tinyxml2, I will make my own research: look for includes for tinyxml2 in the code, or try to remove the dependency and see where the build breaks. This comment doesn't add value to me, as in the process I will realize that tinyxml2 is used for the camera, indeed.

cmake/compiler_flags.cmake
```diff
@@ -29,8 +29,8 @@ else()
     endif()
 endif()
 
-# We need a define if on APPLE
 if(APPLE)
+    # We need a define if on APPLE
```
JonasVautherin: Same thing here: `if(APPLE) add_definitions("-DAPPLE")` tells me that "if `APPLE` is defined, we add the definition called 'APPLE'". And I assume it is needed, otherwise we would not do it.

The comment represents 25% of this piece of code and just says exactly the same :P. Even worse: maybe at some point, the comment will get disconnected from the `add_definitions` call, and it will be misguiding.

cmake/compiler_flags.cmake
```diff
@@ -40,3 +40,6 @@ set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -DDEBUG")
 set(CMAKE_CXX_FLAGS_COVERAGE "${CMAKE_CXX_FLAGS_COVERAGE} --coverage")
 set(CMAKE_EXE_LINKER_FLAGS_COVERAGE "${CMAKE_EXE_LINKER_FLAGS_COVERAGE} --coverage")
 set(CMAKE_LINKER_FLAGS_COVERAGE "${CMAKE_LINKER_FLAGS_COVERAGE} --coverage")
+
+# Otherwise tinyxml2 complains.
```
JonasVautherin: And I believe this one is useful, because that's a bit peculiar.

CMakeLists.txt
julianoes: I suppose so.

cmake/compiler_flags.cmake
julianoes: I guess I wrote these comments because I'm a cmake noob so for me it wasn't trivial that the variable `APPLE` would not result in a define `APPLE` while compiling. Now I know :smile: .

cmake/compiler_flags.cmake
julianoes: Right, anyway I appreciate the discussion :+1:

plugins/camera/camera_impl.h
```diff
@@ -0,0 +1,123 @@
...
+        void *timeout_cookie {nullptr};
+    } _status;
+
+    static constexpr double DEFAULT_TIMEOUT_S = 3.0;
+
+    struct {
+        std::mutex mutex {};
+        Camera::mode_callback_t callback {nullptr};
+        void *timeout_cookie {nullptr};
+    } _get_mode;
```
shakthi-prashanth-m: I think we should store camera mode here. We can use this mode to validate whether we're in Photo or Video mode in APIs like start_video(), take_photo(), etc. Isn't it ?

shakthi-prashanth-m:
@julianoes please rebase to develop. Thanks.

plugins/camera/camera_impl.cpp
```diff
@@ -0,0 +1,766 @@
...
+{
+    // TODO: check whether video capture is already in progress.
+    // TODO: check whether we are in video mode.
+
+    return camera_result_from_command_result(
+               _parent->send_command_with_ack(
+                   MAV_CMD_VIDEO_START_CAPTURE,
+                   MAVLinkCommands::Params {0.0f, // Reserved, set to 0
+                                            NAN, // fps not set yet
+                                            NAN, // resolution not set yet
```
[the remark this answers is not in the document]
julianoes: oh you're right, that changed.

shakthi-prashanth-m:
@julianoes I re-based to the latest changes in the core and added video stream interface.
I have to add integration tests yet. Kindly review. Thanks.

shakthi-prashanth-m:
Appveyor build fails. Is this the issue from Gtest repo ?
```
error: Server does not allow request for unadvertised object 9dde7c4e8c46a65ad6215440bb396ad219e06ed4
Fetched in submodule path 'third_party/gtest', but it did not contain 9dde7c4e8c46a65ad6215440bb396ad219e06ed4. Direct fetching of that commit failed.
Command exited with code 1
```

julianoes:
@shakthi-prashanth-m try increasing the depth here:
https://github.com/dronecore/DroneCore/blob/develop/appveyor.yml#L36

julianoes:
@shakthi-prashanth-m @JonasVautherin I think this is now in a state where it could be reviewed and get merged for now.

plugins/camera/camera.h
```diff
@@ -0,0 +1,381 @@
...
+namespace dronecore {
+
+class CameraImpl;
+class System;
+
+/**
+ * @brief The Camera class can be used to manage cameras that implement the
+ * MAVLink Camera Protocol: https://mavlink.io/en/protocol/camera.html.
+ *
+ * The plugin needs to be instantiated multiple times for multiple cameras.
```
hamishwillee: Suggest change:
```
 * The plugin needs to be instantiated multiple times for multiple cameras.
 * However, this is not supported yet.
```
To
```
 * Currently only a single camera is supported.
 * When multiple cameras are supported the plugin will need to instantiated separately for every camera.
```

plugins/camera/camera.h
```diff
@@ -0,0 +1,381 @@
...
+     *
+     * @return Result of request.
+     */
+    Result take_photo();
+
+    /**
+     * @brief Start photo interval (synchronous).
+     *
+     * Starts a photo timelapse given an interval.
+     *
```
hamishwillee: Maybe add see also to the corresponding stop method (and visa versa).

```
* @sa stop_photo_interval
```

plugins/camera/camera.h
```diff
@@ -0,0 +1,381 @@
...
+     * This takes one photo immediately.
+     *
+     * @return Result of request.
+     */
+    Result take_photo();
+
+    /**
+     * @brief Start photo interval (synchronous).
+     *
+     * Starts a photo timelapse given an interval.
```
hamishwillee: Starts a photo timelapse with a given interval.

plugins/camera/camera.h
```diff
@@ -0,0 +1,381 @@
...
+     *
+     * @param interval_s The interval between photos in seconds.
+     * @return Result of request.
+     */
+    Result start_photo_interval(float interval_s);
+
+    /**
+     * @brief Stop photo interval (synchronous).
+     *
+     * Stops a photo timelapse again.
```
hamishwillee: Again is redundant. Either "Stops a photo timelapse" or "Stops a photo timelapse, previously started with `start_photo_interval`"

plugins/camera/camera.h
hamishwillee: Lazy :-)

plugins/camera/camera.h
```diff
@@ -0,0 +1,381 @@
...
+     *
+     * @return Result of request.
+     */
+    Result stop_photo_interval();
+
+    /**
+     * @brief Start video capture (synchronous).
+     *
+     * This starts a video recording immediately.
```
hamishwillee: Immediately is somewhat redundant - so as for image, I'd refer "This starts video recording/capture." or if necessary  "This starts video recording/capture (immediately)."

plugins/camera/camera.h
```diff
@@ -0,0 +1,381 @@
...
+
+    /**
+     * @brief Callback type for asynchronous Camera calls.
+     */
+    typedef std::function<void(Result)> result_callback_t;
+
+    /**
+     * @brief Take photo (synchronous).
+     *
+     * This takes one photo immediately.
```
hamishwillee: Immediately is somewhat redundant, I'd prefer "This takes one photo" or if necessary  "This captures a single photo (immediately)."

plugins/camera/camera.h
```diff
@@ -0,0 +1,294 @@
...
+     * This starts a video recording immediately.
+     *
+     * @return Result of request.
+     */
+    Result start_video();
+
+    /**
+     * @brief Stop video capture (synchronous).
+     *
+     * This stops a video recording again.
```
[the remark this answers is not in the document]
hamishwillee: See comments for stopping image capture.

plugins/camera/camera.h
```diff
@@ -0,0 +1,381 @@
...
+     *
+     * @return Result of request.
+     */
+    Result stop_video();
+
+    /**
+     * @brief Take photo (asynchronous).
+     *
```
hamishwillee: Comments for async methods as for sync methods.

plugins/camera/camera.h
```diff
@@ -0,0 +1,381 @@
...
+     * @param callback Function to call with result of request.
+     */
+    void stop_video_async(const result_callback_t &callback);
+
+
+    /**
+     * @brief Camera mode type.
+     */
+    enum class Mode {
+        PHOTO,
```
hamishwillee: String for all enum values. Yes pointless, but does render output better.

plugins/camera/camera.h
```diff
@@ -0,0 +1,381 @@
...
+     * @param callback Function to call with result of request.
+     */
+    void get_mode_async(const mode_callback_t &callback);
+
+    /**
+     * @brief Information about a picture just captured.
+     */
+    struct CaptureInfo {
+        /**
+         * @brief Position type.
```
hamishwillee: Didn't position and Quaternion recently get pulled out of telemetry class? If so, are they reusable?

plugins/camera/camera.h
```diff
@@ -0,0 +1,381 @@
...
+     * @brief Type for video stream settings.
+     *
+     * Application may call set_video_stream_settings() before starting
+     * video streaming to tell Camera server to use these settings during video streaming.
+     *
+     * @sa set_video_stream_settings(), get_video_stream_info(), start_video_streaming().
+     */
+    struct VideoStreamSettings {
+        float frame_rate_hz = 0.f; /**< @brief Frames per second. */
+        uint16_t resolution_h_pix = 0u; /**< @brief Resolution horizontal in pixels. */
```
hamishwillee: Horizontal resolution ...

plugins/camera/camera.h
```diff
@@ -0,0 +1,381 @@
...
+        uint16_t resolution_v_pix = 0u; /**< @brief Resolution verticle in pixels. */
```
hamishwillee: "Vertical resolution in pixels". Note spelling of vertical.

plugins/camera/camera.h
```diff
@@ -0,0 +1,381 @@
...
+        uint32_t bit_rate_b_s = 0u; /**< @brief Bit rate in bits per second. */
+        uint16_t rotation_deg = 0u; /**< @brief Video image rotation clockwise (0-359 degrees) */
+        std::string uri {}; /**< @brief Video stream URI */
```
hamishwillee: Terminate consistently with full stops (across all of these).

BTW, I am not sure this bit of the spec works yet (RTSP stream setting via MAVLink)

plugins/camera/camera.h
```diff
@@ -0,0 +1,381 @@
...
+
+        constexpr static const float FRAME_RATE_HIGHEST = -1.0f;
+        constexpr static const uint16_t RESOLUTION_H_HIGHEST = UINT16_MAX;
+        constexpr static const uint16_t RESOLUTION_V_HIGHEST = UINT16_MAX;
+        constexpr static const uint32_t BIT_RATE_AUTO = UINT32_MAX;
+    };
+
+    /**
+     * @brief Sets video stream settings.
+     *
```
hamishwillee: Do we need to give any guidance about what the receiving system will do with invalid information - ie invalid for the specific camera? Does spec state that a -1.0f framerate is treated as "max"?

FYI, if these are setting parameters as defined in the camera definition file, it is probably that this code won't work yet.

plugins/camera/camera.h
```diff
@@ -0,0 +1,381 @@
...
+     */
+    void set_video_stream_settings(const VideoStreamSettings &settings);
+
+    /**
+     * @brief Information about video stream.
+     */
+    struct VideoStreamInfo {
+        VideoStreamSettings settings; /**< @brief Video stream settings. */
+        enum class Status {
+            NOT_RUNNING = 0, IN_PROGRESS
```
hamishwillee: Odd to see them both in one line. Need docs?

plugins/camera/camera.h
```diff
@@ -0,0 +1,381 @@
...
+    /**
+     * @brief Subscribe to capture info updates (asynchronous).
+     *
+     * @param callback Function to call with updates.
+     */
+    void capture_info_async(capture_info_callback_t callback);
+
+    /**
+     * @brief Information about camera status.
+     */
```
hamishwillee: Docs for rest of this?

plugins/camera/camera.h
```diff
@@ -0,0 +1,381 @@
...
+
+
+    /**
+     * @brief Get camera status.
+     *
+     * @param callback Function to call with camera status.
+     */
+    void get_status_async(get_status_callback_t callback);
+
+    bool get_possible_settings(std::vector<std::string> &settings);
```
hamishwillee: Are docs needed for this and following. Ideally everything in public API should be documented.

plugins/camera/camera_definition.h
```diff
@@ -0,0 +1,89 @@
+#pragma once
```
hamishwillee: Not part of the public API right?

julianoes:
@hamishwillee thanks for the comments. I'll work on these.

appveyor.yml
```diff
@@ -33,7 +33,7 @@ clone_folder: C:\dronecore
 before_build:
 
   - cd C:\dronecore
-  - git submodule update --init --recursive --depth 20
+  - git submodule update --init --recursive --depth 100
```
JonasVautherin: I don't wanna know :D

core/call_every_handler.cpp
```diff
@@ -81,6 +82,13 @@ void CallEveryHandler::run_once()
                 _entries_mutex.lock();
             }
         }
+
+        // We leave the loop.
+        // FIXME: there should be a nicer way to do this.
```
JonasVautherin: (Just trying, maybe it's not better)

```
size_t i = 0;
while (i < _entries.size() && !_iterator_invalidated) {
    ...
    i++;
}
```

... instead of the `for` loop of line 68.

plugins/camera/camera.h
shakthi-prashanth-m: Yes the spec says `framerate = -1.0f` is treated as "max".

shakthi-prashanth-m: >if these are setting parameters as defined in the camera definition file

I'm not sure of this. @julianoes might know.

shakthi-prashanth-m: http://mavlink.org/messages/common#SET_VIDEO_STREAM_SETTINGS

integration_tests/camera_test_helpers.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+void set_mode(std::shared_ptr<Camera> camera, Camera::Mode mode)
+{
+    // FIXME: this should not be required.
+    std::this_thread::sleep_for(std::chrono::seconds(1));
+
+    auto prom = std::make_shared<std::promise<void>>();
+    auto ret = prom->get_future();
+
+    // Let's get the mode first
```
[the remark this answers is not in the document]
julianoes: Thanks, fixed in 3bba620.

plugins/camera/camera.h
```diff
@@ -0,0 +1,294 @@
...
+        UNKNOWN
+    };
+
+    /**
+     * @brief Returns a human-readable English string for Camera::Result.
+     *
+     * @param result The enum value for which a human readable string is required.
+     * @return Human readable string for the Camera::Result.
+     */
+    static const char *result_str(Result result);
```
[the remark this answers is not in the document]
julianoes: I agree. Will change it.

plugins/camera/camera_definition.h
```diff
@@ -0,0 +1,64 @@
...
+namespace dronecore {
+
+class CameraDefinition
+{
+public:
+    CameraDefinition();
+    ~CameraDefinition();
+
+
+    void load_file(const char *filename);
```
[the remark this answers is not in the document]
julianoes: That's nicer, I'll change it.

plugins/camera/camera_definition.h
```diff
@@ -0,0 +1,64 @@
...
+    void load_string(const std::string &content);
+
+    const char *get_vendor() const;
+    const char *get_model() const;
```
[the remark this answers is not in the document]
julianoes: Done but not with a reference but a copy.

plugins/camera/camera_impl.cpp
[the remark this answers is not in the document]
julianoes: So as long as we don't cache the file there is no value in having the definition version.

integration_tests/camera_mode.cpp
julianoes: I think I removed most of these.

plugins/camera/camera.h
julianoes: I have a gazillion things I could do every day and I'm always trying to figure out what's most relevant :confused:.

plugins/camera/camera_definition.h
julianoes: I tried to make a few of these methods `const` but could not because when the `map[key]` is used it can insert an entry in the map if not available and thus change the map. Presumably `map[key]` is the wrong way and `map.at(key)` should be used instead.

plugins/camera/camera_definition_test.cpp
julianoes: Changed.

cmake/compiler_flags.cmake
julianoes: Removed it :)

plugins/camera/camera_impl.h
julianoes: We actually store the camera mode in the camera_definition, so I'm not sure if we need it here as well.

integration_tests/camera_test_helpers.cpp
```diff
@@ -0,0 +1,126 @@
...
+    auto status = ret.wait_for(std::chrono::seconds(10));
+
+    EXPECT_EQ(status, std::future_status::ready);
+
+    if (status == std::future_status::ready) {
+        ret.get();
+    }
+
+    // FIXME: this should not be required.
+    std::this_thread::sleep_for(std::chrono::seconds(1));
```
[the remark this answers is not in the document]
julianoes: No, that's probably going to be there for a bit until the E90 behaves as it should.

plugins/camera/camera.h
julianoes: Thx!

plugins/camera/camera.h
julianoes: It can however be confusing because it would mean that the `async` versions go together as well as the sync versions but that's not the case and they can be mixed and matched.

plugins/camera/camera.h
julianoes: It got pulled out of the telemetry class but not out of the plugin. And we don't want dependencies between plugins, so for now this is duplicated. If we decide to pull such types out of the plugins into something common, then we can indeed re-use it.
It's always a trade-off between dependencies and duplication.

plugins/camera/camera.h
julianoes: Fixed.

plugins/camera/camera.h
julianoes: Fixed.

plugins/camera/camera.h
julianoes: I don't know the mavlink streaming stuff.

plugins/camera/camera.h
julianoes: Added.

plugins/camera/camera.h
julianoes: Will do.

plugins/camera/camera_definition.h
julianoes: No, this is internal.

core/call_every_handler.cpp
julianoes: The problem with your suggestion is that the `i++` still happens without check because during the loop while the callback is called, the remove can happen.

julianoes:
Ok, integration tests work against E90, E50, ET, and I've tried to address all comments so this should be ok to merge, finally.

plugins/camera/camera.h
hamishwillee: Just teasing (I thought obviously?). You do plenty of good.

plugins/camera/camera.h
hamishwillee: What I *know* is that RTSP video stream advertising via MAVLink does not work.  Everything else should work apparently (i.e. video capture and storage just got implemented) but I don't know how much testing has gone on.

plugins/camera/camera.h
```diff
@@ -0,0 +1,446 @@
...
+
+class CameraImpl;
+class System;
+
+/**
+ * @brief The Camera class can be used to manage cameras that implement the
+ * MAVLink Camera Protocol: https://mavlink.io/en/protocol/camera.html.
+ *
+ * Currently only a single camera is supported.
+ * When multiple cameras are supported the plugin will need to instantiated
```
hamishwillee: need to be

plugins/camera/camera.h
```diff
@@ -0,0 +1,446 @@
...
+    };
+
+    /**
+     * @brief Callback type to get status.
+     */
+    typedef std::function<void(Result, const Status &)> get_status_callback_t;
+
+
+    /**
+     * @brief Get camera status.
```
hamishwillee: > Get camera status (asynchronous).

(just for consistency with all other methdos)

plugins/camera/camera.h
```diff
@@ -0,0 +1,447 @@
...
+     */
+    struct VideoStreamSettings {
+        float frame_rate_hz = 0.f; /**< @brief Frames per second. */
+        uint16_t horizontal_resolution_pix = 0u; /**< @brief Horizontal resolution in pixels. */
+        uint16_t vertical_resolution_pix = 0u; /**< @brief Vertical resolution in pixels. */
+        uint32_t bit_rate_b_s = 0u; /**< @brief Bit rate in bits per second. */
+        uint16_t rotation_deg = 0u; /**< @brief Video image rotation clockwise (0-359 degrees). */
+        std::string uri {}; /**< @brief Video stream URI. */
+
+        void set_highest()
```
hamishwillee: Missing documentation

plugins/camera/camera.h
```diff
@@ -0,0 +1,447 @@
...
+        {
+            frame_rate_hz = FRAME_RATE_HIGHEST;
+            horizontal_resolution_pix = RESOLUTION_H_HIGHEST;
+            vertical_resolution_pix = RESOLUTION_V_HIGHEST;
+            // FIXME: Should bit_rate_b_s be part of set_highest() ?
+            bit_rate_b_s = BIT_RATE_AUTO;
+        }
+
+        constexpr static const float FRAME_RATE_HIGHEST = -1.0f;
```
hamishwillee: Missing documentation for these ones.

hamishwillee:
Looks good. I added some minor docs fixes. There is still a method and a few constants to document (marked up).

shakthi-prashanth-m:
MAVLink Video stream spec needs to be changed as raised here: https://github.com/mavlink/mavlink/issues/883. So, we've a dependency on https://github.com/mavlink/mavlink/pull/885.

julianoes:
@hamishwillee thanks for making the doc fixes, appreciated!

PR #346, opened 2018-04-04
269 update core proto

### LINKED ISSUE #269: Update backend to new core proto

JonasVautherin:
This PR implements `SubscribeDiscover` and `SubscribeTimeout` for the core module.

Fixes #269.


backend/test/core_service_impl_test.cpp
```diff
@@ -0,0 +1,278 @@
...
+    std::vector<uint64_t> uuids;
+    auto uuids_stream_future = subscribeDiscoverAsync(uuids);
+    subscription_future.wait();
+    event_callback(uuid0);
+    event_callback(uuid1);
+    event_callback(uuid2);
+    _core_service->stop();
+    uuids_stream_future.wait();
+
+    EXPECT_EQ(3, uuids.size());
```
julianoes: Nitpick, this would have to be an `ASSERT_EQ` because the `at()` below will fail otherwise.

backend/test/core_service_impl_test.cpp
```diff
@@ -0,0 +1,278 @@
...
+
+    _core_service->stop();
+    timeout_stream_future.wait();
+
+    EXPECT_EQ(0, count);
+}
+
+std::future<void> CoreServiceImplTest::subscribeTimeoutAsync(int *stream_count)
+{
+    return std::async(std::launch::async, [this, stream_count]() {
```
julianoes: Hm, I didn't know `std::launch::async.

backend/test/core_service_impl_test.cpp
JonasVautherin: Good point :)

JonasVautherin:
Well, 13 unit tests for 3 rpc functions, that doesn't sound excessive to me :-). For gRPC streams, it requires quite some code because I run an in-process server.

Also, only the public interface is unit tested, so that should be quite stable. If the tests ever need big changes, it will mean that the public interface changed, and in this case I find it reasonable to have to change the tests.

I know it may seem like a lot of tests, but I am quite confident now that the code is doing what I expect it to do (even though it doesn't mean it is completely bug-free). If somebody ever refactors the code without breaking the tests, I will be confident that the refactoring is correct. And finally, I believe the tests serve as documentation: one can learn how to interact with the code by reading the unit tests.

PR #352, opened 2018-04-06
backend: add packaging scripts for iOS

JonasVautherin:
Used from macOS to assemble the fat binary `backend.framework` and push it to S3.


backend/tools/push_backend_framework_to_s3.bash
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
julianoes: Would be good to add a comment what this requires, e.g. the aws cli tool, and how to set the key, etc.

JonasVautherin: Good point, I added a README explaining the purpose of the scripts and how to run them. I did not add documentation on how to actually use `awscli`, as the amazon documentation is very good for that (once you know which commands you want to use) :-).

PR #357, opened 2018-04-10
backend: improve backend api and initialization

JonasVautherin:
This PR adds a callback notifying when the gRPC server is started to the backend C api. This is necessary when the frontend starts the backend on the phone, because it needs to wait for the gRPC server to be started before it can initiate a connection.

### DIFF
--- a/backend/src/backend.cpp
+++ b/backend/src/backend.cpp
@@ -1,22 +1,10 @@
 #include "backend.h"
 
-#include <grpc/grpc.h>
-#include <grpc++/server.h>
-#include <grpc++/server_builder.h>
-#include <grpc++/server_context.h>
-#include <grpc++/security/server_credentials.h>
 #include <memory>
-#include <mutex>
 
-#include "action/action.h"
-#include "action/action_service_impl.h"
 #include "connection_initiator.h"
-#include "core/core_service_impl.h"
 #include "dronecore.h"
-#include "log.h"
-#include "mission/mission.h"
-#include "mission/mission_service_impl.h"
-#include "telemetry/telemetry_service_impl.h"
+#include "grpc_server.h"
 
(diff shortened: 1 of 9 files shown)


backend/src/backend_api.h
```diff
@@ -2,7 +2,7 @@
 extern "C" {
 #endif
 
-__attribute__((visibility("default"))) int runBackend(int mavlink_listen_port);
+__attribute__((visibility("default"))) void runBackend(int mavlink_listen_port, void (*onServerStarted)(void*), void *context);
```
julianoes: Missing `#pragma once` in this file.

JonasVautherin: Good catch! Thanks!

JonasVautherin: That's fixed.

backend/src/backend.cpp
```diff
@@ -27,53 +15,35 @@ class DroneCoreBackend::Impl
...
-        builder.AddListeningPort(server_address, grpc::InsecureServerCredentials());
-        LogInfo() << "Server set to listen on " << server_address;
+        _server->wait();
     }
 
-    dronecore::DroneCore _dc;
-    dronecore::backend::ConnectionInitiator<dronecore::DroneCore> _connection_initiator;
-    std::unique_ptr<grpc::Server> _server;
+private:
+    DroneCore _dc;
```
shakthi-prashanth-m: Not reference ? I don't know, I'm just asking.

JonasVautherin: No, because `backend.cpp` is actually the one owning the object. It has to be created somewhere, right? :-)

shakthi-prashanth-m: Yes :)

PR #359, opened 2018-04-10
Improvements to core

shakthi-prashanth-m:
- Avoids code duplication in `MAVLinkSystem::set_flight_mode()`, `MAVLinkSystem::set_msg_rate()` and their async versions.
- `MAVLinkCommands::send_command(...)`, `MAVLinkCommands::queue_command_async(...)` parameters are made `const` as they won't be modified inside these methods.


core/mavlink_system.h
```diff
@@ -172,6 +177,20 @@ class MAVLinkSystem
     static void system_thread(MAVLinkSystem *self);
     static void send_heartbeat(MAVLinkSystem &self);
 
+    // FIXME: Is the method name OK ? OR Should it be make_command_for_flight_mode(...) ?
```
JonasVautherin: This should be fixed before we merge this PR, right?

shakthi-prashanth-m: Yes, I wanted to get feedback about naming. :)

shakthi-prashanth-m:
@julianoes @JonasVautherin 
I shall submit next commit by removing FIXME if both you are OK with the naming and rest of the changes.

JonasVautherin:
I am fine with `make_command_flight_mode` :-)

shakthi-prashanth-m:
@julianoes if you're ok with changes (including naming), the you can approve so that I shall submit a new commit by removing FIXME comment.

core/mavlink_system.cpp
```diff
@@ -553,7 +553,10 @@ void MAVLinkSystem::get_param_float_async(const std::string &name,
                                             callback));
 }
 
-MAVLinkCommands::Result MAVLinkSystem::set_flight_mode(FlightMode system_mode)
+MAVLinkCommands::Result
+MAVLinkSystem::make_command_flight_mode(FlightMode flight_mode,
+                                        uint8_t component_id,
+                                        MAVLinkCommands::CommandLong &command)
```
julianoes: What is the reason to only make the command but not send it yet?

JonasVautherin: I like the separation of concerns: one function makes it, the other one sends it. For instance, `set_flight_mode(...)` reads "make the command, and if it results in a success, then send it". When reading `set_flight_mode(...)`, I might not be interested at all in the detail of `make_command_flight_mode`. And I can go there if I am.

shakthi-prashanth-m: @JonasVautherin You worded my thought. :) 
Yeah, `MAVLinkCommands` class provides `send_command(..)` so, I thought lets directly call that with the command prepared from `make_command_*` stuff.

shakthi-prashanth-m:
@JonasVautherin @julianoes Shall we merge this ?

PR #360, opened 2018-04-10
core: notify already-connected systems on register_on_discover

JonasVautherin:
This PR changes `register_on_discover` in such a way that it directly publishes the uuids of the systems that are already connected. 

Before that, one had query the list of already connected systems, but it probably never makes sense to register to the callback without wanting to know which ones are already connected.

Working on an integration test from Swift.

### DIFF
--- a/core/dronecore.h
+++ b/core/dronecore.h
@@ -152,6 +152,9 @@ public:
      *
      * This sets a callback that will be notified if a new system is discovered.
      *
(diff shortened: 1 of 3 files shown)


JonasVautherin:
I don't have a strong opinion on the naming here. However, it is out of the scope of this PR :-).

hamishwillee:
You also need to updated the in-source docs for `DroneCore::register_on_discover()` to reflect this behaviour change.

PR #361, opened 2018-04-12
Minor fixes to in-source docs following #310

hamishwillee:


core/system.h
```diff
@@ -17,41 +17,56 @@ class PluginImplBase;
...
      * @brief Checks whether the system has an autopilot.
-     * @return true if its an autopilot, false otherwise.
+     * @return `true` if it has an autopilot, `false` otherwise.
      */
     bool has_autopilot() const;
 
     /**
-     * @brief Checks whether the system is a standalone (Non-autopilot).
-     * @return true if stand alone, false otherwise.
+     * @brief Checks whether the system is a standalone (non-autopilot).
```
hamishwillee: How does this differ from `!has_autopilot()`?

shakthi-prashanth-m: Its same. Its a convenience API for standalone systems.
`if (system.is_standalone()) { ... }` is intuitive than calling `!has_autopilot()`.
What do you think ?

core/system.h
```diff
@@ -17,41 +17,56 @@ class PluginImplBase;
 class System
 {
 public:
+    /**
+     * @brief Constructor.
+     * @param system_id MAVLink system id.
+     * @param comp_id MAVLink component id.
```
shakthi-prashanth-m: To avoid using `MAVLinkSystem` terminology, I think we've to provide `SystemImpl` may be in future. Isn't it ?

hamishwillee: I don't know how that question relates to this section of code :-0

shakthi-prashanth-m: I meant that we'd to mention about MAVLink in DroneCore's public header file `system.h`. :)
Anyway app developer won't be able to use this constructor, so I was thinking is it necessary to mention about it.

core/system.h
hamishwillee: As a reader  `has_autopilot()` and `!has_autopilot()` are unambigous. Standalone is not - perhaps it means a system with just an autopilot but no camera. Unfortunately I don't think there is any better word for a system that doesn't have an autopilot associated, which is why it might be better just to remove this method.

If you do keep this, it needs a more full explanation. Maybe

Perhaps. 
```
@brief Checks whether this is a non-autopilot system (e.g. a standalone camera).

This is a convenience function. It returns `!has_autopilot()`.
```

core/system.h
hamishwillee: It appears in the docs because it is a public API, so best to say something ...

Perhaps in the class overview doc we should add:

> System objects are not created directly by application code, but are returned by the DroneCore class. 

In the  constructor we could have:
```
 /**
 * @brief Constructor.
 *
 * This constructor is not (and should not be) directly called by application code.
 * 
 * @param system_id System id.
 * @param comp_id Component id.
*/
```

Thoughts?

core/system.h
shakthi-prashanth-m: OK. I have some thoughts about this. I need your too. :-) Its little lengthy but gives clarity, I guess.

How about using `is_stationary()` for robots or systems that don't move. They may have movable arms and could be called "autonomous" without a "pilot" software for moving the robot itself.
So, I see 2 categories:
1. Drone which is a Mobile robot or autopilot (with or without any components) => `has_autopilot()` is true and Action, Mission plugins are applicable for them. If `has_camera()` is true then Camera plugin is also applicable.
`has_autopilot()` is `true`. `has_camera()`, `has_gimbal()` may or may not be true.

2. Stationary robots or Stationary camera (with or without any components) Action, Mission plugins won't be applicable, but camera may be if it has any.
`has_autopilot()` is `false`. `is_stationary()` is `true`. `has_camera()`, `has_gimbal()` may or may not be true.

Please share your thoughts.

hamishwillee: I think this is a solution without a problem (yet) :-). I think we should limit ourselves to deciding if is_standalone is useful only.

shakthi-prashanth-m: In that sense `is_standalone` is misleading right ? OR we could rename (if possible now) to `is_stationary` ? @julianoes @JonasVautherin What shall we do ? I initially thought standalone means stationary, so I named like that..

JonasVautherin: I tend to follow @hamishwillee when it comes to the wording :-).

TBH, when I first read it, I thought `is_standalone` meant `has only an autopilot and no other components`. So `!has_autopilot()` is clearer for me in that sense.

shakthi-prashanth-m: Yeah @JonasVautherin I too agree. But I was thinking aloud for clarity.
So, @hamishwillee you mean "standalone" is understood as **a device/system standing alone without any companion or components** ? 
But what is wrong with `is_stationary()` for Land Robots or Just plain MAVLink cameras ?
Its all about answering which plugins to be used when, isn't it ?

JonasVautherin: Maybe we should keep it simple and stay with `has_autopilot`, as suggested by Hamish. Later, if we see the need for that when we have real-world applications using the SDK for other things than drones, then it will still be time to add this. What do you think?

shakthi-prashanth-m: Sure. So shall I raise a PR for that ?

JonasVautherin: Sounds good to me. Thanks @shakthi-prashanth-m :-).

JonasVautherin:
@hamishwillee: style checker is failing. Would you mind fixing this? Then we can merge it :-).

Thanks!

hamishwillee:
Thanks guys. I've fixed style. Merge at will :-)

PR #363, opened 2018-04-12
Reorganize telemetry structs

JonasVautherin:
* Moved some structs out in their own file, so that they can be included standalone
* Reordered a few includes on the way

You'll see mainly that some structs (and corresponding `operator==`s moved in `telemetry_structs`, and the rest is about removing `Telemetry::` where it is not needed anymore.


hamishwillee:
Feels (still) very odd moving these outside of the associated class. I'll doc when merged.

JonasVautherin:
And I still don't understand why it feels odd :D. To me it's very clear:

* There is no single bad point in having them separate.
* `Position`, `Battery` and the likes are concepts that generalize to dronecore, not implementation details of the `Telemetry`. If the goal of keeping them in `Telemetry` is for namespacing, then that's wrong. Namespaces are made just for that.
* Having them inside `Telemetry` forces everybody who wants to use a `Position` to include the telemetry (we have examples where we don't need it).
* When using templates (which is exactly what I am doing for mocking here), you don't want to mix the different implementations together. Having the datastructs inside `telemetry.h` forces me to mix `telemetry.h` with my other template implementations, and I definitely don't want that.

shakthi-prashanth-m:
@JonasVautherin I had the exact doubt in my mind while working on FollowMe configuration class: whether FollowMe config class should be part of FollowMe class or outside.  I see that everybody outside are following the latter style.  I think we should also move FollowMe struct as well.  Thanks for clarifying

JonasVautherin:
@shakthi-prashanth-m: note that we don't necessarily need to move them for the sake of it. Here I move stuff because I am using `Telemetry` in a way that requires it (i.e. I'm mocking it with templates).

So if it works for `FollowMe` for now, there is not necessarily a need to change it. We can always do it later, when we actually feel we need it :).

shakthi-prashanth-m:
Got it. But one thing. The class interface should be easily readable. If we have define nested structs inside it, it disturbs the reading flow. Also, we can definitely have a class outside.
For eg:
```cpp
class FollowMe {
    ...
    FollowMeSettings get_settings() const;
    ...
};

class FollowMeSettings {
    ...
};
```
So FollowMe composes of FollowMeSettings. It is not necessary to have `FollowMeSettings` inside `FollowMe` class.

JonasVautherin:
Should I merge this PR then? :-)

plugins/telemetry/math_conversions.cpp
```diff
@@ -21,7 +22,7 @@ Telemetry::EulerAngle to_euler_angle_from_quaternion(Telemetry::Quaternion quate
     return euler_angle;
 }
 
-Telemetry::Quaternion to_quaternion_from_euler_angle(Telemetry::EulerAngle euler_angle)
+Quaternion to_quaternion_from_euler_angle(EulerAngle euler_angle)
```
julianoes: I miss the namespacing of these structs now and I fear they will then be included in other plugins as well, and not just inside telemetry.

plugins/telemetry/telemetry_impl.cpp
```diff
@@ -1,37 +1,39 @@
...
-    _position(Telemetry::Position {double(NAN), double(NAN), NAN, NAN}),
+    _position(Position {double(NAN), double(NAN), NAN, NAN}),
     _home_position_mutex(),
-    _home_position(Telemetry::Position {double(NAN), double(NAN), NAN, NAN}),
+    _home_position(Position {double(NAN), double(NAN), NAN, NAN}),
     _in_air(false),
     _armed(false),
     _attitude_quaternion_mutex(),
-    _attitude_quaternion(Telemetry::Quaternion {NAN, NAN, NAN, NAN}),
+    _attitude_quaternion(Quaternion {NAN, NAN, NAN, NAN}),
```
julianoes: While you're at it I suggest to do all these default values in the header file instead of the constructor initializer list.

plugins/telemetry/math_conversions.cpp
JonasVautherin: One question could also be: should they go in `core`? Are they not datastructs that we'd like to share publicly at the library level? Or is it really a concept of the telemetry module?

julianoes: Yer, we could have them shared as well, that's true.

julianoes:
> My latest opinion is that it makes sense to keep them in `telemetry`, but in a `dronecore::telemetry` namespace instead of nested into the `Telemetry` class.

That's also a good way. I was using `Telemetry::` sort of like a namespace, or rather I added the types as a public interface for the class and I thought that was nicest but if they need to go in a separate file, then the `dronecore::telemetry` namespace makes sense.

Issue #350, opened 2018-04-05
Implement mission features in backend


thomasgubler:

JonasVautherin:
`SubscribeMissionProgress` is in progress, but has a bug.

PR #369, opened 2018-04-23
350 mission backend

JonasVautherin:
Add `UploadMission` and `StartMission` to the mission backend.

Connects to #350.


backend/test/mission_service_impl_test.cpp
```diff
@@ -1,55 +1,358 @@
...
+{
+    *mission = arg0;
+    *callback = arg1;
+    callback_saved_promise->set_value();
+}
+
+TEST_F(MissionServiceImplUploadTest, doesNotFailWhenArgsAreNull)
 {
-    // TODO
+    auto upload_handle = uploadMissionAndSaveParams(nullptr, nullptr);
```
julianoes: :+1:

backend/test/mission_service_impl_test.cpp
```diff
@@ -1,55 +1,358 @@
...
+    _result_callback(dc::Mission::Result::UNKNOWN);
```
julianoes: Why would the result be `UNKNOWN`? I think the result should only be unknown if something in the enum translation went wrong. If it is an actual error, it should be `ERROR` or some specific error code, right?

Or is it that the error is unknown because with `nullptr` the callback has not actually been called?

backend/test/mission_service_impl_test.cpp
```diff
@@ -1,55 +1,358 @@
...
+}
+
+TEST_P(MissionServiceImplUploadTest, uploadResultIsTranslatedCorrectly)
+{
+    auto response = std::make_shared<UploadMissionResponse>();
+    std::vector<std::shared_ptr<dc::MissionItem>> mission_items;
+    auto request = generateUploadRequest(mission_items);
+    auto upload_handle = uploadMissionAndSaveParams(request, response);
+
+    _result_callback(GetParam().second);
```
julianoes: Help me please. What's `GetParam()` here?

backend/test/mission_service_impl_test.cpp
```diff
@@ -1,55 +1,358 @@
...
+}
+
+std::vector<std::shared_ptr<dc::MissionItem>> MissionServiceImplUploadTest::generateListOfOneItem()
+{
+    std::vector<std::shared_ptr<dc::MissionItem>> mission_items;
+
+    auto mission_item = std::make_shared<dc::MissionItem>();
+    mission_item->set_position(41.848695, 75.132751);
+    mission_item->set_relative_altitude(50.4);
+    mission_item->set_speed(8.3);
```
julianoes: Some of the fields are actually floats and so it should be `8.3f`, however, I wonder if we shouldn't just make everything `double` in DroneCore, especially in the public interface since DroneCore is not run on embedded platforms generally and everything is becoming 64 bits these days.

plugins/mission/mission_item.cpp
```diff
@@ -117,4 +131,59 @@ std::string MissionItem::to_str(MissionItem::CameraAction camera_action)
     }
 }
 
+bool operator==(const MissionItem &lhs, const MissionItem &rhs)
+{
+    return (lhs.get_latitude_deg() == rhs.get_latitude_deg()
+            || isnan(lhs.get_latitude_deg()) && isnan(rhs.get_latitude_deg()))
```
julianoes: I think it would be better to make it explicit what the operator order is.

plugins/mission/mission_item.cpp
```diff
@@ -117,4 +131,59 @@ std::string MissionItem::to_str(MissionItem::CameraAction camera_action)
...
+           << "\tlon: " << mission_item.get_longitude_deg() << std::endl
+           << "\trel_alt: " << mission_item.get_relative_altitude_m() << std::endl
+           << "\tis_fly_through: " << (mission_item.get_fly_through() ? "true" : "false") << std::endl
+           << "\tspeed: " << mission_item.get_speed_m_s() << std::endl
+           << "\tgimbal_pitch: " << mission_item.get_gimbal_pitch_deg() << std::endl
+           << "\tgimbal_yaw: " << mission_item.get_gimbal_yaw_deg() << std::endl
+           << "\tloiter_time: " << mission_item.get_loiter_time_s() << std::endl
+           << "\tcamera_action: " << mission_item.get_camera_action() << std::endl
+           << "\tcamera_photo_interval: " << mission_item.get_camera_photo_interval_s()
+           << std::endl << "]" << std::endl;
```
julianoes: Nice!

plugins/mission/mission_item.cpp
```diff
@@ -117,4 +131,59 @@ std::string MissionItem::to_str(MissionItem::CameraAction camera_action)
...
+        case MissionItem::CameraAction::START_PHOTO_INTERVAL:
+            return str << "CameraAction::START_PHOTO_INTERVAL";
+        case MissionItem::CameraAction::STOP_PHOTO_INTERVAL:
+            return str << "CameraAction::STOP_PHOTO_INTERVAL";
+        case MissionItem::CameraAction::START_VIDEO:
+            return str << "CameraAction::START_VIDEO";
+        case MissionItem::CameraAction::STOP_VIDEO:
+            return str << "CameraAction::STOP_VIDEO";
+        case MissionItem::CameraAction::NONE:
+            return str << "CameraAction::NONE";
```
julianoes: We should probably do the same kind of thing instead of the `result_str()` methods.

plugins/mission/mocks/mission_mock.h
```diff
@@ -8,13 +8,14 @@
 namespace dronecore {
 namespace testing {
 
-typedef std::function<void(Mission::Result)> result_callback_t;
+typedef std::function<void(Mission::Result)> mission_result_callback_t;
```
julianoes: Can't you get this typedef from somwhere else? Or is it specific to testing?

plugins/mission/mission_item.cpp
JonasVautherin: Oh, right. macOS is warning about that, and apparently I did not fix it on the right computer. I'll change it!

julianoes: Haha, it's a warning. The operator precedence is clear in theory but it's nice for the reader to make it explicit in my opinion.

backend/test/mission_service_impl_test.cpp
JonasVautherin: It is just that I need to synchronize the call in the service (see [here](https://github.com/dronecore/DroneCore/blob/350-mission-backend/backend/src/plugins/mission/mission_service_impl.h#L30) and [here](https://github.com/dronecore/DroneCore/blob/350-mission-backend/backend/src/plugins/mission/mission_service_impl.h#L102)). So the upload call is blocked until the callback is actually called (this assumes that the callback is always called, eventually).

The value of it in this unit test has not importance. It just needs to trigger. Maybe I should make it a variable called `ARBITRARY_RESULT`, to make this clear!

plugins/mission/mocks/mission_mock.h
JonasVautherin: You're right. I did not want to take `Mission::result_callback_t`, but for wrong reasons :-).

PR #372, opened 2018-04-25
Serial connection support for macOS

julianoes:
It recently came up on Slack that it would be good to have support for the serial connection on macOS.

### DIFF
--- a/cmake/compiler_flags.cmake
+++ b/cmake/compiler_flags.cmake
@@ -38,6 +38,10 @@ if(APPLE)
     add_definitions("-DAPPLE")
 endif()
 
+if(LINUX)
+    add_definitions("-DLINUX")
+endif()
+
 # Add DEBUG define for Debug target because that is not done automatically.
 set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -DDEBUG")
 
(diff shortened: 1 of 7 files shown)


hamishwillee:
@julianoes This concerns me - I was not aware that any of our APIs were "platform specific". If we don't support an API on a particular platform then that should be part of the documentation.

So does this work on Windows? Are there any other APIs that we know only work on Linux, Windows or Mac?

JonasVautherin:
@hamishwillee: For what I understand from the code, the serial connection is now supported on Linux and macOS, and not on Windows.

hamishwillee:
> @hamishwillee: For what I understand from the code, the serial connection is now supported on Linux and macOS, and not on Windows.

Thanks. Then we should probably add a warning to `DroneCore::add_any_connection()` and `add_serial_connection()` docs. Something like

> @warning This method is not supported on Windows (it is supported on Linux and macOS).

I am making the assumption that everything else works everywhere.

JonasVautherin:
> I am making the assumption that everything else works everywhere.

I wouldn't know :grinning:.

julianoes:
@hamishwillee you're right.
Let's add that comment about Windows.

UDP and TCP connections are supported on all 3 operating systems.

PR #375, opened 2018-04-29
Serial connection support for Windows

julianoes:
This adds support for serial connections on Windows.

Todos:
- [x] make it work on Windows.
- [x] remove comments that this is not supported.
- [x] fix `add_any_connection` to work on Windows.

### DIFF
--- a/core/dronecore.h
+++ b/core/dronecore.h
@@ -59,9 +59,11 @@ public:
     ConnectionResult add_any_connection(const std::string &connection_url);
 
     /**
-     * @brief Adds a UDP connection to the specified port number on 0.0.0.0 on any local interface.
+     * @brief Adds a UDP connection to the specified port number.
      *
-     * @param local_port The local UDP port to listen to (defaults to 14540, the same as mavros).
+     * Any incoming connections are accepted (0.0.0.0).
+     *
+     * @param local_port The local UDP port to listen to (defaults to 14540, the same as MAVROS).
      * @return The result of adding the connection.
      */
     ConnectionResult add_udp_connection(int local_port = DEFAULT_UDP_PORT);
@@ -69,11 +71,11 @@ public:
     /**
      * @brief Adds a UDP connection to the specified port number and local interface.
      *
-     * To accept only connections local connections of the machine, use 127.0.0.1, for any
-     * incoming connections, use 0.0.0.0.
+     * To accept only local connections of the machine, use 127.0.0.1.
+     * For any incoming connections, use 0.0.0.0.
      *
      * @param local_bind_ip The local UDP ip to listen to.
-     * @param local_port The local UDP port to listen to (defaults to 14540, the same as mavros).
+     * @param local_port The local UDP port to listen to (defaults to 14540, the same as MAVROS).
      * @return The result of adding the connection.
      */
     ConnectionResult add_udp_connection(const std::string &local_ip,
--- a/core/dronecore.h
+++ b/core/dronecore.h
@@ -23,11 +23,10 @@ class System;
 class DroneCore
 {
 public:
-    static constexpr auto DEFAULT_UDP_CONNECTION_URL = "udp://:14540";
+    static constexpr auto DEFAULT_UDP_BIND_IP = "0.0.0.0";
     static constexpr int DEFAULT_UDP_PORT = 14540;
     static constexpr auto DEFAULT_TCP_REMOTE_IP = "127.0.0.1";
     static constexpr int DEFAULT_TCP_REMOTE_PORT = 5760;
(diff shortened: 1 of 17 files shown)


core/dronecore.h
```diff
@@ -48,26 +47,45 @@ class DroneCore
...
+     * - Default Bind host IP is any local interface (0.0.0.0)
      *
      * @param connection_url connection URL string.
      * @return The result of adding the connection.
      */
-    ConnectionResult add_any_connection(const std::string &connection_url = DEFAULT_UDP_CONNECTION_URL);
+    ConnectionResult add_any_connection(const std::string &connection_url);
+
+    /**
+     * @brief Adds a UDP connection to the specified port number on 0.0.0.0 on any local interface.
```
hamishwillee: Does "any local interface" mean any network? (ie I am wondering what this means as opposed to if you did not have this text)

core/dronecore.h
```diff
@@ -48,26 +47,45 @@ class DroneCore
...
+     * @return The result of adding the connection.
+     */
+    ConnectionResult add_udp_connection(int local_port = DEFAULT_UDP_PORT);
 
     /**
-     * @brief Adds a UDP connection to the specified port number.
+     * @brief Adds a UDP connection to the specified port number and local interface.
      *
-     * @param local_port_number The local UDP port to listen to (defaults to 14540, the same as mavros).
+     * To accept only connections local connections of the machine, use 127.0.0.1, for any
```
hamishwillee: "127.0.0.1, for any" - make that comma into a full stop and split into 2 sentences.

core/dronecore.h
```diff
@@ -48,26 +47,45 @@ class DroneCore
...
+     * incoming connections, use 0.0.0.0.
+     *
+     * @param local_bind_ip The local UDP ip to listen to.
+     * @param local_port The local UDP port to listen to (defaults to 14540, the same as mavros).
```
hamishwillee: mavros is the package name, MAVROS is the product name. I would say this is an instance of MAVROS.

core/log.h
```diff
@@ -50,6 +43,18 @@ class NullStream
 
 namespace dronecore {
 
+enum class Color {
```
hamishwillee: This is all much nicer :-)

core/dronecore.h
JonasVautherin: "0.0.0.0" does means "listen on any interface". So "on any local interface" here does not add information, but explicits what "0.0.0.0" means.

Here it is [from Wikipedia](https://en.wikipedia.org/wiki/0.0.0.0):

> In the context of servers, 0.0.0.0 can mean "all IPv4 addresses on the local machine". If a host has two IP addresses, 192.168.1.1 and 10.1.2.1, and a server running on the host is configured to listen on 0.0.0.0, it will be reachable at both of those IP addresses.

core/dronecore.h
```diff
@@ -48,26 +47,47 @@ class DroneCore
...
-     * - TCP - tcp://[Server_host][:Server_port]
-     * - Serial - serial://[Dev_Node][:Baudrate]
+     * - TCP - tcp://[Remote_host][:Remote_port]
+     * - Serial - serial://Dev_Node[:Baudrate]
      *
-     * Default URL : udp://:14540.
-     * - Default Bind host IP is localhost(127.0.0.1)
-     *
-     * @warning Serial connections are not supported on Windows (they are supported on Linux and macOS).
+     * Default URL : udp://0.0.0.0:14540.
```
hamishwillee: @julianoes The new interface **must** take a `connection_url` string so presumably this is no longer a default value. Why have we chosen to not have a default anymore?

julianoes: No, the host is not a must, see:
```
- TCP - tcp://[Remote_host][:Remote_port]
- Serial - serial://[Dev_Node][:Baudrate]
- Serial - serial://Dev_Node[:Baudrate]
```

core/dronecore.h
```diff
@@ -23,11 +23,10 @@ class System;
 class DroneCore
 {
 public:
-    static constexpr auto DEFAULT_UDP_CONNECTION_URL = "udp://:14540";
+    static constexpr auto DEFAULT_UDP_BIND_IP = "0.0.0.0";
```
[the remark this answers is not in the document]
julianoes: I see, I'll add it next time.

core/dronecore.h
julianoes: Hm, you're right. There is no default. You need to specify it, and that shouild usually be done using a command line argument I suggest.

hamishwillee: @julianoes Before I do that, this is now inconsistent - the other methods (e.g. for adding UDP) do have defaults. Can you either add the default back OR remote all the other defaults?

PR #386, opened 2018-05-04
build_system: build tinyxml2 statically

JonasVautherin:
Build tinyxml2 statically (to fix the build for iOS).

It may look a bit hacky, but I did not want to pollute the whole scope with options set only for tinyxml2. One alternative would be to fork tinyxml2 and modify its `CMakeLists.txt`, and another one would be to not build tinyxml2 as a submodule (I would like to do that at some point, some kind of "super-build" scheme).

At least it is contained into `cmake/tinyxml2.cmake` for now.

@julianoes let me know if you think we should link it dynamically for some platforms.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,8 +12,9 @@ project(dronecore)
 option(BUILD_TESTS "Build tests" ON)
 
 include(cmake/compiler_flags.cmake)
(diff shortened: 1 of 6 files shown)


JonasVautherin:
Haha, now the tests are all inverted. Just need to keep the best of both solutions :sweat_smile:.

PR #395, opened 2018-05-16
fly_qgc_mission: link to json11

### LINKED ISSUE #381: fly_qgc_mission error loading shared libraries: libjson11.so
```
/DroneCore/example/fly_qgc_mission/build$ ./fly_qgc_mission 
./fly_qgc_mission: error while loading shared libraries: libjson11.so: cannot open shared object file: No such file or directory
``

julianoes:
We need to link to libjson11 as well because libdronecore_mission.so doesn't include it.

@JonasVautherin any better idea? Or should we include json11 inside `libdronecore.a` like tinyxml2 to prevent this?

Fixes #381.

### DIFF
--- a/example/fly_mission/CMakeLists.txt
+++ b/example/fly_mission/CMakeLists.txt
@@ -19,5 +19,4 @@ target_link_libraries(fly_mission
     dronecore_action
     dronecore_mission
     dronecore_telemetry
-    json11
 )
--- a/example/fly_qgc_mission/CMakeLists.txt
+++ b/example/fly_qgc_mission/CMakeLists.txt
@@ -18,6 +18,5 @@ target_link_libraries(fly_qgc_mission
     dronecore
     dronecore_action
     dronecore_mission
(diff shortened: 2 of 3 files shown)


hamishwillee:
Is json present by default on machines or will library need to be installed (so it can be linked). Or to put it another way, is there an extra step that needs to be added to run this example?

JonasVautherin:
Isn't `json11` [supposed to be statically linked](https://github.com/dronecore/json11/blob/master/CMakeLists.txt#L25) to `libdronecore_mission`? At least it is for iOS. For which platform is this fix?

julianoes:
@JonasVautherin this fix is for macOS.

JonasVautherin:
Are you sure you updated your submodules? It should be in `github.com/dronecore/json11`. I believe it works for me, but I need to try it again.

julianoes:
> Are you sure you updated your submodules?

Let me check.

PR #398, opened 2018-05-17
Various fixes found doing camera tests

julianoes:
This includes a couple of fixes that I found while running the camera tests against the E90.

### DIFF
--- a/core/mavlink_commands.cpp
+++ b/core/mavlink_commands.cpp
@@ -137,7 +137,7 @@ void MAVLinkCommands::receive_command_ack(mavlink_message_t message)
     mavlink_command_ack_t command_ack;
     mavlink_msg_command_ack_decode(&message, &command_ack);
 
-    LogDebug() << "We got an ack: " << command_ack.command;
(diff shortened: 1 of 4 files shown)


core/mavlink_commands.cpp
```diff
@@ -137,7 +137,7 @@ void MAVLinkCommands::receive_command_ack(mavlink_message_t message)
     mavlink_command_ack_t command_ack;
     mavlink_msg_command_ack_decode(&message, &command_ack);
 
-    LogDebug() << "We got an ack: " << command_ack.command;
+    // LogDebug() << "We got an ack: " << command_ack.command;
```
hamishwillee: Just FMI, shouldn't useful debug logs be left active, and non useful ones get deleted from the codeline?

JonasVautherin: :grin: that's subject to debate :grin:.

I tend to be against commented code, but for a log like this, I personally don't have a very strong opinion.

julianoes: I have re-added this debug line many many times, so I thought I might as well leave it in case I need it yet again.

PR #403, opened 2018-05-21
Fix VTOL example

### LINKED ISSUE #401: transition_vtol_fixed_wing example never ready to arm
This example is never ready to arm. Running it sometimes gets a segmentation fault. 

```
$ ./transition_vtol_fixed_wing udp://:14540
Waiting to discover system...
[03:18:46|Info ] New device on: 127.0.0.1:14557 (udp_connection.cpp:208)
[03:18:46|Debug] New: System ID: 1 Comp ID: 1 (dronecore_impl.cpp:286)
[03:18:46|Debug] Component Autopilot added. (mavlink_system.cpp:330)
[03:18:46|Debug] MAVLink: info: data link #1 lost (mavlink_system.cpp:267)
[03:18:46|Debug] MAVLink: info: data link #1 regained (mavlink_system.cpp:267)
[03:18:46|Debug] Found 1 component(s). (mavlink_system.cpp:462)
[03:18:46|Debug] Discovered 4294967298 (mavlink_system.cpp:464)
Discovered system with UUID: 4294967298
[03:18:46|Debug] We got an ack: 520 (mavlink_commands.cpp:140)
[03:18:48|Debug] We got an ack: 511 (mavlink_commands.cpp:140)
Vehicle not ready to arm
```

I changed the arm test code such that it retries the health check:

```
// Check if vehicle is ready to arm
while (telemetry->health_all_ok() != true) {
std::cout << ERROR_CONSOLE_TEXT << "Vehicle not ready to arm" << NORMAL_CONSOLE_TEXT << std::endl;
sleep_for(std::chrono::seconds(1));
}
```

In this case I got it ready to arm after 1 second. But then it fails with landed state unknown.

```
Vehicle not ready to arm
Altitude: 0.044 m
Arming...
Arming failed:Command denied, landed state is unknown
```

BUT, if I change nothing and just run the takeoff_and_land example it will takeoff and land. No problems with arming at all.

julianoes:
Fixes #401.

### DIFF
--- a/core/mavlink_system.cpp
+++ b/core/mavlink_system.cpp
@@ -91,6 +91,7 @@ void MAVLinkSystem::unregister_all_mavlink_message_handlers(const void *cookie)
 
         if (it->cookie == cookie) {
             it = _mavlink_handler_table.erase(it);
+            _iterator_invalidated = true;
         } else {
             ++it;
         }
@@ -126,7 +127,7 @@ void MAVLinkSystem::process_mavlink_message(const mavlink_message_t &message)
 #if MESSAGE_DEBUGGING==1
     bool forwarded = false;
 #endif
-    for (auto it = _mavlink_handler_table.begin(); it != _mavlink_handler_table.end(); ++it) {
+    for (auto it = _mavlink_handler_table.begin(); it != _mavlink_handler_table.end(); /* ++it */) {
         if (it->msg_id == message.msgid) {
 #if MESSAGE_DEBUGGING==1
             LogDebug() << "Forwarding msg " << int(message.msgid) << " to " << size_t(it->cookie);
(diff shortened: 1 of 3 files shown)


core/mavlink_system.cpp
```diff
@@ -136,6 +137,16 @@ void MAVLinkSystem::process_mavlink_message(const mavlink_message_t &message)
             it->callback(message);
             _mavlink_handler_table_mutex.lock();
         }
+
+        if (_iterator_invalidated) {
+            // Someone messed with the map while we were doing the callback.
```
JonasVautherin: How can that happen? Don't we have the `_mavlink_handler_table_mutex` lock for preventing that?

julianoes: The problem is that we have to release the lock during the callback. If we didn't release the lock during the callback, we could lock ourselves out if the code in the callback tries to register or unregister a message.

An alternative would be to use a recursive_lock, so we wouldn't lock ourselves out, however, we would still change the map and invalidate the iterator.

julianoes: The proper solution is to have events in queues instead of callbacks, I think but we're far away from that :disappointed:.

JonasVautherin: I see. Well, that's fine for now :-).

hamishwillee:
@julianoes  Is there a test of some kind that we don't have that we should have had to catch this?

julianoes:
> @julianoes Is there a test of some kind that we don't have that we should have had to catch this?

We should have integration tests in CI to catch this in the future. For now I try to eliminate one segfault or deadlock after the other when I run into it. This way I've been able to fix a couple in the past weeks.

hamishwillee:
Thanks @julianoes 
It would be good to add those tests. For every segfault you fix I think I could inject 3 :-)

PR #404, opened 2018-05-21
Add fake system

julianoes:
This is some refactoring and a suggestion how to implement a fake system which responds to commands.

This is more a suggestion, something to start the discussion. Maybe the borders are drawn incorrectly, and we should be testing at the interface of the (mavlink) system instead of going all the way through UDP connections. At that point it's more of an integration test than a unit test.

@JonasVautherin what do you think?

### DIFF
--- a/cmake/unit_tests.cmake
+++ b/cmake/unit_tests.cmake
@@ -14,6 +14,7 @@ set_target_properties(unit_tests_runner
 
 target_link_libraries(unit_tests_runner
     dronecore
+    dronecore_action
     dronecore_mission
     dronecore_camera
     gtest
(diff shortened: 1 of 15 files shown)


hamishwillee:
Why not create an `add_fake_connection()` and integrate at that level?

julianoes:
> @hamishwillee: Why not create an add_fake_connection() and integrate at that level?

Yer that's what I thought as well initially. That would be another way.

JonasVautherin:
But then shouldn't it go in its own folder, instead of in `core/`?

julianoes:
@JonasVautherin I tried to continue here a bit but I'm not sure it's the right testing level and architecture.

julianoes:
@JonasVautherin suggests that we make actual "fake" plugins, e.g. a `fake_mission_acceptor` and then have two SDK instances talking to each other to test each other.

hamishwillee:
That's a clever idea. So one side might be an API implementing autopilot side mission handling. NIce.

PR #405, opened 2018-05-22
Mission pause

Sushma7785:
Connects to #350.

### DIFF
--- a/backend/test/mission_service_impl_test.cpp
+++ b/backend/test/mission_service_impl_test.cpp
@@ -29,6 +29,9 @@ using RPCCameraAction = dc::rpc::mission::MissionItem::CameraAction;
 using StartMissionRequest = dc::rpc::mission::StartMissionRequest;
 using StartMissionResponse = dc::rpc::mission::StartMissionResponse;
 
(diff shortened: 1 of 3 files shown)


backend/test/mission_service_impl_test.cpp
```diff
@@ -386,6 +389,53 @@ TEST_F(MissionServiceImplIsFinishedTest, isMissionFinishedDoesNotCrashWithNullRe
...
+{
+protected:
+    std::future<void> pauseMissionAndSaveParams(std::shared_ptr<PauseMissionResponse> response);
+};
+
+INSTANTIATE_TEST_CASE_P(MissionResultCorrespondences,
+                        MissionServiceImplPauseTest,
+                        ::testing::ValuesIn(generateInputPairs()));
+
+ACTION_P2(SaveResult, callback, callback_saved_promise)
```
JonasVautherin: This `ACTION_P2` is already defined above. I believe you don't need to redefine it here. Can you test without it? :)

Sushma7785: okay sure.

PR #406, opened 2018-05-23
System refactor

julianoes:
@JonasVautherin this is the working progress of refactoring the system class towards supporting out-of-process plugins.


julianoes:
@JonasVautherin I think we can get this in for now.

PR #410, opened 2018-05-24
Switch to clang-format

### LINKED ISSUE #358: Its time to switch Coding style to Clang format
Clang format is most popular C++ coding style: http://clang.llvm.org/docs/ClangFormat.html
Benefits:
- We can also integrate this in git using [git-clang-format](https://github.com/llvm-mirror/clang/blob/master/tools/clang-format/git-clang-format) or add git pre commit hook.
- Clang has a [static analyzer](https://clang-analyzer.llvm.org/).
- It will rearrange header file inclusion order alphabetically (@JonasVautherin was this earlier).
- Supports various styles.
  - LLVM A style complying with the LLVM coding standards
  - Google A style complying with Google’s C++ style guide
  - Chromium A style complying with Chromium’s style guide
  - Mozilla A style complying with Mozilla’s style guide
  - WebKit A style complying with WebKit’s style guide

Reference:
http://clang.llvm.org/docs/ClangFormat.html
http://clang.llvm.org/docs/ClangFormatStyleOptions.html
https://clang-analyzer.llvm.org/
https://clangformat.com/ - Interactive guide and builder.

@julianoes @JonasVautherin @hamishwillee .

julianoes:
While the clang-format style tries to reflect the previous astyle style
as close as possible, there is, unfortuntately, still a considerable diff.

Fixes #358.


hamishwillee:
@julianoes We mention astyle here: https://docs.dronecore.io/en/contributing/code_style.html

Obviously we're now using clang. 
- What are the style settings defined by (previously it said *astyle settings defined by astylerc*).
- Is it still called by `make fix_style`?
- Are the explanations accurate? (other than typo "Large enough TO allow ...)

I will have to remove astyle from the dependencies - e.g. : https://docs.dronecore.io/en/contributing/build.html#linux
What do have to add for Clang?

core/mavlink_parameters.cpp
```diff
@@ -111,7 +111,7 @@ MAVLinkParameters::do_work()
         _state = State::SET_PARAM_BUSY;
 
         char param_id[PARAM_ID_LEN] = {};
-        STRNCPY(param_id, set_param_work->param_name.c_str(), sizeof(param_id));
+        STRNCPY(param_id, set_param_work->param_name.c_str(), sizeof(param_id)-1);
```
JonasVautherin: I don't get that :-/

JonasVautherin: Oh, forget it, that's in the commit description :sweat_smile:

julianoes: GCC 8 doesn't like the fact that we could potentially copy 17 chars without null termination into `param_id`. At that point `param_id` would not be null terminated. That's how I understand it.

julianoes:
> @hamishwillee 
> @julianoes We mention astyle here: https://docs.dronecore.io/en/contributing/code_style.html

Yes, we are switching.

> Obviously we're now using clang.
> 
> - What are the style settings defined by (previously it said astyle settings defined by astylerc).

The settings are defined in .clang-format (added in https://github.com/dronecore/DroneCore/pull/410/commits/7269ce549525a39bfd30f82d07951fb511a45e7e)

> - Is it still called by make fix_style?

Yes, the usage is the same.

> Are the explanations accurate? (other than typo "Large enough TO allow ...)

Yes, that's still the same.

> I will have to remove astyle from the dependencies - e.g. : https://docs.dronecore.io/en/contributing/build.html#linux
> What do have to add for Clang?

Correct. Developers need to have `clang-format` version >= 6.

For Ubuntu 16.04: (https://github.com/dronecore/DroneCore/blob/switch-to-clang-format/docker/Dockerfile-Ubuntu-16.04#L37-L42)
```
wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
sudo apt-add-repository "deb http://apt.llvm.org/xenial/ llvm-toolchain-xenial-6.0 main"
sudo apt-get update
sudo apt-get install -y clang-format-6.0
sudo update-alternatives --install /usr/bin/clang-format clang-format /usr/bin/clang-format-6.0 1000
```

Ubuntu 18.04: (https://github.com/dronecore/DroneCore/blob/switch-to-clang-format/docker/Dockerfile-Ubuntu-18.04#L21)
```
sudo apt-get install clang-format
sudo update-alternatives --install /usr/bin/clang-format clang-format /usr/bin/clang-format-6.0 1000
```

Fedora 28: (https://github.com/dronecore/DroneCore/blob/switch-to-clang-format/docker/Dockerfile-Fedora-28#L14)
```
sudo dnf install clang
```

core/mavlink_parameters.cpp
shakthi-prashanth-m: I was just thinking, why don't we a hire a local `std::string` for doing this. Applicable elsewhere in such scenarios as well. Makes sense ?

plugins/telemetry/telemetry_impl.cpp
```diff
@@ -8,47 +8,28 @@
...
-    _gps_info_subscription(nullptr),
-    _battery_subscription(nullptr),
-    _flight_mode_subscription(nullptr),
-    _health_subscription(nullptr),
-    _health_all_ok_subscription(nullptr),
-    _rc_status_subscription(nullptr),
-    _ground_speed_ned_rate_hz(0.0),
-    _position_rate_hz(-1.0)
+TelemetryImpl::TelemetryImpl(System &system)
+    : PluginImplBase(system), _position_mutex(),
```
shakthi-prashanth-m: At some places like these (constructor member initialization), I feel old style is better readable. But, thats unavoidable I think.

julianoes: Could call, I was going to look at the changes and figure out how to prevent some of them.

core/mavlink_parameters.cpp
julianoes: @shakthi-prashanth-m I don't think I understand what you mean.

julianoes:
@shakthi-prashanth-m @JonasVautherin I changed some of the clang-format options to match the previous style closer. Have another look please if you see anything too ugly.

PR #413, opened 2018-05-25
core: fix macOS warning

julianoes:
Clang and MSVC are not agreeing on this issue, so we'll ignore
the clang warning for now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,7 +69,7 @@ matrix:
       env:
       - MATRIX_EVAL="BUILD_TARGET=docker_build_fedora_28"
     - os: osx
-      osx_image: xcode9.2
+      osx_image: xcode9.3
       env:
         - BUILD_TARGET=osx_build
 
--- a/cmake/compiler_flags.cmake
+++ b/cmake/compiler_flags.cmake
@@ -26,9 +26,9 @@ else()
 
         set(warnings "${warnings} -Wlogical-op")
     elseif(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
-        set(warnings "${warnings} -Wno-missing-braces")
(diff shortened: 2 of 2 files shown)


core/thread_pool_test.cpp
```diff
@@ -57,6 +57,13 @@ TEST(ThreadPool, SimpleTaskWithArgs)
     EXPECT_EQ(second, sum);
 }
 
+// We need to capture first because otherwise MSVC complains about
+// an implicit capture. However, this makes clang sad, so we have to
+// ignore its warning.
+#if defined(__clang__)
+#pragma clang diagnostic ignored "-Wunused-lambda-capture"
```
JonasVautherin: Why not doing that at the CMake level?

julianoes: I wasn't sure whether to apply this everywhere. I was hoping it only ever comes up here.

JonasVautherin: As you want :).

As a principle, I don't like macros in the code, and I prefer to keep my code independent from the build system (as much as possible). But I'm fine with this as well.

julianoes: Convinced :smile:

JonasVautherin:
Note that you've changed it only for AppleClang, now (not sure if it is necessary for Clang, too).

julianoes:
> Note that you've changed it only for AppleClang, now (not sure if it is necessary for Clang, too).

You're right:
```
export CC=/usr/bin/clang
export CXX=/usr/bin/clang++
make
-- The C compiler identification is Clang 6.0.0
-- The CXX compiler identification is Clang 6.0.0
...
/home/julianoes/src/DroneCore/core/thread_pool_test.cpp:70:17: error: lambda capture 'first' is not required to be captured for this use
      [-Werror,-Wunused-lambda-capture]
    tp.enqueue([first, &second]() {
                ^
```

PR #418, opened 2018-06-05
Fix camera set_mode lockup

julianoes:
This fixes a lockup that happened when running `camera::set_mode()` in rapid succession in the integration test. The timing was only right to trigger it when run using valgrind.

The command to reproduce this with E90 was:
```
make && valgrind build/default/integration_tests/integration_tests_runner --gtest_filter="CameraTest.SetMode" --gtest_catch_exceptions=0
```

### DIFF
--- a/core/mavlink_commands.cpp
+++ b/core/mavlink_commands.cpp
@@ -149,9 +149,7 @@ void MAVLinkCommands::receive_command_ack(mavlink_message_t message)
         case MAV_RESULT_ACCEPTED:
             _state = State::NONE;
             _state_mutex.unlock();
-            if (work->callback) {
-                work->callback(Result::SUCCESS, 1.0f);
-            }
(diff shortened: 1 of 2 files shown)


core/mavlink_commands.cpp
```diff
@@ -144,10 +144,11 @@ void MAVLinkCommands::receive_command_ack(mavlink_message_t message)
         return;
     }
 
-    std::lock_guard<std::mutex> lock(_state_mutex);
+    _state_mutex.lock();
     switch (command_ack.result) {
         case MAV_RESULT_ACCEPTED:
             _state = State::NONE;
+            _state_mutex.unlock();
             if (work->callback) {
```
JonasVautherin: Shouldn't it also use `call_callback`?

julianoes: It should, scandalous. Thx.

PR #423, opened 2018-06-08
mission: import null as NaN

### LINKED ISSUE #422: Check how mission plan import handles `null`
For waypoints and takeoff items, param4 of the mavlink command can be set to `NaN` which means no change. However, when importing a QGC .plan, unchanged is represented as `null`.

We need to check how `null` is handled, presumably it gets translated to `0` instead of `NaN`.

Reported by @nicovanduijn.

julianoes:
When QGC sets a param as `null` it means unchanged which should
translate to NaN for mavlink mission items.

Fixes #422.

### DIFF
--- a/plugins/mission/mission_impl.cpp
+++ b/plugins/mission/mission_impl.cpp
@@ -1203,7 +1203,12 @@ Mission::Result MissionImpl::import_mission_items(Mission::mission_items_t &all_
         // Extract parameters of each mission item
         std::vector<double> params;
         for (auto &p : json_mission_item["params"].array_items()) {
-            params.push_back(p.number_value());
+            if (p.is_null()) {
+                // QGC sets params as `null` if they should be unchanged.
+                params.push_back(NAN);
+            } else {
+                params.push_back(p.number_value());
+            }
         }
 
         result = build_mission_items(command, params, new_mission_item, all_mission_items);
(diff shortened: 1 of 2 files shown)


julianoes:
According to @nicovanduijn this still doesn't work.

plugins/mission/mission_impl.cpp
```diff
@@ -1203,7 +1203,12 @@ Mission::Result MissionImpl::import_mission_items(Mission::mission_items_t &all_
         // Extract parameters of each mission item
         std::vector<double> params;
         for (auto &p : json_mission_item["params"].array_items()) {
-            params.push_back(p.number_value());
+            if (p.is_null()) {
```
shakthi-prashanth-m: Does this check for `null` rightly ?

shakthi-prashanth-m:
I think QGC should stick to using NaN rather. Isn't it ?

julianoes:
> I think QGC should stick to using NaN rather. Isn't it ?

I think `NaN` is not supported by the JSON format.

plugins/mission/mission_impl.cpp
julianoes: From my tests, yes.

julianoes:
@shakthi-prashanth-m agreed?

plugins/mission/mission_impl.cpp
shakthi-prashanth-m: Then, its fine.

shakthi-prashanth-m:
>I think NaN is not supported by the JSON format.
Oh yeah. Sure agreed.

PR #430, opened 2018-06-14
core: fix lockup of action::takeoff_async

julianoes:
This fixes the takeoff async integration test where we locked ourselves
out because the takeoff command actually does two commands in
succession, first a mode change, then the takeoff command. What happened
is that we locked ourselves out because we sent the second command in
the callback of the first command which still had the lock.


julianoes:
An alternative would probably be to use a recursive lock.

PR #431, opened 2018-06-20
Parallelize the builds of dependencies

thomasetter:
This commit just sets the number of jobs used by make to the number of cpus in the system. This greatly accelerated builds on my system.
I might have more cores than most, but with all current-gen laptops having at least 8 (hyper-/SMT) threads, this should give a nice boost for most users.

TESTED:
`git clone https://github.com/thomasetter/DroneCore.git --recurse-submodules -j 20 && cd DroneCore/ && mkdir build && cd build && cmake .. && make -j12`

### DIFF
--- a/backend/cmake/helpers/build_target.cmake
+++ b/backend/cmake/helpers/build_target.cmake
@@ -1,3 +1,5 @@
+include(ProcessorCount)
+
 function(build_target SRC_DIR TARGET_DIR)
     file(MAKE_DIRECTORY ${TARGET_DIR}/build)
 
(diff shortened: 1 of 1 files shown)


julianoes:
Thanks @thomasetter. However, the recommended way to build is just `make` which creates the cmake build folder and does `make -j8`.

https://github.com/dronecore/DroneCore/blob/d1a5342637e49c204a96ed7aba74eba112117405/Makefile#L22
https://github.com/dronecore/DroneCore/blob/d1a5342637e49c204a96ed7aba74eba112117405/Makefile#L47-L67

Since we were considering to move away from the magic Makefile to plain cmake your change is great though.

JonasVautherin:
Oh, I had not realized that. Thanks Julian!

PR #433, opened 2018-06-28
WIP telemetry: add local position subscription

Stifael:
In progress:
ToDo:

- [ ]  integration / unit tests


integration_tests/telemetry_async.cpp
```diff
@@ -232,3 +237,15 @@ void print_rc_status(Telemetry::RCStatus rc_status)
               << std::endl;
     _received_rc_status = true;
 }
+
+void print_position_velocity_local(Telemetry::PositionVelocityLocal position_velocity_local)
+{
+    std::cout << "Got position north_m: " << position_velocity_local.north_m << " m, "
```
julianoes: There is the unit `m` at the end so you don't need to write `north_m` as well.

integration_tests/telemetry_async.cpp
```diff
@@ -232,3 +237,15 @@ void print_rc_status(Telemetry::RCStatus rc_status)
...
+              << "east_m: " << position_velocity_local.east_m << " m, "
+              << "donw_m: " << position_velocity_local.down_m << " m" << std::endl
```
julianoes: `donw` -> `down`.

plugins/telemetry/telemetry.h
```diff
@@ -43,6 +43,21 @@ class Telemetry : public PluginBase {
         float relative_altitude_m; /**< @brief Altitude relative to takeoff altitude in metres */
     };
 
+    /**
+     * @brief Position and velocity local type.
```
julianoes: Please describe what it is for someone that doesn't know or understand PX4.

plugins/telemetry/telemetry.h
```diff
@@ -43,6 +43,21 @@ class Telemetry : public PluginBase {
...
+     */
+    struct PositionVelocityLocal {
+        float north_m; /**< @brief Position along north-direction of local frame in meters. */
```
julianoes: I don't think it's clear what is meant with `north_m` and `north_m_s`. 

I think it would make more sense to have the structure like this:
```
struct LocalPositionNED {
    float north_m;
    float east_m;
    float down_m
};
struct VelocityNED {
    float north_m_s;
    float east_m_s;
    float down_m_s;
};
struct LocalPositionAndVelocity {
    LocalPositionNED position;
    VelocityNED velocity;
};
```
and maybe you can even share some of the types with other structs.

plugins/telemetry/telemetry.h
```diff
@@ -250,6 +273,14 @@ class Telemetry : public PluginBase {
      */
     Result set_rate_rc_status(double rate_hz);
 
+    /**
+     * @brief Set rate of position velocity local updates (asynchronous).
```
julianoes: "position velocity local updates" is not clear what it means. The comments need to be proper sentences.

integration_tests/telemetry_async.cpp
Stifael: resolved

integration_tests/telemetry_async.cpp
Stifael: resolved

plugins/telemetry/telemetry.h
Stifael: resolved

plugins/telemetry/telemetry.h
Stifael: I added the structure you suggested, but decided against using one of the other structures, despite the available velocity structure (`GroundSpeedNED`). The name GroundSpeedNED is contradicting because `Speed` implies that there is no direction information and the values always have to be positive (in contrast to `velocity`). However, `NED` implies the opposite that the quantity in question is velocity and not speed.

plugins/telemetry/telemetry.h
Stifael: resolved

PR #434, opened 2018-06-28
Additional offboard setpoints

Stifael:
a couple more sets of offboard-setpoints. The added sets are as follow:
position-yaw, position-velocity-yaw, velocity-climbrate-yaw, velocity-climbrate-yawspeed,

Note that the firmware does not support it yet. However, once this [PR](https://github.com/PX4/Firmware/pull/9731) gets in together with some changes of this [branch](https://github.com/Stifael/Firmware/tree/offboard_enable_feedforward), any kind of mixing of setpoints should be allowed.

~You don't have to review it yet.~

### DIFF
--- a/plugins/offboard/include/plugins/offboard/offboard.h
+++ b/plugins/offboard/include/plugins/offboard/offboard.h
@@ -74,11 +74,11 @@ public:
      * @brief Type for Position commands in NED (North East Down) coordinates and yaw.
      */
     struct PositionNEDYaw {
-        float north_m; /**< @brief Position along local north in meters. */
-        float east_m; /**< @brief Position along local east in meters. */
-        float down_m; /**< @brief Position along local down in meters. */
-        float yaw_deg; /**< @brief Yaw in degrees relative to local reference (positive is
-                          clock-wise looking from above). */
+        float north_m; /**< @brief Position along north in meters. */
+        float east_m; /**< @brief Position along east in meters. */
+        float down_m; /**< @brief Position along down in meters. */
+        float yaw_deg; /**< @brief Yaw in degrees (0 North, positive is clock-wise looking from
+                          above). */
     };
 
     /**
@@ -93,19 +93,20 @@ public:
     };
 
     /**
-     * @brief Type for Velocity commands in NE (North East Down) and Altitude in N coordinates and
+     * @brief Type for Velocity commands in NE (North East) and Altitude in N (Down) coordinate and
      * yaw.
      */
     struct VelocityAltitudeNEDYaw {
         float north_m_s; /**< @brief Velocity North in meters/second. */
         float east_m_s; /**< @brief Velocity East in meters/second. */
-        float down_m; /**< @brief Velocity Down in meters. */
+        float down_m; /**< @brief Altitude Down in meters. */
         float yaw_deg; /**< @brief Yaw in degrees (0 North, positive is clock-wise looking from
                           above). */
     };
 
     /**
-     * @brief Type for Velocity commands in NED (North East Down) coordinates and yaw.
+     * @brief Type for Position commands in NE (North East) and Velocity in N (Down) coordinate and
+     * yaw.
      */
     struct PositionClimbRateNEDYaw {
         float north_m; /**< @brief Position North in meters. */
@@ -116,13 +117,14 @@ public:
     };
 
     /**
-     * @brief Type for Velocity commands in NED (North East Down) coordinates and yaw.
+     * @brief Type for Position commands in NE (North East) and Velocity in N (Down) coordinate and
+     * yawspeed.
      */
     struct PositionClimbRateNEDYawspeed {
         float north_m; /**< @brief Position North in meters. */
         float east_m; /**< @brief Position East in meters. */
         float down_m_s; /**< @brief Velocity Down in meters/second. */
-        float yawspeed_deg_s; /**< @brief Yawspeed in degrees / seconds (0 North, positive is
+        float yawspeed_deg_s; /**< @brief Yawspeed in degrees per seconds (positive is
                           clock-wise looking from above). */
     };
 
@@ -185,8 +187,9 @@ public:
     bool is_active() const;
 
     /**
-     * @brief Set the position in NED coordinates and yaw.
+     * @brief Set the position in NED coordinate and yaw.
      *
+     * @see PositionNEDYaw
      * @param position_ned_yaw Position and yaw `struct`.
      */
     void set_position_ned_yaw(PositionNEDYaw position_ned_yaw);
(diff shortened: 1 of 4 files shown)


Stifael:
To write proper integration tests, the telemetry PR for [local position](https://github.com/dronecore/DroneCore/pull/433) is required. Other than that ready for review.

Please note again: Px4 master does not yet support the mixing of setpoints yet (but this [branch](https://github.com/Stifael/Firmware/tree/offboard_enable_feedforward) does)

plugins/offboard/offboard.h
```diff
@@ -81,6 +92,42 @@ class Offboard : public PluginBase {
...
     };
 
+    /**
+     * @brief Type for Velocity commands in NE (North East) and Altitude in N (Down) coordinate and
+     * yaw.
+     */
+    struct VelocityAltitudeNEDYaw {
+        float north_m_s; /**< @brief Velocity North in meters/second. */
+        float east_m_s; /**< @brief Velocity East in meters/second. */
+        float down_m; /**< @brief Altitude Down in meters. */
```
julianoes: I find `Altitude Down` confusing but I don't have a better idea.

plugins/offboard/offboard.h
```diff
@@ -81,6 +92,42 @@ class Offboard : public PluginBase {
...
+        float yaw_deg; /**< @brief Yaw in degrees (0 North, positive is clock-wise looking from
+                          above). */
+    };
+
+    /**
+     * @brief Type for Position commands in NE (North East) and Velocity in N (Down) coordinate and
```
julianoes: `N (Down)`?

plugins/offboard/offboard.h
```diff
@@ -81,6 +92,42 @@ class Offboard : public PluginBase {
...
+        float north_m; /**< @brief Position North in meters. */
+        float east_m; /**< @brief Position East in meters. */
+        float down_m_s; /**< @brief Velocity Down in meters/second. */
+        float yaw_deg; /**< @brief Yaw in degrees (0 North, positive is clock-wise looking from
+                          above). */
+    };
+
+    /**
+     * @brief Type for Position commands in NE (North East) and Velocity in N (Down) coordinate and
+     * yawspeed.
```
julianoes: Is this an actual use case?

plugins/offboard/offboard_impl.cpp
```diff
@@ -106,6 +106,88 @@ void OffboardImpl::receive_command_result(MAVLinkCommands::Result result,
...
+    Offboard::VelocityAltitudeNEDYaw velocity_altitude_ned_yaw)
+{
+    TranslationNED translation_ned = {NAN,
+                                      NAN,
+                                      velocity_altitude_ned_yaw.down_m,
+                                      velocity_altitude_ned_yaw.north_m_s,
+                                      velocity_altitude_ned_yaw.east_m_s,
+                                      NAN,
+                                      velocity_altitude_ned_yaw.yaw_deg,
+                                      NAN};
```
julianoes: The way this is written is actually hard to read because it is not clear what the fields are and what is set to `NAN`.
I think it's not self-explanatory and can lead to mistakes. What about this way?
```
TranslationNED translation_ned{NAN, NAN, NAN, NAN, NAN, NAN, NAN, NAN};
translation_ned.down_m = position_ned_yaw.down_m;
...
```

plugins/offboard/offboard_impl.cpp
```diff
@@ -160,6 +242,101 @@ void OffboardImpl::set_velocity_body(Offboard::VelocityBodyYawspeed velocity_bod
...
+    float yawspeed = 0.0f;
+    const float afx = 0.0f;
+    const float afy = 0.0f;
+    const float afz = 0.0f;
+
+    _mutex.lock();
+    if (std::isfinite(_translation_ned.yaw_deg)) {
+        yaw = to_rad_from_deg(_translation_ned.yaw_deg);
+        // enable yaw
+        IGNORE_YAW &= ~(1 << 10);
```
julianoes: So this is a bit inverted in how it was planned. The goal was to create readable constants for the ignore flags, so that we don't have to use these magic `<< 10` shifters anywhere else.

However, what you're doing is more bitwise operations that are hard to check down here.

I'd suggest to do something like that:
```
const uint16_t IGNORE_YAW = (1 << 10);
...

uint16_t ignore_flags = 0;

if (std::isfinite(_translation_ned.yaw_deg)) {
    yaw = to_rad_from_deg(_translation_ned.yaw_deg);
} else {
    ignore_flags |= IGNORE_YAW;
}
```

plugins/offboard/offboard_impl.cpp
```diff
@@ -160,6 +242,101 @@ void OffboardImpl::set_velocity_body(Offboard::VelocityBodyYawspeed velocity_bod
     send_velocity_body();
 }
 
+void OffboardImpl::send_translation_ned()
```
julianoes: I like the code sharing here!

julianoes:
@Stifael when you correct issues from the review, please add additional commits instead of rewriting the existing ones. That way it's easier to see what has been addressed.

plugins/offboard/offboard.h
Stifael: since i called the other two members `velocity`, I can call the third one `position down`

plugins/offboard/offboard.h
Stifael: error: D (down)

plugins/offboard/offboard.h
Stifael: it was mainly added for @darioxz. He has a version where he uses his own "yaw-controller" that spits out yawspeed setpoints.

julianoes:
@Stifael make sure to rebase. There have been some changes :smile:

Stifael:
the Autopilot changes are/will be in this PR https://github.com/PX4/Firmware/pull/10031#issuecomment-407798323

plugins/offboard/include/plugins/offboard/offboard.h
```diff
@@ -139,13 +186,46 @@ class Offboard : public PluginBase {
...
+     */
+    void set_velocity_altitude_ned_yaw(VelocityAltitudeNEDYaw velocity_altitude_ned_yaw);
+
+    /**
+     * @brief Set the position in NE and climbrate in D coordinate and yaw.
+     *
+     * @see PositionClimbRateNEDYaw
+     * @param position_climbrate_ned_yaw Position/Climbrate and yaw `struct`.
+     */
+    void set_position_climbrate_ned_yaw(PositionClimbRateNEDYaw position_climbrate_ned_yaw);
```
julianoes: If the type is `ClimbRate` it should be `climb_rate`. I would probably go for `Climbrate` and `climbrate`.

plugins/offboard/offboard_impl.cpp
```diff
@@ -160,6 +234,101 @@ void OffboardImpl::set_velocity_body(Offboard::VelocityBodyYawspeed velocity_bod
...
+{
+    // by default ingore all setpoints
```
julianoes: I think this comment is wrong because that's not what we're doing.
Also, `ingore` -> `ignore`.

julianoes:
@Stifael any time to wrap this up sometime? Would be nice!

Stifael:
I will come back to you once I know if I get the time to work on it or not.

julianoes:
There is demand for this :)

https://github.com/Dronecode/DronecodeSDK/pull/434

hamishwillee:
As an aside, it still frustrates me that we don't have or publish a clear picture on what is supported PX4 side with respect to MAVLink messages/frames. I'm hoping the test suite will eventually allow some sort of automated testing to verify that different behaviour is supported/not supported.

PR #435, opened 2018-06-29
Cmake public vs private header

Stifael:
I moved all public headers in an include directory of the specific plugin or core. The header within include directories are made public. 
Related issue/request: https://github.com/dronecore/DroneCore/issues/424


julianoes:
I'm not sure I agree with this proposal. I don't like the fact that we need to do `#include "plugins/telemetry/include/telemetry.h"` now which is ugly. At least all `include` directories should be include directories for cmake.

I think first we need to specify what the including should look like both inside the library and outside.

Right now it is from the outside (installed):
```
#include <dronecore/dronecore.h>
#include <dronecore/action.h>
```

And from the inside:
```
#include "dronecore.h"
#include "plugins/action/action.h"
```

I think in the future we should get the APIs to match, so from the outside:
```
#include <dronecore/dronecore.h>
#include <dronecore/plugins/action/action.h>
```

And inside:
```
#include "dronecore.h"
#include "plugins/action/action.h"
```

Stifael:
@julianoes I think there is a confusion about what I am trying to solve.

right now what is supported:

## NOW

### installation: 
```
#include <dronecore/dronecore.h>
#include <dronecore/action.h>
```
I think this also makes sense. Not sure if it makes sense to have the additional `plugins` folder in the directory. If another directory is required, then you need to change the installation directory, but that is independent of this PR. 

### Inside
From inside I think you are referring to `example` directory. I don't think that works without adding some include_directories in the CMakeLists.txt, which works on the directory level and should be avoided in general. [here](https://gist.github.com/mbinna/c61dbb39bca0e4fb7d1f73b0d66a4fd1) is a good cmake tutorial.

## WITH THIS PR

### installation
should be still the same as before because the installation directory did not change:
https://github.com/dronecore/DroneCore/pull/435/files#diff-7c907cb05df9fabfc3030ae65643da3bR20

### inside
to me, inside is a bit a weird concept, because I don't want to create a folder within DroneCore (such as the example directory). What I want is to add DroneCore as subdirectory in my own CMakeLists.txt and link the core and the plugins to my own target. 
Example:
```
My-Project:
Top CMakeLists.txt
 add_subdirectory("path to DroneCore")
-- src
---CMakeLists.txt
--- target_link_libraries(my-project dronecore dronecore_[plugin-name])
```
this way, in my-project.cpp file, I can just add 

```
#include "dronecore.h"
#include " plugin-name.h"
```

julianoes:
Ok, but internally this is very ugly:
```
#include "plugins/action/include/action.h"
```

Stifael:
> Ok, that's why I suggest to include with plugins/action/action.h

i used target_include_directories, where you can specify which directories are public and which are private. this is also the recommended way.

> but internally this is very ugly
What do you mean with internal? are you referring to example or in general. By the way, this kind of [inldue](https://github.com/dronecore/DroneCore/pull/435/files#diff-61258fc66251d6743064005d856182eaR5) are not necessary and can be replaced with:
```
#include [plugin-name].h
```

I forgot to replace it.

julianoes:
Ok I understand what you mean. 

However, don't you think that it would be nicer to have the same paths for the install and the inside (example or your submodule) case?

Stifael:
I suggest to specify first what the include paths should be and then go from there.

JonasVautherin:
I need to have a closer look at that before I can join the discussion. I'll do that ASAP!

Stifael:
any updates?
I just need to know what the desire path should be.

Stifael:
I will now go with: 

```
#include <dronecore/dronecore.h>
#include <dronecore/plugins/action/action.h>

And inside:

#include "dronecore.h"
#include "plugins/action/action.h"
```

JonasVautherin:
On my side, I am fine with @Stifael's [latest proposition](https://github.com/Dronecode/DronecodeSDK/pull/435#issuecomment-405236980).

But I would not be against having the "external" scheme for everything (i.e. instead of having a different scheme for "internal"):

```
#include <dronecore/dronecore.h>
#include <dronecore/plugins/action/action.h>
```

julianoes:
> But I would not be against having the "external" scheme for everything (i.e. instead of having a different scheme for "internal"):
> 
> ```
> #include <dronecore/dronecore.h>
> #include <dronecore/plugins/action/action.h>
> ```

@JonasVautherin is that usual to use this internal as well? I wouldn't think so.

JonasVautherin:
@julianoes I don't know. I've seen it and it did not look weird to me. But again, both are fine for me ;-). Let's go for [the other one](https://github.com/Dronecode/DronecodeSDK/pull/435#issuecomment-405236980) then?

julianoes:
Ok, I agree that this is how internally the includes can happen:
```
#include "dronecode_sdk.h"
#include "system.h" (if only system is required)
#include "plugins/<name>/<name>.h"
```

If you're actually in another project and you have a submodule of dronecode_sdk, then you should also be able to include it like this to prevent clashes with your own project:
```
#include "dronecode_sdk/dronecode_sdk.h"
#include "dronecode_sdk/system.h" (if only system is required)
#include "dronecode_sdk/plugins/<name>/<name>.h"
```

Stifael:
> If you're actually in another project and you have a submodule of dronecode_sdk, then you should also be able to include it like this to prevent clashes with your own project:

I don't know how to do that yet or if that is even possible to have a different include path if the project is included as a submodule.

Stifael:
I mean it would be possible, but I think that would require to move the entire Dronecore project within a folder `dronecode_sdk` that contains one CMakLists.txt that only contains `add_subdirectory(dronecde_sdk)`

julianoes:
Oh ok, yer nevermind then.

Stifael:
@JonasVautherin @julianoes how should we proceed? 
In terms of SDK installation nothing has changed. In terms of internal include, it changed according to https://github.com/PX4/Firmware/pull/10035. 
I would prefer to change the installation path in another PR.

julianoes:
I find this a bit ugly, is it really required with the `include` directory?
```
tree plugins/info   
plugins/info
├── CMakeLists.txt
├── include
│   └── plugins
│       └── info
│           └── info.h
├── info.cpp
├── info_impl.cpp
└── info_impl.h
```

Stifael:
@julianoes 
I would'nt know how to do it differently. Somehow you need to have a different directory than the current CMAKE_SOURCE_DIR to be able to distinguish between PUBLIC and PRIVATE headers. If I remove the include, then I don't know how to make the distinction without losing the namespace.

examples from tutorials:
https://pabloariasal.github.io/2018/02/19/its-time-to-do-cmake-right/
https://rix0r.nl/blog/2015/08/13/cmake-guide/

Stifael:
This tree could also work:

``` 
plugins/info
├── CMakeLists.txt
├── include
│   └── plugins
│       └── info
│           └── info.h
├── src/ 
       |____    info.cpp
       |____ info_impl.cpp
       ...
```

julianoes:
That's also ugly, isn't it?

Stifael:
In my opinion the only reason why it is ugly is because of the long path that is required for plugins. 
But again, don't know how to do it differently given the desired path:
```
#include "plugins/info/info.h"

JonasVautherin:
I don't find it weird to have `include` and `src` in a project (and only `include` like now is also fine for me, since we don't have so many files).

In our case, it may seem weird because we have many small libraries, but those libraries are somehow independent, so they count as "projects", right?

julianoes:
Yer it makes sense, it's just a bit strange how the directory `include` again needs to have the full path but I understand how it needs to be that way technically.

julianoes:
Thanks @Stifael!

PR #436, opened 2018-06-29
Added calibration plugin

julianoes:
This adds a calibration plugin which right now supports gyro, accelerometer, and magnetometer calibration.


julianoes:
Note to self about CI:
```
/root/DroneCore/install/include/dronecore/calibration.h:11: warning: Compound dronecore::Calibration is not documented.
```

PR #439, opened 2018-07-06
Add plugin warnings

julianoes:
By oversight we did not have the special warnings enabled for the plugins.

This enables the warnings again and fixes a few that crept in. And one was actually a bug so that confirms that warnings are important!

### DIFF
--- a/core/global_include.h
+++ b/core/global_include.h
@@ -5,14 +5,15 @@
 #include <chrono>
 #include <thread>
 
-#ifdef WINDOWS
-#ifndef _USE_MATH_DEFINES
-#define _USE_MATH_DEFINES
+// Instead of using the constant from math.h or cmath we define it ourselves. This way
+// we don't import all the other C math functions and make sure to use the C++ functions
+// from the standard library (e.g. std::abs() instead of abs()).
+#ifndef M_PI
+constexpr double M_PI = 3.14159265358979323846;
 #endif
-// cmath doesn't contain M_PI
-#include <math.h>
-#else
-#include <cmath>
+
+#ifndef M_PI_F
+constexpr float M_PI_F = float(M_PI);
 #endif
 
 #define MIN(x_, y_) ((x_) > (y_)) ? (y_) : (x_)
@@ -22,10 +23,6 @@
 #define STRNCPY strncpy_s
 #else
 #define STRNCPY strncpy
-#endif
-
-#ifndef M_PI_F
-#define M_PI_F float(M_PI)
 #endif
 
 namespace dronecode_sdk {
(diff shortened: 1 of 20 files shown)


plugins/calibration/calibration_impl.cpp
```diff
@@ -330,7 +330,7 @@ void CalibrationImpl::process_statustext(const mavlink_message_t &message)
             // FALLTHROUGH
         case CalibrationStatustextParser::Status::CANCELLED:
             _calibration_callback = nullptr;
-            _state == State::NONE;
+            _state = State::NONE;
```
JonasVautherin: Good catch :)

plugins/telemetry/telemetry.cpp
```diff
@@ -319,17 +319,17 @@ const char *Telemetry::result_str(Result result)
...
 {
-    return abs(lhs.position.north_m - rhs.position.north_m) <=
-               std::numeric_limits<double>::epsilon() &&
-           abs(lhs.position.east_m - rhs.position.east_m) <=
-               std::numeric_limits<double>::epsilon() &&
-           abs(lhs.position.down_m - rhs.position.down_m) <=
+    return fabs(lhs.position.north_m - rhs.position.north_m) <=
                std::numeric_limits<float>::epsilon() &&
-           abs(lhs.velocity.north_m_s - rhs.velocity.north_m_s) <=
+           fabs(lhs.position.east_m - rhs.position.east_m) <=
```
JonasVautherin: Right.

core/global_include.h
```diff
@@ -5,14 +5,15 @@
 #include <chrono>
 #include <thread>
 
-#ifdef WINDOWS
-#ifndef _USE_MATH_DEFINES
-#define _USE_MATH_DEFINES
+// Instead of using the constant from math.h or cmath we define it ourselves. This way
```
JonasVautherin: :+1:

plugins/camera/camera_impl.cpp
```diff
@@ -509,8 +510,13 @@ Camera::Result CameraImpl::set_mode(const Camera::Mode mode)
...
 void CameraImpl::save_camera_mode(const float mavlink_camera_mode)
 {
+    if (!std::isfinite(mavlink_camera_mode)) {
+        LogWarn() << "Can't save NAN as camera mode";
+        return;
+    }
+
     MAVLinkParameters::ParamValue value;
-    value.set_uint32(mavlink_camera_mode);
+    value.set_uint32(uint32_t(mavlink_camera_mode));
```
JonasVautherin: What about `static_cast<uint32_t>(mavlink_camera_mode)`? Looks more modern to me :sweat_smile:

julianoes: True.

julianoes:
I'm merging this and ignoring the jenkins CI for the branch. I think it fails because it greps for warning which is the name of the branch and therefore triggers when it should no.

PR #441, opened 2018-07-06
[WIP] Update to new UUID

julianoes:
At some point we need to update to the new UUID (`uid2`) mavlink field.

This is an old branch where I started this.


julianoes:
I'm tempted to scrap this and just use the MAVLink system ID 1-255 instead of the complicated UID addressing.

hamishwillee:
I can see why. The only value in it is that you can write code that is resistant to the case where people change Sys ID on a system, and in future we could move to a different transport than MAVLink. For the former the benefits seem minimal for the complication, and for the later we'd probably rethink  a lot of things anyway.

julianoes:
Right but we don't have that now. And this UID only works for autopilots but not form companion computers, cameras, etc. anyway.

hamishwillee:
Yes. So if I was the boss, we would move to just SysIds.

PR #442, opened 2018-07-06
[WIP] Multi mission checks

julianoes:
This is to keep @shakthi-prashanth-m's old  branch.

This is an example, however, we should make an integration test that does several missions in a row.


julianoes:
We now do two mission in succession in the integration test, at least once #642 is in.

PR #445, opened 2018-07-09
Fix docker working directory

julianoes:
As brought up by @hamishwillee we need to change the docker working directory. This required to rebuild the docker images and therefore also update the tags in the Jenkinsfile.


julianoes:
I was hoping to fix the Jenkins build today and then merge this. And after that we can rebase the other PRs and get them in as well.

Issue #440, opened 2018-07-06
MAVLink parameter implementation should handle retrying, buffering


julianoes:
The plugins should not need to retry getting parameters but the parameter implementation should handle all the nitty gritty. Plugins should only get or set something and then wait on the result.

julianoes:
That's been improved.

PR #446, opened 2018-07-09
action: add RTL return altitude setting

julianoes:
This adds the setting to change the RTL return altitude which is the altitude at which the drone will return if RTL is engaged if it is lower (if higher, it will return at the current higher altitude).

The param system needs some refactoring in order to get this to work properly, see #440.

Todo: an integration test for this would be nice.

### DIFF
--- a/plugins/action/action.h
+++ b/plugins/action/action.h
@@ -228,7 +228,7 @@ public:
     /**
      * @brief Get the takeoff altitude.
      *
-     * @return takeoff Takeoff altitude relative to ground/takeoff location, in meters.
+     * @return Takeoff altitude relative to ground/takeoff location, in meters.
      */
     float get_takeoff_altitude_m() const;
 
@@ -246,7 +246,28 @@ public:
      */
     float get_max_speed_m_s() const;
 
-    // Non-copyable
+    /**
+     * @brief Set the return to launch minimum return altitude.
+     *
+     * When return to launch is initiated, the vehicle climbs to the return altitude if it is lower
+     * and stays at the current altitude if higher than the return altitude. Then it returns to the
+     * home location and lands there.
+     *
+     * @param relative_altitude_m Return altitude relative to takeoff location, in meters.
+     */
+    void set_return_to_launch_return_altitude(float relative_altitude_m);
+
+    /**
+     * @brief Get the return to launch minimum return altitude.
+     *
+     * When return to launch is initiated, the vehicle climbs to the return altitude if it is lower
+     * and stays at the current altitude if higher than the return altitude. Then it returns to the
+     * home location and lands there.
+     *
+     * @return Return altitude relative to takeoff location, in meters.
+     */
+    float get_return_to_launch_return_altitude() const;
+
     /**
      * @brief Copy constructor (object is not copyable).
      */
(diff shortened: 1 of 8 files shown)


plugins/action/action.h
```diff
@@ -246,7 +246,28 @@ class Action : public PluginBase {
...
+     * home location and lands there.
+     *
+     * @param relative_altitude_m Return altitude relative to takeoff location, in meters.
+     */
+    void set_return_to_launch_return_altitude(float relative_altitude_m);
+
+    /**
+     * @brief Get the return to launch minimum return altitude.
+     *
+     * When return to launch is initiated, the vehicle climbs to the return altitude if it is lower
```
hamishwillee: I might just remove this and do a see also to the `get_max_speed_m_s` (or visa versa)

plugins/action/action.h
```diff
@@ -246,7 +246,28 @@ class Action : public PluginBase {
      */
     float get_max_speed_m_s() const;
 
-    // Non-copyable
+    /**
+     * @brief Set the return to launch minimum return altitude.
+     *
+     * When return to launch is initiated, the vehicle climbs to the return altitude if it is lower
```
hamishwillee: 1. The naming in PX4 for RTL is now "Return Mode". We should perhaps rename the other API
2. The behaviour on return is configurable based on parameters - e.g. it might not land. In particular if you call this on a Fixed Wing vehicle. I tend to not talk about the landing if I don't need to.
3. Would it make sense to embed the setter into the return command function? Probably not, but we should certainly do a see also between this and the return command function.

Perhaps:

> When Return mode is engaged the vehicle will ascend to this altitude (if below) for the return flight. The vehicle will maintain its current altitude if above this altitude.

plugins/action/action_impl.cpp
```diff
@@ -411,21 +411,18 @@ void ActionImpl::set_takeoff_altitude(float relative_altitude_m)
...
     if (success) {
-        // TODO: This should not be buffered like this, the param system
-        //       needs some refactoring.
         _relative_takeoff_altitude_m = new_relative_altitude_m;
     }
 }
 
 float ActionImpl::get_takeoff_altitude_m() const
 {
+    // FIXME: we never actually get the param, as we should.
```
JonasVautherin: So that means that you need to set it from the user side in order to be able to get it?

plugins/action/action.h
julianoes: Right.

plugins/action/action.h
julianoes: 1. Good point, we should think about the naming.

2. I think the SDK needs to assume default params for everything it has not changed. Otherwise we're opening a whole can of worms and can't guarantee at all what happens when you call something. The firmware has way too many parameters that you need to know to be safe.

3. My aim is to have simple functions with as little arguments as possible. Otherwise, the function signature soon would look like this:
```
void Action::return_to_launch_async(action_callback_t callback, float return_altitude_m, bool include_land, bool do_smart_rtl);
```
```

plugins/action/action_impl.cpp
julianoes: Exactly. I'll refactor the param interface soon, see #440.

PR #447, opened 2018-07-09
Gimbal accelerometer calibration

julianoes:
This adds an API to calibrate the accelerometer of a gimbal.
The only difference to the normal accelerometer calibration is that the command is sent to the component ID of the gimbal.


julianoes:
Note, needs testing on E90.

PR #448, opened 2018-07-10
334 finish camera backend

### LINKED ISSUE #334: Implement backend for camera
- [x] TakePhoto
- [x] StartPhotoInterval
- [x] StopPhotoInterval
- [x] StartVideo
- [x] StopVideo
- [x] StartVideoStreaming
- [x] StopVideoStreaming
- [x] SetMode
- [x] SubscribeMode
- [x] SetVideoStreamSettings
- [x] SubscribeVideoStreamInfo
- [x] SubscribeCaptureInfo
- [x] SubscribeCameraStatus
- [x] SubscribeCurrentSettings
- [x] SubscribePossibleSettings
- [x] SetSetting

JonasVautherin:
Fixes #334.

### DIFF
--- a/backend/src/plugins/camera/camera_service_impl.h
+++ b/backend/src/plugins/camera/camera_service_impl.h
@@ -314,6 +314,113 @@ public:
         return grpc::Status::OK;
     }
 
+    grpc::Status
+    SubscribeCaptureInfo(grpc::ServerContext * /* context */,
+                         const rpc::camera::SubscribeCaptureInfoRequest * /* request */,
+                         grpc::ServerWriter<rpc::camera::CaptureInfoResponse> *writer) override
+    {
+        std::promise<void> stream_closed_promise;
+        auto stream_closed_future = stream_closed_promise.get_future();
+
+        bool is_finished = false;
+
+        _camera.subscribe_capture_info([&writer, &stream_closed_promise, &is_finished](
+                                           const dronecore::Camera::CaptureInfo capture_info) {
+            rpc::camera::CaptureInfoResponse rpc_capture_info_response;
+            auto rpc_capture_info = translateCaptureInfo(capture_info);
+            rpc_capture_info_response.set_allocated_capture_info(rpc_capture_info.release());
+
+            if (!writer->Write(rpc_capture_info_response) && !is_finished) {
+                is_finished = true;
+                stream_closed_promise.set_value();
+            }
+        });
+
+        stream_closed_future.wait();
+
+        return grpc::Status::OK;
+    }
+
+    static std::unique_ptr<rpc::camera::CaptureInfo>
+    translateCaptureInfo(const dronecore::Camera::CaptureInfo &capture_info)
+    {
+        auto rpc_capture_info =
+            std::unique_ptr<rpc::camera::CaptureInfo>(new rpc::camera::CaptureInfo());
+        rpc_capture_info->set_allocated_position(
+            translatePosition(capture_info.position).release());
+        rpc_capture_info->set_allocated_quaternion(
+            translateQuaternion(capture_info.quaternion).release());
+        rpc_capture_info->set_time_utc_us(capture_info.time_utc_us);
+        rpc_capture_info->set_is_success(capture_info.success);
+        rpc_capture_info->set_index(capture_info.index);
+        rpc_capture_info->set_file_url(capture_info.file_url);
+
+        return rpc_capture_info;
+    }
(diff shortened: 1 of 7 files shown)


backend/src/plugins/camera/camera_service_impl.h
```diff
@@ -121,6 +123,542 @@ class CameraServiceImpl final : public rpc::camera::CameraService::Service {
...
+    }
+
+    grpc::Status SubscribeMode(grpc::ServerContext * /* context */,
+                               const rpc::camera::SubscribeModeRequest * /* request */,
+                               grpc::ServerWriter<rpc::camera::ModeResponse> *writer) override
+    {
+        std::promise<void> stream_closed_promise;
+        auto stream_closed_future = stream_closed_promise.get_future();
+
+        bool is_finished = false;
```
julianoes: This flag is used to avoid setting the promise twice, right? Why is that happening in the first place?

backend/src/plugins/camera/camera_service_impl.h
```diff
@@ -121,6 +123,542 @@ class CameraServiceImpl final : public rpc::camera::CameraService::Service {
...
+                fillResponseWithResult(response, dronecode_sdk::Camera::Result::WRONG_ARGUMENT);
+            }
+        } else {
+            const std::string setting_id = request->setting().setting_id();
+            dronecode_sdk::Camera::Option option;
+            option.option_id = request->setting().option().option_id();
+
+            _camera.set_option_async(
+                [this, response, &set_option_called_promise](
+                    dronecode_sdk::Camera::Result camera_result) {
```
julianoes: Don't you need to ignore results such as `IN_PROGRESS` and maybe others before setting the response here?

backend/test/camera_service_impl_test.cpp
```diff
@@ -174,6 +293,995 @@ TEST_F(CameraServiceImplTest, stopsVideoStreamingEvenWhenArgsAreNull)
...
+    result_callback(ARBITRARY_CAMERA_RESULT);
+}
+
+TEST_F(CameraServiceImplTest, setSettingReturnsWrongArgumentErrorIfRequestIsNull)
+{
+    dronecode_sdk::rpc::camera::SetSettingResponse response;
+
+    _camera_service.SetSetting(nullptr, nullptr, &response);
+
+    EXPECT_EQ("WRONG_ARGUMENT", CameraResult::Result_Name(response.camera_result().result()));
```
julianoes: Nice!

backend/src/plugins/camera/camera_service_impl.h
JonasVautherin: Oh, I don't know ^^. So `set_option_async()` may return `IN_PROGRESS`?

julianoes: Yes, I think that can potentially happen. Just ignore that result.

backend/src/plugins/camera/camera_service_impl.h
JonasVautherin: I had the issue in the unit tests, actually. But it's true that there, I am manually triggering the callback, which may never happen from the plugin.

I think this is the first example of a stream that gets disconnected from the client (e.g. the client calls `tryCancel()`). And I need an extra message after cancel has been called (supposedly because of how gRPC works), which was triggering this issue.

I'd say we keep it like this for now, I could investigate further into it later.

JonasVautherin:
The branch was out of date, but it still builds now that it is up-to-date with `develop`.

backend/src/plugins/camera/camera_service_impl.h
julianoes: I suppose so yes. Not all have that result but for calibration it's used because the process can take a while.

JonasVautherin: Are you sure I need to do that? It seems to me that `IN_PROGRESS` is some kind of error (like a specialization of `BUSY`). It happens if you call `SetSetting` again before the previous call finished, right?

julianoes: You receive `IN_PROGRESS` when the camera is working on the command. So it's not an error but rather an information.

julianoes: I suggest you add it and then merge it.

JonasVautherin: I see. Will do!

PR #476, opened 2018-07-31
core: don't use colors for iOS

julianoes:
The iOS console does not support colors, therefore we should omit the color magic.


julianoes:
Tested, merging.

PR #477, opened 2018-07-31
core: allow queue items to be changed

julianoes:
With the old implementation, it was not possible to change the values of an item borrowed from the queue because every borrow received a new shared_ptr of a copy of the item.

By storing the items as shared_ptr from the beginning, we can solve this.


julianoes:
Urgent bugfix, therefore merging it.

PR #479, opened 2018-08-02
Embed camera definition files for Yuneec camera

julianoes:
This is an interim workaround where the camera definition xml files are embedded in the library.

The tool to embed the xml files is [cpp_rsc](https://github.com/orex/cpp_rsc).


julianoes:
Note that this does not work for the iOS build yet because `cpp_rsc` compiles a binary that runs on the host as part of the build process. We need to work our way around this somehow, e.g. by re-using the `cpp_rsc` binary from the `default` build.

julianoes:
> So Yuneec MAVLink camera doesn't supply the camera definition file URL - as per the MAVLink Camera API spec?

No, they do supply it but the problem is that the file is not accessible currently through the MFi transport layer (mavlink is transmitted but no http yet), so this is an interim workaround.

CMakeLists.txt
```diff
@@ -16,6 +16,17 @@ include(cmake/curl.cmake)
...
 include(cmake/zlib.cmake)
 
+if(NOT MSVC)
+    include(third_party/cpp_rsc/cmake/modules/cpp_resource.cmake)
+    if(NOT ANDROID AND NOT IOS)
+        # If not cross-compiling, we can build cpp_rsc.
+        add_subdirectory(third_party/cpp_rsc/src)
+    else()
+        # For cross-compilation, re-use the binary from default.
+        set_property(GLOBAL PROPERTY CPPRSC_CMD ${CMAKE_CURRENT_BINARY_DIR}/../default/third_party/cpp_rsc/src/cpp_rsc)
```
JonasVautherin: I don't get where that is used :/

julianoes: It's used in https://github.com/orex/cpp_rsc/blob/928ad84e11d768b1c6a3aa427277234b49b53cb5/cmake/modules/cpp_resource.cmake#L71.

JonasVautherin: I see. And `add_resource` and `link_resource_file` come from cpp_rsc as well.

PR #480, opened 2018-08-02
core: retry longer to get autopilot version msg

julianoes:
This should resolve an error where we give up to early and therefore
never receive a autopilot version message and therefore we don't know
whether we can use the mission int mode.


julianoes:
Need this ASAP, merging.

PR #485, opened 2018-08-08
Add Goto Location Action

mzahana:
This PR adds `goto_location()` action which allows to send the vehicle to a specific WGS84 global position. It uses [MAV_DO_REPOSITION](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_REPOSITION)

### DIFF
--- a/plugins/action/action.cpp
+++ b/plugins/action/action.cpp
@@ -32,7 +32,10 @@ ActionResult Action::land() const
     return _impl->land();
 }
 
-ActionResult Action::goto_location(double latitude_deg, double longitude_deg, float altitude_amsl_m, float yaw_deg)
+ActionResult Action::goto_location(double latitude_deg,
+                                   double longitude_deg,
+                                   float altitude_amsl_m,
+                                   float yaw_deg)
 {
     return _impl->goto_location(latitude_deg, longitude_deg, altitude_amsl_m, yaw_deg);
 }
--- a/plugins/action/action_impl.cpp
+++ b/plugins/action/action_impl.cpp
@@ -133,9 +133,11 @@ ActionResult ActionImpl::return_to_launch() const
         _parent->set_flight_mode(SystemImpl::FlightMode::RETURN_TO_LAUNCH));
 }
 
-ActionResult ActionImpl::goto_location(double latitude_deg, double longitude_deg, float altitude_amsl_m, float yaw_deg)
+ActionResult ActionImpl::goto_location(double latitude_deg,
+                                       double longitude_deg,
+                                       float altitude_amsl_m,
+                                       float yaw_deg)
 {
-
     MAVLinkCommands::CommandInt command{};
 
     command.command = MAV_CMD_DO_REPOSITION;
--- a/plugins/action/include/plugins/action/action.h
+++ b/plugins/action/include/plugins/action/action.h
@@ -117,7 +117,8 @@ public:
      *
      * @return ActionResult of request.
      */
-    ActionResult goto_location(double latitude_deg, double longitude_deg, float altitude_amsl_m, float yaw_deg);
+    ActionResult
+    goto_location(double latitude_deg, double longitude_deg, float altitude_amsl_m, float yaw_deg);
 
     /**
      * @brief Send command to transition the drone to fixedwing.
--- a/plugins/action/action_impl.h
+++ b/plugins/action/action_impl.h
@@ -26,7 +26,8 @@ public:
     ActionResult takeoff() const;
     ActionResult land() const;
     ActionResult return_to_launch() const;
-    ActionResult goto_location(double latitude_deg, double longitude_deg, float altitude_amsl_m, float yaw_deg);
+    ActionResult
+    goto_location(double latitude_deg, double longitude_deg, float altitude_amsl_m, float yaw_deg);
     ActionResult transition_to_fixedwing() const;
     ActionResult transition_to_multicopter() const;
 


plugins/action/include/plugins/action/action.h
```diff
@@ -106,6 +106,18 @@ class Action : public PluginBase {
      */
     ActionResult return_to_launch() const;
 
+    /**
+     * @brief Send command to reposition the vehicle to a specific WGS84 global position
+     *
+     * This sends the vehicle to a specified lattitude/longitude/altitude coordinates.
+     * @param latitude deg * 1E7
```
TSC21: Why not pass the latitude and longitude as normal coordinates and then multiply by `1E7` on the `ActionResult` code?

mzahana: Possible. I guess that requires to use `double` instead of `float` and then multiply by 1E7, then cast to float when passing to `command.params`, no?

hamishwillee:
How does this differ from the intention of [offboard control](https://sdk.dronecode.org/en/guide/offboard.html)? ie should this be the unimplemented position setting part of offboard control?

mzahana:
@hamishwillee Offboard does not have GPS setpoints. It's in local frame. So, conversions need to happen. Also, the flight controller already provides this functionality  without much coding.

plugins/action/include/plugins/action/action.h
mzahana: lat/long can be entered in degrees now.

plugins/action/action_impl.cpp
```diff
@@ -133,17 +133,20 @@ ActionResult ActionImpl::return_to_launch() const
         _parent->set_flight_mode(SystemImpl::FlightMode::RETURN_TO_LAUNCH));
 }
 
-ActionResult ActionImpl::goto_location(float latitude, float longitude, float altitude)
+ActionResult ActionImpl::goto_location(double latitude, double longitude, double altitude)
 {
+    float lat1e7 = (float) (latitude * 1E7);
```
TSC21: you don't need this. You can use a floating point literal:
```c++
ActionResult ActionImpl::goto_location(float latitude, float longitude, float altitude)
{
        ....
        command.params.param5 = latitude * 1E7F;
        command.params.param6 =  longitude * 1E7F;
        ....
}
```

mhkabir:
It would be more correct to use double precision for lat/lon.

TSC21:
> It would be more correct to use double precision for lat/lon.

@mhkabir that's only possible in [`COMMAND_INT`](https://mavlink.io/en/messages/common.html#COMMAND_INT). This is using [`COMMAND_LONG`](https://mavlink.io/en/messages/common.html#COMMAND_LONG).

plugins/action/include/plugins/action/action.h
```diff
@@ -106,6 +106,18 @@ class Action : public PluginBase {
      */
     ActionResult return_to_launch() const;
 
+    /**
+     * @brief Send command to reposition the vehicle to a specific WGS84 global position
+     *
+     * This sends the vehicle to a specified lattitude/longitude/altitude coordinates.
+     * @param latitude in degrees
+     * @param longitude in degrees
+     * @param altitude Altitude AMSL (meters)
```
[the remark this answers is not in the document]
TSC21: Here defined: https://github.com/PX4/Firmware/blob/bgat-master/src/modules/navigator/navigator_main.cpp#L414

mzahana:
@julianoes I am not sure if PX4 uses `COMMAND_INT` to parse `MAV_CMD_DO_REPOSITION`.

So, in summary, what is required to merger this PR?

plugins/action/action_impl.cpp
mzahana: if I keep `latitude` and `longitude` as `float`, I get warnings when I enter big values like `48.1234567` for loosing precision.

mzahana:
So, this is what I have so far.
```bash
ActionResult ActionImpl::goto_location(double latitude_deg, double longitude_deg, float altitude_amsl_m)
{

    MAVLinkCommands::CommandLong command{};

    command.command = MAV_CMD_DO_REPOSITION;
    command.target_component_id = _parent->get_autopilot_id();
    command.params.param1 = -1; /* default speed */
    command.params.param2 = MAV_DO_REPOSITION_FLAGS_CHANGE_MODE;
    // command.param4 = NaN /* Yaw Unchanged */
    command.params.param5 = (float)latitude_deg * 1E7F;
    command.params.param6 = (float)longitude_deg * 1E7F;
    command.params.param7 = altitude_amsl_m;

    return action_result_from_command_result(_parent->send_command(command));
}
```

Let me know your feedback.

julianoes:
@mzahana it looks to me like command int should be supported:
https://github.com/PX4/Firmware/blob/96995852cb6a2e2e23894c7ea05dd8385c595172/src/modules/navigator/navigator_main.cpp#L349-L351

plugins/action/action_impl.cpp
```diff
@@ -133,6 +133,23 @@ ActionResult ActionImpl::return_to_launch() const
         _parent->set_flight_mode(SystemImpl::FlightMode::RETURN_TO_LAUNCH));
 }
 
+ActionResult ActionImpl::goto_location(double latitude_deg, double longitude_deg, float altitude_amsl_m)
+{
+
+    MAVLinkCommands::CommandLong command{};
```
julianoes: I suggest you change it to `CommandInt` and try it in SITL.

mzahana:
@julianoes `CommandInt` is used and tested in SITL. Works fine.

plugins/action/action.cpp
```diff
@@ -32,6 +32,11 @@ ActionResult Action::land() const
     return _impl->land();
 }
 
+ActionResult Action::goto_location(double latitude_deg, double longitude_deg, float altitude_amsl_m, float yaw_rad)
```
julianoes: Please expose yaw in degrees and convert it to rad internally.

plugins/action/include/plugins/action/action.h
```diff
@@ -106,6 +106,19 @@ class Action : public PluginBase {
      */
     ActionResult return_to_launch() const;
 
+    /**
+     * @brief Send command to reposition the vehicle to a specific WGS84 global position
+     *
+     * This sends the vehicle to a specified lattitude/longitude/altitude coordinates.
+     * @param latitude_deg Latitude in degrees
+     * @param longitude_deg Longitude in degrees
+     * @param altitude_amsl_m Altitude AMSL in meters
```
julianoes: Are you sure that this command is in absolute altitude and not relative? Could it be that we need to specify the frame when sending the command int?

plugins/action/action.cpp
mzahana: Done.

plugins/action/include/plugins/action/action.h
mzahana: Also, I think [MAV_FRAME_GLOBAL_RELATIVE_ALT_INT](https://mavlink.io/en/messages/common.html#MAV_FRAME_GLOBAL_RELATIVE_ALT_INT) should be used in (https://github.com/Dronecode/DronecodeSDK/blob/27027d4eae32267616d6016cfebdcd13b0703719/core/mavlink_commands.h#L35)

mzahana: My guess is yes. It needs to be fixed on the PX4 side. I just have not figured out yet where it is exactly consumed.

julianoes: You're right, I filed an issue: https://github.com/PX4/Firmware/issues/10246.

julianoes:
@mzahana would be good if you could rebase this on develop and force push it.

plugins/action/include/plugins/action/action.h
julianoes: @mzahana oh yes that should be `_INT`, thanks for the note!

julianoes:
@mzahana I hope you don't mind that I rebased and force pushed your branch. I also fixed the style issues and added an integration test.

julianoes:
@TSC21 are you ok if we merge this for now, and then follow up with the firmware issues.

plugins/action/include/plugins/action/action.h
```diff
@@ -106,6 +106,20 @@ class Action : public PluginBase {
      */
     ActionResult return_to_launch() const;
 
+    /**
+     * @brief Send command to reposition the vehicle to a specific WGS84 global position
+     *
+     * This sends the vehicle to a specified lattitude/longitude/altitude coordinates.
+     * @param latitude_deg Latitude in degrees
```
hamishwillee: FYI, I wouldn't necessarily change this now, but one of the benefits of naming with units is that you don't necessarily have to spell everything out in the description. So normally I'd just have `Latitude.` (and yes, arguably you don't even need that, but good to have something because it shows that you have thought about docs).

PR #500, opened 2018-08-19
Fix embedding

julianoes:
Just two small fixes.


CMakeLists.txt
```diff
@@ -16,6 +16,7 @@ include(cmake/tinyxml2.cmake)
 include(cmake/zlib.cmake)
 
 if(NOT MSVC)
+    cmake_policy(SET CMP0058 NEW)
```
JonasVautherin: Do we use the `BYPRODUCT` anywhere?

JonasVautherin:
Why is `embed_files.cmake` not necessary anymore?

julianoes:
> Why is embed_files.cmake not necessary anymore?

This was just a leftover from an earlier try.

CMakeLists.txt
julianoes: No idea. It's somewhere in the cpp_rsc cmake code.

PR #501, opened 2018-08-19
[WIP/draft] Bring Python wrapper using pybind11 back

julianoes:
This is another try with pybind11 and based on the work from  #37.


hamishwillee:
> I do like the gRPC Python wrapper as well but it just seems like such a crazy undertaking.

Why crazy - or crazy "just for Python". I'm biased, but from what I can see the end result of the gRPC/Rx pipeline should be very consistent - which is a happy thing for a tech writer.

julianoes:
This is now replaced with https://github.com/Dronecode/DronecodeSDK-Python

julianoes:
@JonasVautherin we might want to cherry-pick https://github.com/Dronecode/DronecodeSDK/pull/501/commits/164801a275be8deb25c98bdd145d572a3f9f7e0d

JonasVautherin:
Yep that would be fine for me!

PR #504, opened 2018-08-21
backend: bump version to 0.1.0

JonasVautherin:


julianoes:
Sorry, I didn't realize I needed to review and merge this.

I think for something like this that needs to go out the door as soon as possible, it makes sense for you (as an admin) to just merge it.

PR #508, opened 2018-08-23
Another try to fix mission progress crash

julianoes:
The filtering was done to early and meant that we would send mission progress with equal values in rapid succession. Presumably this triggered a crash in the backend.

The error in GRPC was: `GRPC_CALL_ERROR_TOO_MANY_OPERATIONS`.


julianoes:
Already built and uploaded to s3, so time to merge this as well :smile: .

PR #510, opened 2018-08-29
mission: added RTL after a mission is complete

julianoes:
This adds the option to set/get whether RTL should automatically be executed at the end of a mission.


JonasVautherin:
Doc generation is failing, I'll fix that.

julianoes:
Thx!

PR #515, opened 2018-08-29
camera: notify after subscription to settings

julianoes:
It would be handy to get an update right after subscribing.


JonasVautherin:
> It would be handy to get an update right after subscribing.

Makes me feel like I already reviewed that :sweat_smile:... Deja vu?

julianoes:
Not sure :smile:

PR #527, opened 2018-09-01
backend: fix crash in SubscribeCameraStatus

JonasVautherin:
Tested by @douglaswsilva.

### DIFF
--- a/backend/src/plugins/camera/camera_service_impl.h
+++ b/backend/src/plugins/camera/camera_service_impl.h
@@ -705,7 +705,7 @@ public:
 
 private:
     Camera &_camera;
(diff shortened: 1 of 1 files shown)


backend/src/plugins/camera/camera_service_impl.h
```diff
@@ -704,6 +706,7 @@ class CameraServiceImpl final : public rpc::camera::CameraService::Service {
 
 private:
     Camera &_camera;
+    mutable std::mutex _subscribe_camera_status_mutex{};
```
julianoes: Why `mutable`? The method `SubscribeCameraStatus` is not `const`, or is it?

PR #530, opened 2018-09-03
Move SDK/backend version to version file

julianoes:
This adds a file version which contains the semantic version and copies it into source at build time.

Todo: automatically create `git tag` when version changes.


julianoes:
Created issues for todos: #541, #542.

PR #531, opened 2018-09-03
Fix race in camera backend

JonasVautherin:
You know what I mean.

### DIFF
--- a/backend/src/plugins/camera/camera_service_impl.h
+++ b/backend/src/plugins/camera/camera_service_impl.h
@@ -171,11 +171,12 @@ public:
 
         bool is_finished = false;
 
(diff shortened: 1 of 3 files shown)


backend/src/plugins/camera/camera_service_impl.h
```diff
@@ -626,6 +630,7 @@ class CameraServiceImpl final : public rpc::camera::CameraService::Service {
                     translateSettingOptions(setting_option, rpc_setting_option);
                 }
 
+                std::lock_guard<std::mutex> lock(_subscribe_camera_status_mutex);
```
julianoes: I guess you should rename the mutex at this point.

PR #539, opened 2018-09-13
Fix settings notification again

julianoes:
This resolves the issue where the current settings were not notified
after color mode was updated. This happened because color mode does
not lead to any other setting changes except the color mode itself.
In this case `get_unknown_params()` would not find any unknown params
but still return true. The right solution is to get rid of the confusing
return argument and just check the size of the params vector directly.

This is the correct fix instead of #524.


julianoes:
@JonasVautherin I'm gonna release the backend 0.2.5 and test it.

julianoes:
Tested, works.

Issue #254, opened 2018-02-08
Build in docker as user


julianoes:
Currently when we build in docker it uses root and therefore leaves build artefacts with root as the owner. This requires `sudo make distclean` to clean up afterwards outside of docker which is inconvenient.

As @hamishwillee suggests we should build something similar to:
https://dev.px4.io/en/test_and_ci/docker.html
https://github.com/PX4/containers/blob/master/docker/px4-dev/Dockerfile_base#L64

dagar:
Docker ENTRYPOINT with this script seems to work well.
https://github.com/PX4/containers/blob/master/docker/px4-dev/scripts/entrypoint.sh

Dockerfile

    # create and start as LOCAL_USER_ID
    COPY scripts/entrypoint.sh /usr/local/bin/entrypoint.sh
    ENTRYPOINT ["/usr/local/bin/entrypoint.sh"]

When running the container you need to pass - eLOCAL_USER_ID=$UID and it creates an internal user matching your UID. Otherwise it uses UID 9001, which is still better than root.

julianoes:
@dagar last time I tried to use that it did not work for me, I remember. I'll have to try it again.

PR #549, opened 2018-09-19
Docker as user

dbowerman:
Fix issue #254. Run docker build environment as normal user. Add README to docker directory explaining how to create and run. Add convenience script to root directory.


docker/Dockerfile-Ubuntu-18.04
```diff
@@ -38,6 +38,10 @@ RUN update-alternatives --install /usr/bin/clang-format clang-format /usr/bin/cl
...
 RUN gem install --no-ri --no-rdoc fpm;
 
+RUN useradd -d /home/dronecode -m -s /bin/bash dronecode
+
+USER dronecode
+
 CMD ["/bin/bash"]
 
-WORKDIR "/root/DronecodeSDK"
+WORKDIR "/home/dronecode/DronecodeSDK"
```
julianoes: I guess that should work.

docker/README
```diff
@@ -0,0 +1,20 @@
+These Dockerfiles will create a DronecodeSDK development container image based on whichever flavour of Ubuntu or Fedora you prefer.
```
julianoes: Would be nice to have a markdown file README.md.

docker/README
```diff
@@ -0,0 +1,20 @@
...
+
+To create an image :-
+
+cd /wherever/DronecodeSDK/docker
```
julianoes: And then put code blocks into ```.

docker/README
```diff
@@ -0,0 +1,20 @@
...
+
+To use the image :-
+
+cd /wherever/DronecodeSDK
+./run-docker
+
+This will create and run a container based on your image.
+/path/DronecodeSDK will be mounted inside the container at /home/dronecode/DronecodeSDK
+
+dronecode@dronecode-docker:~/DronecodeSDK$ ls
```
julianoes: This `ls` output will be out of sync quickly, I'd rather not commit it here.

run-docker.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
```
julianoes: Nitpick I suggest `#!/usr/bin/env sh`

run-docker.sh
```diff
@@ -0,0 +1,3 @@
...
+
+docker run -it --rm --hostname dronecode-docker -v $(pwd):/home/dronecode/DronecodeSDK dronecode
```
julianoes: Do we need the hostname? And does that not conflict if you have multiple on the same machine, e.g. when testing with the different containers?

docker/Dockerfile-Ubuntu-18.04
dbowerman: Yes it does. You are running as user dronecode inside the container so best to have the workdir in the users home directory rather than roots. Also allows for further enhancement to say mount an application at say /home/dronecode/app and have the sdk and application external to the container.

docker/README
dbowerman: ok will fix.

docker/README
dbowerman: ok

docker/README
dbowerman: ok

run-docker.sh
dbowerman: sure

run-docker.sh
dbowerman: its not necessary. leaving it out means a bash prompt of dronecode@<randomstringhere>. dronecode@dronecode-docker was a bit more intuitive.

docker/README.md
```diff
@@ -0,0 +1,24 @@
+## Building and using Docker development images.
+
+These Dockerfiles will create a DronecodeSDK development container image based on whichever flavour of Ubuntu or Fedora you prefer.
+
+To create an image :-
```
julianoes: `:-` -> `:`

docker/README.md
```diff
@@ -0,0 +1,24 @@
...
+```bash
+cd /whereever/DronecodeSDK/docker
+docker build -t dronecode -f Dockerfile-Ubuntu-18.04 .
+```
+
+To use the image :-
+
+```bash
+cd /whereever/DronecodeSDK
+./run-docker
```
julianoes: `run-docker` -> `run-docker.sh`

docker/README.md
```diff
@@ -0,0 +1,24 @@
+## Building and using Docker development images.
+
+These Dockerfiles will create a DronecodeSDK development container image based on whichever flavour of Ubuntu or Fedora you prefer.
+
+To create an image :-
+
+```bash
+cd /whereever/DronecodeSDK/docker
+docker build -t dronecode -f Dockerfile-Ubuntu-18.04 .
```
julianoes: If you tag it to `dronecode/dronecode-sdk-ubuntu-18.04`, you don't actually need to build this yourself and can just pull it from dockerhub:
https://hub.docker.com/r/dronecode/dronecode-sdk-ubuntu-18.04/

I suggest also to use this tag in the run_docker.sh script. It makes it easier, I think.

docker/README.md
dbowerman: oops

docker/README.md
dbowerman: Are you saying that the docker build is not needed? is ci building and pushing to docker hub?

julianoes: I'm currently doing it manually using: https://github.com/Dronecode/DronecodeSDK/blob/develop/docker/build_and_push_docker_images.sh.

dbowerman: ah ok, so do you want to leave the reference to building it locally (with the tag changed) in the README in case someone wants to build locally? or remove altogether?

JonasVautherin:
Thanks @dbowerman!

PR #550, opened 2018-09-19
fix SIGSEGV in is_connected() when no system is connected

gamoreno:
Fixes an issue in which if an application (e.g., offboard_velocity) is started
before the simulator is started, a segmentation fault is caused because it
tries to dereference a null pointer.

Signed-off-by: Gabriel Moreno <gabrielm@cs.cmu.edu>

### DIFF
--- a/core/dronecode_sdk_impl.cpp
+++ b/core/dronecode_sdk_impl.cpp
@@ -255,6 +255,10 @@ bool DronecodeSDKImpl::is_connected() const
 {
     std::lock_guard<std::recursive_mutex> lock(_systems_mutex);
 
+    if (_systems.empty()) {
+        return false;
+    }
+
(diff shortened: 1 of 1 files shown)


julianoes:
I'm not sure why jenkins chas somehow changed from branch to pr-head again, I'll ignore it for now. @dagar thoughts?

dagar:
You're probably only seeing branches for PRs that have a branch off the main repo.
It's a setting here. http://ci.px4.io:8080/job/Dronecode/configure

julianoes:
Thanks for your response @dagar. It looks like the Jenkins settings have changed quite a bit as well, maybe that's why it stopped working. Or maybe the difference is that this comes from a fork. I'll play around and see.

PR #552, opened 2018-09-20
mission: hack to set gimbal angle absolute for now

julianoes:
This is a hack to hardcode the gimbal yaw angle to absolute for now,
basically custom for the 3DR and H520 integration.


julianoes:
> Should we open an issue for the clean fix?

Done

PR #555, opened 2018-09-22
Try to prevent command lockup

julianoes:
This is an attempt to fix a lockup issue that we see in the Swift
example app when the camera capture status subscription is used.
Presumably, the lockup has nothing to do with the actual camera capture
status but it just leads to the right timing to trigger something else.

What happens is that `receive_timeout()` is called but then the command
queue turns out to be empty (returns `nullptr`). After that, no more
commands can be processed and we keep getting the warning:
"Command ack not matching our current command: 257".

This makes sense because we stop doing work, so send any commands as
long as `_state` is `State::WAITING` and there is no other timeout that
would reset this.

Therefore, this change should resolve that case by reseting `_state`, so
that we can keep going.

However, this change does not fix or explain why the queue is empty in
the first place and out of sync with `_state`.


julianoes:
I tested this and it doesn't break the integration test, so I suggest it's worth a try in the iOS example.

core/mavlink_commands.cpp
```diff
@@ -233,7 +231,15 @@ void MAVLinkCommands::receive_timeout()
     auto work = _work_queue.borrow_front();
 
     if (!work) {
-        // Nevermind, there is nothing to do.
+        // FIXME:
+        // There is nothing to do.
+        // However, we probably shouldn't be in State::WAITING if nothing is in the queue.
+        // Otherwise, we might wait forever because we don't register a new timeout.
```
JonasVautherin: And you observed that we do end here, right? I reviewed the code quickly, I still don't have a full understanding but from what I've seen, it's clear that whenever you set the state to `WAITING`, you register a timeout (which should _always_ be inserted into a map (i.e. I don't see why this insertion would fail)).

The thing I don't understand yet is what `_work_queue.return_front()` does, but I'll have a closer look.

core/mavlink_commands.cpp
```diff
@@ -149,55 +147,51 @@ void MAVLinkCommands::receive_command_ack(mavlink_message_t message)
         case MAV_RESULT_ACCEPTED:
             _state = State::NONE;
             _state_mutex.unlock();
-            call_callback(work->callback, Result::SUCCESS, 1.0f);
-            _work_queue.pop_front();
             _parent.unregister_timeout_handler(_timeout_cookie);
+            _work_queue.pop_front();
```
JonasVautherin: What makes you think that those two calls should be inverted? Just to be sure I understand it :-).

core/mavlink_commands.cpp
```diff
@@ -210,6 +204,10 @@ void MAVLinkCommands::receive_command_ack(mavlink_message_t message)
                                              work->retries_to_do * work->timeout_s,
                                              &_timeout_cookie);
             _work_queue.return_front();
+            // FIXME: We can only call callbacks with promises once, so let's not do it
```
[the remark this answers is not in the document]
julianoes: That's true actually, unless you want progress feedback to the user, e.g. for a UI.

core/mavlink_commands.cpp
julianoes: Yes, I observed that based on printfs. And once this happened, it was locked up which makes sense because now we're in `WAITING` state when really there is nothing going on and we have to reset to `NONE.

Make sure to also read the commit messages, it might make it clearer.

Correct I'm assuming the registered timeout is always triggered (correctly).

`borrow` leaves the lock locked, so then you need to return it again, so it's unlocked.

core/mavlink_commands.cpp
julianoes: See commit message but basically you could have popped the work item from the queue and then the timeout still triggers and wants to access this work item.

core/mavlink_commands.cpp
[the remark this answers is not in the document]
JonasVautherin: Oh right, that's at the lower level, and some `IN_PROGRESS` are actually useful (e.g. for the calibration plugin).

PR #556, opened 2018-09-23
Fix #547 Add callback for components discovered.

### LINKED ISSUE #547: Add callbacks for components discovered
It would be handy to have callbacks when a component is added, so you could e.g. just wait until a camera is found before trying to interact with it.

Requested in http://discuss.px4.io/t/subscribe-to-a-callback-for-components-being-added/7843

dbowerman:
Add the ability to register a callback so the application can be notified when individual components are added.
See example/takeoff_land for usage example.


core/system_impl.cpp
```diff
@@ -341,6 +347,11 @@ size_t SystemImpl::total_components() const
     return _components.size();
 }
 
+void SystemImpl::register_component_discovered_callback(discover_callback_t callback)
```
JonasVautherin: What happens if the callback is registered after the component is discovered?

dbowerman: In the current implementation the callback won't be called in that case.

JonasVautherin: To be consistent with the system discovery, shouldn't we add that here, too? Or is it too complex because we then need to detect when a component times out?

dbowerman: Only if the component timeout before the callback is registered. I'll take another look at it.

JonasVautherin: In my understanding, [`auto res_pair = _components.insert(component_id);`](https://github.com/Dronecode/DronecodeSDK/pull/556/files#diff-780b2b438d1b4c2c3d32e3bc7b4f1d15R335) will be `true` only when a new component is inserted. And at that very moment, if a callback is registered (i.e. [`component_discovered_callback != nullptr`](https://github.com/Dronecode/DronecodeSDK/pull/556/files#diff-780b2b438d1b4c2c3d32e3bc7b4f1d15R337), then it will be called.

Which means that if events happen in this order:

1. User registers the callback
2. Component is discovered
3. Callback is called

Then everything is fine. But if the component has already been discovered when the user registers the callback, i.e.:

1. Component is discovered
2. Callback is `nullptr`, so it is not called
3. User registers the callback

Then the user just "missed" the callback, as the component will not be discovered again. However, I believe that in this scenario, the user actually wants to know that a component has been discovered, right?

dbowerman: Yes, that could be true. In my case I know that I want to use the component discovery feature, so I place my callback registration appropriately. i.e. very early as in takeoff_land example.
I can see however that someone may use it differently.

core/include/system.h
```diff
@@ -76,6 +82,14 @@ class System {
      */
     uint64_t get_uuid() const;
 
+    /**
+     * @brief Register a callback to be called when a component is discovered.
+     *
+     * @param callback a function of type void(uint8_t) which will be called with the COMPONENT_ID
```
julianoes: I would prefer if we don't "leak" the mavlink component ID outside of the SDK but only the notion of autopilot/camera/gimbal. Therefore, in my opinion, it should be something like `register_camera_discovered()`.

JonasVautherin: Is there a convention on how components should be called? I mean how do you know, from MAVLink, that it is a camera?

julianoes: The camera has `MAV_COMP_ID_CAMERA[x]`, it's something known to the SDK internals.

https://mavlink.io/en/messages/common.html#MAV_COMP_ID_CAMERA

core/system_impl.cpp
dbowerman: @JonasVautherin I've just push an update to fix this.

core/include/system.h
dbowerman: @julianoes Can we declare an enum like ::Camera ::Gimbal ::AutoPilot ::Unknown etc (maybe this already exists and pass that to the callback? that way there'd only be one callback function instead of one function per component.

julianoes: Yes an enum for components makes sense, I agree.

dbowerman: @julianoes change pushed.

core/include/system.h
```diff
@@ -1,9 +1,20 @@
...
 
+/**
+ * @brief Component Types
+ */
+enum ComponentType { UNKNOWN = 0, AUTOPILOT, CAMERA, GIMBAL };
+
+/**
+ * @brief type for component discovery callback
+ */
+typedef std::function<void(ComponentType)> discover_callback_t;
```
julianoes: Yes, I like it.

core/include/system.h
julianoes: Cool, the comment needs to be adapted though.

core/system_impl.cpp
```diff
@@ -341,6 +366,37 @@ size_t SystemImpl::total_components() const
...
+        ComponentType component_type;
+        for (const auto &elem : _components) {
+            switch (elem) {
+                case MAV_COMP_ID_AUTOPILOT1:
+                    component_type = AUTOPILOT;
+                    break;
+                case MAV_COMP_ID_CAMERA:
+                case MAV_COMP_ID_CAMERA2:
+                case MAV_COMP_ID_CAMERA3:
+                case MAV_COMP_ID_CAMERA4:
```
julianoes: Next time we'll have to refactor this switch case into a function.

core/system_impl.cpp
```diff
@@ -332,6 +334,29 @@ void SystemImpl::add_new_component(uint8_t component_id)
 {
     auto res_pair = _components.insert(component_id);
     if (res_pair.second) {
+        if (component_discovered_callback != nullptr) {
+            ComponentType component_type;
```
JonasVautherin: Small readability thing: I would put that in a function, e.g. `getComponentType(const uint8_t component_id)` (and here do `const auto component_type = getComponentType(component_id)`. What do you think?

core/system_impl.cpp
JonasVautherin: Thanks, looking at it!

core/system_impl.cpp
JonasVautherin: Oh, just seing it after my comment above. So that is duplicated from above, and should become something like `const auto component_type = getComponentType(component_id);`.

dbowerman:
@julianoes @JonasVautherin comments addressed :)

core/system_impl.cpp
```diff
@@ -328,10 +330,33 @@ std::string SystemImpl::component_name(uint8_t component_id)
...
+        case MAV_COMP_ID_CAMERA4:
+        case MAV_COMP_ID_CAMERA5:
+        case MAV_COMP_ID_CAMERA6:
+            return CAMERA;
+        case MAV_COMP_ID_GIMBAL:
+            return GIMBAL;
+        default:
+            return UNKNOWN;
+    }
+}
```
julianoes: Awesome!

JonasVautherin:
Testing right now, will merge ASAP.

PR #557, opened 2018-09-23
Add sudo to Docker development image.

dbowerman:
Add sudo so that root commands can be used now docker image defaults to normal user.

Davids-MacBook-Pro:DronecodeSDK dave$ ./run-docker.sh
dronecode@host:~/DronecodeSDK$ make default

dronecode@host:~/DronecodeSDK$ sudo make default install

dronecode@host:~/DronecodeSDK$ sudo ldconfig

dronecode@host:~/DronecodeSDK$ cd example/takeoff_land/build

dronecode@host:~/DronecodeSDK/example/takeoff_land/build$ make

PR #558, opened 2018-09-24
Jenkinsfile: update to latest docker tag

julianoes:
This includes sudo.

This is a follow up to #557.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -105,20 +105,20 @@ script:
     make BUILD_TYPE=Coverage run_unit_tests -j4;
   fi
 - if [[ "${BUILD_TARGET}" = "docker_build_ubuntu_16.04" ]]; then
-    docker run -it --rm -v $TRAVIS_BUILD_DIR:/root/DronecodeSDK:rw dronecode/dronecode-sdk-ubuntu-16.04 ./travis-docker-build.sh;
-    docker run -it --rm -v $TRAVIS_BUILD_DIR:/root/DronecodeSDK:rw dronecode/dronecode-sdk-ubuntu-16.04 ./create_packages.sh;
+    docker run -it --rm -v $TRAVIS_BUILD_DIR:/home/dronecode/DronecodeSDK:rw dronecode/dronecode-sdk-ubuntu-16.04 ./travis-docker-build.sh;
+    docker run -it --rm -v $TRAVIS_BUILD_DIR:/home/dronecode/DronecodeSDK:rw dronecode/dronecode-sdk-ubuntu-16.04 ./create_packages.sh;
   fi
 - if [[ "${BUILD_TARGET}" = "docker_build_ubuntu_18.04" ]]; then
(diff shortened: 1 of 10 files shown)


julianoes:
@dbowerman do you know why travis-ci ends up without permission:
https://travis-ci.org/Dronecode/DronecodeSDK/jobs/432376354#L575-L577

dbowerman:
@julianoes only that somehow $TRAVIS_BUILD_DIR at that point is not valid. ie the container mounted some dud directory.

JonasVautherin:
@julianoes: About the examples, I believe that it should be possible to build them without `sudo`:

1. Install the SDK locally (using `INSTALL_PREFIX`).
2. Build examples with `CPLUS_INCLUDE_PATH=/path/to/installed/include` and `LIBRARY_PATH=/path/to/installed/lib`.
3. Run the examples (possibly with `LD_LIBRARY_PATH`, but I'm not even sure that's necessary here).

julianoes:
@JonasVautherin yes that would work but it means the setup is a bit different from the real use case but I guess that's ok.

dbowerman:
Is the issue the 'user' option being passed to docker. ie. -u `stat -c "%u:%g" $TRAVIS_BUILD_DIR` ?
Inside the container the users are 'dronecode' and 'root' or more correctly 'dronecode' and 'dronecode using sudo'.

I reckon remove the -u option.

julianoes:
I don't understand the error that now happens on Fedora:
```
dirname: missing operand
Try 'dirname --help' for more information.
/bin/sh: line 0: cd: build/default: Not a directory
Build dir: build/default
/bin/sh: line 0: cd: build/default: Not a directory
```

julianoes:
@dbowerman the problem is that the solution with the default `dronecode` user only works if you're using docker as the user with ID 1000. For everything else you seem to run into troubles.

In my opinion the solution before with just root just worked even though it would sometimes require a `sudo make clean` to clean up.

dbowerman:
@julianoes Ah yes. thinking about it, that makes sense. I didn't pick that because I'm the only user (usually uid 1000) on my systems. Give me a day or so to see if theres a solution, otherwise revert I guess.

dbowerman:
@julianoes this article describes the exact problem and offers a solution.
https://denibertovic.com/posts/handling-permissions-with-docker-volumes/
Do you want to go down that path? or revert?

julianoes:
@dbowerman oh yes nice, that should work.

julianoes:
@dbowerman hm, so I tried this and got:
```
Starting with UID : 9001
useradd: Permission denied.
useradd: cannot lock /etc/passwd; try again later.
su: user user does not exist
```

julianoes:
Note: I'm trying to come by without `gosu` for now, we'll see if that's good enough.

JonasVautherin:
What's the status on that? I believe it is currently breaking the travis build on all the PRs, which is a bit blocking :/.

julianoes:
Forgot to update the docker images, re-running travis.

julianoes:
Yay, it finally passed. @JonasVautherin please review and merge, so CI is fixed.

JonasVautherin:
Thanks!

PR #560, opened 2018-09-26
calibration: return only one struct (and the result)

JonasVautherin:
Because of the Rx language bindings, we cannot have more than one return type (that is, one `Result` that says when there is an error and when the stream completes, and one actual struct containing data). This has to be reflected in the C++ API if we want to auto-generate those headers, too.

Moreover, some languages don't have the notion of `NaN` and default to `0` (which is an issue for e.g. a `progress`). For this reason, the resulting struct has booleans to explicitly say if a value is defined or not.

Finally, I tried to remove some code duplication (see [here](https://github.com/Dronecode/DronecodeSDK/compare/calibration-callback-one-result?expand=1#diff-e2a195cc797105ca574a291666fd69ebR77)). It should behave the same in my understanding.

### DIFF
--- a/integration_tests/calibration.cpp
+++ b/integration_tests/calibration.cpp
@@ -7,7 +7,7 @@
 using namespace dronecode_sdk;
 using namespace std::placeholders; // for `_1`
 
-static void receive_calibration_callback(const Calibration::Result &result,
+static void receive_calibration_callback(const Calibration::Result result,
                                          const Calibration::ProgressData &progress_data,
                                          const std::string &calibration_type,
                                          bool &done);
@@ -108,7 +108,7 @@ TEST(HardwareTest, CalibrationGimbalAccelerometer)
     }
 }
 
-void receive_calibration_callback(const Calibration::Result &result,
+void receive_calibration_callback(const Calibration::Result result,
                                   const Calibration::ProgressData &progress_data,
                                   const std::string &calibration_type,
                                   bool &done)
--- a/integration_tests/calibration.cpp
+++ b/integration_tests/calibration.cpp
@@ -7,10 +7,9 @@
 using namespace dronecode_sdk;
 using namespace std::placeholders; // for `_1`
 
-static void receive_calibration_callback(Calibration::Result result,
-                                         float progress,
-                                         const std::string text,
-                                         const std::string calibration_type,
+static void receive_calibration_callback(const Calibration::Result &result,
+                                         const Calibration::ProgressData &progress_data,
+                                         const std::string &calibration_type,
                                          bool &done);
 
 TEST(HardwareTest, CalibrationGyro)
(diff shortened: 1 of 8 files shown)


integration_tests/calibration.cpp
```diff
@@ -7,9 +7,8 @@
 using namespace dronecode_sdk;
 using namespace std::placeholders; // for `_1`
 
-static void receive_calibration_callback(Calibration::Result result,
-                                         float progress,
-                                         const std::string text,
+static void receive_calibration_callback(const Calibration::Result result,
+                                         const Calibration::ProgressData &progress_data,
```
julianoes: Why are some references and some not?

plugins/calibration/include/plugins/calibration/calibration.h
```diff
@@ -33,7 +35,7 @@ class Calibration : public PluginBase {
     ~Calibration();
 
     /**
-     * @brief Possible results returned for camera commands.
+     * @brief Possible results returned for calibration commands.
```
julianoes: Oops.

plugins/calibration/include/plugins/calibration/calibration.h
```diff
@@ -49,6 +51,33 @@ class Calibration : public PluginBase {
...
+        std::string status_text;
+
+        ProgressData(const bool _has_progress,
+                     const float _progress,
+                     const bool _has_status_text,
+                     const std::string &_status_text) :
+            has_progress(_has_progress),
+            progress(_progress),
+            has_status_text(_has_status_text),
+            status_text(_status_text)
```
julianoes: I think that's fine.

integration_tests/calibration.cpp
JonasVautherin: Good point, I guess they should all be references... unless you have a different opinion?

julianoes: I would say normal by-value for all.

JonasVautherin: Disagreed, [see here](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f16-for-in-parameters-pass-cheaply-copied-types-by-value-and-others-by-reference-to-const): "For “in” parameters, pass cheaply-copied types by value and others by reference to const".

julianoes: Right, unless you use `std::move`.

JonasVautherin: I'm not sure what you mean. So you make the function take them by value, and assume that the caller moves them? Or you take them by value, but move them in the implementation of the function?

In any case I am not sure it behaves exactly the same as a const reference, does it? And a const ref is actually extremely clear on the intention ("I will just read that thing").

julianoes: Ok, you're right. So I vote `Calibration::Result result` and `Calibration::ProgressData &progress_data`.

JonasVautherin: Deal. Done :laughing:

julianoes: Oh I now realize this is just the callback used in the integration test :man_facepalming:.

PR #561, opened 2018-09-26
backend: add calibration plugin

JonasVautherin:
@julianoes @thomasgubler: just FYI, as `calibration_service_impl.h` is the first file that has been autogenerated (it did not require any manual editing).

I'm not generating the unit tests yet, so I'll have to test that manually.


backend/src/plugins/calibration/calibration_service_impl.h
```diff
@@ -0,0 +1,167 @@
...
+                std::lock_guard<std::mutex> lock(_subscribe_mutex);
+                if (!writer->Write(rpc_response) && !is_finished) {
+                    is_finished = true;
+                    stream_closed_promise.set_value();
+                }
+            });
+
+        stream_closed_future.wait();
+        return grpc::Status::OK;
+    }
```
julianoes: Looks all correct!

PR #562, opened 2018-09-30
Request Camera Info

dbowerman:
Add Camera::request_info()

Currently the request camera info Mavlink message is sent in the CameraImpl constructor only.
This means you need to get the timing right (using manual delays) before trying to use camera options as the camera definition file may or may not be loaded.

Adding this function means you can subscribe to component discovered callback, if the component is the camera, call request_info() triggering downstream handling of available camera options etc.


julianoes:
Hm, fair point. Every plugin has an `enable()` and `disable` method to be run when a system is connected. Presumably there should be an option to wait for a specific component, e.g. the camera for the camera plugin.

julianoes:
@dbowerman good question, not sure. I just looked at the diff and saw that it has some commits from your other PR? Could you clean that up and then force push?

JonasVautherin:
Yeah sorry, I actually forgot it. Good point that you pinged us :+1:.

What is the difference between instantiating the camera object after the component is discovered, and instantiating the camera object before but calling `request_info()` after it is discovered? [Looking at the code](https://github.com/Dronecode/DronecodeSDK/blob/develop/plugins/camera/camera_impl.cpp#L85), it seems to me that it is exactly the same. Which would mean we can already achieve that with the current implementation. Is that correct?

My concern is that this way, if we ended up adding more stuff in the `init()` of `camera_impl.cpp`, then people doing the "manual init" proposed in this PR would need to add that to their code. Whereas if they wait for the camera to be discovered before they instantiate the plugin, then they would be fine.

What do you think?

dbowerman:
@JonasVautherin The problem is that currently the mavlink message to request the camera info is *only* sent in the Camera:: constructor. There is no other way to cause it to be sent (that I can see).
This means you need to put delays in your code "sleep(n)" to ensure that the camera component has been discovered before the camera object is instantiated otherwise you will never be able to use camera settings etc.

I'm working in iOS and its not good to have "delays" at all. I want to be able to use component discovery to be notified when the camera component is discovered, then send the request info mavlink message knowing that it will succeed irrespective of when the camera object was instantiated.

hope this makes sense ? :)

JonasVautherin:
@dbowerman: I did not mean to necessarily close the PR, I just want to make sure we need it.

My concern is the following: we need language bindings on top of C++ (Swift, Python, Java), and we are working on autogenerating them (from the definitions [here](https://github.com/Dronecode/DronecodeSDK-Proto/tree/master/protos)). So there are tricks we could do in the C++ interface that may not work well in the other languages. Adding `request_info()` to the language bindings doesn't seem so elegant to me (even though it works in the C++ API).

I am definitely convinced that the API should not force you to `sleep` to get the timings right, but if you cannot do it with the current API, I feel like maybe there is another way to improve that in a nice way for the language bindings.

dbowerman:
@JonasVautherin Yes I agree completely. I'll work on achieving what I need with the current api.

PR #564, opened 2018-09-30
More docker follow-up

julianoes:
I forgot to add the local user ID in run-docker.sh.

And we should try to compile the examples in CI.


julianoes:
@dbowerman the trick with `-e LOCAL_USER_ID=`id -u` does not work with Jenkins, not sure why:
```
Failed to run image 'dronecode/dronecode-sdk-fedora-27:2018-09-28'. Error: docker: Error response from daemon: linux spec user: unable to find user `: no matching entries in passwd file.
```

dbowerman:
hmm

julianoes:
This is good to be merged.

PR #566, opened 2018-09-30
Some follow-me fixes

### LINKED ISSUE #408: Follow Me example behaviour has changed in latest firmware
Previously flew spirals as shown here: https://sdk.dronecode.org/en/examples/follow_me.html

Now it flies a straight line somewhat south, while spinning a bit: 
image

I tried this for gazebo and jmavsim using current firmware and DroneCore, on a real machine and in VM.

julianoes:
This fixes #408 and makes the example a bit easier to understand. The path is changed from a semi-circular path (which I had broken in the meantime) to just a square. Also, we're now following from behind which is most intuitive to start with.

### DIFF
--- a/example/follow_me/fake_location_provider.cpp
+++ b/example/follow_me/fake_location_provider.cpp
@@ -37,39 +37,30 @@ void FakeLocationProvider::stop()
     }
 }
 
-// Rudimentary location provider whose successive lat, lon combination
-// makes Drone revolve in a semi-circular path.
+// Rudimentary location provider to draw a square.
 void FakeLocationProvider::compute_locations()
 {
     while (!should_exit_) {
-        if (count_++ < 10) {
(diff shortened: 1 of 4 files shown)


example/follow_me/fake_location_provider.cpp
```diff
@@ -37,39 +37,30 @@ void FakeLocationProvider::stop()
...
-        }
-        if (count_++ < MAX_LOCATIONS) {
-            location_callback_(latitude_deg_, longitude_deg_);
-            longitude_deg_ += LONGITUDE_DEG_PER_METER * 3;
+        } else {
+            // We're done.
+            should_exit_ = true;
         }
         sleep_for(seconds(1));
+        ++count_;
```
JonasVautherin: Yep!

PR #568, opened 2018-10-03
external_example: fix paths

julianoes:
This fixes the paths for me in a try according to:
https://sdk.dronecode.org/en/guide/sdk_extensions.html


hamishwillee:
I've been away a couple of days, so sorry for late response.
Anything which makes the docs correct (without me having to change them) makes me happy. 

The whole plugins and extensions sections haven't been maintained for a while because of the implementation churn. Would be worth your review when you have time.

PR #571, opened 2018-10-18
Fix landed detection

hdiethelm:
The examples report landed already before the vehicle is landed which is not so nice... ;-)

### DIFF
--- a/example/offboard_velocity/offboard_velocity.cpp
+++ b/example/offboard_velocity/offboard_velocity.cpp
@@ -238,7 +238,7 @@ int main(int argc, char **argv)
     action_error_exit(land_result, "Landing failed");
 
     // Check if vehicle is still in air
-    while (telemetry->in_air() == true) {
+    while (telemetry->in_air()) {
         std::cout << "Vehicle is landing..." << std::endl;
         sleep_for(seconds(1));
     }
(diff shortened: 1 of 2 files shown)


example/offboard_velocity/offboard_velocity.cpp
```diff
@@ -237,9 +237,16 @@ int main(int argc, char **argv)
     const ActionResult land_result = action->land();
     action_error_exit(land_result, "Landing failed");
 
+    // Check if vehicle is still in air
+    while (telemetry->in_air() == true) {
```
JonasVautherin: What about removing `== true`?

```suggestion
    while (telemetry->in_air()) {
```

example/takeoff_land/takeoff_and_land.cpp
```diff
@@ -128,8 +128,16 @@ int main(int argc, char **argv)
         return 1;
     }
 
+    // Check if vehicle is still in air
+    while (telemetry->in_air() == true) {
```
JonasVautherin: Same as above:

```suggestion
    while (telemetry->in_air()) {
```

example/offboard_velocity/offboard_velocity.cpp
hdiethelm: Copy paste from while (telemetry->health_all_ok() != true) frome above. Just remove it.

example/takeoff_land/takeoff_and_land.cpp
hdiethelm: Same as above

PR #578, opened 2018-10-26
Fix backend connection issues

julianoes:
This fixes a couple of connection issues:
1. We start the discovery before the vehicle is actually connected.
2. We miss the discovery because the GRPC server doesn't run yet when we try to register the discovery.
3. The camera tried to early (before being connected) to get the camera information.
4. A UDP_BRIDGE would mean that we always think the vehicle is connected, when we actually only see the bridge.


backend/src/backend_api.cpp
```diff
@@ -5,8 +5,8 @@
 void runBackend(const int mavlink_listen_port, void (*onServerStarted)(void *), void *context)
 {
     dronecode_sdk::backend::DronecodeSDKBackend backend;
-    backend.connect(mavlink_listen_port);
     backend.startGRPCServer();
+    backend.connect(mavlink_listen_port);
```
JonasVautherin: Hmm, then that could be done after the callback (line 11) is called, I believe.

As discussed, this change makes the callback superfluous. But because `backend.connect()` is waiting for the discovery, calling `backend.connect()` before or after the callback (line 11) will change the behavior on the user side. So let's try to first call the callback, and then wait for the connection, so that the callback is truly superfluous :D.

Not sure if that makes sense at all :sweat_smile:.

core/system_impl.cpp
```diff
@@ -187,11 +187,13 @@ void SystemImpl::process_heartbeat(const mavlink_message_t &message)
     if (is_autopilot(message.compid) && !have_uuid()) {
         request_autopilot_version();
 
+#if defined(ENABLE_FALLBACK_TO_SYSTEM_ID)
```
JonasVautherin: Not in the scope of this PR (because it affects other parts), but as a general comment: I'm always nervous with `#ifdef`s. We could design this to use a boolean instead, couldn't we?

julianoes: You're right.

backend/src/backend_api.cpp
julianoes: But it works? So what should I change?

JonasVautherin: Well nevermind, maybe that's actually clearer like this, because on the client side we call it `Core.connect()` and not `Core.start()`.

PR #579, opened 2018-10-26
Fix camera status race

JonasVautherin:
There was a race condition that made the backend crash when the CameraStatus stream was closed by the client. This seems to solve it (or at least makes it much more robust, as I couldn't reproduce the crash anymore).


plugins/camera/camera_impl.cpp
```diff
@@ -953,9 +953,14 @@ void CameraImpl::check_status()
 
 void CameraImpl::notify_status(Camera::Status status)
 {
-    if (_subscribe_status_callback) {
-        _parent->call_user_callback([this, status]() { _subscribe_status_callback(status); });
+    // Make a copy because it is passed to the thread pool
+    const auto status_callback = _subscribe_status_callback;
+
+    if (status_callback == nullptr) {
```
julianoes: Ah, yes I guess that makes sense.

PR #585, opened 2018-10-31
Add plugin to set params

julianoes:
This adds a plugin to set/get "raw" mavlink params.

This has been tested against the E90.

When setting int parameters, we get an error but the param change still works. Hopefully, this will be fixed on the Firmware side: https://github.com/PX4/Firmware/pull/10792

### DIFF
--- a/core/mavlink_parameters.cpp
+++ b/core/mavlink_parameters.cpp
@@ -58,10 +58,10 @@ void MAVLinkParameters::set_param_async(const std::string &name,
 MAVLinkParameters::Result
 MAVLinkParameters::set_param(const std::string &name, const ParamValue &value, bool extended)
 {
-    auto prom = std::make_shared<std::promise<Result>>();
-    auto res = prom->get_future();
+    auto prom = std::promise<Result>();
+    auto res = prom.get_future();
 
-    set_param_async(name, value, [&prom](Result result) { prom->set_value(result); }, extended);
+    set_param_async(name, value, [&prom](Result result) { prom.set_value(result); }, extended);
 
     return res.get();
 }
@@ -103,13 +103,13 @@ void MAVLinkParameters::get_param_async(const std::string &name,
 std::pair<MAVLinkParameters::Result, MAVLinkParameters::ParamValue>
 MAVLinkParameters::get_param(const std::string &name, ParamValue value_type, bool extended)
 {
-    auto prom = std::make_shared<std::promise<std::pair<Result, MAVLinkParameters::ParamValue>>>();
-    auto res = prom->get_future();
+    auto prom = std::promise<std::pair<Result, MAVLinkParameters::ParamValue>>();
+    auto res = prom.get_future();
 
     get_param_async(name,
                     value_type,
                     [&prom](Result result, ParamValue value) {
-                        prom->set_value(std::make_pair<>(result, value));
+                        prom.set_value(std::make_pair<>(result, value));
                     },
                     extended);
 
--- a/core/mavlink_parameters.cpp
+++ b/core/mavlink_parameters.cpp
@@ -38,7 +38,7 @@ void MAVLinkParameters::set_param_async(const std::string &name,
     //     LogDebug() << "setting param " << name << " to " << value.get_int();
     // }
 
-    if (name.size() > PARAM_ID_LEN) {
+    if (name.size() >= PARAM_ID_LEN) {
         LogErr() << "Error: param name too long";
         if (callback) {
             callback(false);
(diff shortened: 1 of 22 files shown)


core/mavlink_parameters.cpp
```diff
@@ -38,10 +38,10 @@ void MAVLinkParameters::set_param_async(const std::string &name,
     //     LogDebug() << "setting param " << name << " to " << value.get_int();
     // }
 
-    if (name.size() > PARAM_ID_LEN) {
+    if (name.size() >= PARAM_ID_LEN) {
```
JonasVautherin: Was that a bug before?

core/mavlink_parameters.cpp
```diff
@@ -324,9 +336,17 @@ void MAVLinkParameters::process_param_ext_value(const mavlink_message_t &message
             if (strncmp(work->param_name.c_str(), param_ext_value.param_id, PARAM_ID_LEN) == 0) {
                 ParamValue value;
                 value.set_from_mavlink_param_ext_value(param_ext_value);
-                _cache[work->param_name] = value;
-                if (work->callback) {
-                    work->callback(true, value);
+                if (value.is_same_type(work->param_value_type)) {
```
JonasVautherin: Here we check that `value.set_from_mavlink_param_ext_value` above did not fail (by silently not updating the value), right?

But then I don't get why the error is called "WRONG_TYPE" :thinking:

core/system_impl.cpp
```diff
@@ -631,95 +631,131 @@ void SystemImpl::set_param_ext_int_async(const std::string &name, int32_t value,
     _params.set_param_async(name, param_value, callback, true);
 }
 
-std::pair<bool, float> SystemImpl::get_param_float(const std::string &name)
+std::pair<MAVLinkParameters::Result, float> SystemImpl::get_param_float(const std::string &name)
 {
-    auto prom = std::make_shared<std::promise<std::pair<bool, float>>>();
+    auto prom = std::make_shared<std::promise<std::pair<MAVLinkParameters::Result, float>>>();
```
JonasVautherin: I don't think it has to be a pointer, as it is used only within the scope of this function. Unless the callback can live longer than the function?

core/mavlink_parameters.cpp
julianoes: Only if you tried setting/getting a param longer than allowed.

MAVLink params are allowed to have a name of 16 chars + termination character.

The thing is that `PARAM_ID_LEN` is 17, so if `name.size()` is 17 it's already too long because `size()` does not include the termination character.

core/mavlink_parameters.cpp
julianoes: When you do the call to get a param you need to specify which type you expect, and that's what we compare here.
So one type comes from the mavlink message and one from the work item from the user.

core/system_impl.cpp
julianoes: Hm, you're right. I guess that's just what the very first example of promise/future looked like. I'll fix it.

core/mavlink_parameters.cpp
JonasVautherin: > Only if you tried setting/getting a param longer than allowed.

:rofl: 

Got it, thanks for the explanation!

core/mavlink_parameters.cpp
JonasVautherin: I guess I'll need to learn about how the protocol works at some point :sweat_smile:.

core/system_impl.cpp
JonasVautherin: Yeah, we have many places that use pointers where that wouldn't be necessary (typically the examples or integration tests). Not a very big deal, but we can think about changing that when we see them :-).

core/mavlink_parameters.cpp
hamishwillee: > MAVLink params are allowed to have a name of 16 chars + termination character.

Not quite, they are sent with null character if less than 16 chars, but are not sent with termination char at length 16 chars. https://mavlink.io/en/messages/common.html#PARAM_VALUE

hamishwillee:
How did you test PARAM_EXT? I didn't think anything other than QGC and Camera Manager used them.

hamishwillee:
And just For my information, is the message flow for PARAM_EXT-* the same as for the normal param protocol (though obviously with the new messages)?

julianoes:
> How did you test PARAM_EXT?

Yuneec E90 :smile: 

> And just For my information, is the message flow for PARAM_EXT-* the same as for the normal param protocol (though obviously with the new messages)?

It's pretty much the same except that you also have the ack/nack that you can return. @dogmaphobic would know the details.

PR #586, opened 2018-11-06
Improve VTOL example

julianoes:
This improves the readability of the VTOL example, and also adds some
more action to it by sending it South using `goto_location`.

selection_003

### DIFF
--- a/example/transition_vtol_fixed_wing/transition_vtol_fixed_wing.cpp
+++ b/example/transition_vtol_fixed_wing/transition_vtol_fixed_wing.cpp
@@ -1,46 +1,36 @@
-
 #include <chrono>
 #include <cstdint>
+#include <iostream>
+#include <thread>
+#include <cmath>
 #include <dronecode_sdk/action.h>
 #include <dronecode_sdk/dronecode_sdk.h>
 #include <dronecode_sdk/telemetry.h>
-#include <iostream>
-#include <thread>
 
 using std::this_thread::sleep_for;
+using std::chrono::seconds;
 using std::chrono::milliseconds;
 using namespace dronecode_sdk;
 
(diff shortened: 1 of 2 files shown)


example/transition_vtol_fixed_wing/transition_vtol_fixed_wing.cpp
```diff
@@ -1,72 +1,50 @@
-
 #include <chrono>
 #include <cstdint>
+#include <iostream>
```
JonasVautherin: Just to mention it: my preference for headers is to include them in alphabetical order. `dronecode` is a dependency of the example, just as `thread`. Alphabetical order is always consistent, where the preference of the developer, well, depends on the developer :-).

example/transition_vtol_fixed_wing/transition_vtol_fixed_wing.cpp
```diff
@@ -1,72 +1,50 @@
...
 
 using std::this_thread::sleep_for;
+using std::chrono::seconds;
 using std::chrono::milliseconds;
 using namespace dronecode_sdk;
 
-#define ERROR_CONSOLE_TEXT "\033[31m" // Turn text on console red
-#define TELEMETRY_CONSOLE_TEXT "\033[34m" // Turn text on console blue
-#define NORMAL_CONSOLE_TEXT "\033[0m" // Restore normal console colour
+static constexpr auto ERROR_CONSOLE_TEXT = "\033[31m";
```
JonasVautherin: Always happy to have less macros :+1:

example/transition_vtol_fixed_wing/transition_vtol_fixed_wing.cpp
julianoes: Right, but I'm the correct developer :smile:

PR #592, opened 2018-11-16
Fix yaw to North in LOITER_TIME

julianoes:
I forgot to signal no change to yaw during a LOITER_TIME.


version
```diff
@@ -1 +1 @@
-0.7.3
+0.8.0
```
JonasVautherin: Isn't it a bugfix? Why not 0.7.4?

julianoes: Cause this release includes the param_raw addition.

PR #595, opened 2018-11-19
Move ActionResult back to Action:Result

julianoes:
Cherry-picked from #501.


hamishwillee:
Is this the beginning of moving everything or a one-off? Note, requires a bit of change in the docs.

julianoes:
> Is this the beginning of moving everything or a one-off? Note, requires a bit of change in the docs.

So actually, `Action::Result`, `Telemetry::Result`, etc. was the default for all videos but we had changed `Action::Result` to `ActionResult` as a one off and realized later it was not necessarily needed and not consistent with the rest, so now we moved it back.

PR #597, opened 2018-11-19
Add -Weffc++

julianoes:
I want to add the effc++ warning because we had a bug where a member variable was not initalized and this resulted in random behavior. See #596.

Really, this is something that the compiler should catch, no matter what!

### DIFF
--- a/plugins/telemetry/telemetry_impl.h
+++ b/plugins/telemetry/telemetry_impl.h
@@ -83,6 +83,9 @@ public:
     void health_all_ok_async(Telemetry::health_all_ok_callback_t &callback);
     void rc_status_async(Telemetry::rc_status_callback_t &callback);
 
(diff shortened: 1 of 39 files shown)


julianoes:
@JonasVautherin please review so we can get it in and avoid conflicts.

plugins/telemetry/telemetry_impl.h
```diff
@@ -197,7 +197,7 @@ class TelemetryImpl : public PluginImplBase {
 
     // The ground speed and position are coupled to the same message, therefore, we just use
     // the faster between the two.
-    double _ground_speed_ned_rate_hz{1.0};
+    double _ground_speed_ned_rate_hz{0.0};
```
JonasVautherin: Mouaha I had missed that.

PR #604, opened 2018-11-22
backend: fail fast when dependency build fails

JonasVautherin:
So that people don't have to scroll up to find the error message when the build fails.


JonasVautherin:
And somehow it fails on the CI.

PR #606, opened 2018-11-23
backend: fail fast when dependency build fails

JonasVautherin:
Trying again, in case it was a Jenkins issue.


JonasVautherin:
I have absolutely no clue why it is doing that. I'm reading the output of a process, and maybe cmake uses a temporary file to save the variable.

Well, too bad.

PR #619, opened 2018-12-06
Fix for fly-through weirdness

julianoes:
Presumably we set loiter items with time 0 which lead to strange flight behaviour together with the non-fly-through option.


Stifael:
Tested in sitl:
fly-through = true, time-inside = 0/NAN -> passes waypoint
fly-through = false, time-inside =0/NAN -> stop
fly-through = true, time-inside > 0.0 -> stop

Without clear definition, case three might still be confusing. I personally would expect that fly-through has higher priority than time-inside. However, at the end it boils down to definition

julianoes:
@Stifael I've added a warning for the 3rd confusing case.

PR #622, opened 2018-12-10
Resolves #610

### LINKED ISSUE #610: Remove hard-coded `-j8` argument for `make`
The current Makefile is sort of a wrapper script around cmake and that's a bit confusing.

Especially annoying is actually the hard-coded `-j8` argument for make. It would be better if this was taken from the CPU count or your settings for make (`MAKEFLAGS`).

https://github.com/Dronecode/DronecodeSDK/blob/ef45e4ddb7a8c5dcd0ac20df9a330afdaa9a7be6/Makefile#L22

More thoughts: https://stackoverflow.com/questions/10567890/parallel-make-set-j8-as-the-default-option

simrankedia:
Resolves #610

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,6 @@ ifdef NINJA_BUILD
 else
 	CMAKE_GENERATOR ?= "Unix Makefiles"
 	MAKE = make
-	# Use 8 processes in parallel for the usual make
 	MAKE_ARGS = -j$(NPROCS)
 endif
 
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,16 @@ endif
 # Check for ninja and use it if available
 NINJA_BUILD := $(shell ninja --version 2>/dev/null)
 
+NPROCS := 1
+OS := $(shell uname -s)
+
+ifeq ($(OS),Linux)
+	NPROCS := $(shell grep -c ^processor /proc/cpuinfo)
+endif
+ifeq ($(OS),Darwin) #Mac OS
+	NPROCS := $(shell sysctl -n hw.ncpu || echo 1)
+endif
+
 ifdef NINJA_BUILD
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -19,7 +29,7 @@ else
 	CMAKE_GENERATOR ?= "Unix Makefiles"
 	MAKE = make
 	# Use 8 processes in parallel for the usual make
-	MAKE_ARGS = -j8
+	MAKE_ARGS = -j$(NPROCS)
```
julianoes: Please remove the comment above since it is no longer true.

Makefile
```diff
@@ -11,6 +11,16 @@ endif
...
 NINJA_BUILD := $(shell ninja --version 2>/dev/null)
 
+NPROCS := 1
+OS := $(shell uname -s)
+
+ifeq ($(OS),Linux)
+	NPROCS := $(shell grep -c ^processor /proc/cpuinfo)
+endif
+ifeq ($(OS),Darwin) #Mac OS
+	NPROCS := $(shell sysctl -n hw.ncpu || echo 1)
```
julianoes: Tested, works.

Makefile
```diff
@@ -11,6 +11,16 @@ endif
 # Check for ninja and use it if available
 NINJA_BUILD := $(shell ninja --version 2>/dev/null)
 
+NPROCS := 1
+OS := $(shell uname -s)
+
+ifeq ($(OS),Linux)
+	NPROCS := $(shell grep -c ^processor /proc/cpuinfo)
```
julianoes: Works on my Arch Linux.

Makefile
simrankedia: Removed!

simrankedia:
Hi,
I have updated the changes. Please have a look. 
Thanks

PR #629, opened 2019-01-07
info: expose new UUID (uid2)

julianoes:
This adds support to query the new UUID using the info plugin.
Vehicles only sending the old UUID (uid) are still supported if the uid2 is all zeros as specified.

@bkueng or @thomasetter please review.

### DIFF
--- a/plugins/info/include/plugins/info/info.h
+++ b/plugins/info/include/plugins/info/info.h
@@ -82,14 +82,21 @@ public:
     };
 
     /**
+     * @brief Type containing identification.
+     */
+    struct Identification {
+        uint8_t hardware_uuid[18]; /**< @brief UUID of hardware. */
+    };
+
+    /**
      * @brief Gets the UUID of the system.
      *
      * If possible this will be a unique identifier provided by hardware.
      *
      * @return a pair containing the result of the request and if successful,
-     * the UUID of the system.
+     * the identification information of the system.
      */
-    std::pair<Result, uint64_t> uuid() const;
+    std::pair<Result, Identification> get_identification() const;
 
     /**
      * @brief Get system version information.
--- a/plugins/info/include/plugins/info/info.h
+++ b/plugins/info/include/plugins/info/info.h
@@ -85,7 +85,7 @@ public:
      * @brief Type containing identification.
      */
     struct Identification {
-        uint8_t hardware_uuid[18]; /**< @brief UUID of hardware. */
+        uint8_t hardware_uid[18]; /**< @brief UID of hardware. */
     };
 
     /**
--- a/plugins/info/include/plugins/info/info.h
+++ b/plugins/info/include/plugins/info/info.h
@@ -85,7 +85,8 @@ public:
      * @brief Type containing identification.
      */
     struct Identification {
-        uint8_t hardware_uid[18]; /**< @brief UID of hardware. */
+        uint8_t hardware_uid[18]; /**< @brief UID of hardware. This refers to uid2 of MAVLink. If
+                                     the system does not support uid2 yet, this will be all zero. */
     };
 
     /**
--- a/plugins/info/include/plugins/info/info.h
+++ b/plugins/info/include/plugins/info/info.h
@@ -90,7 +90,7 @@ public:
     };
 
     /**
-     * @brief Gets the UUID of the system.
+     * @brief Gets the identification of the system.
      *
      * If possible this will be a unique identifier provided by hardware.
      *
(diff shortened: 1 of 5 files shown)


plugins/info/info_impl.cpp
```diff
@@ -116,9 +117,28 @@ void InfoImpl::process_autopilot_version(const mavlink_message_t &message)
     const char *product_name = product_id_str(autopilot_version.product_id);
     STRNCPY(_product.product_name, product_name, sizeof(_product.product_name) - 1);
 
+    if (is_array_zero(autopilot_version.uid2, sizeof(autopilot_version.uid2))) {
+        std::memcpy(_identification.hardware_uuid,
+                    reinterpret_cast<uint8_t *>(&autopilot_version.uid),
+                    sizeof(autopilot_version.uid));
+    } else {
+        std::memcpy(
+            _identification.hardware_uuid, autopilot_version.uid2, sizeof(autopilot_version.uid2));
```
bkueng: I'd make sure the rest of `hardware_uuid` is set to zero in case `process_autopilot_version` is called several times (mostly a precaution).

plugins/info/include/plugins/info/info.h
```diff
@@ -81,15 +81,22 @@ class Info : public PluginBase {
...
      * @brief Gets the UUID of the system.
      *
      * If possible this will be a unique identifier provided by hardware.
      *
      * @return a pair containing the result of the request and if successful,
-     * the UUID of the system.
+     * the identification information of the system.
      */
-    std::pair<Result, uint64_t> uuid() const;
+    std::pair<Result, Identification> get_identification() const;
```
thomasetter: As a user, how do I know whether a uid2 or a uid1 has been read?
If we do not want to provide a method to access the old version, could we document this here?
Maybe we changing the comment to mention uid and uid2 is enough.

plugins/info/include/plugins/info/info.h
```diff
@@ -81,15 +81,22 @@ class Info : public PluginBase {
         char product_name[32]; /**< @brief Name of product. */
     };
 
+    /**
+     * @brief Type containing identification.
+     */
+    struct Identification {
+        uint8_t hardware_uuid[18]; /**< @brief UUID of hardware. */
```
thomasetter: can we call this uid instead of uuid?
UUID is generally understood to be a 16 byte identifier: https://en.wikipedia.org/wiki/UUID

plugins/info/include/plugins/info/info.h
julianoes: Yes I would say we don't support the old version. And we could document it.

plugins/info/info_impl.cpp
julianoes: Good catch.

plugins/info/include/plugins/info/info.h
julianoes: Hm, so what is it actually?
UUID: universally unique ID
UID: unique ID

According to the MAVLink spec it's a UID, so I guess you're right.
https://mavlink.io/en/messages/common.html#AUTOPILOT_VERSION

plugins/info/include/plugins/info/info.h
julianoes: Right now it just falls back to the old `uid`, is that a problem?

thomasetter: It means that the id will silently change after an upgrade from e.g. PX4 stable to master which is suboptimal.
Maybe we can just remove the fallback and call the function get_uid2(). What do you think?

julianoes: Ok removing the fallback makes sense. For the naming, we try to abstract the underlying MAVLink API, so therefore just UID. But I'll mention it in the comment.

plugins/info/include/plugins/info/info.h
```diff
@@ -81,15 +81,22 @@ class Info : public PluginBase {
...
 
+    /**
+     * @brief Type containing identification.
+     */
+    struct Identification {
+        uint8_t hardware_uid[18]; /**< @brief UID of hardware. */
+    };
+
     /**
      * @brief Gets the UUID of the system.
```
JonasVautherin: So we call it UID, right?

JonasVautherin:
```suggestion
     * @brief Gets the UID of the system.
```

plugins/info/info_impl.cpp
```diff
@@ -116,6 +117,10 @@ void InfoImpl::process_autopilot_version(const mavlink_message_t &message)
     const char *product_name = product_id_str(autopilot_version.product_id);
     STRNCPY(_product.product_name, product_name, sizeof(_product.product_name) - 1);
 
+    std::memset(_identification.hardware_uid, 0, sizeof(_identification.hardware_uid));
```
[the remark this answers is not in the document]
julianoes: Thanks, yes that's better. Fixed in 8da2fe55693ab0ed51f0172c2eb53de2cb561d0f.

julianoes:
Well, somehow Jenkins decided this branch is "disabled" and not worth testing. Screw you Jenkins I'm merging it anyway!

PR #635, opened 2019-01-10
[WIP] Connection result enhance

shrit:
Fix ConnectionResult -> Connection::Result


integration_tests/.#.#.#camera_settings.cpp
```diff
@@ -0,0 +1 @@
+omar@linux-4-102.722:1545258157
```
julianoes: Please remove this file.

And to ignore them in the future you can setup a global gitignore file for your machine:
https://stackoverflow.com/questions/7335420/global-git-ignore

integration_tests/.#.#camera_settings.cpp
```diff
@@ -0,0 +1 @@
+omar@linux-4-102.722:1545258157
```
julianoes: And this.

core/include/connection.h
```diff
@@ -0,0 +1,65 @@
+#pragma once
```
julianoes: The problem with moving `Connection` to the public header files is that we expose the `Connection` base class which doesn't need to be public. Also, this means that `mavlink_receiver.h` needs to be included because symbols of it are required in `connection.h`, however, `mavlink_receiver.h` is not public either.

So this means that we need to have some separation for this and I believe that's why we had the connection result defined in `dronecode_sdk.h` previously (and then later `connection_result.h`).

core/include/connection.h
```diff
@@ -0,0 +1,65 @@
...
+  virtual Result start() = 0;
+  virtual Result stop() = 0;
+  
+  virtual bool send_message(const mavlink_message_t &message) = 0;
+
+  inline const char *result_str(const Result result);
+  
+  // Non-copyable
+  Connection(const Connection &) = delete;
+  const Connection &operator=(const Connection &) = delete;
```
julianoes: The indenting is wrong in this file.

core/include/connection.h
shrit: I fixed all the other small problems also the one with the indentations.
Thanks for pointing to me this issue, I understand that users should not be able to use the Connection class directly. Thus, we can move it to the core directory. 
Are you ok with this modification??
I am going to send you a pull request with the above modifications.

julianoes: @JonasVautherin I'm sure you have an opinion about that? :smile:

JonasVautherin: Haha, I was the one making `connection_result.h` in the first place, so I am not sure if I have a better solution :sweat_smile:. I had two reasons for that:

1. A `ConnectionResult` is a status, and not an internal object of `Connection`. We expose statuses, and not `Connection`, so I was in favor of not having them in the same header.
2. When mocking (e.g. [here](https://github.com/Dronecode/DronecodeSDK/blob/develop/core/mocks/dronecode_sdk_mock.h)), I don't need the `Connection` header, precisely because only the statuses are exposed. So it was looking cleaner to me to not have the full header there as well.

This said, I don't have a super strong opinion about that, and I am open to other solutions. But I can't say I have a better idea :/.

shrit: I will try to find a solution in the next few days. Now, I need to find a way to reset my EKF2
during the simulation as we have discussed in the forum :smiley:

julianoes:
@shrit do you have more ideas how to solve this? I myself don't have a good answer to the problems that we found above.

shrit:
Well personally, I did not find anythin different from the original code.

Maybe we can go back to the original solution, but change the name of the struct connection_result to 
Connection and rename the class Connection to Connections.
From the public point of view that is going to remove the confusion with other plugins, and make the user more comfortable.  However in this case, we need to change the class Connection to Connections everywhere it has been used in the SDK

julianoes:
> Maybe we can go back to the original solution, but change the name of the struct connection_result to
Connection and rename the class Connection to Connections.

`I would rather have something like `Connection::Result` but just `Connection` for a result seems odd.

And why would you call it `Connections`? You sort of add one connection at a time and you get one `ConnectionResult` in response.

shrit:
Hm, you are right. But if you really prefer Connection::Result we will have to execute some major modification on the code, For example, look at this scenario:

In fact, I notice that the Connection class handle the mavlink protocol rather that handling the connection. For me, the Connection class should manage the tcp, udp and the serial, or even any future networking protocol. In this case they are manged by the DronecodeSDK class.

In this case, we can create a new class Connection that handle the networking part of the dronecodeSDK.Then for each class TcpConnection, or UdpConnection we can create a local enum Result that handle locally the connection state.

I understand that mavlink is the communication protocol rather that the connection protocol. Thus, 
we can create a different class to handle the mavlink and make it accessible in the TcpConnection class or Udp.

Please, let me know what do you think?

julianoes:
Hm, good point. So I tried to have as little as possible mavlink in the actual connections. All I can see is how the bytes are passed to `_mavlink_receiver` e.g. in udp_connection:
https://github.com/Dronecode/DronecodeSDK/blob/develop/core/udp_connection.cpp#L208

Is that what you mean? And how would you propose to improve it?

shrit:
Maybe I did not understand perfectly your last answer

I have created a branch with some of the modification that I have talked about in the last comment,
I did not even compile it, there is still a lot of work to do, but please tell me what do you thing,

https://github.com/shrit/DronecodeSDK/tree/connection_result

julianoes:
Closing in favor of #667

PR #636, opened 2019-01-11
travis: remove CI git tag mechanism again

julianoes:
We don't need this anymore since we changed the releasing back.


julianoes:
Ignoring that Jenkins failed because `java.io.IOException: No space left on device`.

PR #643, opened 2019-01-18
Remove Fedora 27 CI build and add Fedora 29

julianoes:
Fedora 27 has already been end-of-lifed last year, so we should remove it and add Fedora 28 instead.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,13 +61,13 @@ matrix:
       services:
         - docker
       env:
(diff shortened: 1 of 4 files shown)


julianoes:
We can ignore appveyor CI, the gtest checkout fails and it's already fixed in https://github.com/Dronecode/DronecodeSDK/pull/638/commits/c808e122e2f1aab847e28db27682a55ee8f4cc9c.

PR #644, opened 2019-01-18
Add mission raw plugin

julianoes:
This will be a plugin which enables direct/raw access to MAVLink MISSION_ITEM_INT messages.

See progress here @sanderux.

### DIFF
--- a/integration_tests/mission_raw_mission_changed.cpp
+++ b/integration_tests/mission_raw_mission_changed.cpp
@@ -1,0 +1,63 @@
+#include "integration_test_helper.h"
+#include "dronecode_sdk.h"
+#include "plugins/mission/mission.h"
+#include "plugins/mission_raw/mission_raw.h"
+#include <memory>
+#include <future>
+
+using namespace dronecode_sdk;
+using namespace std::placeholders; // for `_1`
+
+TEST_F(SitlTest, MissionRawMissionChanged)
+{
+    DronecodeSDK dc;
+
+    ConnectionResult ret = dc.add_udp_connection();
+    ASSERT_EQ(ret, ConnectionResult::SUCCESS);
+
+    // Wait for system to connect via heartbeat.
+    std::this_thread::sleep_for(std::chrono::seconds(2));
+    ASSERT_TRUE(dc.is_connected());
+
+    System &system = dc.system();
+    ASSERT_TRUE(system.has_autopilot());
+
+    auto mission = std::make_shared<Mission>(system);
+    auto mission_raw = std::make_shared<MissionRaw>(system);
+
+    std::promise<void> prom_changed{};
+    std::future<void> fut_changed = prom_changed.get_future();
+
+    LogInfo() << "Subscribe for mission changed notification";
+    mission_raw->subscribe_mission_changed([&prom_changed]() { prom_changed.set_value(); });
+
+    // The mission change callback should not trigger yet.
+    EXPECT_EQ(fut_changed.wait_for(std::chrono::milliseconds(500)), std::future_status::timeout);
+
+    auto new_item = std::make_shared<MissionItem>();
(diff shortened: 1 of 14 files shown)


plugins/mission/mission.cpp
```diff
@@ -65,18 +65,20 @@ const char *Mission::result_str(Result result)
...
             return "Success";
-        case Result::BUSY:
-            return "Busy";
         case Result::ERROR:
             return "Error";
         case Result::TOO_MANY_MISSION_ITEMS:
             return "Too many mission items";
-        case Result::INVALID_ARGUMENT:
-            return "Invalid argument";
+        case Result::BUSY:
```
JonasVautherin: What's the link between `mission.cpp` and the `mission_raw` plugin?

sanderux: @JonasVautherin the mission plugin only returns values known to the SDK. The raw plugin returns all of them.

JonasVautherin: Right, but I see a `Result` defined in `plugins/mission_raw`. This one is from `plugins/mission`, which is a different plugin. Do we need those changes in the `mission` plugin, too? And is that part of this PR?

julianoes: Oh right. I just found that one case was missing as part of copying from `mission`.

integration_tests/mission_raw_mission_changed.cpp
```diff
@@ -2,11 +2,19 @@
...
 #include <future>
 
 using namespace dronecode_sdk;
-using namespace std::placeholders; // for `_1`
+
+static constexpr double SOME_LATITUDES[] = {47.398170, 47.398175};
+static constexpr double SOME_LONGITUDES[] = {8.545649, 8.545654};
+static constexpr float SOME_ALTITUDES[] = {5.0f, 7.5f};
+static constexpr float SOME_SPEEDS[] = {4.0f, 5.0f};
+static constexpr unsigned NUM_SOME_ITEMS = sizeof(SOME_LATITUDES) / sizeof(SOME_LATITUDES[0]);
```
JonasVautherin: Could we use a `vector` instead? Would be clearer to me :confused:

integration_tests/mission_raw_mission_changed.cpp
```diff
@@ -34,30 +42,67 @@ TEST_F(SitlTest, MissionRawMissionChanged)
...
+
+    {
+        std::promise<void> prom{};
+        std::future<void> fut = prom.get_future();
+        LogInfo() << "Download raw mission items.";
+        mission_raw->download_mission_async(
+            [&prom](MissionRaw::Result result,
+                    std::vector<std::shared_ptr<MissionRaw::MavlinkMissionItemInt>> items) {
+                EXPECT_EQ(result, MissionRaw::Result::SUCCESS);
+                // TODO: validate items
```
JonasVautherin: That TODO is outdated, isn't it?

julianoes:
I'm ignoring Jenkins because I don't know what he's trying to tell me.

julianoes:
@JonasVautherin also I forgot to fix what you raised, I'll do that in #642.

plugins/mission_raw/include/plugins/mission_raw/mission_raw.h
```diff
@@ -0,0 +1,133 @@
...
+     *
+     * @param result Enum for which string is required.
+     * @return Human readable string for the MissionRaw::Result.
+     */
+    static const char *result_str(Result result);
+
+    /**
+     * @brief Mission item identical to MAVLink MISSION_ITEM_INT.
+     */
+    struct MavlinkMissionItemInt {
```
hamishwillee: No support for MISSION_ITEM?

julianoes: Nope. The SDK only does the new shiny things.

PR #646, opened 2019-01-21
core: fix segfault on plugin init

julianoes:
This fixes the case where on plugin initialization, the `std::vector _mavlink_handler_table` is growing and needs to be re-located, however, the iterator is not invalidated as it should because everything moved and we're accessing garbage.


julianoes:
@bkueng hm, you're right, thanks!

PR #648, opened 2019-01-22
Jenkins enable CCACHE, post cleanup, and add ninja-build to docker

dagar:
Here are a few minor things that might help improve build times on Jenkins.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,6 +8,25 @@ cmake_policy(SET CMP0025 NEW)
 
 project(dronecode_sdk)
 
+# ccache
+option(CCACHE "Use ccache if available" ON)
+find_program(CCACHE_PROGRAM ccache)
+if (CCACHE AND CCACHE_PROGRAM AND NOT DEFINED ENV{CCACHE_DISABLE})
+
+	get_filename_component(ccache_real_path ${CCACHE_PROGRAM} REALPATH)
+	get_filename_component(cxx_real_path ${CMAKE_CXX_COMPILER} REALPATH)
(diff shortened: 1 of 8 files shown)


dagar:
I'll add ninja-build and ccache to the containers separately.  https://github.com/Dronecode/DronecodeSDK/pull/649

dagar:
TODO: put ccache links on PATH.

julianoes:
> TODO: put ccache links on PATH.

Are you sure that's what's wrong? Because locally it works, I think it's just the Jenkins docker cache because I accidentally pushed to this tag without ccache first. I can push to a new tag and try again.

dagar:
Yes, to actually be active it needs to be the first thing in the path like this (or explicitly handled by the build system).
https://github.com/PX4/containers/blob/master/docker/px4-dev/Dockerfile_base#L81

image

dagar:
Safe to merge now, but won't have a big impact until we do another pass to enable CCACHE.

dagar:
Actually, here's another way to do it via cmake. fba0855e55f5c

julianoes:
> Actually, here's another way to do it via cmake. fba0855

Unfortunately that doesn't work on travis:
`ccache: error: Failed to create directory /home/user/.ccache/tmp: Permission denied`

dagar:
One last thing to note is that we're still only getting CCACHE for the SDK build itself. Setting it up on the system path would also catch the dependencies, which I think is where you spent the majority of the build time.

julianoes:
Which dependency would that be? Because they get built from the same CMakeLists.txt...

Issue #651, opened 2019-01-23
Expose temperature


sanderux:
I would like to be able to read HIGHRES_IMU.temperature

JonasVautherin:
@julianoes: is that `Info` or `Telemetry`? Or something else?

julianoes:
I'd say that's for telemetry.

sanderux:
I might be completely off in the idea behind the SDK, but why would you not expose all mavlink topics and allow getting updates? You can still create wrappers around standard things but you are not limiting devs to access all telemetry information.

Same goes for sending stuff, its good to have wrappers for everything the SDK understands, but a generic message factory such as dronekit/pymavlink would serve as a fall-back for anything unknown to the SDK or custom added to Mavlink.

julianoes:
Yes, a "backdoor" like that is something we probably should do. But that's orthogonal to this request.

julianoes:
This has been added to the C++ API:
https://github.com/mavlink/MAVSDK/blob/221bd80fd96f5524da2a14f4f5f46eb667e411e4/src/plugins/telemetry/include/plugins/telemetry/telemetry.h#L177
but it's missing in Python, so it needs to be added to the proto file here:
https://github.com/mavlink/MAVSDK-Proto/blob/master/protos/telemetry/telemetry.proto

PR #657, opened 2019-02-05
Add MAVLink pass-through plugin

### LINKED ISSUE #652: Allow sending STATUS_TEXT message
To inform the user from an onboard secondary computer i would like to be able to send STATUS_TEXT messages to the flightcontroller. PX4 is configured to relay these to other devices. I would like to be able to set the SOURCE_SYSTEM to the MAV_SYS_ID of the flight controller so QGC doesnt see the message as a different vehicle.

julianoes:
This adds a backdoor to send and receive direct/raw MAVLink messages.

> "With great power comes great responsibility." 
> This plugin allows you to send and receive MAVLink messages. There is no checking or safe-guards, you're on your own, and you have been warned.

This should close #652, #651.

To compile this in you need to actually do:
```
make ENABLE_MAVLINK_PASSTHROUGH=1
```

TODO:  add this build to CI


plugins/mavlink_backdoor/include/plugins/mavlink_backdoor/mavlink_backdoor.h
```diff
@@ -0,0 +1,125 @@
...
+
+    /**
+     * @brief Subscribe to messages using message ID.
+     *
+     * This means that all future messages being received will trigger the
+     * callback to be called. To stop the subscription, call this method with
+     * `nullptr` as the argument.
+     *
+     * @param message_id The MAVLink message ID.
+     * @param callback Callback to be called for message subscription.
```
bkueng: I would like this to be more generic:
- A callback for all received messages independent from the ID
- Ability to modify the message
- The callback should return a bool: if false, the message will be dropped.

These are requirements for the testing suite.

julianoes: > - Ability to modify the message

Modify before it goes to other plugins? Ok, that should be possible to add.

> - The callback should return a bool: if false, the message will be dropped.

I don't get it. Dropped from where? Ah, so we need a hook that is called before any other plugins (or the core) get the message delivered.

> These are requirements for the testing suite.

I was planning to come up with higher level hooks for testing such as "drop 50% of the incoming messages" but that might not yield consistent results, so I need to think a bit more about that.

julianoes:
The `raw` plugins are closer to MAVLink but don't expose the MAVLink headers, so they are still somewhat encapsulated, however, the passthrough is actual MAVLink types being exported.

I've renamed it to passthrough.

julianoes:
@bkueng I've added an API to intercept incoming and outgoing MAVLink messages, check it out!

hamishwillee:
@julianoes 
- Do you still need to do `make ENABLE_MAVLINK_PASSTHROUGH=1` 
  - why?
  - If not already done, can we add that to the class docs so that this info is included in the API ref.
- How then, do I generate docs for this using the `./generate_docs.sh` ? YOu might have some issues if this was not actually built for docs in testing.

Also, I think this file is in the wrong place:

image

julianoes:
@hamishwillee thanks for catching these things. I'm checking the points.

PR #659, opened 2019-02-07
Some doc and install fixes

julianoes:
Addresses @hamishwillee's notes in #657.


julianoes:
Ok, we should probably not try to generate docs for all of mavlink :see_no_evil:. I need to fix that somehow.

hamishwillee:
Thanks! Corresponding ref added in https://github.com/dronecore/sdk_docs/pull/152

PR #662, opened 2019-02-07
cmake: fix tag extraction, again

JonasVautherin:
I'm having issues with this CMake thing...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,11 +63,13 @@ set(dronecode_sdk_install_include_dir "include/dronecode_sdk")
 set(dronecode_sdk_install_lib_dir ${lib_path})
 
 execute_process(
-    COMMAND git describe --abbrev=8 --dirty --always --tags | sed 's/v\\([0-9]*\\.[0-9]*\\.[0-9]*.*$\\)/\\1/'
+    COMMAND git describe --abbrev=8 --dirty --always --tags
     WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
     OUTPUT_VARIABLE VERSION_STR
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -63,12 +63,14 @@ set(dronecode_sdk_install_include_dir "include/dronecode_sdk")
...
 
 execute_process(
-    COMMAND git describe --abbrev=8 --dirty --always --tags | sed 's/v\\([0-9]*\\.[0-9]*\\.[0-9]*.*$\\)/\\1/'
+    COMMAND git describe --abbrev=8 --dirty --always --tags
     WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
     OUTPUT_VARIABLE VERSION_STR
     OUTPUT_STRIP_TRAILING_WHITESPACE
 )
 
+STRING(REGEX REPLACE v\([0-9]+.[0-9]+.[0-9]+.*$\) \\1 VERSION_STR ${VERSION_STR})
```
julianoes: There should be a `VERSION_STR_VERSION_ONLY` and the normal `VERSION_STR` because for the library printf on loading it should display everything.

JonasVautherin: I just remove the first "v" but keep all the rest (just because release tags look like "v0.14.0" and I need "0.14.0")... Do you want another variable just for that? I thought it was not that important :sweat_smile:

JonasVautherin: Just to make it clear, `v0.14.0-10-gd7337e2s3` will become `0.14.0-10-gd7337e2s3`. Do we actually need to keep the `v` for `printf`?

julianoes: Ah no we don't

PR #666, opened 2019-02-12
Plugins: don't remove pending timeouts on disconnect

julianoes:
We  should not get rid of pending timeouts just because the system is reported to be timed out which can happen if we miss a couple of heartbeats in a row due to a lossy connection. We still want to get the notification that our current transaction timed out, otherwise we'll wait for an answer indefinitely (unless the user of the API checks for a timeout again which shouldn't be required).

Found as part of https://github.com/Auterion/mavlink-testing-suite/pull/10.

### DIFF
--- a/plugins/mission/mission_impl.cpp
+++ b/plugins/mission/mission_impl.cpp
@@ -58,13 +58,11 @@ void MissionImpl::init()
 
 void MissionImpl::enable() {}
 
-void MissionImpl::disable()
-{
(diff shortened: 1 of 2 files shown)


hamishwillee:
Merging a demonic PR (666). Very brave.

JonasVautherin:
Isn't it "999" in Australia? I thought everything was inverted there :trollface:

hamishwillee:
No, -666

PR #667, opened 2019-02-12
[WIP] Connection result

shrit:
Please do not merge,
This is the solution I thought about, please have a look,


shrit:
For example, for users I can imagine something similar to this code :
PS: you are right, I need to fix the style each time.
For the `Mavlink` class,  you can choose any another name that you find appropriate. Now, it is hidden from the user side.  However, if you have any other ideas about how to expose the `Mavlink` class, (for example merge it with another classes) I would like to hear from you

    
```
int main(int argc, char **argv)                                                                                                                                                                                   
{                                                                                                                                                                                                                 
    DronecodeSDK dc;                                                                                                                                                                                              
    Connection connect;                                                                                                                                                                                           
    std::string connection_url;                                                                                                                                                                                   
    Connection::Result connection_result;                                                                                                                                                                         
                                                                                                                                                                                                                  
    if (argc == 2) {                                                                                                                                                                                              
        connection_url = argv[1];                                                                                                                                                                                 
        connection_result = connect.add_any_connection(connection_url);                                                                                                                                           
        //we can rename the above function to the following                                                                                                                                                       
        connection_result = connect.create_connection(connection_url);                                                                                                                                            
        connection_result = connect.create_tcp_connection(connection_url);                                                                                                                                        
        connection_result = connect.create_udp_connection(connection_url);                                                                                                                                        
                                                                                                                                                                                                                  
    } else {                                                                                                                                                                                                      
        usage(argv[0]);                                                                                                                                                                                           
        return 1;                                                                                                                                                                                                 
    }                                                                                                                                                                                                             
                                                                                                                                                                                                                  
    if (connection_result != Connection::Result::SUCCESS) {                                                                                                                                                       
        std::cout << ERROR_CONSOLE_TEXT                                                                                                                                                                           
                  << "Connection failed: " << connect.result_str(connection_result)                                                                                                                               
                  << NORMAL_CONSOLE_TEXT << std::endl;                                                                                                                                                            
        return 1;                                                                                                                                                                                                 
    }                                                                                                                                                                                                             
                                                                                                                                                                                                                  
    // Wait for the system to connect via heartbeat                                                                                                                                                               
    while (!dc.is_connected()) {                                                                                                                                                                                  
        std::cout << "Wait for system to connect via heartbeat" << std::endl;                                                                                                                                     
        sleep_for(seconds(1));                                                                                                                                                                                    
    }
   // users can get the system discovered and use the plugin..
  // users code.....
}  
         
```

julianoes:
So how is `dc` connected with `connect`?

To me it seems like we want something like this:
```
DronecodeSDK dc;
UdpConnection udp_connection(14540);
dc.add_connection(udp_connection);
```

However, the problem I see here is that we have no return argument for the constructor so we can't communicate that sometimes creating a connection fails. (We could use exceptions but I decided not to use exceptions in this library.)

julianoes:
Also, you should start off with latest `develop`, otherwise you get conflicts to merge this.

shrit:
Hmm, Actually I though about moving the connection functions from `DronecodeSDK` class to another 
class that is called `Connection` as you can see in the merge request.

If you want really to have something like 
```
DronecodeSDK dc;
UdpConnection udp_connection(14540);
dc.add_connection(udp_connection);

```

I think I need to rethink another solution, because in this case we are implementing the pimpl for the class connection and we are no using it.
Maybe I think I did not understand well your proposal.

julianoes:
> I think I need to rethink another solution, because in this case we are implementing the pimpl for the class connection and we are no using it.

We are not using it? I don't understand how `dronecode_sdk_impl.cpp` is talking to the connection in your case.

shrit:
I am separating `dronecode_sdk_impl.cpp `from `connection.cpp`. Thus, a user is going to use dronecodeSDK only to verify if the system is connected or not, the dronecodeSDK will not manage at all 
the connection, even it is going to be a connection Independent.

julianoes:
@shrit ok and how does the SDK get the messages delivered?

shrit:
Ah, you are right. We need to think about it..

julianoes:
I'm not sure what to do here. Maybe at some point in the future we can revisit the API but I'm closing this for now. Thanks for your contributions nevertheless @shrit!

shrit:
No problem, I also did not have free time to look at it into details,

PR #668, opened 2019-02-13
Fix mission gRPC server crash

julianoes:
This should fix https://github.com/Dronecode/DronecodeSDK-Python/issues/46 as well as a couple of other potential mission crashes.

### DIFF
--- a/plugins/mission/mission_impl.cpp
+++ b/plugins/mission/mission_impl.cpp
@@ -1167,9 +1167,11 @@ void MissionImpl::report_progress()
     }
 
     if (should_report) {
-        _parent->call_user_callback([this, current, total]() {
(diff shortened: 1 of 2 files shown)


plugins/mission/mission_impl.cpp
```diff
@@ -1167,9 +1159,11 @@ void MissionImpl::report_progress()
     }
 
     if (should_report) {
-        _parent->call_user_callback([this, current, total]() {
+        std::lock_guard<std::recursive_mutex> lock(_mission_data.mutex);
+        auto temp_callback = _mission_data.progress_callback;
```
JonasVautherin: Yep, good catch!

PR #670, opened 2019-02-14
Fix param set issue

julianoes:
This fixes an issue reported by @douglaswsilva.

As part of this pull request I've also updated to the latest E90 camera definition file.

@Stifael or @potaito maybe you can send me the latest E50, ET, etc. camera definition files, so I can add them here.

### DIFF
--- a/core/mavlink_parameters.cpp
+++ b/core/mavlink_parameters.cpp
@@ -264,7 +264,7 @@ void MAVLinkParameters::process_param_value(const mavlink_message_t &message)
         return;
     }
 
-    if (strncmp(work->param_name.c_str(), param_value.param_id, PARAM_ID_LEN) != 0) {
+    if (work->param_name.compare(param_value.param_id) != 0) {
         // No match, let's just return the borrowed work item.
         return;
     }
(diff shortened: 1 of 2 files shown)


core/mavlink_parameters.cpp
```diff
@@ -264,7 +264,7 @@ void MAVLinkParameters::process_param_value(const mavlink_message_t &message)
         return;
     }
 
-    if (strncmp(work->param_name.c_str(), param_value.param_id, PARAM_ID_LEN) != 0) {
+    if (work->param_name.compare(param_value.param_id) != 0) {
```
JonasVautherin: I'm not sure I get the difference. What was the issue with the previous one?

julianoes: Have you read the commit description?

JonasVautherin: I didn't :sweat_smile:. I should get used to that, but I'm unfortunately not used to finding useful information in commit descriptions. Will work on it, thanks!

julianoes: Hehe, I try to write make them explanatory.

potaito:
@julianoes That would be nice, thanks. But where do I find these definition files and why would they change?

julianoes:
They change as cameras get new features or fixes.

julianoes:
Ideally the SDK downloads the files via http but that's currently broken through MFi on ST10C.

PR #671, opened 2019-02-14
cmake: fix bug in STRING REPLACE

JonasVautherin:
The last argument of STRING REPLACE, which is a variable, has to be
inside double quotes otherwise it may be misinterpreted.


JonasVautherin:
Well, if `VERSION_STR`'s value is "0.14.0", that's fine, but if it's "something with spaces", suddenly the CMAKE command ends up having 3 parameters instead of one ([something, with, spaces] instead of "something with spaces").

julianoes:
Right ok.

PR #673, opened 2019-02-15
Alpine dcsdk docker

JonasVautherin:
Proposition building on top of @mrpollo's PR (#669).

This is making a multi-stage build out of it, so that the end image is much smaller (~90M instead of almost 3G).

@mrpollo: what do you think?


mrpollo:
Much much better, this is awesome, can we also create a `dronecode-sdk-backend` entry on docker hub?

docker/Dockerfile-backend
```diff
@@ -0,0 +1,29 @@
...
+            cmake \
+            curl-dev \
+            git \
+            go \
+            linux-headers \
+            perl
+
+WORKDIR /root/dronecode_sdk
+
+RUN git clone https://github.com/dronecode/dronecodesdk --recursive .
```
mrpollo: Can we use an `ARG` here to determine the branch/tag? we could also do a shallow clone with depth 1

Example:
```dockerfile
...
ARG branch=master
RUN git clone --branch $branch --depth 1  https://github.com/dronecode/dronecodesdk --recursive .
...
```

This will make it easier to do builds per releases on CI

```bash
docker build --build-arg branch="v0.14.1" -f Dockerfile-backend -t dronecode/dronecode-sdk-base .
```

Although I'm not sure if this messes up your multistage build and how caching works due to the ARG

JonasVautherin: I am not sure about the ARG and the deployment on docker hub. I've never automated that... But if it helps I'm open to it.

About the shallow clone, that won't work because of the submodules.

JonasVautherin: Actually I'm thinking that instead of an `ARG` and `git clone`, we should probably `COPY` the sources. Because that will get built by Travis, and it has a clone of the repo already :+1:.

JonasVautherin: What about the newer version? Note that when building from the `docker` directory, the command is:

```
docker build -f Dockerfile-backend ..
```

i.e. it needs the root of the repo as context. When building from the root, it would be:

```
docker build -f docker/Dockerfile-backend .
```

julianoes: I would probably favor this:
```
docker build -f Dockerfile-backend ..
```

JonasVautherin: Right. But that's for the CI, right? Should we merge this PR in the meantime, or should it come with the CI integration already?

julianoes: Ok, we can merge this for now.

PR #675, opened 2019-02-19
example: added example for multiple drones

julianoes:
This adds the example started by Shayaan Haider and shared via Slack.
The current state doesn't seem to be functional just yet, only one of
two drones responds.


julianoes:
Presumably this needs some work/refactoring in the connection layer. Right now all messages are sent to all links.

julianoes:
Ok, refactoring done. This now works for me.

example/multiple_drones/multiple_drones.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+    // the loop below adds the number of ports the sdk monitors.
+    for (int i = 1; i < argc; ++i) {
+        ConnectionResult connection_result = dc.add_any_connection(argv[i]);
+        if (connection_result != ConnectionResult::SUCCESS) {
+            std::cout << ERROR_CONSOLE_TEXT
+                      << "Connection error: " << connection_result_str(connection_result)
+                      << NORMAL_CONSOLE_TEXT << std::endl;
+            ;
+            return 1;
```
shakthi-prashanth-m: So if any of the port is already busy, the example exits ? Lets say I am controlling 3 drones, and 1 of them is QGC port! So, can't I just continue with those two of them ?

julianoes: Good point, thanks. That can be improved!

example/multiple_drones/multiple_drones.cpp
```diff
@@ -0,0 +1,153 @@
...
+        return 1;
+    }
+
+    DronecodeSDK dc;
+
+    // the loop below adds the number of ports the sdk monitors.
+    for (int i = 1; i < argc; ++i) {
+        ConnectionResult connection_result = dc.add_any_connection(argv[i]);
+        if (connection_result != ConnectionResult::SUCCESS) {
+            std::cout << ERROR_CONSOLE_TEXT
```
shakthi-prashanth-m: Its better `std::cerr` for error messages.

julianoes:
Thanks for the review @shakthi-prashanth-m !

julianoes:
Coveralls.io API status was orange, so let's hope it's just that.

JonasVautherin:
Still seems to fail because of coveralls...

example/multiple_drones/multiple_drones.cpp
julianoes: I think I'd leave it like that because it's really something unexpected and it will make it harder to debug why things are not working after that.

PR #681, opened 2019-02-21
Various camera improvements

julianoes:
As part of testing against the camera-manager, various issues are coming up.

### DIFF
--- a/plugins/camera/camera_impl.cpp
+++ b/plugins/camera/camera_impl.cpp
@@ -2,7 +2,6 @@
 #include "camera_definition.h"
 #include "system.h"
 #include "global_include.h"
-#include "mavlink_include.h"
 #include "http_loader.h"
 #if !defined(WINDOWS)
 #include "camera_definition_files.h"
@@ -107,13 +106,31 @@ void CameraImpl::disable()
     _parent->remove_call_every(_flight_information_call_every_cookie);
 }
 
+Camera::Result CameraImpl::select_camera(unsigned id)
+{
+    if (id > 6) {
+        return Camera::Result::WRONG_ARGUMENT;
+    }
+
+    // camera component IDs go from 100 to 105.
+    _component_id = MAV_COMP_ID_CAMERA + id;
+
+    // We should probably reload everything to make sure the
+    // correct  camera is initialized.
+    disable();
+    enable();
+
+    return Camera::Result::SUCCESS;
+}
+
 MAVLinkCommands::CommandLong CameraImpl::make_command_request_camera_info()
 {
     MAVLinkCommands::CommandLong command_camera_info{};
 
     command_camera_info.command = MAV_CMD_REQUEST_CAMERA_INFORMATION;
     command_camera_info.params.param1 = 1.0f; // Request it
-    command_camera_info.target_component_id = MAV_COMP_ID_CAMERA;
+    command_camera_info.target_component_id = _component_id;
+    ;
 
     return command_camera_info;
 }
@@ -128,7 +145,7 @@ MAVLinkCommands::CommandLong CameraImpl::make_command_take_photo(float interval_
     cmd_take_photo.params.param2 = interval_s;
     cmd_take_photo.params.param3 = no_of_photos;
     cmd_take_photo.params.param4 = float(_capture.sequence++);
-    cmd_take_photo.target_component_id = MAV_COMP_ID_CAMERA;
+    cmd_take_photo.target_component_id = _component_id;
 
     return cmd_take_photo;
 }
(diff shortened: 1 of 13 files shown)


plugins/camera/camera_impl.cpp
```diff
@@ -107,13 +106,31 @@ void CameraImpl::disable()
...
+
 MAVLinkCommands::CommandLong CameraImpl::make_command_request_camera_info()
 {
     MAVLinkCommands::CommandLong command_camera_info{};
 
     command_camera_info.command = MAV_CMD_REQUEST_CAMERA_INFORMATION;
     command_camera_info.params.param1 = 1.0f; // Request it
-    command_camera_info.target_component_id = MAV_COMP_ID_CAMERA;
+    command_camera_info.target_component_id = _component_id;
+    ;
```
shakthi-prashanth-m: Maybe typo

plugins/camera/camera_impl.cpp
```diff
@@ -107,13 +106,31 @@ void CameraImpl::disable()
     _parent->remove_call_every(_flight_information_call_every_cookie);
 }
 
+Camera::Result CameraImpl::select_camera(unsigned id)
+{
+    if (id > 6) {
```
shakthi-prashanth-m: We support Ids's 0 to 5 right ? If so, it should be 
```suggestion
    const size_t MAX_SUPPORTED_IDS = 5; // inside class
    if (id > MAX_SUPPORTED_IDS) {
```

plugins/camera/camera_impl.cpp
julianoes: Thanks!

plugins/camera/camera_impl.cpp
julianoes: Good catch! I'll fix it.

plugins/camera/include/plugins/camera/camera.h
```diff
@@ -65,6 +66,22 @@ class Camera : public PluginBase {
      */
     static std::string result_str(Result result);
 
+    /**
+     * @brief Select camera to interact with.
+     *
+     * @param id The camera ID from 0 to 5.
+     *
+     * Currently supported are IDs 0 to 5 mapping to MAVLink component IDs
```
shakthi-prashanth-m: Is it necessary to mention MAVLink ?

julianoes: I think for now this can be helpful.

plugins/camera/camera_impl.cpp
```diff
@@ -87,24 +92,76 @@ void CameraImpl::deinit()
...
+            _camera_found = true;
+            manual_enable();
+        }
+    }
 }
 
 void CameraImpl::enable()
+{
+    // FIXME: We check for the connection status manually because
+    // we're not interested int the connection state of the autopilot
```
JonasVautherin:
```suggestion
    // we're not interested in the connection state of the autopilot
```

plugins/camera/camera_impl.cpp
```diff
@@ -87,24 +92,76 @@ void CameraImpl::deinit()
...
+    request_flight_information();
+
     _parent->add_call_every(
         [this]() { request_flight_information(); }, 10.0, &_flight_information_call_every_cookie);
 }
 
 void CameraImpl::disable()
+{
+    // FIXME: We check for the connection status manually because
+    // we're not interested int the connection state of the autopilot
```
JonasVautherin:
```suggestion
    // we're not interested in the connection state of the autopilot
```

plugins/camera/camera_impl.cpp
```diff
@@ -87,24 +92,76 @@ void CameraImpl::deinit()
...
 {
     invalidate_params();
     _parent->remove_call_every(_flight_information_call_every_cookie);
+
+    _camera_found = false;
+}
+
+Camera::Result CameraImpl::select_camera(unsigned id)
+{
+    static constexpr std::size_t MAX_SUPPORTED_ID = 5;
```
JonasVautherin: Why is that?

plugins/camera/camera_impl.cpp
```diff
@@ -87,24 +92,76 @@ void CameraImpl::deinit()
...
+
+void CameraImpl::manual_disable()
 {
     invalidate_params();
     _parent->remove_call_every(_flight_information_call_every_cookie);
+
+    _camera_found = false;
+}
+
+Camera::Result CameraImpl::select_camera(unsigned id)
```
JonasVautherin: Is the idea the following?

1. We instantiate a `camera` object
2. We select the desired camera
3. If needed, we can instantiate another `camera` object and set it to another camera

julianoes: I guess so. I agree it's ugly.

plugins/camera/camera_impl.cpp
julianoes: 'Cause MAVLink definition.

JonasVautherin: Fair enough :D

PR #682, opened 2019-02-22
Add codes to support cancel calibration, cancel mission download, and cancel mission upload.

douglaswsilva:
- Add support for cancel calibration in the SDK wrapper.
- Add support for cancel mission download in the SDK wrapper.
- Add support for cancel mission upload in the SDK wrapper.

### DIFF
--- a/backend/src/plugins/calibration/calibration_service_impl.h
+++ b/backend/src/plugins/calibration/calibration_service_impl.h
@@ -170,8 +170,8 @@ public:
     }
 
     grpc::Status Cancel(grpc::ServerContext * /* context */,
-                                const rpc::calibration::CancelCalibrationRequest * /* request */,
-                                rpc::calibration::CancelCalibrationResponse * /* response*/) override 
+                        const rpc::calibration::CancelCalibrationRequest * /* request */,
+                        rpc::calibration::CancelCalibrationResponse * /* response*/) override
     {
         _calibration.cancel_calibration();
         return grpc::Status::OK;
--- a/backend/src/plugins/calibration/calibration_service_impl.h
+++ b/backend/src/plugins/calibration/calibration_service_impl.h
@@ -170,8 +170,8 @@ public:
     }
 
     grpc::Status Cancel(grpc::ServerContext * /* context */,
-                                const rpc::calibration::CancelRequest * /* request */,
-                                rpc::calibration::CancelResponse * /* response */) override 
+                        const rpc::calibration::CancelRequest * /* request */,
+                        rpc::calibration::CancelResponse * /* response */) override
     {
         _calibration.cancel_calibration();
         return grpc::Status::OK;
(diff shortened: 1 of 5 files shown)


backend/src/plugins/calibration/calibration_service_impl.h
```diff
@@ -169,6 +169,14 @@ class CalibrationServiceImpl final : public rpc::calibration::CalibrationService
         return grpc::Status::OK;
     }
 
+    grpc::Status Cancel(grpc::ServerContext * /* context */,
+                                const rpc::calibration::CancelCalibrationRequest * /* request */,
```
JonasVautherin: I believe the indentation is wrong here. You probably want to run `make fix_style` to fix the indentation, otherwise the CI will fail :sweat_smile:.

backend/src/plugins/calibration/calibration_service_impl.h
```diff
@@ -169,6 +169,14 @@ class CalibrationServiceImpl final : public rpc::calibration::CalibrationService
         return grpc::Status::OK;
     }
 
+    grpc::Status Cancel(grpc::ServerContext * /* context */,
```
julianoes: I vote to stay as close as possible to the C++ API (or otherwise change the C++ API).
Therefore, I'd change this to `CancelCalibration`.

JonasVautherin: We will have a chance to slightly update the C++ API when we start auto-generating it (because that will more or less break the API anyway). So I would go for the better API rather than consistency with C++.

That's negociable to me, but I feel like `calibration.cancel()` works and `calibration.cancelCalibration()` is more verbose.

JonasVautherin: But that's something to be discussed [here](https://github.com/Dronecode/DronecodeSDK-Proto/pull/74).

JonasVautherin:
Now that [the API PR](https://github.com/Dronecode/DronecodeSDK-Proto/pull/74) has been merged, you need to use the `master` branch of `DronecodeSDK-Proto` here. For that, just update the `proto/` submodule to point to master.

Something like this:

```
cd path/to/dronecodesdk/proto
git fetch
git checkout master
cd ..
git add proto
```

douglaswsilva:
Done. 👍🏼

backend/src/plugins/calibration/calibration_service_impl.h
JonasVautherin: @julianoes: Are you fine with the solution merged [here](https://github.com/Dronecode/DronecodeSDK-Proto/pull/74) eventually?

julianoes:
You need to do a `make fix_style`.

JonasVautherin:
For some reason it doesn't work for him, I'll do it. I seem to have broken something on latest `develop`, I am working on it now and will merge this branch then.

douglaswsilva:
@JonasVautherin `clang-format version 8.0.0 (tags/google/stable/2018-12-18)`

PR #683, opened 2019-02-24
cmake: fix android build and update grpc dependencies

JonasVautherin:
* For some reason, Android now needs zlib to be compiled with -fPIC.
* Update grpc dependencies (BoringSSL was failing to build on Android, and we were stuck on 1 year old dependencies :sweat_smile:)

Builds on linux (default + android), macOS (default, ios, ios_simulator), passes the unit tests and apparently runs the examples fine on linux and iPad (e.g. telemetry, takeoff and land, mission, ...).

Notes:

* Android-NDK r19b requires [this fix](https://android-review.googlesource.com/c/platform/ndk/+/907834) to build. It will come with r19c.
* Still pointing to https://github.com/dronecore/grpc, but [I opened a PR](https://github.com/grpc/grpc/pull/18151) that may help us go back to using upstream.

### DIFF
--- a/backend/cmake/third_party/grpc/CMakeLists.txt
+++ b/backend/cmake/third_party/grpc/CMakeLists.txt
@@ -5,7 +5,7 @@ project(external-grpc)
 include(ExternalProject)
 
 set(ARG_GIT_REPOSITORY https://github.com/dronecore/grpc.git)
-set(ARG_GIT_TAG enable-android-ios-build)
+set(ARG_GIT_TAG 16856-fix-codegen-off-install)
 
 # This answer probably saved me from destroying my computer:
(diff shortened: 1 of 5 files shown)


backend/cmake/third_party/grpc/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@ project(external-grpc)
 include(ExternalProject)
 
 set(ARG_GIT_REPOSITORY https://github.com/dronecore/grpc.git)
-set(ARG_GIT_TAG enable-android-ios-build)
+set(ARG_GIT_TAG 16856-fix-codegen-off-install)
```
julianoes: :+1:

PR #684, opened 2019-02-25
fix for mission_raw plugin:  donwload mission handling 

A-l-e-k-s-G:
When there is no mission stored inside PX4 and the received mission count is 0, the process_mission_count() handler should respond with a mission ack instead of requesting a mission item.


JonasVautherin:
Oh, nice catch!

Would you mind writing an integration test for that?

PR #687, opened 2019-03-04
Add gimbal support 

douglaswsilva:
- Add gimbal support to SetPitchAndYaw.


JonasVautherin:
You forgot to fix style (which is why the CI is failing), but other than that, it looks good to me :+1:.

julianoes:
Cool, thx!

PR #692, opened 2019-03-08
backend: remove warnings from protobuf-generated files

JonasVautherin:
NOTE: That's apparently required only for the unit tests target, I'm not sure why it's not failing for the actual binary.


JonasVautherin:
@julianoes: I admin-merged that so that I can merge Doug's PRs.

PR #694, opened 2019-03-12
API changes: UID, discovery callbacks

### LINKED ISSUE #984: Switch to uid2
We should get away from the uid field which can have duplicates ASAP.

WIP is here: https://github.com/mavlink/MAVSDK/pull/694

julianoes:
This is work in progress to get some feedback and start a discussion on how the API to get systems and be notified about connected systems and components should look like.

Fixes #984.


gamoreno:
It looks fine, however it forces one to use callbacks for any case with more than one system. In my case, I wait for two drones to connect before anything else, so I do something like:
    while (dc.system_uuids().size() < 2) {
        cout << "Waiting for drones to connect" << endl;
        this_thread::sleep_for(chrono::seconds(1));
    }
With the proposed API, I'd be forced to use callbacks. However, that's ok. It's not that much problem.

What I think is a must is to add the method System::uint8_t get_sysid() const. Using the current UUIDs is a pain for me, so I'm glad they're being removed. However, I need to be able to identify the drones, which have been assigned different SYSIDs. I have to be able to tell which one is which.

Thanks

cmake/compiler_flags.cmake
```diff
@@ -10,7 +10,7 @@ else()
     # We are not using exceptions to make it easier to write wrappers.
     add_definitions(-fno-exceptions)
 
-    set(warnings "-Wall -Wextra -Werror -Wshadow -Wno-strict-aliasing -Wold-style-cast -Wdouble-promotion -Wformat=2 -Weffc++")
+    set(warnings "-Wall -Wextra -Wshadow -Wno-strict-aliasing -Wold-style-cast -Wdouble-promotion -Wformat=2 -Weffc++")
```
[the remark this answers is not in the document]
julianoes: No, thx.

julianoes:
@gamoreno thanks for the comment. You're right we need sync APIs without callbacks as well. I'll add them.

src/integration_tests/action_hover_sync.cpp
```diff
@@ -38,14 +38,19 @@ void takeoff_and_hover_at_altitude(float altitude_m)
...
 
-    System& system = dc.system();
-    auto telemetry = std::make_shared<Telemetry>(system);
+    auto systems = dc.systems();
+
+    ASSERT_EQ(systems.size(), 1);
+
+    auto system = systems.at(0);
+
+    auto telemetry = std::make_shared<Telemetry>(*system.lock());
```
julianoes: I'm pretty sure this is wrong because we create a temporary `std::shared_ptr` using `std::weak_ptr::lock()`. The temporary pointer is deleted again, so the reference count goes back to 1 (for MavsdkImpl who owns it).
The plugin is given a reference to the actual `System` and assumes it's valid when that's theoretically not true.

src/core/mavsdk_impl.h
```diff
@@ -29,6 +29,8 @@ class MavsdkImpl {
     ConnectionResult add_tcp_connection(const std::string& remote_ip, int remote_port);
     ConnectionResult add_serial_connection(const std::string& dev_path, int baudrate);
 
+    std::vector<std::weak_ptr<System>> systems() const;
```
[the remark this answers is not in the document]
julianoes: > At this point QGC could not connect on port 14540, for some reason that I don't know for sure

MAVSDK is bound to port 14540, therefore QGC can't also bind to it.

> I believe that QGC sends heartbeats, is that correct?

Only once it receives heartbeats from a drone.

> Does the spec say that the GCS must send heartbeats?

Probably but a GCS doesn't know where to send heartbeats to, so it needs to wait for a heartbeat arriving first.

> Wouldn't it work if Mavsdk discovered the drone, and if then System established the connection?

You could pass on the connection but again, the connection can be shared between multiple systems. You can have drone 1 sending heartbeats to 14540 and drone 2 sending heartbeats to 14540. When you answer back to the "remotes" you use the IP and port to address the correct one of the two.

> Why is it that QGC cannot connect after MAVSDK? Does the drone stop broadcasting and start sending directed UDP message once it receives the first message?

As long as the port is bound you can't have something else bind to it. Once MAVSDK is closed, QGC should be able to bind and connect again.

julianoes:
@JonasVautherin I've continued a bit here and I think that the API to get a system is now cleaner. I agree it's not what we eventually might want but I think it's nicer.

coderkalyan:
https://github.com/mavlink/MAVSDK/issues/1023 This is probably relevant? It's trying to achieve the same goals, in a generic way as well.

https://docs.google.com/document/d/1UKXxG3Ph_9Nqws4hybiA0PPDo6L2PUYLB7mp_EwvI54/edit doc here detailing proposal

julianoes:
@coderkalyan right this is also in the back of my mind.

src/plugins/action/include/plugins/action/action.h
```diff
@@ -38,6 +38,8 @@ class Action : public PluginBase {
      */
     explicit Action(System& system);
 
+    explicit Action(std::shared_ptr<System> system);
```
JonasVautherin, line 41: I'm not completely sure about the ownership story. So `Mavsdk` creates a list of `shared_ptr<System>`s that can outlive it. To each of those systems, `Mavsdk` gives a reference to a `unique_ptr<MavsdkImpl>`, meaning that a `System` must not outlive `Mavsdk`.

By giving a `System&` to the plugins, we were essentially saying: "Systems are owned by `Mavsdk`, so the plugins cannot outlive `Mavsdk` otherwise their reference is getting invalid". But by giving a `shared_ptr<System>`, we say "you own your destiny, do whatever you please", right? And that's not true, because a plugin cannot outlive `Mavsdk`, because each `System` still has a reference to `MavsdkImpl` :sweat_smile:.

Do I make any sense at all, or am I confused?

JonasVautherin:
@julianoes: I think that the CI fails because of some deprecation (`register_on_discover` should be `register_on_change`), but the new version is not implemented yet. Does that seem correct?

julianoes (commented):
Nice, thanks for doing all these changes!

src/core/mavsdk_impl.cpp
```diff
@@ -406,7 +417,7 @@ void MavsdkImpl::make_system_with_component(uint8_t system_id, uint8_t comp_id)
     // Make a system with its first component
     auto new_system = std::make_shared<System>(*this, system_id, comp_id, _is_single_system);
 
-    _systems.insert(system_entry_t(system_id, new_system));
+    _systems.insert(std::pair<uint8_t, std::shared_ptr<System>>(system_id, new_system));
```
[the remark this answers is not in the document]
JonasVautherin: @julianoes: ping on that :slightly_smiling_face:

julianoes:
@JonasVautherin we still need to change the examples, or ignore the deprecation warnings. I've added `subscribe_on_change`, would you mind checking if that makes sense to you?

JonasVautherin (changes requested):
@julianoes: It seems to me that we call [`set_connected`](https://github.com/mavlink/MAVSDK/blob/develop/src/core/system_impl.cpp#L37) in the `SystemImpl` constructor, which is created in the `System` constructor.

However, `set_connected` [calls `notify_on_discover()`](https://github.com/mavlink/MAVSDK/blob/develop/src/core/system_impl.cpp#L521), which calls `subscribe_on_change`.

As a result, when a new system is discovered, `subscribe_on_change` is called before the System is actually created and added to the `_systems` list. And that goes against what I ended doing naturally with this API, which is exactly querying the `mavsdk.systems()` in the `subscribe_on_change` callback, e.g.:

```c++
// Assuming that there is only one system anyway:
mavsdk.subscribe_on_change([&mavsdk]() {
    const auto system = mavsdk.systems().at(0);
});
```

This throws an exception because `mavsdk.systems()` is empty when the callback is triggered (those are the errors we see in the unit tests in CI).

Does that make sense?

julianoes:
> Does that make sense?

Yes, I would use it in the same way. That's definitely not good as is.

src/core/mavsdk_impl.cpp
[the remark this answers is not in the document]
julianoes: Isn't it the other way round? If a system with the same sysid already exists even though the uid is different, there won't be a notification.

julianoes:
@JonasVautherin I'm still trying to figure this out. It only calls `is_connected()` when `connected` is `true`, however, that's only when you add it with `add_udp_remote()`, I think. :thinking:

JonasVautherin:
We have 3 unit tests that are failing because of the way the mocks are initiated: the `Mavsdk` mock does not implement `systems()`, which crashes. But even if it did (I tried), the `Mavsdk` mock does not contain any `MavsdkImpl`, which needs to be passed to the `System` such that the notification can happen through `_parent.trigger_on_discover`.

I don't see a clean way to fix that, other than by adding a few mocks (`MavsdkImpl`, `System`, `SystemImpl`), but it feels a bit convoluted. @julianoes what do you think?

JonasVautherin (dismissed):

julianoes:
Note to self: pending docs.

JonasVautherin (dismissed):

src/core/system.h
```diff
@@ -103,6 +103,19 @@ class System {
      */
     uint8_t get_system_id() const;
 
+    /**
+     * @brief type for is connected callback.
+     */
+    using IsConnectedCallback = std::function<void(bool)>;
+
+    /**
+     * @brief Subscribe to callback to be called when system is discovered
```
JonasVautherin, line 112:
```suggestion
     * @brief Subscribe to callback to be called when system connection state changes
```

JonasVautherin (approved):
I think one comment is wrong, but LGTM otherwise :+1:

src/core/system.h
julianoes: What about the line below?

src/core/system.h
```diff
@@ -98,6 +103,19 @@ class System {
...
+     *        or times out.
```
JonasVautherin, line 113:
```suggestion
     * @brief Subscribe to callback to be called when system connection state changes.
```

JonasVautherin (approved):
I went through all the changes again, haven't found anything to complain about :+1:.

julianoes:
Thanks @JonasVautherin!

PR #696, opened 2019-03-14
plugin offboard: support for position-ned

elbarto1980:
Following up on our little chat about the opportunity to add offboard position control in the Local Cartesian (NED) frame (see: [here](http://discuss.px4.io/t/offboard-frames-question/8181))

### DIFF
--- a/integration_tests/offboard_position.cpp
+++ b/integration_tests/offboard_position.cpp
@@ -19,6 +19,8 @@ TEST_F(SitlTest, OffboardPositionNED)
     // Wait for system to connect via heartbeat.
     std::this_thread::sleep_for(std::chrono::seconds(2));
 
+    ASSERT_TRUE(dc.system().has_autopilot());
+
     System &system = dc.system();
     auto telemetry = std::make_shared<Telemetry>(system);
     auto action = std::make_shared<Action>(system);
--- a/integration_tests/offboard_position.cpp
+++ b/integration_tests/offboard_position.cpp
@@ -1,0 +1,109 @@
+#include <iostream>
+#include <cmath>
+#include "integration_test_helper.h"
+#include "dronecode_sdk.h"
+#include "plugins/action/action.h"
+#include "plugins/telemetry/telemetry.h"
+#include "plugins/offboard/offboard.h"
+#include "plugins/mission/mission.h"
+
+using namespace dronecode_sdk;
+
+TEST_F(SitlTest, OffboardPositionNED)
+{
+    DronecodeSDK dc;
+
+    ConnectionResult ret = dc.add_udp_connection();
+    ASSERT_EQ(ConnectionResult::SUCCESS, ret);
+
+    // Wait for system to connect via heartbeat.
+    std::this_thread::sleep_for(std::chrono::seconds(2));
+
+    System &system = dc.system();
+    auto telemetry = std::make_shared<Telemetry>(system);
+    auto action = std::make_shared<Action>(system);
+    auto offboard = std::make_shared<Offboard>(system);
+    auto mission = std::make_shared<Mission>(system);
+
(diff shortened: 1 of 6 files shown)


julianoes:
Great, thank you @elbarto1980. I'll give it a test in the next days.

JonasVautherin:
@elbarto1980: The CI is failing because of indentation issues. Can you run `make fix_style` and push the changes?

elbarto1980:
@JonasVautherin make fix_style returns this:
> Clang version 3 too old (required >= 6)
> Makefile:140: recipe for target 'fix_style' failed
> make: *** [fix_style] Error 1
And I don't know why, because I have gcc (GCC) 6.3.1 20161221 (Red Hat 6.3.1-1)

julianoes:
I can try to run it and add it to the PR, let me try.

julianoes:
Yay, I was able to push the style fixes to your fork @elbarto1980, that's great.

julianoes:
@elbarto1980 sorry for the delay here. I'm still hoping to test this in the next days!

JonasVautherin:
> Preprocessing /home/user/DronecodeSDK/install/include/dronec/home/user/DronecodeSDK/install/include/dronecode_sdk/plugins/offboard/offboard.h:85: warning: Compound dronecode_sdk::Offboard::PositionNEDYaw is not documented.

I think the CI is failing because of the documentation step.

elbarto1980:
Now "coverage/coveralls" is failing because of coverage decrease... but it doesn't seem to me that any of the files I modified are in the "changed" tab. Is that ok?

julianoes:
Yes, that's fine. Let me try to use it.

julianoes:
@elbarto1980 cool I tested this and it seems to fly around a do a circle :smile: so that's good enough for now, let's merge it.

In the future it would be nice to have one test that does the circle, one that does the interruption, and one for a simple square.

plugins/offboard/offboard_impl.cpp
```diff
@@ -106,6 +106,33 @@ void OffboardImpl::receive_command_result(MAVLinkCommands::Result result,
     }
 }
 
+void OffboardImpl::set_position_ned(Offboard::PositionNEDYaw position_ned_yaw)
+{
+    _mutex.lock();
```
julianoes: We should be using lock guards instead of manual lock and unlock but this needs to be changed in the whole plugin anyway.

integration_tests/offboard_position.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+    EXPECT_EQ(offboard_result, Offboard::Result::SUCCESS);
+
+    offboard->set_position_ned({0.0f, 0.0f, -10.0f, 90.0f});
+    std::this_thread::sleep_for(std::chrono::seconds(1));
+
+    {
+        const float radius = 10.0f;
+        const float step = 0.01f;
+        for (float i = 0.0f; i <= 2 * static_cast<float>(M_PI); i += step) {
```
julianoes: Two changes I would suggest here:

1. Rename `i` to `angle` because `i` is usually an integer, so this can be confusing.
2. And I would get rid of the cast, all you need is:
```
2.0f * M_PI_F
```
    or:
```
constexpr float full_circle = 2.0f * M_PI_F;
for (...; angle <= full_circle; ++angle) {
```
    or:
```
static_cast<float>(M_2_PI)
```

julianoes:
@elbarto1980 should I do the suggestions or do you have time? I don't mind either way.

elbarto1980:
> @elbarto1980 should I do the suggestions or do you have time? I don't mind either way.

Sure, I will push them in 5 minutes... :+1:

elbarto1980:
> Great, thank you @elbarto1980.

Thank you! :wink:

Issue #698, opened 2019-03-15
Cancel mission upload and download lead to crashes


julianoes:
According to @douglaswsilva: 

> cancel mission upload and download, which are working but sometimes crash the app. I had to add a 5 second delay between canceling and uploading mission in order to prevent the crash, so it might be some async weird behavior.

Stacktrace:

```
#1        0x000000012729b07c in bool dronecode_sdk::Any::is<unsigned char>() const at /Users/douglas/Documents/DronecodeSDK/build/ios/../../core/any.h:27
#2        0x000000012729adbc in dronecode_sdk::MAVLinkParameters::ParamValue::is_same_type(dronecode_sdk::MAVLinkParameters::ParamValue const&) const at /Users/douglas/Documents/DronecodeSDK/build/ios/../../core/mavlink_parameters.h:356
#3        0x0000000127295430 in dronecode_sdk::MAVLinkParameters::ParamValue::operator==(dronecode_sdk::MAVLinkParameters::ParamValue const&) const at /Users/douglas/Documents/DronecodeSDK/build/ios/../../core/mavlink_parameters.h:377
#4        0x00000001272c3ed0 in dronecode_sdk::CameraDefinition::get_possible_options(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, std::__1::vector<dronecode_sdk::MAVLinkParameters::ParamValue, std::__1::allocator<dronecode_sdk::MAVLinkParameters::ParamValue> >&) at /Users/douglas/Documents/DronecodeSDK/build/ios/../../plugins/camera/camera_definition.cpp:506
#5        0x00000001272944c0 in dronecode_sdk::CameraImpl::get_possible_options(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, std::__1::vector<dronecode_sdk::Camera::Option, std::__1::allocator<dronecode_sdk::Camera::Option> >&) at /Users/douglas/Documents/DronecodeSDK/build/ios/../../plugins/camera/camera_impl.cpp:1129
#6        0x000000012729768c in dronecode_sdk::CameraImpl::notify_possible_setting_options() at /Users/douglas/Documents/DronecodeSDK/build/ios/../../plugins/camera/camera_impl.cpp:1327
[...]

PR #699, opened 2019-03-18
camera: thread-safety fixes

julianoes:
This tries to robustify the thread-safety of the camera plugin. Some of the callbacks and some shared data were not protected by mutexes.
Also, all callbacks are not called from the thread pool workers reducing the chances for deadlocks.

This work is in response to the stacktrace in #698.

### DIFF
--- a/plugins/camera/camera_impl.cpp
+++ b/plugins/camera/camera_impl.cpp
@@ -70,12 +70,14 @@ void CameraImpl::deinit()
 
     {
         std::lock_guard<std::mutex> lock(_status.mutex);
-        _status.callback = nullptr;
+        _status.status_callback = nullptr;
+        _status.subscription_callback = nullptr;
     }
 
     {
-        std::lock_guard<std::mutex> lock(_get_mode.mutex);
-        _get_mode.callback = nullptr;
+        std::lock_guard<std::mutex> lock(_mode.mutex);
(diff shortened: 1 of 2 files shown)


plugins/camera/camera_impl.cpp
```diff
@@ -86,6 +88,22 @@ void CameraImpl::deinit()
     {
         std::lock_guard<std::mutex> lock(_video_stream_info.mutex);
         _video_stream_info.callback = nullptr;
+        _video_stream_info.subscription_callback = nullptr;
+    }
+
+    {
+        std::lock_guard<std::mutex> lock(_information.mutex);
```
JonasVautherin: I guess that can be commented out / removed too, can't it?

plugins/camera/camera_impl.h
```diff
@@ -199,12 +153,67 @@ class CameraImpl : public PluginImplBase {
...
+        void *timeout_cookie{nullptr};
+
+        Camera::subscribe_video_stream_info_callback_t subscription_callback{nullptr};
+        void *call_every_cookie{nullptr};
+    } _video_stream_info{};
+
+    struct {
+        std::mutex mutex{};
+        Camera::Information data{};
+        // Camera::get_information callback{nullptr};
```
JonasVautherin:
```suggestion
```

PR #700, opened 2019-03-19
core: get rid of termios2, use termios on Linux

julianoes:
In order to simplify and to enable the build in manylinux1 it makes sense to go back to termios instead of termios2.


JonasVautherin:
Do we lose something by doing that? e.g. the ability to use non-standard baudrates?

julianoes:
I don't think so.

julianoes:
Tested this on Linux and baudrate change doesn't seem to be working anymore. Needs more investigation.

julianoes:
Tested on Linux and macOS, now works for both.

JonasVautherin:
> [  FAILED  ] CurlTest.Curl_DownloadText_HTTPS_Success

Seems to me like it is failing on appveyor, and only in Debug, because of an unrelated curl test.

PR #704, opened 2019-03-21
Jenkins: bring Jenkinsfile back

julianoes:
Let's get Jenkins up and running again, FYI @dagar :smile:


julianoes:
@dagar any idea what this means?
```
Failed to run top '2919b615ed9770b2ec8a958788de323a6712074e6589fe6c5f9dcdd8c043ee08'. Error: Error response from daemon: rpc error: code = Unknown desc = containerd: container not found
```

dagar:
Long story, but newer Jenkins doesn't like the way we were hanlding entrypoint.sh. https://github.com/Dronecode/DronecodeSDK/blob/develop/docker/Dockerfile-Ubuntu-16.04#L54

See (aka copy) what was done in PX4/containers for detail. https://github.com/PX4/containers/blob/master/docker/px4-dev/scripts/entrypoint.sh
https://github.com/PX4/containers/commit/64154790cc53ea598d6222d28abf36358487cc4a#diff-8fe21b75f1d50e84b43f3c2125b20d20

@lamping7 FYI

julianoes:
Oh, right, so I need to change all of the docker stuff, ok.

Jenkinsfile
```diff
@@ -0,0 +1,379 @@
...
+            }
+          }
+          steps {
+            sh 'export'
+            sh 'git submodule deinit -f .'
+            sh 'git clean -ff -x -d .'
+            sh 'git submodule sync --recursive'
+            sh 'git submodule update --init --recursive --force'
+            sh 'ccache -z'
+            sh 'export INSTALL_PREFIX=`pwd`/install && make BUILD_TYPE=Release BUILD_BACKEND=1 INSTALL_PREFIX=$INSTALL_PREFIX dENABLE_MAVLINK_PASSTHROUGH=1 efault install'
```
JonasVautherin:
```suggestion
            sh 'export INSTALL_PREFIX=`pwd`/install && make BUILD_TYPE=Release BUILD_BACKEND=1 INSTALL_PREFIX=$INSTALL_PREFIX ENABLE_MAVLINK_PASSTHROUGH=1 default install'
```

julianoes:
@dagar this seems to work now but I'd appreciate a review regarding ccache. I don't think it's working.

julianoes:
According to @dagar we are lucky if ccache works because the ec2 instance has not been recycled yet. Otherwise that's what we have for now.

PR #706, opened 2019-03-22
Allow users to test health component separately

shrit:
Since the only way to know which prefilight check is
causing the problem is tp print the health struct, this way
will allow user to test each one separately without
print out the struct

Signed-off-by: Omar Shrit <shrit@lri.fr>


shrit:
OK, convinced !!

PR #715, opened 2019-04-02
Overload operator<< to print position and velocity ned

shrit:
Signed-off-by: Omar Shrit <shrit@lri.fr>


shrit:
You are welcome !

PR #722, opened 2019-04-11
Fix Windows build

julianoes:
This fixes the missing include and curl linking issues. FYI: @MaEtUgR


MaEtUgR:
Looks good, the include was missing like I wrote here: https://github.com/Dronecode/DronecodeSDK/issues/718#issue-429710976
I used the old curl version from the guide but it's better to use the new one and maybe we could even get a prebuilt library...

PR #725, opened 2019-04-18
mavlink_parameters: fix timeout for 16 char params

julianoes:
This fixes a char array overflow in strlen in string compare. This happened because the case where the MAVLink param name has length 16 and is not handled correctly. The MAVLink field is not 0 terminated which means that we need to copy to bytes to a 0 terminated buffer first before using it.

Thanks for reporting @douglaswsilva!

### DIFF
--- a/core/mavlink_parameters.cpp
+++ b/core/mavlink_parameters.cpp
@@ -1,5 +1,6 @@
 #include "mavlink_parameters.h"
 #include "system_impl.h"
+#include <cstring>
 #include <future>
 
 namespace dronecode_sdk {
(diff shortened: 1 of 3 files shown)


hamishwillee:
According to the [PARAM_VALUE](https://mavlink.io/en/messages/common.html#PARAM_VALUE) id is terminated by NULL if the length is less than 16 human-readable chars and WITHOUT null termination (NULL) byte if the length is exactly 16 chars. I think you're saying it is not NULL terminated int he 16 char case.

julianoes:
Right, it's at least not guaranteed to be null terminated.

Issue #719, opened 2019-04-07
Windows build is broken


julianoes:
Currently the Windows build seems broken, I tried in "x64 Native Tools Command Prompt for VS 2017". I'm not sure why this is not caught in appveyor.

```
"C:\DroneCore\build\ALL_BUILD.vcxproj" (default target) (1) ->
"C:\DroneCore\build\core\dronecode_sdk.vcxproj" (default target) (3) ->
(ClCompile target) ->
  c:\dronecore\core\locked_queue.h(27): error C2220: warning treated as error - no 'object' file generated [C:\DroneCor
e\build\core\dronecode_sdk.vcxproj]
  c:\dronecore\core\locked_queue.h(27): error C2061: syntax error: identifier 'iterator' [C:\DroneCore\build\core\drone
code_sdk.vcxproj]
  c:\dronecore\core\locked_queue.h(27): error C2334: unexpected token(s) preceding '{'; skipping apparent function body
 [C:\DroneCore\build\core\dronecode_sdk.vcxproj]
  c:\dronecore\core\locked_queue.h(29): error C2061: syntax error: identifier 'iterator' [C:\DroneCore\build\core\drone
code_sdk.vcxproj]
  c:\dronecore\core\locked_queue.h(29): error C2334: unexpected token(s) preceding '{'; skipping apparent function body
 [C:\DroneCore\build\core\dronecode_sdk.vcxproj]
  c:\dronecore\core\locked_queue.h(31): error C2061: syntax error: identifier 'iterator' [C:\DroneCore\build\core\drone
code_sdk.vcxproj]
  c:\dronecore\core\locked_queue.h(31): error C2334: unexpected token(s) preceding '{'; skipping apparent function body
 [C:\DroneCore\build\core\dronecode_sdk.vcxproj]
  c:\dronecore\core\locked_queue.h(27): error C2220: warning treated as error - no 'object' file generated [C:\DroneCor
e\build\core\dronecode_sdk.vcxproj]
  c:\dronecore\core\locked_queue.h(27): error C2061: syntax error: identifier 'iterator' [C:\DroneCore\build\core\drone
code_sdk.vcxproj]
  c:\dronecore\core\locked_queue.h(27): error C2334: unexpected token(s) preceding '{'; skipping apparent function body
 [C:\DroneCore\build\core\dronecode_sdk.vcxproj]
[...]

PR #729, opened 2019-04-24
core: fix MSVC compiler error

julianoes:
Somehow, on one machine these changes were needed in order to avoid MSVC complaining about:
```
warning C4346: 'iterator': dependent name is not a type
```


julianoes:
Fixed #719.

MaEtUgR:
I have no idea why, good that you found this out.

julianoes:
Neither do I!

PR #730, opened 2019-04-24
Enable backend to accept connection URL

### LINKED ISSUE #587: Enable serial and TCP connections for backend
The backend currently only supports UDP and the port is hardcoded for the backend_bin.

We should use `add_any_connection`, so that we can pass any connection URI.

julianoes:
This allows the backend_bin to be started with any connection URL.

Resolves #587.

### DIFF
--- a/backend/src/backend.cpp
+++ b/backend/src/backend.cpp
@@ -14,9 +14,9 @@ public:
     Impl() {}
     ~Impl() {}
 
-    void connect(const int mavlink_listen_port)
+    void connect(const std::string &connection_url)
     {
-        _connection_initiator.start(_dc, 14540);
+        _connection_initiator.start(_dc, connection_url);
         _connection_initiator.wait();
     }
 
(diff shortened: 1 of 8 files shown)


julianoes:
Ignoring CI error, took note in #731.

julianoes:
@JonasVautherin can we merge it?

JonasVautherin:
Wait, what's this error on Fedora? :sweat_smile: 

```
../../backend/test/camera_service_impl_test.cpp:436: Failure

  modes.size()
    Which is: 6
  received_modes.size()
    Which is: 7
Sending interrupt signal to process
```

I guess that's independent from this PR.

douglaswsilva:
@JonasVautherin Hey, it is indeed breaking the IOS SDK. Any Idea how we can fix it?

JonasVautherin:
Good point! Let's track this in the [Swift SDK repo](https://github.com/Dronecode/DronecodeSDK-Swift/issues/121).

PR #733, opened 2019-04-26
Add Status Text

douglaswsilva:
Get status text through the SDK.

### DIFF
--- a/core/mavlink_parameters.cpp
+++ b/core/mavlink_parameters.cpp
@@ -114,13 +114,14 @@ MAVLinkParameters::get_param(const std::string &name, ParamValue value_type, boo
     auto prom = std::promise<std::pair<Result, MAVLinkParameters::ParamValue>>();
     auto res = prom.get_future();
 
-    get_param_async(name,
-                    value_type,
-                    [&prom](Result result, ParamValue value) {
-                        prom.set_value(std::make_pair<>(result, value));
-                    },
-                    this,
-                    extended);
+    get_param_async(
+        name,
+        value_type,
+        [&prom](Result result, ParamValue value) {
+            prom.set_value(std::make_pair<>(result, value));
+        },
+        this,
+        extended);
 
     return res.get();
 }
--- a/core/mavlink_parameters.cpp
+++ b/core/mavlink_parameters.cpp
@@ -114,14 +114,13 @@ MAVLinkParameters::get_param(const std::string &name, ParamValue value_type, boo
     auto prom = std::promise<std::pair<Result, MAVLinkParameters::ParamValue>>();
     auto res = prom.get_future();
 
-    get_param_async(
-        name,
-        value_type,
-        [&prom](Result result, ParamValue value) {
-            prom.set_value(std::make_pair<>(result, value));
-        },
-        this,
-        extended);
+    get_param_async(name,
+                    value_type,
+                    [&prom](Result result, ParamValue value) {
+                        prom.set_value(std::make_pair<>(result, value));
+                    },
+                    this,
+                    extended);
 
     return res.get();
 }
--- a/core/mavlink_parameters.cpp
+++ b/core/mavlink_parameters.cpp
@@ -30,6 +30,7 @@ MAVLinkParameters::~MAVLinkParameters()
 void MAVLinkParameters::set_param_async(const std::string &name,
                                         const ParamValue &value,
                                         set_param_callback_t callback,
+                                        const void *cookie,
                                         bool extended)
 {
     // if (value.is_float()) {
@@ -52,6 +53,7 @@ void MAVLinkParameters::set_param_async(const std::string &name,
     new_work.param_name = name;
     new_work.param_value = value;
     new_work.extended = extended;
+    new_work.cookie = cookie;
 
     _work_queue.push_back(new_work);
 }
@@ -62,7 +64,8 @@ MAVLinkParameters::set_param(const std::string &name, const ParamValue &value, b
     auto prom = std::promise<Result>();
     auto res = prom.get_future();
 
-    set_param_async(name, value, [&prom](Result result) { prom.set_value(result); }, extended);
+    set_param_async(
+        name, value, [&prom](Result result) { prom.set_value(result); }, this, extended);
 
     return res.get();
 }
@@ -70,6 +73,7 @@ MAVLinkParameters::set_param(const std::string &name, const ParamValue &value, b
 void MAVLinkParameters::get_param_async(const std::string &name,
                                         ParamValue value_type,
                                         get_param_callback_t callback,
+                                        const void *cookie,
                                         bool extended)
 {
     // LogDebug() << "getting param " << name << ", extended: " << (extended ? "yes" : "no");
@@ -98,6 +102,7 @@ void MAVLinkParameters::get_param_async(const std::string &name,
     new_work.param_name = name;
     new_work.param_value = value_type;
     new_work.extended = extended;
+    new_work.cookie = cookie;
 
     _work_queue.push_back(new_work);
 }
@@ -113,9 +118,23 @@ MAVLinkParameters::get_param(const std::string &name, ParamValue value_type, boo
                     [&prom](Result result, ParamValue value) {
                         prom.set_value(std::make_pair<>(result, value));
                     },
+                    this,
                     extended);
 
     return res.get();
+}
+
+void MAVLinkParameters::cancel_all_param(const void *cookie)
+{
+    LockedQueue<WorkItem>::Guard work_queue_guard(_work_queue);
+
+    for (auto item = _work_queue.begin(); item != _work_queue.end(); /* manual incrementation */) {
+        if (item->get()->cookie == cookie) {
+            item = _work_queue.erase(item);
+        } else {
+            ++item;
+        }
+    }
 }
 
 void MAVLinkParameters::do_work()
--- a/core/mavlink_parameters.cpp
+++ b/core/mavlink_parameters.cpp
@@ -162,8 +162,8 @@ void MAVLinkParameters::do_work()
                 work->param_value.get_128_bytes(param_value_buf);
 
                 // FIXME: extended currently always go to the camera component
-                mavlink_msg_param_ext_set_pack(GCSClient::system_id,
-                                               GCSClient::component_id,
+                mavlink_msg_param_ext_set_pack(_parent.get_own_system_id(),
+                                               _parent.get_own_component_id(),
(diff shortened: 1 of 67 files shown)


douglaswsilva:
Btw, will wait until https://github.com/Dronecode/DronecodeSDK-Proto/pull/82 get's merged. 
Also, there are a couple of Keywords `Anotacao` in the code which means I'd like your guys' input.

plugins/telemetry/telemetry_impl.cpp
```diff
@@ -152,6 +156,13 @@ Telemetry::Result TelemetryImpl::set_rate_in_air(double rate_hz)
         _parent->set_msg_rate(MAVLINK_MSG_ID_EXTENDED_SYS_STATE, rate_hz));
 }
 
+Telemetry::Result TelemetryImpl::set_rate_status_text(double rate_hz)
+{
+    return telemetry_result_from_command_result(_parent->set_msg_rate(
+        MAVLINK_MSG_ID_EXTENDED_SYS_STATE,
+        rate_hz)); // Anotacao: What rate to set here? MAVLINK_MSG_ID_STATUSTEXT?
```
JonasVautherin: @julianoes

plugins/telemetry/telemetry_impl.cpp
```diff
@@ -228,6 +239,15 @@ void TelemetryImpl::set_rate_in_air_async(double rate_hz, Telemetry::result_call
         std::bind(&TelemetryImpl::command_result_callback, std::placeholders::_1, callback));
 }
 
+void TelemetryImpl::set_rate_status_text_async(double rate_hz,
+                                               Telemetry::result_callback_t callback)
+{
+    _parent->set_msg_rate_async(
+        MAVLINK_MSG_ID_EXTENDED_SYS_STATE, // Anotacao: What rate to set here?
```
JonasVautherin: @julianoes

plugins/telemetry/telemetry_impl.cpp
```diff
@@ -489,6 +509,39 @@ void TelemetryImpl::process_heartbeat(const mavlink_message_t &message)
...
+            break;
+        case MAV_SEVERITY_CRITICAL:
+            type = Telemetry::StatusText::StatusType::CRITICAL;
+            break;
+        default:
+            type = Telemetry::StatusText::StatusType::INFO;
+            break;
+    }
+
+    // statustext.text is not null terminated, therefore we copy it first to
```
JonasVautherin: Are we sure about that? Looks like something that could be inspired from somewhere else, so just wanting to confirm for `statustext.text` :)

core/mavlink_parameters.cpp
```diff
@@ -114,13 +114,14 @@ MAVLinkParameters::get_param(const std::string &name, ParamValue value_type, boo
...
     auto res = prom.get_future();
 
-    get_param_async(name,
-                    value_type,
-                    [&prom](Result result, ParamValue value) {
-                        prom.set_value(std::make_pair<>(result, value));
-                    },
-                    this,
-                    extended);
+    get_param_async(
```
julianoes: The formatting by the newer clang-format version is annoying but actually nicer :smile:

integration_tests/telemetry_async.cpp
```diff
@@ -40,6 +41,7 @@ static bool _received_ground_speed = false;
 static bool _received_gps_info = false;
 static bool _received_battery = false;
 static bool _received_rc_status = false;
+static bool _received_status_text = false;
```
julianoes: I don't know if we should check for this because it's not really guaranteed that we receive any.

integration_tests/telemetry_async.cpp
```diff
@@ -132,6 +136,7 @@ TEST_F(SitlTest, TelemetryAsync)
     EXPECT_TRUE(_received_gps_info);
     EXPECT_TRUE(_received_battery);
     EXPECT_TRUE(_received_rc_status);
+    EXPECT_TRUE(_received_status_text);
```
julianoes: Therefore, I'd remove this check.

plugins/telemetry/include/plugins/telemetry/telemetry.h
```diff
@@ -129,6 +129,11 @@ class Telemetry : public PluginBase {
                                      5: RTK float, 6: RTK fixed). */
     };
 
+    struct StatusText {
```
julianoes: This needs a docstring.

plugins/telemetry/include/plugins/telemetry/telemetry.h
```diff
@@ -129,6 +129,11 @@ class Telemetry : public PluginBase {
...
+        enum class StatusType { INFO, WARNING, CRITICAL } type;
```
julianoes: Also the enum needs a doc string.

plugins/telemetry/include/plugins/telemetry/telemetry.h
```diff
@@ -246,6 +251,14 @@ class Telemetry : public PluginBase {
      */
     Result set_rate_in_air(double rate_hz);
 
+    /**
+     * @brief Set rate of status text updates (synchronous).
+     *
+     * @param rate_hz Rate in Hz.
+     * @return Result of request.
+     */
+    Result set_rate_status_text(double rate_hz);
```
julianoes: I would not implement this. Statustext come when they happen. They don't really have a rate.

plugins/telemetry/telemetry_impl.cpp
julianoes: As written above, not necessary at all.

plugins/telemetry/telemetry_impl.cpp
julianoes: Not needed.

plugins/telemetry/telemetry_impl.cpp
```diff
@@ -489,6 +509,39 @@ void TelemetryImpl::process_heartbeat(const mavlink_message_t &message)
...
+
+    switch (statustext.severity) {
+        case MAV_SEVERITY_WARNING:
+            type = Telemetry::StatusText::StatusType::WARNING;
+            break;
+        case MAV_SEVERITY_CRITICAL:
+            type = Telemetry::StatusText::StatusType::CRITICAL;
+            break;
+        default:
+            type = Telemetry::StatusText::StatusType::INFO;
```
julianoes: I guess that's a good default for now to just use `INFO` if we don't know it. However, I could catch `INFO` properly in a `case` and then for default emit a `LogWarn() << "Unknown StatusText severity"` so that we are warned if we ever get a severity that we we don't expect.

plugins/telemetry/telemetry_impl.cpp
julianoes: Unfortunately that's true:
https://mavlink.io/en/messages/common.html#STATUSTEXT

And even if it was not true, it's a good precaution anyway.

plugins/telemetry/telemetry_impl.cpp
```diff
@@ -489,6 +509,39 @@ void TelemetryImpl::process_heartbeat(const mavlink_message_t &message)
...
+            break;
+    }
+
+    // statustext.text is not null terminated, therefore we copy it first to
+    // an array big enough that is zeroed.
+    char text_with_null[sizeof(statustext.text) + 1]{};
+    memcpy(text_with_null, statustext.text, sizeof(statustext.text));
+
+    std::string text = text_with_null;
```
julianoes: You can make this:
```
const std::string text = text_with_null;
```

plugins/telemetry/telemetry_impl.cpp
douglaswsilva: Cool. I thought about leaving a UNKNOWN status type as a default, but then I'd have to change the proto file. For now I'm doing something like this:
```
    switch (statustext.severity) {
        case MAV_SEVERITY_WARNING:
            type = Telemetry::StatusText::StatusType::WARNING;
            break;
        case MAV_SEVERITY_CRITICAL:
            type = Telemetry::StatusText::StatusType::CRITICAL;
            break;
        case MAV_SEVERITY_INFO:
            type = Telemetry::StatusText::StatusType::INFO;
            break;
        default:
            LogWarn() << "Unknown StatusText severity";
            type = Telemetry::StatusText::StatusType::INFO;
            break;
    }
```
I still set type to info if we get an unknown value, since we'd like to see it in our app. 
Let me know if I should leave it like this, or actually set the type as UNKNOWN.

julianoes: Yes exactly, that's what I meant.

plugins/telemetry/telemetry_impl.h
```diff
@@ -30,7 +30,7 @@ class TelemetryImpl : public PluginImplBase {
     Telemetry::Result set_rate_position(double rate_hz);
     Telemetry::Result set_rate_home_position(double rate_hz);
     Telemetry::Result set_rate_in_air(double rate_hz);
-    Telemetry::Result set_rate_status_text(double rate_hz);
+    // Telemetry::Result set_rate_status_text(double rate_hz); Anotacao
```
julianoes: I think you can delete these things.

douglaswsilva: Hm, I did. Not sure why it's still showing here. 🤔

douglaswsilva: Now it seems to show all my old "Anotacao" messages. Will check how to fix it.

douglaswsilva:
@julianoes I just tested locally and it builds fine. Just updated everything from develop, let's see if it will build.

julianoes:
@douglaswsilva ok I've squashed your commits and removed the unrelated style changes (I have the same problem on Arch Linux though).

Let's see if CI is happy.

julianoes:
@douglaswsilva ok it doesn't build yet if you include the backend:
```
make BUILD_BACKEND=1
```

julianoes:
Presumably the proto submodule was not updated. I've checked out master of the submodule and commited that.

JonasVautherin:
Oh, still missing documentation:

> Generating docs for nested compound dronecode_sdk::Offboard::Attit/tmp/jenkins/workspace/Dronecode_DronecodeSDK_PR-733/install/include/dronecode_sdk/plugins/telemetry/telemetry.h:147: warning: Member type (variable) of class dronecode_sdk::Telemetry::StatusText is not documented.

douglaswsilva:
@JonasVautherin I added the description (actually it seems like one of you guys had done it already), but let me know if still doesn't work.

PR #735, opened 2019-04-29
Add gRPC backend support for Offboard plugin

### LINKED ISSUE #702: Add offboard plugin to backend and language wrappers
It would be nice to add support for offboard to the language bindings.

cswkim:
Also add plugin mock and backend test (which is currently missing - WIP):

`backend/test/CMakeLists.txt` (to be modified)
`backend/test/offboard_service_impl_test.cpp` (to be added)

Resolves #702.

### DIFF
--- a/backend/src/core/core_service_impl.h
+++ b/backend/src/core/core_service_impl.h
@@ -43,16 +43,16 @@ public:
                        const rpc::core::ListRunningPluginsRequest * /* request */,
                        dronecode_sdk::rpc::core::ListRunningPluginsResponse *response) override
     {
-        std::string plugin_names[9] = {"action",
-                                       "calibration",
-                                       "camera",
-                                       "core",
-                                       "gimbal",
-                                       "info",
-                                       "mission",
-                                       "offboard",
-                                       "param",
-                                       "telemetry"};
+        std::string plugin_names[10] = {"action",
+                                        "calibration",
+                                        "camera",
+                                        "core",
+                                        "gimbal",
+                                        "info",
+                                        "mission",
+                                        "offboard",
+                                        "param",
+                                        "telemetry"};
 
         for (const auto plugin_name : plugin_names) {
             auto plugin_info = response->add_plugin_info();
--- a/backend/src/core/core_service_impl.h
+++ b/backend/src/core/core_service_impl.h
@@ -50,6 +50,7 @@ public:
                                        "gimbal",
                                        "info",
                                        "mission",
+                                       "offboard",
                                        "param",
                                        "telemetry"};
 
--- a/backend/src/plugins/offboard/offboard_service_impl.h
+++ b/backend/src/plugins/offboard/offboard_service_impl.h
@@ -72,8 +72,8 @@ public:
         return grpc::Status::OK;
     }
 
-    static dronecode_sdk::Offboard::AttitudeRate translateRPCAttitudeRate(
-        const rpc::offboard::AttitudeRate &rpc_attitude_rate)
+    static dronecode_sdk::Offboard::AttitudeRate
+    translateRPCAttitudeRate(const rpc::offboard::AttitudeRate &rpc_attitude_rate)
     {
         dronecode_sdk::Offboard::AttitudeRate attitude_rate;
 
@@ -90,16 +90,16 @@ public:
                                 rpc::offboard::SetPositionNedResponse * /* response */) override
     {
         if (request != nullptr) {
-            auto requested_position_ned_yaw = translateRPCPositionNEDYaw(
-                request->position_ned_yaw());
+            auto requested_position_ned_yaw =
+                translateRPCPositionNEDYaw(request->position_ned_yaw());
             _offboard.set_position_ned(requested_position_ned_yaw);
         }
 
         return grpc::Status::OK;
     }
 
-    static dronecode_sdk::Offboard::PositionNEDYaw translateRPCPositionNEDYaw(
-        const rpc::offboard::PositionNEDYaw &rpc_position_ned_yaw)
+    static dronecode_sdk::Offboard::PositionNEDYaw
+    translateRPCPositionNEDYaw(const rpc::offboard::PositionNEDYaw &rpc_position_ned_yaw)
     {
         dronecode_sdk::Offboard::PositionNEDYaw position_ned_yaw;
 
@@ -116,8 +116,8 @@ public:
                                  rpc::offboard::SetVelocityBodyResponse * /* response */) override
     {
         if (request != nullptr) {
-            auto requested_velocity_body_yawspeed = translateRPCVelocityBodyYawspeed(
-                request->velocity_body_yawspeed());
+            auto requested_velocity_body_yawspeed =
+                translateRPCVelocityBodyYawspeed(request->velocity_body_yawspeed());
             _offboard.set_velocity_body(requested_velocity_body_yawspeed);
         }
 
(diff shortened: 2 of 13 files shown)


cswkim:
The backend unit test is still in progress, but I've just added a few items to that and would like some feedback before continuing. Disclaimer - I have very limited experience with `C++` (and that was MANY years ago) and absolutely no experience using Googletest. I tried piecing together the test file for offboard by looking at the other backend test files, which might have confused me even more hah. I have some general questions, I'm not sure a PR comments thread is the appropriate place for them (probably better suited for the forum), but I'll leave them here for now:

- If a function does not take any arguments, is it required (or common) to have a test for a null request like [so](https://github.com/Dronecode/DronecodeSDK/blob/develop/backend/test/action_service_impl_test.cpp#L60-L67)?
- Likewise, if a function does not return anything, is it required/common to have a test for a null response like [so](https://github.com/Dronecode/DronecodeSDK/blob/develop/backend/test/camera_service_impl_test.cpp#L448-L456)? This and the above seem like common patterns.
- When dealing with custom type arguments, like for `set_video_stream_settings(VideoStreamSettings)` in the camera service, there is a test [here](https://github.com/Dronecode/DronecodeSDK/blob/develop/backend/test/camera_service_impl_test.cpp#L473-L483) that involves translating from rpc. I'm pretty clueless as to why this translation is needed. The backend camera service defines the translations [here](https://github.com/Dronecode/DronecodeSDK/blob/develop/backend/src/plugins/camera/camera_service_impl.h#L205-L236), does every struct for Offboard (AttitudeRate, PositionNEDYaw, VelocityBodyYawspeed, VelocityNEDYaw) need to/from translate functions?
- I'm also unsure about the use of `_allocated_` and `.release()` syntax [here](https://github.com/Dronecode/DronecodeSDK/blob/develop/backend/test/camera_service_impl_test.cpp#L480). Are these only used with rpc translations?

JonasVautherin:
> If a function does not take any arguments, is it required (or common) to have a test for a null request like so?

My opinion is that if the code needs to handle "invalid" arguments, it has to be tested. There is logic specific to receiving a `nullptr`, so I write a test for it. So that if somebody ever removes this logic, the test will fail.

> Likewise, if a function does not return anything, is it required/common to have a test for a null response like so? This and the above seem like common patterns.

Same as above. If not handling that in the implementation may result a crash, then we want to handle it. And if we want to handle it, we want a test to ensure that we actually do :-).

> When dealing with custom type arguments, like for set_video_stream_settings(VideoStreamSettings) in the camera service, there is a test here that involves translating from rpc. I'm pretty clueless as to why this translation is needed. The backend camera service defines the translations here, does every struct for Offboard (AttitudeRate, PositionNEDYaw, VelocityBodyYawspeed, VelocityNEDYaw) need to/from translate functions?

That's a good point. The thing is that we basically convert an enum into another one, using their `int` value. If the C++ implementation suddenly inverts two elements of the enum, their `int` value will change and the conversion will be wrong. So there is this test that `UNKNOWN` stays `UNKNOWN` and doesn't become, say, `DENIED` during the conversion.

> I'm also unsure about the use of _allocated_ and .release() syntax here. Are these only used with rpc translations?

`_allocated_` comes from protobuf: that's related to how they manage the memory. Basic types like booleans are not `_allocated_`, but structs are. And that's how we deal with them [in the autogeneration](https://github.com/Dronecode/DronecodeSDK-Proto/blob/master/pb_plugins/protoc_gen_dcsdk/type_info.py#L98-L101): we hardcode what is a "primitive" type and what is "allocated" ("11" is a [`TYPE_MESSAGE`](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/descriptor.proto#L159) and "14" is a [`TYPE_ENUM`](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/descriptor.proto#L164)).

`.release()` is probably for the [unique_ptr](https://en.cppreference.com/w/cpp/memory/unique_ptr/release): we release the ownership and pass the pointer to protobuf, which takes a raw pointer.

cswkim:
Thanks for all the help @JonasVautherin ! I've been so wrapped up in trying to understand the code that I haven't been able to test yet 😆 

Yeah I need to update the protos submodule to include the `offboard.proto`. After I've added the remaining testing code, I'll make sure the backend unit test succeeds and play around with the python wrappers before moving this PR from draft to ready for review.

backend/src/core/core_service_impl.h
```diff
@@ -50,6 +50,7 @@ class CoreServiceImpl final : public dronecode_sdk::rpc::core::CoreService::Serv
                                        "gimbal",
                                        "info",
                                        "mission",
+                                       "offboard",
```
julianoes: You need to bump the size of the array to 10.
@JonasVautherin we might as well change this to a `std::vector` with initializer list.

backend/src/plugins/offboard/offboard_service_impl.h
```diff
@@ -0,0 +1,116 @@
+#include "plugins/offboard/offboard.h"
+#include "offboard/offboard.grpc.pb.h"
+
+namespace dronecode_sdk {
+namespace backend {
+
+template<typename Offboard = Offboard>
+class OffboardServiceImpl final : public rpc::offboard::OffboardService::Service {
```
julianoes: I think you need the full namespace here:
```
dronecode_sdk::rpc::offboard::OffboardService::Service
```

backend/src/plugins/offboard/offboard_service_impl.h
```diff
@@ -0,0 +1,116 @@
...
+public:
+    OffboardServiceImpl(Offboard &offboard) : _offboard(offboard) {}
+
+    template<typename ResponseType>
+    void fillResponseWithResult(ResponseType *response,
+                                dronecode_sdk::Offboard::Result &offboard_result) const
+    {
+        auto rpc_result = static_cast<rpc::offboard::OffboardResult::Result>(offboard_result);
```
julianoes: Same here, and everywhere below.

backend/src/plugins/offboard/offboard_service_impl.h
```diff
@@ -0,0 +1,116 @@
...
+        auto *rpc_offboard_result = new rpc::offboard::OffboardResult();
+        rpc_offboard_result->set_result(rpc_result);
+        rpc_offboard_result->set_result_str(dronecode_sdk::Offboard::result_str(offboard_result));
+
+        response->set_allocated_offboard_result(rpc_offboard_result);
+    }
+
+    grpc::Status Start(grpc::ServerContext * /* context */,
+                       const rpc::Offboard::StartRequest * /* request */,
+                       rpc::Offboard::StartResponse *response) override
```
julianoes: And: `Offboard` -> `offboard`.

backend/src/plugins/offboard/offboard_service_impl.h
```diff
@@ -0,0 +1,116 @@
...
+        if (response != nullptr) {
+            auto is_active = _offboard.is_active();
+            response->set_is_active(is_active);
+        }
+
+        return grpc::Status::OK;
+    }
+
+    grpc::Status SetAttitudeRate(grpc::ServerContext * /* context */,
+                                 const rpc::Offboard::SetAttitudeRateRequest * /* request */,
```
julianoes: And then you need to follow the compiler warnings here. E.g. `request` is required in this method.

backend/src/plugins/offboard/offboard_service_impl.h
JonasVautherin: That's weird, it is not required e.g. [here](https://github.com/Dronecode/DronecodeSDK/blob/develop/backend/src/plugins/action/action_service_impl.h#L8), for some reason :thinking:

julianoes: Ok, maybe it works without and the problem was just the `Offboard` vs `offboard`.

JonasVautherin: I would try that first :ok_hand:

PR #738, opened 2019-05-01
Add NO_SETPOINT_SET result to Offboard plugin

### LINKED ISSUE #732: Add `NO_SETPOINT_SET` To Offboard Results
`Offboard::result_str` and `OffboardImpl::offboard_result_from_command_result` are missing `NO_SETPOINT_SET`.

cswkim:
I think the 2 changed files are the only places that need updating, but I'm not 100% sure. If someone could please verify.

Resolves #732

### DIFF
--- a/core/mavlink_commands.h
+++ b/core/mavlink_commands.h
@@ -24,6 +24,7 @@ public:
         COMMAND_DENIED,
         TIMEOUT,
         IN_PROGRESS,
+        NO_SETPOINT_SET,
         UNKNOWN_ERROR
     };
 
--- a/core/mavlink_commands.h
+++ b/core/mavlink_commands.h
@@ -24,7 +24,6 @@ public:
         COMMAND_DENIED,
         TIMEOUT,
         IN_PROGRESS,
-        NO_SETPOINT_SET,
         UNKNOWN_ERROR
     };
 
--- a/plugins/offboard/offboard.cpp
+++ b/plugins/offboard/offboard.cpp
@@ -67,6 +67,8 @@ const char *Offboard::result_str(Result result)
             return "Command denied";
         case Result::TIMEOUT:
             return "Timeout";
+        case Result::NO_SETPOINT_SET:
+            return "No setpoint set";
         case Result::UNKNOWN:
         default:
             return "Unknown";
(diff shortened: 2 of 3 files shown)


plugins/offboard/offboard_impl.cpp
```diff
@@ -459,6 +459,8 @@ Offboard::Result OffboardImpl::offboard_result_from_command_result(MAVLinkComman
             return Offboard::Result::COMMAND_DENIED;
         case MAVLinkCommands::Result::TIMEOUT:
             return Offboard::Result::TIMEOUT;
+        case MAVLinkCommands::Result::NO_SETPOINT_SET:
```
julianoes: But `MAVLinkCommands::Result::NO_SETPOINT_SET` does not exist:
https://github.com/Dronecode/DronecodeSDK/blob/4332994c55c9ac8357954ccd1afcc5fa4a87172a/core/mavlink_commands.h#L19-L28

plugins/offboard/offboard.cpp
```diff
@@ -67,6 +67,8 @@ const char *Offboard::result_str(Result result)
             return "Command denied";
         case Result::TIMEOUT:
             return "Timeout";
+        case Result::NO_SETPOINT_SET:
+            return "No setpoint set";
```
julianoes: This one is correct.

plugins/offboard/offboard_impl.cpp
cswkim: OK, just added.

core/mavlink_commands.h
```diff
@@ -24,6 +24,7 @@ class MAVLinkCommands {
         COMMAND_DENIED,
         TIMEOUT,
         IN_PROGRESS,
+        NO_SETPOINT_SET,
```
julianoes: Nono, a MAVLink command can't ever have that result. It's an invention of the SDK :)

cswkim: Ah I see. OK I removed from mavlink command results and `offboard_result_from_command_result()`.

JonasVautherin:
Yay! Thanks @cswkim!

PR #747, opened 2019-05-14
core: hack to ignore messages from QGC for now

julianoes:
If we don't ignore the heartbeats from QGC we keep getting the warning that the remote changed on every heartbeat. Eventually we probably need a proper routing fix but this should do for now.

FYI @DanielePettenuzzo


julianoes:
@JonasVautherin do you undestand why Jenkins fails?

JonasVautherin:
I believe that's a timeout (related to AWS?). I restarted the jobs.

PR #724, opened 2019-04-17
Reorganize deps

JonasVautherin:
This is moving the dependencies out of the project and adding them using `find_package` instead. Because we still want to provide a way to compile the dependencies from sources, it is organized as follows:

```
- superbuild
    |
    --- third_party
    |
    --- src
```

The superbuild adds both. Note that `third_party` builds the dependencies at configure time. 

To be done before merging this PR:
- [x] Fix all CI scripts
- [x] Add build for iOS
- [x] Update documentation (by adding a Makefile showing how to use CMake)
- [x] Test iOS build with Swift SDK
- [ ] ~Add build for Android~
- [ ] ~Test Android build~

Resolves #718, resolves #703

PR #753, opened 2019-05-29
integration_tests: add attitude rate flips

julianoes:
I wanted to improve the attitude and altitude control but realized that altitude control was not working well with just the P controller.

Therefore, instead of implementing attitude control using attitude rate using offboard mode it probably makes more sense to implement flips which is what's likely used for rate control.

For attitude or altitude control it's generally easier to actually use the attitude control or altitude (velocity) control API directly.


JonasVautherin:
(Are you trying to get that in before #724? :laughing:)

julianoes:
Hah, don't care, I can rebase too.

PR #759, opened 2019-06-04
Fix dependencies when building core as shared libs

### LINKED ISSUE #758: Missing dependencies when linking SDK as shared libraries
This was reported by @elbarto1980 on [discuss](https://discuss.px4.io/t/curl-package-not-found/11250): 

When building and installing the SDK as shared libs, like so:

```sh
mkdir -p build/default
cd build/default
cmake -DBUILD_SHARED_LIBS=ON …/…
make
sudo make install
```

And then linking against an example, cmake complains about missing dependencies:

```
CMake Error at /usr/local/lib/cmake/DronecodeSDK/DronecodeSDKConfig.cmake:4 (find_package):
Could not find a package configuration file provided by “CURL” with any of
the following names:

CURLConfig.cmake
curl-config.cmake
```

JonasVautherin:
When building the core as shared libraries, the dependencies (e.g. curl) are linked statically and therefore are not required to be linked further.

This PR fixes the build system such that those dependencies are linked using `PRIVATE`, meaning that they don't propagate to `DronecodeSDKTargets.cmake`.

Resolves #758.


JonasVautherin:
@julianoes: So I set `BUILD_SHARED_LIBS=ON` by default, because it seems like it is what people want to do usually (and then install that). It also makes sense for the backend. In our case, we tend to link our libs statically against the backend, so the CI gets `-DBUILD_SHARED_LIBS=OFF` when `-DBUILD_BACKEND=ON`, but that's our choice. Does that make sense to you?

julianoes:
>  set BUILD_SHARED_LIBS=ON by default

Yes I think that makes sense.

JonasVautherin:
Always surprised to see that the coverage can decrease by 3.4% with such changes :sweat_smile:

PR #764, opened 2019-06-10
install backend_bin with libs and includes

JonasVautherin:
Let's install `backend_bin` in `<INSTALL_PREFIX>/bin`, alongside `libs` and `include`. So that later it will be possible to fetch that as release artifacts.


JonasVautherin:
There isn't any, for `backend_bin`. But of course that reminds me that I forgot to install all the other objects coming from the backend :sweat_smile:. Let me fix that.

PR #766, opened 2019-06-11
Create dependencies for the .proto files to regenerate the generated files on changes

thomasetter:
Before this, changes in the `.proto` files do not propagate if they have been compiled already


src/backend/src/CMakeLists.txt
```diff
@@ -6,19 +6,23 @@ include(cmake/compile_proto.cmake)
 
 foreach(COMPONENT_NAME ${COMPONENTS_LIST})
     compile_proto_pb(${COMPONENT_NAME} PB_COMPILED_SOURCE)
-    list(APPEND PB_COMPILED_SOURCES ${PB_COMPILED_SOURCE})
-
     compile_proto_grpc(${COMPONENT_NAME} GRPC_COMPILED_SOURCE)
-    list(APPEND GRPC_COMPILED_SOURCES ${GRPC_COMPILED_SOURCE})
+    add_library(${COMPONENT_NAME}_proto_gens ${GRPC_COMPILED_SOURCE} ${PB_COMPILED_SOURCE})
```
JonasVautherin: But then that means that we create a new library for each plugin, right? `libaction_proto_gens.so`, `libtelemetry_proto_gens.so`, ...

JonasVautherin: And if yes, should we always generate them as static archives? It feels a bit weird to me to have a shared lib just for the proto file (but maybe it's just me :sweat_smile:)

thomasetter: Sg, made them static

PR #769, opened 2019-06-15
Improve travis deployment

JonasVautherin:
- Deploy manylinux backend binary
- Deploy macos backend binary
- Fix rpm/deb deployment (that I broke when reorganizing the deps)

Notes:
- The rpm/deb packages now include `backend_bin` (which I believe we should rename, e.g. to `mavsdk_backend`, but I will open another PR to discuss that).
- I removed a few `Debug` builds, as it doubles the build time and that's important when building with `-DBUILD_BACKEND=ON`. If we really need to build in Debug mode, we could make a nightly build, IMO.
- `create_packages.sh` is now called by `travis-docker-build.sh`, which IMO makes it cleaner since the later also calls `fix_style.sh` and `generate_docs.sh`.
- `create_packages.sh` takes arguments now. The intent is to make it easier to use without relying on a convention like: "install into `build/default/install/usr`".
- `create_packages.sh` copies the files into a temp dir, and adds `/usr/` there. The benefit of this is that we don't have to build from scratch again just for packaging, improving the CI build time.

### DIFF
--- a/tools/create_packages.sh
+++ b/tools/create_packages.sh
@@ -3,80 +3,73 @@
 set -e
 
 echo "Packaging"
+echo "---------"
+
+if [ "$#" -ne 2 ]; then
+  echo "Error: missing path to the 'install' directory (i.e. what was specified as -DCMAKE_INSTALL_PREFIX)" >&2
+  echo "       or output path (where the package will be generated)!" >&2
+  echo "Usage: $0 <path/to/install> <output/path>" >&2
+  exit 1
+fi
+
+script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+install_dir=$1
+output_dir=$2
+
+working_dir=$(mktemp -d)
+mkdir -p ${working_dir}/install
+cp -r ${install_dir} ${working_dir}/install/usr
+
+# We need the relative path of all files.
+library_files=`find ${working_dir}/install -type f | cut -sd / -f 5-`
+echo "Files to be packaged:"
+echo ${library_files}
 
 # This creates a version such as "v1.2.3-5-g123abc".
 version=`git describe --always --tags`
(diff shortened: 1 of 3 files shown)


JonasVautherin:
Trying it on a tag ([v0.17.1](https://github.com/mavlink/MAVSDK/releases/tag/v0.17.1)) and it seems to work :tada:

tools/create_packages.sh
```diff
@@ -3,80 +3,73 @@
...
+echo "---------"
 
-# This creates a version such as "v1.2.3-5-g123abc".
-version=`git describe --always --tags`
-# We want to extract 1.2.3 from it.
-version=`echo ${version} | sed 's/v\([0-9]*\.[0-9]*\.[0-9]*\).*$/\1/'`
+if [ "$#" -ne 2 ]; then
+  echo "Error: missing path to the 'install' directory (i.e. what was specified as -DCMAKE_INSTALL_PREFIX)" >&2
+  echo "       or output path (where the package will be generated)!" >&2
+  echo "Usage: $0 <path/to/install> <output/path>" >&2
```
julianoes: :+1:

tools/create_packages.sh
```diff
@@ -3,80 +3,73 @@
...
+
+echo "#!/bin/sh" > ${working_dir}/run_ldconfig
+echo "/sbin/ldconfig" >> ${working_dir}/run_ldconfig
+
+common_args="--chdir ${working_dir}/install \
+             --input-type dir  \
+             --name mavsdk \
+             --version ${version} \
+             --maintainer julian@oes.ch \
+             --url https://sdk.dronecode.org \
```
julianoes: This one is probably going to change, right?

JonasVautherin: Yes, I guess as soon as we have the new domain :+1:

PR #770, opened 2019-06-15
backend: rename for packaging

JonasVautherin:
Note: it is conflicting with #769 (which assumes `backend_bin`)!

When installing mavsdk on the system, one will currently have `/usr/lib/libbackend.so` and/or `/usr/bin/backend_bin`. We should probably make that a bit more explicit. This PR proposes `/usr/lib/libmavsdk_backend.so` and `/usr/bin/mavsdk_backend_bin`.

Other ideas could be `mavsdk_server`, `mavsdk_bin`. @julianoes what do you think?

### DIFF
--- a/src/backend/src/CMakeLists.txt
+++ b/src/backend/src/CMakeLists.txt
@@ -33,12 +33,12 @@ if(IOS)
 endif()
 
 if(IOS)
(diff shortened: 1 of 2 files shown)


julianoes:
I like `mavsdk_server`.

PR #771, opened 2019-06-18
Add offboard/set_actuator_control to offboard plugin

irsdkv:
`set_actuator_control` method was added recently to `offboard` plugin and it is not supported in resent offboard grpc backend introduced with [great pull request](https://github.com/Dronecode/DronecodeSDK/pull/735) So this adds set_actuator_control support.

The corresponding MAVSDK-Proto PR is [here](https://github.com/mavlink/MAVSDK-Proto/pull/87).

### DIFF
--- a/src/backend/src/plugins/offboard/offboard_service_impl.h
+++ b/src/backend/src/plugins/offboard/offboard_service_impl.h
@@ -60,6 +60,35 @@ public:
         return grpc::Status::OK;
     }
 
+    grpc::Status SetActuatorControl(grpc::ServerContext * /* context */,
+                             const rpc::offboard::SetActuatorControlRequest *request,
+                             rpc::offboard::SetActuatorControlResponse * /* response */) override
+    {
+        if (request != nullptr) {
+            auto requested_actuator_control = translateRPCActuatorControl(request->actuator_control());
+            _offboard.set_actuator_control(requested_actuator_control);
+        }
+
+        return grpc::Status::OK;
+    }
+
+    static mavsdk::Offboard::ActuatorControl
+    translateRPCActuatorControl(const rpc::offboard::ActuatorControl &rpc_actuator_control)
+    {
+        mavsdk::Offboard::ActuatorControl actuator_control = {};
+
+        actuator_control.actuator_group =
+            static_cast<mavsdk::Offboard::ActuatorControl::ActuatorGroup>(rpc_actuator_control.actuator_group());
+
+        int len = std::min(8, rpc_actuator_control.actuator_values_size());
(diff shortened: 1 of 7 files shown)


irsdkv:
> First of all, thanks a lot for the contribution! This looks great from a coding point of view.
> 
> However, I would like to challenge the interface from a API / user friendlyness point of view. MAVSDK tries to abstract away some of the pecularities of MAVLink and PX4, and I would argue that the control groups is one of these odd things that we would like to simplify.
> 
> Do you actually need access to all 4 control groups and does that work for you? Or do you have a specific use case that you would like to expose?
> For instance I think it's fine to have an interface for direct pass-through of motor commands but I'm reluctant to expose all groups, unless we know that they are all properly supported and we have integration tests for them.

Yes, I understand your point of view.
In our use cases we need an interface for direct pass-through for more than 8 motors at the same time (we maintain our Octoplane VTOL, https://github.com/PX4/Firmware/pull/12303 ). This can be done using, for example, two Control groups (first for MC mode motors, second for Push/pull motor and for high lift devices). px4/Firmware at this time has support for four groups (https://github.com/PX4/Firmware/blob/v1.9.0/src/modules/mavlink/mavlink_receiver.cpp#L1040).

Because of you are right in case of using few groups may create additional complexity I propose the following solution:
1.  Remove Control group enums (https://github.com/mavlink/MAVSDK/pull/771/files#diff-ab1f989e0d05cbfe0ee11813ec8aa194R143, https://github.com/mavlink/MAVSDK-Proto/pull/87/files#diff-d071a0f40909dedf07ff7e03789cb56aR73);
2. Set uint8_t type for control_group fields in structs and (optional) check that group number will never be great than "3";
3. Set control_group field default value to "0" in APIs (mavsdk and mavsdk-proto) due to indicate that this is the preferred use case.

I think that this solution can provide flexibility for those who need it and on the other hand provide a simple solution for others.

Please let us know if you suggest another approach

julianoes:
@irsdkv thanks for the suggestion. This would hide the complexity for a simple use case.

Alternatively would an interface for all PWM outputs work for you as described below:

```
float actuator_outputs[16]; // First 8 will go to control group 0, the following 8 to control group 1.
```

I'm assuming you are not using the PX4 mixers, but direct pass-through, correct?

irsdkv:
@julianoes please consider https://github.com/mavlink/MAVSDK/pull/782

julianoes:
@irsdkv I'm sorry I'm still not completely happy with a "hidden" parameter, what about the suggestion [above](https://github.com/mavlink/MAVSDK/pull/771#issuecomment-507269911)?

PR #774, opened 2019-06-23
Mavlink ftp

MatejFranceskin:
Currently mavlink ftp client has been tested against PX4 mavlink ftp server.
FTP server has not been tested yet.


example/mavlink_ftp/mavlink_ftp.cpp
```diff
@@ -0,0 +1,253 @@
+/**
+ * @file fly_mission.cpp
+ *
+ * @brief Demonstrates how to Add & Fly Waypoint missions using the MAVSDK.
```
JonasVautherin: That would need to be updated :slightly_smiling_face:

src/plugins/mavlink_ftp/include/plugins/mavlink_ftp/mavlink_ftp.h
```diff
@@ -0,0 +1,220 @@
...
+    /**
+     * @brief Destructor (internal use only).
+     */
+    ~MavlinkFTP();
+
+    /**
+     * @brief Possible results returned for requests.
+     */
+    enum Result : uint8_t {
+        kSuccess,
```
JonasVautherin: We tend to write more [`SUCCESS`](https://github.com/mavlink/MAVSDK/blob/develop/src/plugins/mission/include/plugins/mission/mission.h#L46), etc. Which we should change eventually because it caused us issues with e.g. already defined `ERROR` on Windows, but well, that's how it is now.

src/plugins/mavlink_ftp/include/plugins/mavlink_ftp/mavlink_ftp.h
```diff
@@ -0,0 +1,220 @@
...
+     * @param callback Callback to receive result of this request.
+     */
+    void calc_file_crc32_async(std::string path, file_crc32_result_callback_t callback);
+
+    /**
+     * @brief Get our own system ID.
+     *
+     * @return our own system ID.
+     */
+    uint8_t get_our_sysid() const;
```
JonasVautherin: Do we need to expose those in this plugin? I don't see where that's used...

example/mavlink_ftp/mavlink_ftp.cpp
MatejFranceskin: Updated!

src/plugins/mavlink_ftp/include/plugins/mavlink_ftp/mavlink_ftp.h
MatejFranceskin: Changed to the MAVSDK coding style. Previous was copy/pasted from Mavlink FTP implementation in Firmware.

src/plugins/mavlink_ftp/include/plugins/mavlink_ftp/mavlink_ftp.h
MatejFranceskin: True - we don't need them. Copy/paste error.

src/plugins/mavlink_ftp/include/plugins/mavlink_ftp/mavlink_ftp.h
```diff
@@ -0,0 +1,189 @@
...
+#include "mavlink/v2.0/common/mavlink.h"
+#include "plugin_base.h"
+
+namespace mavsdk {
+
+class System;
+class MavlinkFTPImpl;
+
+/**
+ * @brief The MavlinkFTP class provides MAVLink FTP functionality
```
julianoes: Maybe add a comment one step further back.

```suggestion
 * @brief The MavlinkFTP class provides file uploading and downloading using MAVLink FTP.
```

src/plugins/mavlink_ftp/include/plugins/mavlink_ftp/mavlink_ftp.h
```diff
@@ -0,0 +1,189 @@
...
+        ERR_UNKOWN_COMMAND,        ///< Unknown command opcode
+        ERR_FAIL_FILE_EXISTS,      ///< File exists already
+        ERR_FAIL_FILE_PROTECTED,   ///< File is write protected
+
+        // These error codes are returned to client without contacting the server
+        ERR_OP_IN_PROGRESS = 200,  ///< Operation is already in progress
+        ERR_INVALID_PARAMETER,     ///< Invalid parameter
+        ERR_FAIL_LOCAL_OPEN,       ///< Failed to open local file
+        ERR_FAIL_LOCAL_READ,       ///< Failed to read local file
+        ERR_FAIL_LOCAL_WRITE,      ///< Failed to write local file
```
julianoes: I think we should not expose the underlying MAVLink FTP errors but we should wrap the errors that we are actually expecting to something simple that a user implementing against the library can understand.

E.g.
```
CONNECTION_ERROR,
PROTOCOL_ERROR,
UNSUPPORTED,
FILE_EXISTS,
FILE_PROTECTED,
IN_PROGRESS
```
```

File: src/plugins/mavlink_ftp/include/plugins/mavlink_ftp/mavlink_ftp.h [hunk 6]
```diff
@@ -0,0 +1,189 @@
...
+    std::string result_str(Result result);
+
+    /**
+     * @brief Result callback type for async ftp calls.
+     */
+    typedef std::function<void(Result)> result_callback_t;
+
+    /**
+     * @brief Progress callback type for async ftp calls.
+     */
```
julianoes: What are the two arguments?

src/plugins/mavlink_ftp/include/plugins/mavlink_ftp/mavlink_ftp.h
```diff
@@ -0,0 +1,189 @@
...
+    void rename_async(std::string from_path, std::string to_path, result_callback_t callback);
+
+    /**
+     * @brief Calculate CRC32 for file (asynchronous).
+     *
+     * @param path Path to file for which CRC32 is calculated.
+     * @param callback Callback to receive result of this request.
+     */
+    void calc_file_crc32_async(std::string path, file_crc32_result_callback_t callback);
```
julianoes: I don't think we need to expose the CRC check. The file transfer protocol should make sure everything is correct but the user of the library should not have to care in my opinion.

src/plugins/mavlink_ftp/mavlink_ftp_impl.cpp
```diff
@@ -0,0 +1,971 @@
...
+#include "crc32.h"
+#include "mavlink_ftp_impl.h"
+#include "system.h"
+#include "global_include.h"
+
+namespace mavsdk {
+
+using namespace std::placeholders; // for `_1`
+
+namespace fs = std::experimental::filesystem;
```
julianoes: Nice, this seems to make it quite easy without boost, right?

src/plugins/mavlink_ftp/mavlink_ftp_impl.cpp
```diff
@@ -0,0 +1,971 @@
...
+void MavlinkFTPImpl::enable()
+{
+}
+
+void MavlinkFTPImpl::disable()
+{    
+}
+
+void MavlinkFTPImpl::_process_ack(PayloadHeader *payload)
+{
```
julianoes: I think you'll need a mutex around all the member vars, right?

src/plugins/mavlink_ftp/mavlink_ftp_impl.h
```diff
@@ -0,0 +1,166 @@
...
+#include "mavlink_include.h"
+#include "plugins/mavlink_ftp/mavlink_ftp.h"
+#include "plugin_impl_base.h"
+
+namespace mavsdk {
+
+class MavlinkFTPImpl : public PluginImplBase {
+public:
+    /// @brief Command opcodes
+    enum Opcode : uint8_t {
```
julianoes: Please use an `enum class`.

src/plugins/mavlink_ftp/mavlink_ftp_impl.h
```diff
@@ -0,0 +1,166 @@
...
+
+    void init() override;
+    void deinit() override;
+
+    void enable() override;
+    void disable() override;
+
+    void send();
+
+    void download_async(std::string remote_file_path,
```
julianoes: Did you choose to copy the `std::string` on purpose?

src/plugins/mavlink_ftp/mavlink_ftp_impl.cpp
```diff
@@ -0,0 +1,971 @@
...
+
+    close(fd);
+
+    payload->size = sizeof(uint32_t);
+    memcpy(payload->data, &checksum, payload->size);
+
+    return MavlinkFTP::Result::SUCCESS;
+}
+
+void MavlinkFTPImpl::send()
```
julianoes: Is this used anywhere?

src/plugins/mavlink_ftp/include/plugins/mavlink_ftp/mavlink_ftp.h
```diff
@@ -0,0 +1,189 @@
...
+    void remove_file_async(std::string path, result_callback_t callback);
+
+    /**
+     * @brief Truncates file (asynchronous).
+     *
+     * @param path Path of a file to remove
+     * @param offset Offset to which file is truncated
+     * @param callback Callback to receive result of this request.
+     */
+    void truncate_file_async(std::string path, uint32_t offset, result_callback_t callback);
```
julianoes: What do we need this for?

MatejFranceskin: It is weird, but it is defined in Mavlink FTP spec: https://github.com/mavlink/mavlink-devguide/blob/master/ko/services/ftp.md#TruncateFile
Maybe they wanted to use it to reset a log file?

julianoes: But does a MAVSDK user need it? If not I would not expose it. What we expose we need to test.

src/plugins/mavlink_ftp/mavlink_ftp_impl.cpp
MatejFranceskin: This part is unfinished - it is meant to be used in "streaming" download. I propose that we don't implement it for now, there are a lot of potential problems with it.

src/plugins/mavlink_ftp/include/plugins/mavlink_ftp/mavlink_ftp.h
MatejFranceskin: I agree, this function doesn't seem very useful. For all the others that are not tested yet I will write integration tests.

src/plugins/mavlink_ftp/mavlink_ftp_impl.cpp
julianoes: Ok your judgement.

src/plugins/mavlink_ftp/mavlink_ftp_impl.cpp
MatejFranceskin: It is being changed in newer gcc versions to std::filesystem only. I guess I should address this in cmake.

src/plugins/mavlink_ftp/include/plugins/mavlink_ftp/mavlink_ftp.h
MatejFranceskin: I thought to use this function with camera definition xml file transfer so that we check if CRC has changed before we download it.

src/plugins/mavlink_ftp/mavlink_ftp_impl.h
MatejFranceskin: Changed - std::string s are passed by reference.

src/integration_tests/mavlink_ftp.cpp
```diff
@@ -0,0 +1,175 @@
...
+                _bytes_transferred = bytes_transferred;
+                _file_size = file_size;
+            },
+            [prom](MavlinkFTP::Result result)
+            {
+                prom->set_value(result);
+            });
+
+        MavlinkFTP::Result result = future_result.get();
+        EXPECT_TRUE(_file_size > 0);
```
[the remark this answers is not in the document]
MatejFranceskin: Changed.

src/integration_tests/mavlink_ftp.cpp
```diff
@@ -0,0 +1,175 @@
...
+                prom->set_value(result);
+            });
+
+        MavlinkFTP::Result result = future_result.get();
+        EXPECT_TRUE(result == MavlinkFTP::Result::SUCCESS || result == MavlinkFTP::Result::ERR_FAIL_FILE_EXISTS || result == MavlinkFTP::Result::ERR_FAIL_FILE_PROTECTED);
+    }
+
+    const int number_of_files = 100;
+    char data[1024];
+    memset(data, 13, sizeof(data));
```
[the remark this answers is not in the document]
MatejFranceskin: Changed.

src/integration_tests/mavlink_ftp.cpp
```diff
@@ -0,0 +1,175 @@
...
+                _file_size = file_size;
+            },
+            [prom](MavlinkFTP::Result result)
+            {
+                prom->set_value(result);
+            });
+
+        MavlinkFTP::Result result = future_result.get();
+        EXPECT_TRUE(_file_size > 0);
+        EXPECT_TRUE(_bytes_transferred == _file_size && _file_size != 0);
```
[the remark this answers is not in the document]
MatejFranceskin: Changed this and all other occurences.

src/integration_tests/mavlink_ftp.cpp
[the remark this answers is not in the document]
MatejFranceskin: Changed this and all other occurences.

src/plugins/mavlink_ftp/mavlink_ftp_impl.cpp
julianoes: No that's fine with me.

src/plugins/mavlink_ftp/crc32.cpp
```diff
@@ -0,0 +1,123 @@
+/************************************************************************************************
+ * apps/system/zmodem/host/crc32.c
+ *
+ * This file is a part of NuttX:
```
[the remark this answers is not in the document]
MatejFranceskin: I made crc32 a class and retained note about author of the algorithm

src/plugins/mavlink_ftp/crc32.h
```diff
@@ -0,0 +1,83 @@
+/****************************************************************************
+ * apps/system/zmodem/host/crc16.h
```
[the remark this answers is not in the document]
MatejFranceskin: I made crc32 a class and retained note about author of the algorithm

src/plugins/mavlink_ftp/include/plugins/mavlink_ftp/mavlink_ftp.h
MatejFranceskin: First one is bytes transferred, second one is total bytes. Could be simplified to report float "percentage completed", what do you think?

src/plugins/mavlink_ftp/mavlink_ftp_impl.cpp
```diff
@@ -176,8 +176,8 @@ void MavlinkFTPImpl::_process_nak(PayloadHeader *payload)
     _curr_op = kCmdNone;
 }
 
-void MavlinkFTPImpl::download_async(std::string remote_file_path,
-                                    std::string local_folder,
+void MavlinkFTPImpl::download_async(std::string& remote_file_path,
```
julianoes: I would make it `const std::string&`.

MatejFranceskin: Changed

src/plugins/mavlink_ftp/mavlink_ftp_impl.cpp
MatejFranceskin: I implemented and tested it along with retry / timeout mechanism.

example/mavlink_ftp/mavlink_ftp.cpp
```diff
@@ -0,0 +1,255 @@
...
+    auto telemetry = std::make_shared<Telemetry>(system);
+
+    while (!telemetry->health_all_ok()) {
+        std::cout << "Waiting for system to be ready" << std::endl;
+        sleep_for(seconds(1));
+    }
+
+    std::cout << "System ready" << std::endl;
+
+    {
```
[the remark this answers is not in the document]
MatejFranceskin: Agree - moved to functions.

src/plugins/mavlink_ftp/include/plugins/mavlink_ftp/mavlink_ftp.h
MatejFranceskin: Implemented translation between ftp server return codes and SDK user return codes

src/plugins/mavlink_ftp/mavlink_ftp_impl.h
MatejFranceskin: I need that enum values implicitly convert to uint8_t so that they're used according to protocol definition. They need to have the same values as PX4 Mavlink FTP server implementation. Is there a more elegant way to get around that?

JonasVautherin:
Looking very good (and clean) to me! I'm curious about @julianoes' opinion on the changes in the core.

@MatejFranceskin: note that it is failing the CI because of some `-Werror`s :sweat_smile:

src/core/udp_connection.h
```diff
@@ -21,6 +21,8 @@ class UdpConnection : public Connection {
 
     bool send_message(const mavlink_message_t &message);
 
+    void add_remote(const std::string &remote_ip, const int remote_port, const uint8_t remote_sysid = 0);
```
julianoes: If we use this `add_remote` what happens to the `local_ip` and `local_port` that we use in the constructor? I wonder if we could create a remote connection using the remote IP in the constructor but I might be confused about networking.

src/integration_tests/mavlink_ftp.cpp
```diff
@@ -251,3 +252,178 @@ TEST(MavlinkFTPTest, UploadFiles)
...
+    auto mavlink_ftp_server = std::make_shared<MavlinkFTP>(system_cc);
+    mavlink_ftp_server->set_root_dir("/usr/local");
+    uint8_t server_comp_id = mavlink_ftp_server->get_our_compid();
+
+    auto mavlink_ftp_client = std::make_shared<MavlinkFTP>(system_gcs);
+    mavlink_ftp_client->set_target_component_id(server_comp_id);
+    mavlink_ftp_client->set_timeout(10);
+    mavlink_ftp_client->set_retries(10);
+
+    {
```
julianoes: Functions would be nice, I'd say :).

src/core/system.h
```diff
@@ -37,7 +37,7 @@ class System {
      * @param system_id System id.
      * @param comp_id Component id.
      */
-    explicit System(MavsdkImpl &parent, uint8_t system_id, uint8_t comp_id);
+    explicit System(MavsdkImpl &parent, uint8_t system_id, uint8_t comp_id, bool connected);
```
julianoes: The param `connected` is not explained in the docs above.

I don't understand what it exactly means and the problem is that "connected" is an overloaded term in the MAVSDK/MAVLink world already, so I'd be cautious in using it. Please explain what it entails and then maybe we can come up with something better.

src/plugins/mavlink_ftp/crc32.h
```diff
@@ -0,0 +1,20 @@
+#pragma once
+
+#include <sys/types.h>
+#include <stdint.h>
+
+namespace mavsdk {
+
+class crc32 {
+public:
+    crc32() {}
```
julianoes: I think you can get rid of this default ctor.

src/plugins/mavlink_ftp/crc32.h
```diff
@@ -0,0 +1,20 @@
+#pragma once
+
+#include <sys/types.h>
```
julianoes: What's `sys/types.h` for?

src/plugins/mavlink_ftp/crc32.h
```diff
@@ -0,0 +1,20 @@
...
+#include <stdint.h>
```
julianoes: Nitpick: I prefer `cstdint`.

src/plugins/mavlink_ftp/crc32.cpp
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * - The table can be generated at runtime if desired; code to do so is shown later.  It
+ *   might not be obvious, but the feedback terms simply represent the results of eight
+ *   shift/xor operations for all combinations of data and CRC register values
+ *
+ * - The values must be right-shifted by eight bits by the updcrc logic; the shift must
+ *   be u_(bring in zeroes).  On some hardware you could probably optimize the shift in
+ *   assembler by using byte-swap instructions polynomial $edb88320
+  ************************************************************************************************/
+
```
julianoes: :+1:

src/plugins/mavlink_ftp/include/plugins/mavlink_ftp/mavlink_ftp.h
julianoes: No I think that's fine. Is it possible to name them in the declaration and that way the user knows what it is?

src/integration_tests/mavlink_ftp.cpp
MatejFranceskin: Looks so much better with functions. :-)

src/core/system.h
MatejFranceskin: Added explanation.

src/plugins/mavlink_ftp/crc32.h
MatejFranceskin: OK!

src/plugins/mavlink_ftp/crc32.h
MatejFranceskin: It was for size_t. Replaced it with std::size_t.

MatejFranceskin:
Has anybody a better idea to name setup_udp_remote which in contrast to add_udp_connection starts sending heartbeats to remote IP without waiting for other party to connect.

src/core/udp_connection.h
MatejFranceskin: UDP socket is always bound to local_ip and local_port. In case of add_remote the local_ip is localhost and local_port is automatically chosen from one of the free ports.
If we move add_remote to constructor we would have to add a bool flag indicating that it is "remote".

julianoes:
@MatejFranceskin I assume I should review that again to get it in?

MatejFranceskin:
From my side I have to test how it works in real setup with companion computer. Depending on results I might have to make more changes.

src/plugins/mavlink_ftp/include/plugins/mavlink_ftp/mavlink_ftp.h
```diff
@@ -0,0 +1,227 @@
...
+    enum class Result {
+        SUCCESS,
+        TIMEOUT, ///< Timeout
+        IN_PROGRESS, ///< Operation is already in progress
+        FILE_IO_ERROR, ///< File IO operation error
+        FILE_EXISTS, ///< File exists already
+        FILE_PROTECTED, ///< File is write protected
+        INVALID_PARAMETER, ///< Invalid parameter
+        UNSUPPORTED, ///< Unsupported command
+        PROTOCOL_ERROR ///< General protocol error
```
julianoes: I think that's much better!

src/plugins/mavlink_ftp/mavlink_ftp_impl.cpp
```diff
@@ -0,0 +1,1121 @@
...
+
+void MavlinkFTPImpl::_list_directory(uint32_t offset)
+{
+    uint8_t raw_payload[MAVLINK_MSG_FILE_TRANSFER_PROTOCOL_FIELD_PAYLOAD_LEN];
+    PayloadHeader* payload = reinterpret_cast<PayloadHeader*>(raw_payload);
+    payload->seq_number = _seq_number++;
+    payload->session = 0;
+    payload->opcode = _curr_op = CMD_LIST_DIRECTORY;
+    payload->offset = offset;
+    strcpy(reinterpret_cast<char*>(payload->data), _last_path.c_str());
```
julianoes: I'm scared. Can we always use `strncpy`?

example/mavlink_ftp/mavlink_ftp.cpp
```diff
@@ -0,0 +1,399 @@
+/**
+ * @file mavlink_ftp.cpp
```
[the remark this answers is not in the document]
MatejFranceskin: Done.

example/mavlink_ftp/mavlink_ftp.cpp
```diff
@@ -0,0 +1,399 @@
...
+    std::pair<MavlinkFTP::Result, uint32_t> result = future_result.get();
+    if (result.first == MavlinkFTP::Result::SUCCESS) {
+        crc = result.second;
+    }
+    return result.first;
+}
+
+int main(int argc, char** argv)
+{
+    Mavsdk dc;
```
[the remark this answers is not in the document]
MatejFranceskin: Done.

example/mavlink_ftpd/mavlink_ftpd.cpp
```diff
@@ -0,0 +1,64 @@
+/**
+ * @file mavlink_ftpd.cpp
```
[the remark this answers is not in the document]
MatejFranceskin: Done.

src/integration_tests/mavlink_ftp.cpp
```diff
@@ -0,0 +1,367 @@
...
+    System& system_gcs = dc_gcs.system();
+
+    Mavsdk dc_cc;
+    dc_cc.set_configuration(Mavsdk::Configuration::CompanionComputer);
+    ret = dc_cc.setup_udp_remote("127.0.0.1", 24550);
+    ASSERT_EQ(ret, ConnectionResult::SUCCESS);
+    System& system_cc = dc_cc.system();
+
+    auto mavlink_ftp_server = std::make_shared<MavlinkFTP>(system_cc);
+    mavlink_ftp_server->set_root_dir("/usr/local");
```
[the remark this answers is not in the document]
MatejFranceskin: Who knows what was in my mind :-)
Changed it to dot

src/plugins/mavlink_ftp/mavlink_ftp_impl.cpp
MatejFranceskin: Agree!

src/core/mavsdk_impl.cpp
```diff
@@ -181,6 +188,20 @@ ConnectionResult MavsdkImpl::add_udp_connection(const std::string& local_ip, con
     return ret;
 }
 
+ConnectionResult MavsdkImpl::setup_udp_remote(const std::string& remote_ip, int remote_port)
+{
+    auto new_conn = std::make_shared<UdpConnection>(
+        std::bind(&MavsdkImpl::receive_message, this, std::placeholders::_1), "127.0.0.1", 0);
```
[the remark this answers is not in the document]
MatejFranceskin: 0.0.0.0 is better!

src/core/mavsdk_impl.cpp
```diff
@@ -305,7 +325,7 @@ uint8_t MavsdkImpl::get_own_component_id() const
 
         case Mavsdk::Configuration::CompanionComputer:
             // It's at least a possibility that we are bridging MAVLink traffic.
-            return MAV_COMP_ID_UDP_BRIDGE;
+            return 99;
```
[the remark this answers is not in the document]
MatejFranceskin: Server doesn't work because of this:

void SystemImpl::process_heartbeat(const mavlink_message_t& message)
{
    // FIXME: for now we ignore heartbeats from UDP_BRIDGE because that's just
    // confusing since it doesn't mean a vehicle is connected.
    if (message.compid == MAV_COMP_ID_UDP_BRIDGE) {
        return;
    }
...

I think we should be able to set component ID when we create mavlink FTP server. This way it would be possible to have more then one server on the same companion computer.

src/core/udp_connection.h
```diff
@@ -22,6 +22,9 @@ class UdpConnection : public Connection {
 
     bool send_message(const mavlink_message_t& message);
 
+    void
+    add_remote(const std::string& remote_ip, const int remote_port, const uint8_t remote_sysid = 0);
```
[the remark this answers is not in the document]
MatejFranceskin: You mean that this should be moved outside of this class? Connection maintains a list of _remotes, so we would have to move this somewhere else too?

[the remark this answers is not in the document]
julianoes: Ok I see how it is used.

Can you make a private method `add_remote_with_remote_sysid` and a public `add_remote`? I think that would work and not expose the mavlink internal field.

src/plugins/mavlink_ftp/crc32.h
```diff
@@ -0,0 +1,17 @@
+#pragma once
+
+#include <cstdint>
+
+namespace mavsdk {
+
+class crc32 {
```
julianoes: Should be `Crc32` even if it looks ugly :confused:.

src/plugins/mavlink_ftp/mavlink_ftp_impl.h
```diff
@@ -0,0 +1,228 @@
...
+        uint32_t offset; ///< Offsets for List and Read commands
+        uint8_t data[]; ///< command data, varies by Opcode
+    };
+
+    /// @brief Maximum data size in RequestHeader::data
+    static const uint8_t max_data_length =
+        MAVLINK_MSG_FILE_TRANSFER_PROTOCOL_FIELD_PAYLOAD_LEN - sizeof(PayloadHeader);
+
+    struct SessionInfo {
+        int fd;
```
julianoes: defaults should be set here directly, e.g. `{-1}`.

src/plugins/mavlink_ftp/mavlink_ftp_impl.cpp
```diff
@@ -0,0 +1,1121 @@
...
+#include "global_include.h"
+
+namespace mavsdk {
+
+using namespace std::placeholders; // for `_1`
+
+MavlinkFTPImpl::MavlinkFTPImpl(System& system) : PluginImplBase(system)
+{
+    _parent->register_plugin(this);
+    _session_info.fd = -1;
```
julianoes: Default should be set in declaration, not here.

src/plugins/mavlink_ftp/mavlink_ftp_impl.cpp
```diff
@@ -0,0 +1,1121 @@
...
+            } else {
+                _stop_timer();
+                _curr_op_result_callback(_translate(_session_result));
+            }
+            break;
+
+        case CMD_TERMINATE_SESSION:
+            _session_valid = false;
+            _stop_timer();
+            _curr_op_result_callback(_translate(_session_result));
```
julianoes: Can you be sure that the callback is not null?

And you should always use the thread pool for user callbacks:
```
if (_curr_op_result_callback) {
    const auto temp_callback = _curr_op_result_callback;
    _parent->call_user_callback([temp_callback]() { temp_callback(_translate(_session_result)); });
```

And if you make `_translate` `static` you don't need to pass `this` into the lambda.

src/plugins/mavlink_ftp/mavlink_ftp_impl.cpp
```diff
@@ -0,0 +1,1121 @@
...
+        _last_command_timer_running = true;
+        _last_command_timer = std::async(std::launch::async, [&]() {
+            while (true) {
+                _timer_mutex.lock();
+                if (!_last_command_timer_running) {
+                    _timer_mutex.unlock();
+                    return;
+                }
+                auto now = std::chrono::steady_clock::now();
+                auto time_diff = std::chrono::duration_cast<std::chrono::milliseconds>(
```
julianoes: I'd recommend to use the included timer if possible. Have a look at:
https://github.com/mavlink/MAVSDK/blob/f4065b3fb56e65674d045a3406cfb9c2536fcf65/src/core/system_impl.h#L56-L58

src/plugins/mavlink_ftp/mavlink_ftp_impl.cpp
```diff
@@ -0,0 +1,1121 @@
...
+    }
+
+    PayloadHeader* payload = reinterpret_cast<PayloadHeader*>(&ftp_req.payload[0]);
+
+    ServerResult errorCode = ServerResult::SUCCESS;
+
+    // basic sanity checks; must validate length before use
+    if (payload->size > max_data_length) {
+        errorCode = ServerResult::ERR_INVALID_DATA_SIZE;
+        goto out;
```
julianoes: Any chance you can use `return` early instead of `goto`? There is only this one, so an `if` would probably do it.

src/plugins/mavlink_ftp/mavlink_ftp_impl.cpp
```diff
@@ -0,0 +1,1121 @@
...
+{
+    crc32 checksum;
+    ssize_t bytes_read;
+    int fd = ::open(path.c_str(), O_RDONLY);
+
+    if (fd < 0) {
+        return MavlinkFTP::Result::FILE_IO_ERROR;
+    }
+
+    char buffer[18392];
```
julianoes: Why this size? What happens if the file is bigger?

MatejFranceskin:
> Couple more things, sorry to be such a pain confused

You're doing a great job. I know how difficult it is to review in detail work of others.

src/core/udp_connection.h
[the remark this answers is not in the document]
MatejFranceskin: Got it!

src/plugins/mavlink_ftp/mavlink_ftp_impl.cpp
MatejFranceskin: Changed it. (The original code was taken from Firmware)

src/plugins/mavlink_ftp/mavlink_ftp_impl.cpp
MatejFranceskin: This blew my mind. Very good!

src/plugins/mavlink_ftp/mavlink_ftp_impl.cpp
MatejFranceskin: This reads the file in buffer size chunks. Code was taken from Firmware where the buffer was much smaller.

julianoes: Ok, maybe add that as a comment so it's clear for a reader.

example/mavlink_ftp_client/mavlink_ftp_client.cpp
```diff
@@ -0,0 +1,399 @@
...
+              << "put <file> <path>    : Upload one file to remote directory" << std::endl
+              << "get <file> <path>    : Download remote file to local directory" << std::endl
+              << "delete <file>        : Delete remote file" << std::endl
+              << "rename <old> <new>   : Rename file" << std::endl
+              << "dir <path>           : List contents of remote directory" << std::endl
+              << "mkdir <path>         : Make directory on remote machine" << std::endl
+              << "rmdir <path>         : Remove directory on remote machine" << std::endl
+              << "cmp <local> <remote> : Compare local and remote file" << std::endl
+              << "crc32 <path>         : Get remote file crc32" << std::endl
+              << "localcrc32 <path>    : Get local file crc32" << std::endl;
```
bkueng: @MatejFranceskin a couple of things:
- can you generally document what the exit code is?
- mkdir: if the directory already exists, an error EEXIST errno is returned from the server. Adding an argument to ignore that would help, or at least indicate it in the exit code.
- similar for delete/rmdir: I need a way to differentiate/ignore if it does not exist
- rmdir: it's not recursive. A recursive variant would be helpful (I'll need it)
- rename: this also works for directories if they are on the same file system

example/mavlink_ftp_client/mavlink_ftp_client.cpp
```diff
@@ -0,0 +1,399 @@
...
+    future_result.get();
+
+    // We don't need to specify the UUID if it's only one system anyway.
+    // If there were multiple, we could specify it with:
+    // mavsdk.system(uint64_t uuid);
+    System& system = mavsdk.system();
+    auto mavlink_ftp = std::make_shared<MavlinkFTP>(system);
+    mavlink_ftp->set_timeout(200);
+    mavlink_ftp->set_retries(10);
+    mavlink_ftp->set_target_component_id(std::stoi(argv[2]));
```
bkueng: if `argv[2]` is not an int, it will call std::abort. I'd prefer a more graceful handling with proper error message.

example/mavlink_ftp_client/mavlink_ftp_client.cpp
```diff
@@ -0,0 +1,413 @@
...
+        if (argc < 5) {
+            usage(argv[0]);
+            return 1;
+        }
+        res = remove_directory(mavlink_ftp, argv[4]);
+        if (res == MavlinkFTP::Result::SUCCESS) {
+            std::cout << "Directory removed." << std::endl;
+        } else {
+            std::cout << ERROR_CONSOLE_TEXT
+                      << "Create directory error: " << mavlink_ftp->result_str(res)
```
bkueng:
```suggestion
                      << "Remove directory error: " << mavlink_ftp->result_str(res)
```

example/mavlink_ftp_client/mavlink_ftp_client.cpp
MatejFranceskin: Changed.

src/plugins/mavlink_ftp/mavlink_ftp_impl.cpp
MatejFranceskin: Changed to use included timer.

src/core/mavsdk_impl.cpp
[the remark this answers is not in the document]
julianoes: We should actually be able to remove this hack in `process_heartbeat` because we now use `has_autopilot()` to check if something is connected.

src/core/mavsdk_impl.cpp
```diff
@@ -282,8 +312,7 @@ uint8_t MavsdkImpl::get_own_system_id() const
 {
     switch (_configuration.load()) {
         case Mavsdk::Configuration::GroundStation:
-            // FIXME: This doesn't make much sense actually but it seems to work.
-            return 0;
+            return 254;
```
julianoes: I guess this should be: 

MAV_COMP_ID_MISSIONPLANNER

julianoes:
@MatejFranceskin I've rebased and force-pushed this, and fixed the proto submodule.

src/core/mavsdk_impl.cpp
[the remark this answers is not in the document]
julianoes: I'm removing it here: https://github.com/mavlink/MAVSDK/pull/828

bkueng:
One more feature request from me: the commands need a timeout (fixed to a couple of seconds is enough), in all cases (connection discovery and waiting for responses). I noticed that the connection setup can hang infinitely.

julianoes:
@bkueng which methods exactly are you referring to?

bkueng:
I have not debugged the specific case I encountered, and I'm also raising this more as a general requirement.

src/core/mavsdk_impl.cpp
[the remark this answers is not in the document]
MatejFranceskin: Now it works!

src/core/mavsdk_impl.cpp
MatejFranceskin: Yes.

MatejFranceskin:
Beat was referring to mavlink_ftp_client "example" which is was waiting indefinetly for server connection. I changed the behaviour so that it fails if there is no connection in 5 seconds.

bkueng:
Thanks @MatejFranceskin

MatejFranceskin:
I think it is ready for final review. :-)

julianoes:
@MatejFranceskin ok I'll check it today. Have you checked the CI failures?

julianoes:
@JonasVautherin you can just push to this branch, I think.

src/plugins/mavlink_ftp/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+)
+
+add_library(mavsdk_mavlink_ftp
+    mavlink_ftp.cpp
+    mavlink_ftp_impl.cpp
+    crc32.cpp
+)
+
+target_compile_options(mavsdk_mavlink_ftp
+  PRIVATE -fexceptions
```
julianoes: :astonished: do we need this? We generally don't use exceptions.

julianoes: I see that you use `try` `catch` around `std::atoi` but that's in the example and not in this plugin.

src/plugins/mission_raw/mission_raw_impl.cpp
```diff
@@ -252,6 +252,9 @@ void MissionRawImpl::process_mission_item_int(const mavlink_message_t& message)
     }
 
     auto new_item = std::make_shared<MissionRaw::MavlinkMissionItemInt>();
+    if (!new_item) {
+        return;
+    }
```
julianoes: Has this ever happened? How did you find this?

MatejFranceskin: One of the platforms threw warnings as errors if this wasn't checked... I forgot which - I have to check.

src/plugins/mavlink_ftp/CMakeLists.txt
MatejFranceskin: ghc-filesystem library uses exceptions. Maybe it would be better to go back to old ANSI C std functions. std::filesystem doesn't work on all platforms this is why I had to switch to ghc-filesystem. There is too much trouble with it.

JonasVautherin: I am probably wrong, but... greping for `catch(`, it seems to me like the only place using an exception in that library is [here](https://github.com/gulrak/filesystem/blob/master/include/ghc/filesystem.hpp#L4734-L4740). And that's in a `GHC_OS_WINDOWS` `#ifdef`. Could that be?

JonasVautherin: What about using boost filesystem, then? It can apparently be built with `BOOST_NO_EXCEPTION`. Wouldn't it be easier than doing it manually?

julianoes: I guess doing it in C is faster than adding dependencies.

MatejFranceskin: I removed the ghc_filesystem dependency.

src/plugins/mavlink_ftp/fs.cpp
```diff
@@ -0,0 +1,116 @@
...
+#include <fcntl.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <stack>
+#include <stdio.h>
+#include <unistd.h>
+
+#include "fs.h"
+
+#ifndef PATH_MAX
```
julianoes: This requires `#include <linux/limits.h>`, right?

MatejFranceskin: This ifndef was needed only for MacOS... so linux/limits.h was already included from somewhere.

src/plugins/mission_raw/mission_raw_impl.cpp
MatejFranceskin: I removed this change.

PR #779, opened 2019-06-27
Telemetry: add angular velocity and acceleration

### LINKED ISSUE #754: Expose angular velocities and acceleration in telemetry
It would probably be nice if angular velocities and also acceleration would be exposed in telemetry. I have heard people asking for this quite a bit.

This data could be received by the [HIGHRES_IMU](https://mavlink.io/en/messages/common.html#HIGHRES_IMU) message and added to the telemetry plugin in the same way as other data is exposed there.

krady21:
1. adds struct IMUReadingNED encompassing acceleration ,angular speed, magnetic field and temperature
2. adds setters and getters(both synchronous and asynchronous for interacting with acceleration and angular speed) and a process function for getting the desired info from the mavlink message.

Resolves #754 and #651


src/plugins/telemetry/telemetry_impl.cpp
```diff
@@ -424,6 +439,24 @@ void TelemetryImpl::process_mount_orientation(const mavlink_message_t &message)
...
+void TelemetryImpl::process_imu_reading_ned(const mavlink_message_t &message)
+{
+    mavlink_highres_imu_t highres_imu;
+    mavlink_msg_highres_imu_decode(&message, &highres_imu);
+    set_imu_reading_ned(Telemetry::IMUReadingNED({highres_imu.xacc,
+                                                  highres_imu.yacc,
+                                                  highres_imu.zacc,
+                                                  highres_imu.xgyro,
+                                                  highres_imu.ygyro,
+                                                  highres_imu.zgyro}));
```
julianoes: I wonder if we should also add mag readings while we're at it.

src/plugins/telemetry/include/plugins/telemetry/telemetry.h
```diff
@@ -120,6 +120,39 @@ class Telemetry : public PluginBase {
         float velocity_down_m_s; /**< @brief Velocity in Down direction in metres/second. */
     };
 
+    /**
+     * @brief Acceleration type in local coordinates.
+     *
+     * The acceleration is represented in the NED (North East Down) frame and in metres/second^2.
+     */
+    struct AccelerationNED {
+        float north_m_ss; /**< @brief Acceleration in North direction in metres/second^2 */
```
julianoes: I think I would call it `north_m_s2` but that's probably just a matter of taste.

src/plugins/telemetry/telemetry_impl.cpp
krady21: That could be a good idea. I have just found that temperature is also requested as per this issue(#651). That could be a thing too.

julianoes: Yes that would also be helpful.

krady21: I will get right on it!

src/integration_tests/telemetry_async.cpp
```diff
@@ -215,6 +223,20 @@ void print_ground_speed_ned(Telemetry::GroundSpeedNED ground_speed_ned)
     _received_ground_speed = true;
 }
 
+void print_imu_reading_ned(Telemetry::IMUReadingNED imu_reading_ned)
+{
+    std::cout << "Acceleration north: " << imu_reading_ned.acceleration.north_m_s2 << " m/s^2, "
+              << "east: " << imu_reading_ned.acceleration.east_m_s2 << " m/s^2, "
+              << "down: " << imu_reading_ned.acceleration.down_m_s2 << " m/s^2" << std::endl;
```
JonasVautherin:
```suggestion
              << "down: " << imu_reading_ned.acceleration.down_m_s2 << " m/s^2, "
```

src/integration_tests/telemetry_async.cpp
```diff
@@ -215,6 +223,20 @@ void print_ground_speed_ned(Telemetry::GroundSpeedNED ground_speed_ned)
...
+    << "Angular velocity north: " << imu_reading_ned.angular_velocity.north_rad_s << " rad/s, "
+    << "east: " << imu_reading_ned.angular_velocity.east_rad_s << " rad/s, "
+    << "down: " << imu_reading_ned.angular_velocity.down_rad_s << " rad/s" << std::endl;
```
JonasVautherin:
```suggestion
    << "down: " << imu_reading_ned.angular_velocity.down_rad_s << " rad/s, "
```

src/integration_tests/telemetry_async.cpp
```diff
@@ -215,6 +223,20 @@ void print_ground_speed_ned(Telemetry::GroundSpeedNED ground_speed_ned)
...
+    << "Magnetic field north: " << imu_reading_ned.magnetic_field.north_gauss << " G, "
+    << "east: " << imu_reading_ned.magnetic_field.east_gauss << " G, "
+    << "down: " << imu_reading_ned.magnetic_field.down_gauss << " G" << std::endl;
```
JonasVautherin:
```suggestion
    << "down: " << imu_reading_ned.magnetic_field.down_gauss << " G, "
```

krady21: Thanks for the review. I have added the suggestions(for some reason i couldn't commit your suggestions directly so i made another commit to add the desired changes)! The telemetry_async.cpp test still doesn't seem to print the imureading, though.

PR #781, opened 2019-07-01
README: add new MAVSDK logo

julianoes:
I could not find a smaller png, so I used the html tag to resize it.


hamishwillee:
Thanks. I did similar in the docs.

PR #782, opened 2019-07-02
Add offboard/set_actuator_control to offboard plugin without specifying group names

irsdkv:
Implementation of set_actuator_control method due to https://github.com/mavlink/MAVSDK/pull/771#pullrequestreview-256231538.

I think that this solution most corresponds to [mavlink message](https://mavlink.io/en/messages/common.html#SET_ACTUATOR_CONTROL_TARGET) specification compared [first PR](https://github.com/mavlink/MAVSDK/pull/771).

Also in protobuf3 "0" is the default for numeric types so this would hide the complexity.

The corresponding MAVSDK-Proto PR is [here](https://github.com/mavlink/MAVSDK-Proto/pull/88).


irsdkv:
Hi @julianoes 

I tried to implement your suggestion.
Also I tried to avoid unnecessary load on the CPU by passing the length of the array in MAVSDK API method.

If I misunderstand something I will be happy if you will explain it. 
This feature is important for us and we want implement really good solution for community :)

Thank you!

julianoes:
@irsdkv what do you think about this suggestion? https://github.com/mavlink/MAVSDK/pull/771#issuecomment-507982988

So the API would be:

```
std::array<float, 16> actuators = { 0.1, 0.2, 0.3, 0.4, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0 };
```
and if the last 8 entries are zero, we only send the message to control group 0, and we don't need to send the message to control group 1.

Would that work?

irsdkv:
> and if the last 8 entries are zero, we only send the message to control group 0, and we don't need to send the message to control group 1.

Unfortunately, due to px4 uORB subscription subsystem, if we need to fill all controls to zero - we should send all-zero message. If we don't do this then `actuator_controls_1` subscribers will not receive zeroed controls and the actual value will still remain the previous received (non-zero) values.

julianoes:
Ok, what about NAN?

irsdkv:
> Ok, what about NAN?

Hm. I think that this can be a possible solution

src/plugins/offboard/include/plugins/offboard/offboard.h
```diff
@@ -131,6 +131,29 @@ class Offboard : public PluginBase {
...
+        float controls[16]; /**< @brief Actuator controls array.
+                               First 8 will go to control group 0, the following 8 to control group 1.
+                               If you want to send less than 16 controls you should to set first non-specified
+                               control to NAN.
+                               If index of first NAN value will be less or equal than 8 then only one
+                               message for Group 0 will be sent. All controls from first NAN value
+                               to 7 in Group 0 will be set to zero.
+                               If index of first NAN value will be great than 8 then two messages (to
+                               Group 0 and Group 1) will be sent. In this case all controls from first NAN value
+                               to 7 in Group 1 will be set to zero. */
```
irsdkv: Perhaps this description should be more understandable for native English speakers :)

julianoes: :smile: what about:

```
Up to 16 actuator controls can be set. To ignore an output, set it to NaN.
The first 8 actuator controls internally map to control group 0, the latter 8 actuator controls map to control group 1.
Depending on what controls are set (instead of NaN) 1 or 2 MAVLink messages are actually sent.
```

Given that description is a bit awkward, would it be clearer, like that:

```
struct ActuatorControl {
    struct Group {
        float controls[8];
    };
    Group groups[2];
}
```

(Or is that almost what you had suggested in the beginning? Sorry for the long discussion)

irsdkv: All ok, I understand that you just want to construct good solution :)


> ```
> struct ActuatorControl {
>     struct Group {
>         float controls[8];
>     };
>     Group groups[2];
> }
> ```

This is the possible solution, but In this case, in my assumption we need to change [ActuatorControl structure in Proto](https://github.com/mavlink/MAVSDK-Proto/pull/88/files#diff-d071a0f40909dedf07ff7e03789cb56aR72) too due to API "consistency". It can be like that:

```
message ControlGroup {
  repeated float controls = 1;
}
message ActuatorControl {
  repeated ControlGroup control_group = 1;
}
```

But this can be too comlpicated?

-------

If you just don't want to have field like `group_number` in data types then I have this proposal:

```
struct ActuatorControl {
    struct Group {
        float controls[8];
    };
    union {
        Group groups[2];
        float controls[16];
    }
}
```
In this case [current](https://github.com/mavlink/MAVSDK-Proto/pull/88/files#diff-d071a0f40909dedf07ff7e03789cb56aR72) protobuf message can be used.
Also description of usage can be like that:

```suggestion
Up to 16 actuator controls can be set. To ignore an **output group**, set it to NaN. In case of **all** controls in the group is NaN than this group will not be sent (otherwise all NaN controls will sent as zero).
The first 8 actuator controls internally map to control group 0, the latter 8 actuator controls map to control group 1.
Depending on what controls are set (instead of NaN) 1 or 2 MAVLink messages are actually sent. */
```

What do you think about it?

julianoes: > ```
> message ControlGroup {
>  repeated float controls = 1;
>}
>message ActuatorControl {
>  repeated ControlGroup control_group = 1;
>}
>```
>
> But this will be too comlpicated?

I'd say that's fine.

irsdkv: Okay, I will implement it.
In this case the `union` is not required.

irsdkv: Done

src/plugins/offboard/offboard.cpp
```diff
@@ -80,6 +85,43 @@ const char *Offboard::result_str(Result result)
...
+               << ", Command port 0: " << actuator_control.groups[0].controls[0]
+               << ", Command port 1: " << actuator_control.groups[0].controls[1]
+               << ", Command port 2: " << actuator_control.groups[0].controls[2]
+               << ", Command port 3: " << actuator_control.groups[0].controls[3]
+               << ", Command port 4: " << actuator_control.groups[0].controls[4]
+               << ", Command port 5: " << actuator_control.groups[0].controls[5]
+               << ", Command port 6: " << actuator_control.groups[0].controls[6]
+               << ", Command port 7: " << actuator_control.groups[0].controls[7] << "], "
+               << "[group: " << 1
+               << ", Command port 0: " << actuator_control.groups[1].controls[8]
```
julianoes: This will probably segfault, right? :smile:

src/plugins/offboard/offboard_impl.cpp
```diff
@@ -484,6 +511,41 @@ void OffboardImpl::send_attitude_rate()
...
+    for (int i = 0; i < 2; i++) {
+        int nan_count = 0;
+        for (int j = 0; i < 8; j++) {
+            if (std::isnan(actuator_control.groups[i].controls[j])) {
+                nan_count++;
+                actuator_control.groups[i].controls[j] = 0.0f;
+            }
+        }
+        if (nan_count < 8) {
+            send_actuator_control_message(&actuator_control.groups[i].controls[0], i);
```
julianoes: Nice I think that should work!

src/plugins/offboard/offboard.cpp
irsdkv: :man_facepalming:

irsdkv:
It's a pleasure. Thank you too :smile:

julianoes:
CI complains about `std::isnan`:

```
/home/user/DronecodeSDK/src/plugins/offboard/offboard.cpp:92:18: error: '__builtin_isnan' is not a member of 'std'
             if ((std::isnan(lhs.groups[i].controls[j]) && std::isnan(rhs.groups[i].controls[j])) ||
                  ^
```

Try adding `#include <cmath>` in offboard.cpp.

julianoes:
Looks like the grpc service doesn't quite compile (anymore):
https://travis-ci.org/mavlink/MAVSDK/jobs/556753980#L7412

irsdkv:
> Looks like the grpc service doesn't quite compile (anymore):
> https://travis-ci.org/mavlink/MAVSDK/jobs/556753980#L7412

My bad. Seems like just need to update proto submodule rev

julianoes:
I think now you only need:
```
./tools/fix_style.sh .
```

irsdkv:
Done

JonasVautherin:
@irsdkv Thanks for all the work! We recently fixed the clang-format issues, that should be fine now. Sorry for the inconvenience there :sweat_smile:.

Can you try `fix_style` again, from latest `develop`? Just want to make sure it works for you as well. If not I'll have to investigate.

irsdkv:
I think all works ok :)

JonasVautherin:
Hmm somehow it seems like some styling was still wrong. I pushed a fix, let's hope it works this time :-).

irsdkv:
> Hmm somehow it seems like some styling was still wrong. I pushed a fix, let's hope it works this time :-).

Oh, sorry. It was my changes. I did it because I thought it was more readable and similarly to other functions like this. I apologize for not telling you this.

P.S. Whew. Seems that this pull request has a hard fate :smile:

JonasVautherin:
Oh, you're missing some docs (see Jenkins):

```
Parsing file /tmp/jenkins/workspace/mavlink_MAVSDK_PR-782/tools/docs/install/include/mavsdk/plugins/camera/came/tmp/jenkins/workspace/mavlink_MAVSDK_PR-782/tools/docs/install/include/mavsdk/plugins/offboard/offboard.h:151: warning: Compound mavsdk::Offboard::ActuatorControl::Group is not documented.

```

Good news is that this is the last CI test 😆

JonasVautherin:
Argh, somehow having issues with Jenkins. Trying to fix that.

PR #786, opened 2019-07-09
Disable CODEGEN and protobuf binaries when cross-compiling

JonasVautherin:
It is useless (i.e. we don't really want to run protoc on the host machine, but rather at build time on the... well... build machine). And the cross-compilation may fail for those components. So let's just not build them.


JonasVautherin:
Hmmm, and of course it is failing on manylinux, because for now dockcross sets manylinux as "cross-compiling". I have an open PR there, I'll update this one as soon as it is merged and fixed at the dockcross level.

JonasVautherin:
Oh, and it's already released! Awesome!

PR #788, opened 2019-07-11
core/plugins: workaround clang-format bug

julianoes:
We move the callback to the last argument to workaround a bug in clang-format 8.

See bug report:
https://bugs.llvm.org/show_bug.cgi?id=41676


julianoes:
Now it gets more SEO :)

julianoes:
Ok, so the older clang-format is not happy with this just yet :confused:.

julianoes:
Not needed now that #790 went in.

PR #791, opened 2019-07-12
camera: support for range based params

### LINKED ISSUE #695: Camera range options not supported
We should support camera settings that are not an enum option but rather a range.

An example for such a param can be found in the camera manager:

https://github.com/Dronecode/camera-manager/blob/eef47667506eb97e6ee7349489b7738a825dae14/samples/def/camera-def-uvc.xml#L57-L59

julianoes:
This brings initial support in the camera plugin for range based parameters.

Range based parameters are parameters without a predefined option list like an enum but rather a min, max, step size and a type.

FYI @dogmaphobic 

Fixes #695.

### DIFF
--- a/src/backend/src/grpc_server.h
+++ b/src/backend/src/grpc_server.h
@@ -23,6 +23,16 @@
 #include "plugins/offboard/offboard.h"
 #include "offboard/offboard_service_impl.h"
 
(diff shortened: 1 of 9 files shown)


julianoes:
CI error unrelated, merging.

Issue #340, opened 2018-03-25
Rename DroneCore class to dronecode::SDKCore


hamishwillee:
As per https://github.com/dronecore/DroneCore/issues/275#issuecomment-375713530

shakthi-prashanth-m:
We have namespace `dronecore`. So we don't have to say dronecore again in the class name isn't it? We can now rename namespace to `dronecode`. And may keep class as `dronecode::SDKCore` ?

shakthi-prashanth-m:
Or even `dronecode::CoreSDK` ?

julianoes:
I'd vote `dronecode::SDKCore`, that's inline with the repo name.

julianoes:
Well, this has been done a while ago.

PR #798, opened 2019-07-15
Telemetry: async read Unix epoch time (mavlink msg #340) from GPS

elbarto1980:
Hi,
I noticed that in the latest releases of PX4 firmware, GPS time has been added as "microseconds since UNIX epoch"; it is available in the mavlink message #340 (UTM_GLOBAL_POSITION).
I thought that, if you have a companion computer that doesn't have a RTC Backup Battery, it could be very useful to get the time from GPS. It can be used also to sync time in a swarm of drones.


elbarto1980:
I know... Using the QGC Mavlink Inspector widget, I checked the "time" field of every single message coming from the PX4; the UTM message is the only one that contains UTC time. All the other ones have time from  boot. A GPS_TIME message would be great.

julianoes:
> the UTM message is the only one that contains UTC time

Right but looking at the MAVLink spec it is not even specifically documented as being different from the other messages.

elbarto1980:
yes, the descriptions of the time fields are mostly the same in all the messages.

hamishwillee:
@julianoes re [UTM_GLOBAL_POSITION](https://mavlink.io/en/messages/common.html#UTM_GLOBAL_POSITION)

> Right but looking at the MAVLink spec it is not even specifically documented as being different from the other messages. 

This is a broadcast message that acts as a transponder for the drone. It makes no sense for it to be UAS-relative (ie from boot) like most other timestamps in the system, because other systems need to be able to compare it to their own GPS co-ordinates.

That said, there is no requirement that a system implement this, and I wouldn't rely on it. I think something like TIMESYNC might be better ... if its use were documented.

julianoes:
> This is a broadcast message that acts as a transponder for the drone. It makes no sense for it to be UAS-relative (ie from boot) like most other timestamps in the system, because other systems need to be able to compare it to their own GPS co-ordinates.

@hamishwillee ok we should document that then.

thomasetter:
I just saw this and to raise awareness about the problems of leap seconds, I have to mention that UTC and GPS are different times, with UTC being affected by leap seconds, while [GPS time](https://en.wikipedia.org/wiki/Global_Positioning_System#Timekeeping) is not.

We should make sure to distinguish between the two and call things consistently with one of the terms, because e.g. on NTP-synced devices the difference could be noticeable.

hamishwillee:
@thomasetter @julianoes I don't *think* we need to worry too much about this because we've talked consistently about UNIX epoch:

- [UTM_GLOBAL_POSITION.time](https://mavlink.io/en/messages/common.html#UTM_GLOBAL_POSITION): *Time of applicability of position (microseconds since UNIX epoch).* 
- All other timestamps are `time_usec`: *Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.*

So there will have to be conversions done if timestamps are coming from GPS, but at the MAVLink level that should not matter.

Or am I missing something ... in which case, can you guys advise on what change you would like to see.

src/plugins/telemetry/telemetry_impl.h
```diff
@@ -197,6 +201,9 @@ class TelemetryImpl : public PluginImplBase {
     mutable std::mutex _rc_status_mutex{};
     Telemetry::RCStatus _rc_status{false, false, 0.0f};
 
+    mutable std::mutex _utm_epoch_mutex{};
+    uint64_t _utm_epoch{};
```
julianoes:
```suggestion
    uint64_t _utm_epoch_us {};
```

src/plugins/telemetry/telemetry_impl.h
```diff
@@ -115,20 +118,21 @@ class TelemetryImpl : public PluginImplBase {
...
+    void process_attitude_quaternion(const mavlink_message_t &message);
+    void process_mount_orientation(const mavlink_message_t &message);
+    void process_imu_reading_ned(const mavlink_message_t &message);
+    void process_gps_raw_int(const mavlink_message_t &message);
+    void process_extended_sys_state(const mavlink_message_t &message);
+    void process_sys_status(const mavlink_message_t &message);
+    void process_heartbeat(const mavlink_message_t &message);
+    void process_statustext(const mavlink_message_t &message);
+    void process_rc_channels(const mavlink_message_t &message);
+    void process_utm_global_position(const mavlink_message_t &message);
```
julianoes: Please fix the conflicts here.

src/integration_tests/telemetry_async.cpp
```diff
@@ -275,3 +280,10 @@ void print_position_velocity_ned(Telemetry::PositionVelocityNED position_velocit
 
     _received_position_velocity_ned = true;
 }
+
+void print_utm_epoch(uint64_t time_us)
+{
+    std::time_t t = time_us / 10E5;
+    std::cout << "UTC (" << time_us / 10E5 << "): " << std::put_time(std::gmtime(&t), "%c %Z")
```
julianoes: This can be `LogInfo() << ` and then without `std::endl;`.

src/plugins/telemetry/include/plugins/telemetry/telemetry.h
```diff
@@ -855,13 +863,27 @@ class Telemetry : public PluginBase {
      */
     typedef std::function<void(RCStatus rc_status)> rc_status_callback_t;
 
+    /**
+     * @brief Callback type for UTM global position updates.
```
julianoes: Since this only contains the time, I would rename the method to what it is, e.g. `unix_epoch_time`.

elbarto1980:
@julianoes I tried to fix all the conflicts and to rename variables and methods as you suggested... but at a certain point, things got really messy here on github.
Now, even if my code compiles, I am not sure that I fixed everything.
Could you please double-check my commits?
Thank you for your help!

julianoes:
On travis it doesn't find `std::put_time`:

https://travis-ci.org/mavlink/MAVSDK/jobs/568547781#L2393

elbarto1980:
I checked on the Internet, "the std::get_time and std::put_time manipulators are still missing in gcc 4.9" ([see here](https://stackoverflow.com/questions/14136833/stdput-time-implementation-status-in-gcc)). Travis is using gcc (Ubuntu 4.8.4-2ubuntu1~14.04.3) 4.8.4.
I have just pushed a commit that should fix it (I can't test it...Trevis will tell us).

I think I have also fixed the "timeout" issue.

elbarto1980:
I checked on the Internet, "the std::get_time and std::put_time manipulators are still missing in gcc 4.9" ([see here](https://stackoverflow.com/questions/14136833/stdput-time-implementation-status-in-gcc)). Travis is using gcc (Ubuntu 4.8.4-2ubuntu1~14.04.3) 4.8.4.
I have just pushed a commit that should fix it (I can't test it...Travis will tell us).

I think I have also fixed the "timeout" issue.

src/plugins/telemetry/telemetry_impl.h
```diff
@@ -120,19 +120,19 @@ class TelemetryImpl : public PluginImplBase {
...
+    void process_attitude_quaternion(const mavlink_message_t& message);
+    void process_mount_orientation(const mavlink_message_t& message);
+    void process_imu_reading_ned(const mavlink_message_t& message);
+    void process_gps_raw_int(const mavlink_message_t& message);
+    void process_extended_sys_state(const mavlink_message_t& message);
+    void process_sys_status(const mavlink_message_t& message);
+    void process_heartbeat(const mavlink_message_t& message);
+    void process_statustext(const mavlink_message_t& message);
+    void process_rc_channels(const mavlink_message_t& message);
+    void process_unix_epoch_time(const mavlink_message_t& message);
```
julianoes: This is still wrong.

src/integration_tests/telemetry_async.cpp
```diff
@@ -275,3 +280,15 @@ void print_position_velocity_ned(Telemetry::PositionVelocityNED position_velocit
...
 }
+
+void print_unix_epoch_time_us(uint64_t time_us)
+{
+    std::time_t time = time_us / 10E5;
+#if __GNUC__ > 4
+    LogInfo() << "UTC (" << time << "): " << std::put_time(std::gmtime(&time), "%c %Z");
+#else
+    char time_string[24];
+    if (0 < strftime(time_string, sizeof(time_string), "%c %Z", std::gmtime(&time)))
```
julianoes: Good find. Maybe add this:
```
// TODO: remove this once we don't have to build with gcc 4 any longer.
```

src/plugins/telemetry/telemetry_impl.h
elbarto1980: I'm sorry, I can't see it... what is wrong?

elbarto1980: I'm sorry, I can't see it... what is wrong?

julianoes: Sorry, I must have looked at a wrong diff. This is fine.

elbarto1980:
Thank you Julian!

PR #801, opened 2019-07-17
bugfix: camera_photo_interval is not set correctly

JonasVautherin:
Seems like this field was just ignored...

Resolves [this issue in Swift](https://github.com/mavlink/MAVSDK-Swift/issues/137).

@byuarus @douglaswsilva: FYI


JonasVautherin:
Tested to work by @byuarus.

PR #808, opened 2019-07-22
Fix Jenkins CI failures

### LINKED ISSUE #813: Failed to setpoint in offboard mode in SITL
Hello, 

I am using the MAVSDK for gazebo SITL with IRIS quadrotor. I have created automated tests to fly IRIS quadcopter in a specific trajectories.  The quadrotor functions normally in most cases of test. However, during some tests IRIS quadrotors fails to set point, I receive this error rarely:

`Offboard::start() failed: No setpoint set`

I do not understand why this error is occurring from time to time (about once each 30 flights).
You can find the log of one of the flights in the following link. However, I feel it is more related to the SDK rather than PX4.

https://logs.px4.io/plot_app?log=e85cdc2d-f3f1-4907-bddb-08440c3684d4

Do you have any idea??

julianoes:
This is a mixed bag to hopefully remove the spurious Jenkins CI failures.

Also fixes #813 (hopefully).


Jenkinsfile
```diff
@@ -51,7 +51,7 @@ pipeline {
             sh 'git submodule sync --recursive'
             sh 'git submodule update --init --recursive --force'
             sh 'ccache -z'
-            sh 'cmake -DCMAKE_BUILD_TYPE=Release -DBUILD_BACKEND=ON -DBUILD_SHARED_LIBS=OFF -DBUILD_SHARED_LIBS=OFF -DCMAKE_INSTALL_PREFIX=${WORKSPACE}/install -DENABLE_MAVLINK_PASSTHROUGH=ON -Bbuild/release -H.'
```
JonasVautherin: How is that related to Jenkins crashes? And to the cmake cache?

julianoes: No that's not responsible, sorry.

JonasVautherin: So I made it this way (i.e. static libs) in Jenkins because Travis does the shared libs: https://github.com/mavlink/MAVSDK/blob/develop/tools/travis-docker-build.sh#L8.

Do you think we should just build shared libs everywhere in CI?

julianoes:
Well, no matter how many times you restart it, something is always wrong with Jenkins. :disappointed:

Jenkinsfile
julianoes: I didn't change the function here, I only removed the duplicate flag.

JonasVautherin: Oh, I just understood it... damn it took me long enough to realize what you meant :sweat_smile:

PR #809, opened 2019-07-22
core: fix destruction of serial connection

julianoes:
When nothing weas being received over serial the destruction of the serial connection hung waiting for a read to complete.

By using a timeout every second, we can recover after one second. It's not optimal but better than stalling completely.


julianoes:
> I have to trust you on that :blush:

I thought closing it was enough but presumably I was wrong.

The right approach might be to switch to some async serial read/write implementation at some point.

julianoes:
I'm gonna ignore the CI failures which are unrelated.

PR #810, opened 2019-07-22
core: added getter for version

julianoes:
Because that's common for a library to implement.

### DIFF
--- a/src/core/mavsdk_impl.cpp
+++ b/src/core/mavsdk_impl.cpp
@@ -46,24 +46,25 @@ std::string MavsdkImpl::version() const
 
     ++version_counter;
 
-    if (version_counter == 10) {
-        return "You were wondering about the name of this library?";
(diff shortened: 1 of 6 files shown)


src/core/mavsdk_impl.cpp
```diff
@@ -42,7 +42,29 @@ MavsdkImpl::~MavsdkImpl()
 
 std::string MavsdkImpl::version() const
 {
-    return mavsdk_version;
+    static unsigned version_counter = 0;
+
+    ++version_counter;
+
+    if (version_counter == 10) {
```
MaEtUgR: How about a switch case?

julianoes: Yer, might as well.

julianoes: dd6ba97c

PR #815, opened 2019-07-24
Add gimbal mode support

douglaswsilva:
Implementation of set gimbal mode per: https://github.com/mavlink/MAVSDK-Proto/pull/93

@julianoes: 
"`YAW_FOLLOW` means that the gimbal follows the autopilot yaw, so it always looks forward (although with a little bit of damping so the picture is smoother).
`YAW_LOCK` means that the gimbal locks onto a specific heading no matter what direction the drone faces."


src/plugins/gimbal/include/plugins/gimbal/gimbal.h
```diff
@@ -84,6 +84,32 @@ class Gimbal : public PluginBase {
...
 
+    /** 
+     * @brief Set gimbal mode (synchronous).
+     *
+     * This sets the desired mode of a gimbal.
+     * The function will return when the command is accepted, however, it might
+     * take the gimbal longer to actually be set to the new angles.
+     *
+     * @param gimbal_mode The mode to be set. Either lock or follow.
+     * Lock will fix the gimbal poiting to a direction. Follow will follow the vehicle heading.
```
julianoes: Please document that this is for yaw.

src/plugins/gimbal/include/plugins/gimbal/gimbal.h
```diff
@@ -84,6 +84,32 @@ class Gimbal : public PluginBase {
...
+     * @return Result of request.
+     */
+    Result set_gimbal_mode(int gimbal_mode);
```
julianoes: The argument should be an `enum class` similar to the proto definition.

src/plugins/gimbal/gimbal_impl.cpp
```diff
@@ -55,6 +55,43 @@ void GimbalImpl::set_pitch_and_yaw_async(
...
+    return gimbal_result_from_command_result(_parent->send_command(command));
+}
+
+void GimbalImpl::set_gimbal_mode_async(
+    int gimbal_mode, Gimbal::result_callback_t callback)
+{
+    MAVLinkCommands::CommandInt command{};
+
+    command.command = MAV_CMD_DO_MOUNT_CONFIGURE;
+    command.params.param1 = float(MAV_MOUNT_MODE_MAVLINK_TARGETING); ///gimbal_mode;
```
julianoes: Nitpick: `// ` is enough.

src/plugins/gimbal/gimbal_impl.cpp
```diff
@@ -55,6 +55,43 @@ void GimbalImpl::set_pitch_and_yaw_async(
...
+    command.params.param2 = 0.0f;
+    command.params.param3 = 0.0f;
+    command.params.param4 = float(gimbal_mode);
+    command.params.x = 0;
+    command.params.y = 0;
+    command.params.z = 2.0f;
```
julianoes: Maybe add a comment what this is.

src/plugins/gimbal/include/plugins/gimbal/gimbal.h
julianoes: Also, given this can change the behavior of the existing gimbal setting, make sure add a comment there on how the yaw angle is interpreted differently depending on the mode.

JonasVautherin:
It should also update the proto submodule, right? Note that there may be a small typo to fix in offboard, from [this PR in MAVDSK-Proto](https://github.com/mavlink/MAVSDK-Proto/pull/94).

src/plugins/gimbal/include/plugins/gimbal/gimbal.h
douglaswsilva: Can you elaborate. What do you mean by the Yaw angle?

JonasVautherin:
I fixed the typo in #821, while updating the proto submodule. So you won't have to update the proto submodule here :blush:.

src/plugins/gimbal/include/plugins/gimbal/gimbal.h
julianoes: I mean the angle that you pass to `set_pitch_and_yaw`, it's different given the mode, right?

src/plugins/gimbal/gimbal_impl.cpp
douglaswsilva: I was not sure about the extension of the comments, so let me know if I added too much (or too little)

src/plugins/gimbal/gimbal_impl.cpp
```diff
@@ -91,6 +91,18 @@ void GimbalImpl::set_gimbal_mode_async(
...
+float GimbalImpl::to_float_gimbal_mode(const Gimbal::GimbalMode gimbal_mode) const
+{
+    switch (gimbal_mode) {
+        case Gimbal::GimbalMode::YAW_FOLLOW:
+            return 0.0f;
+        case Gimbal::GimbalMode::YAW_LOCK:
+            return 1.0f;
+        default:
+            return 0.0f;
+    }
```
julianoes: Nice!

src/plugins/gimbal/include/plugins/gimbal/gimbal.h
douglaswsilva: So, if I understood your questions correctly, the angle yaw input will be the same for both modes, which is `param7 = 2` from options: `(0 = angle body frame, 1 = angular rate, 2 = angle absolute frame)`. What will change is the `stabilize yaw` between `(1 = yes, 0 = no)` 
0 = Yaw Follow
1 = Yaw Lock

Let me know if you meant something different by angle.

julianoes: But doesn't the gimbal yaw angle mean something else? Basically in lock mode 0 will point North, and in follow mode it will point forward.
Therefore, the yaw variable is relative to the world frame for lock, and relative to the vehicle for follow frame.

src/plugins/gimbal/include/plugins/gimbal/gimbal.h
```diff
@@ -84,6 +92,34 @@ class Gimbal : public PluginBase {
...
+
+    /** 
+     * @brief Set gimbal mode (asynchronous).
+     *
+     * This sets the desired yaw mode of a gimbal.
+     * The function will return when the command is accepted, however, it might
+     * take the gimbal longer to actually be set to the new angles.
+     *
+     * @param gimbal_mode The mode to be set. Either yaw lock or yaw follow.
+     * Yaw lock will fix the gimbal poiting to an absolute direction.
```
douglaswsilva: @julianoes @JonasVautherin What do you guys think of this, should I detail it more?

julianoes: That's ok. I would adapt the description of [set_pitch_and_yaw](https://github.com/mavlink/MAVSDK/blob/f0d8db27473e6db99876fe5201093a8d319c6182/src/plugins/gimbal/include/plugins/gimbal/gimbal.h#L69) and note that the yaw angle reference depends on the gimbal mode.

julianoes: I did the follow up fix for this:  #826

julianoes:
Thanks a lot @douglaswsilva!

PR #819, opened 2019-07-26
Allow messages to be stopped or rate set to default

julianoes:
This restores the functionality for messages to be stopped entirely or the rate set to the default which was lost a while ago. Also, it's now documented in the telemetry API docstrings.


JonasVautherin:
What's the use for that? If you don't want to receive updates anymore, wouldn't you unregister your callback instead? :thinking:

julianoes:
That's true but you might want to configure the messages not to be sent over the link anymore. So this is rather a MAVLink hack and not an SDK thing.

PR #820, opened 2019-07-26
Trying out GitLab CI

julianoes:
GitLab CI presumably allows local runners on Linux, Windows, and macOS, so this is worth a try.


julianoes:
Somehow the git mirroring doesn't work here https://gitlab.com/mavsdk/MAVSDK
The webhooks notifying gitlab are working but GitLab doesn't want to fetch.

julianoes:
Closing until we consider this again.

PR #824, opened 2019-07-30
CMake changes to support MAVSDK as submodule

Stifael:
CMakeLists integration_tests: use relative cmake path
CMakeLists core: include binary dir to get version.h


Stifael:
luckily didn't require too many changes :)

julianoes:
Hehe @Stifael fixing this take two, thanks!

PR #833, opened 2019-08-08
mavlink_parameters: workaround for param mismatch

julianoes:
This could workaround a potential problem with E90 where the MAVLink parameter type does not match the xml type.

@byuarus please test. I ran out of time to test this against an E90 but I thought it was worth a shot.

### DIFF
--- a/src/core/mavlink_parameters.cpp
+++ b/src/core/mavlink_parameters.cpp
@@ -372,6 +372,15 @@ void MAVLinkParameters::process_param_ext_value(const mavlink_message_t& message
                     work->get_param_callback(
                         MAVLinkParameters::Result::SUCCESS, correct_type_value);
                 }
+            } else if (value.is_uint8() && work->param_value.is_uint32()) {
+                // FIXME: workaround for mismatching type uint8_t which should be uint32_t.
+                ParamValue correct_type_value;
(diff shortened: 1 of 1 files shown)


src/core/mavlink_parameters.cpp
```diff
@@ -363,6 +363,14 @@ void MAVLinkParameters::process_param_ext_value(const mavlink_message_t& message
                 if (work->get_param_callback) {
                     work->get_param_callback(MAVLinkParameters::Result::SUCCESS, value);
                 }
+            } else if (value.is_uint8() && work->param_value.is_uint16()) {
+                // FIXME: workaround for mismatching type uint8_t which should be uint16_t.
+                ParamValue correct_type_value;
+                correct_type_value.set_uint16(uint16_t(value.get_uint8()));
```
JonasVautherin: Shouldn't it be a `static_cast` in C++11?

julianoes: Would be nicer yes.

byuarus:
@julianoes @JonasVautherin 
Hey guys, I found the same issue with the E10T camera and made a PR https://github.com/mavlink/MAVSDK/pull/841
It will support mismatch uint8 <-> uint32

PR #837, opened 2019-08-12
core: link to libatomic on arm

### LINKED ISSUE #836: Protobuf fails to build on RPi
Various places in the code fail with references to __atomic_xxx calls.  

```
[ 98%] Built target libprotoc
Scanning dependencies of target protoc
[ 99%] Building CXX object CMakeFiles/protoc.dir/srv/maverick/var/build/dronecore/build/default/third_party/protobuf/protobuf/src/protobuf/src/google/protobuf/compiler/main.cc.o
[100%] Linking CXX executable protoc
/usr/bin/ld: libprotobuf.a(arena.cc.o): in function `google::protobuf::internal::ArenaImpl::Init()':
arena.cc:(.text+0x44): undefined reference to `__atomic_fetch_add_8'
```

This can be worked around by adding the following to CMakeLists.txt:
`set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -latomic")`

```
gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/lib/gcc/arm-linux-gnueabihf/8/lto-wrapper
Target: arm-linux-gnueabihf
Configured with: ../src/configure -v --with-pkgversion='Raspbian 8.3.0-6+rpi1' --with-bugurl=file:///usr/share/doc/gcc-8/README.Bugs --enable-languages=c,ada,c++,go,d,fortran,objc,obj-c++ --prefix=/usr --with-gcc-major-version-only --program-suffix=-8 --program-prefix=arm-linux-gnueabihf- --enable-shared --enable-linker-build-id --libexecdir=/usr/lib --without-included-gettext --enable-threads=posix --libdir=/usr/lib --enable-nls --enable-bootstrap --enable-clocale=gnu --enable-libstdcxx-debug --enable-libstdcxx-time=yes --with-default-libstdcxx-abi=new --enable-gnu-unique-object --disable-libitm --disable-libquadmath --disable-libquadmath-support --enable-plugin --with-system-zlib --with-target-system-zlib --enable-objc-gc=auto --enable-multiarch --disable-sjlj-exceptions --with-arch=armv6 --with-fpu=vfp --with-float=hard --disable-werror --enable-checking=release --build=arm-linux-gnueabihf --host=arm-linux-gnueabihf --target=arm-linux-gnueabihf
Thread model: posix
gcc version 8.3.0 (Raspbian 8.3.0-6+rpi1)
```

julianoes:
It seems like that's required on platforms like Raspberry Pi.

Fixes #836.

### DIFF
--- a/src/core/CMakeLists.txt
+++ b/src/core/CMakeLists.txt
@@ -37,6 +37,13 @@ target_link_libraries(mavsdk
     ${CMAKE_THREAD_LIBS_INIT}
 )
 
+if (${CMAKE_SYSTEM_PROCESSOR} MATCHES "arm")
+    # It seems like on platforms like Raspberry Pi we need to link to libatomic.
+    target_link_libraries(mavsdk
+        atomic
+    )
+endif()
+
 if (IOS)
     target_link_libraries(mavsdk
         PUBLIC


JonasVautherin:
What platform is that for? armv7? If it worked with dockcross without changing our CMakeLists, would that do it? I did compile for armv7 with dockcross successfully, like 2 weeks ago. There is just a small issue because we run `protoc` at cmake configure time, and I'm working on removing that.

TL;DR: it could be that this is not an issue when building with dockcross.

julianoes:
@JonasVautherin no idea!

JonasVautherin:
Note: with #839 and #840, I just built for linux-armv7 with the following, without requiring any change to our CMakeLists:

```
docker run --rm dockcross/linux-armv7 > ./dockcross-linux-armv7
chmod +x dockcross-linux-armv7
./dockcross-linux-armv7 cmake -DBUILD_BACKEND=ON -DBUILD_SHARED_LIBS=OFF -Bbuild/linux-armv7 -S.
./dockcross-linux-armv7 cmake --build build/linux-armv7
```

fnoop:
This is for armv7l platform (Raspberry Pi), running the latest Raspbian Buster, which is based on Debian 10.
Not sure what dockcross has to do with it, it breaks building on a Raspberry.

JonasVautherin:
> Not sure what dockcross has to do with it, it breaks building on a Raspberry.

@fnoop: dockcross is a convenient way to cross-compile. For instance I can just run the docker commands listed above on my computer, and that gives me a `mavsdk_server` binary compiled for armv7 that I can just run on an armv7 device.

I was mentioning that in case it could solve your issue, but I get from your answer that you want to build from the RPi?

It's just that I'm a bit nervous about linking `libatomic` for arm in our CMakeLists to fix an issue with one of our dependencies (it seems like #836 fails to build protoc, and I am not sure how this is related to the `mavsdk` target :thinking:).

fnoop:
@JonasVautherin dockcross sounds cool.  Doesn't solve this issue however.  Also I notice that the CMakeLists.txt in third_party/protobuf contains this:
```
if(CMAKE_CROSSCOMPILING)
    list(APPEND CMAKE_ARGS
        "-Dprotobuf_BUILD_PROTOC_BINARIES=OFF"
        )
endif()
```
If it's not needed, can we turn it off for all builds?

JonasVautherin:
> @JonasVautherin dockcross sounds cool. Doesn't solve this issue however.

Can you be more specific? Did you try on latest the latest MAVSDK `develop` branch? What error do you get?

> If it's not needed, can we turn it off for all builds?

TL;DR: do you actually need to build on the RPi (i.e. can you elaborate on your use-case), or would you be fine cross-compiling for the RPi using dockcross?

Detailed answer:

Well, it is not needed to cross-compile `protoc` because `protoc` is used to generate sources on the build machine. Say you build from your Linux x86_64 machine for the RPi, you will want to run `protoc` on the x86_64 machine, and cross-compile the resulting sources for the RPi.

I can't say that it is not needed when not cross-compiling (sorry for the double negation :sweat_smile:), because you may need it to run `tools/generate_from_protos.sh`.

We could do something like `if(CMAKE_CROSSCOMPILING OR <arch_where_it_fails>)` to disable it when the build machine is the RPi, or we could use @julianoes' fix so that it actually builds on the RPi (but then probably the fix goes in `third_party/protobuf/CMakeLists.txt`. I would vote for the latter :blush:.

fnoop:
@JonasVautherin returning conversation to #836.

JonasVautherin:
Closing this as it does not fix #836, so that we can keep the discussion there.

PR #838, opened 2019-08-12
Enable cmake build option to build without cURL

### LINKED ISSUE #830: Add option to build without curl
For slow compilation (e.g. on embedded boards) it would be nice to be able to build without curl.

julianoes:
This adds the cmake build option BUILD_CURL which can be set to off to disable building and including libcURL in the MAVSDK build.

@JonasVautherin do you think that's bearable?

Fixes #830.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,20 @@ matrix:
             - gcc-5
             - g++-5
     - os: linux
+      name: "ubuntu 14.04 - gcc 6 without curl"
+      dist: trusty
+      sudo: false
+      compiler: gcc
+      env:
+      - MATRIX_EVAL="BUILD_TARGET=no_curl_build && CC=gcc-6 && CXX=g++-6"
+      addons:
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+          packages:
+            - gcc-6
+            - g++-6
+    - os: linux
       name: "ubuntu 14.04 - gcc 6"
       dist: trusty
       sudo: false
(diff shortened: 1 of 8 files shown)


JonasVautherin:
Your call :laughing:.

But I'll be honest: I don't like it, mainly for the following reasons:

* I am not aware of an actual situation where statically linking curl is a no-go for embedded devices (and #830 does not convince me about that :sweat_smile:), but yet we add 7 `if(BUILD_CURL)` statements in our CMakeLists, one build option, one definition and its corresponding `#ifdef`, and some CI building one of the new flavours coming from that (but not all of them, as it would require combining with the `MAVLINK_PASSTHROUGH` ifdefs).
* The reason mentioned in the issue #830 (i.e. "it helps debugging on the device") is IMHO a bad one, because that's not how we solve the fact that devs don't know how to deal with dependencies.

Again, I could live with it and keep it as some "necessary complexity that I don't understand", but in the current state I really don't get the issue with having curl (other than "I don't like compiling it" :sweat_smile:).

Note that we could move curl to the camera plugin, so that those embedded systems could live without it. And if curl is still too much overhead for the camera plugin, maybe we should consider removing completely and relying on e.g. MAVLink FTP.

julianoes:
> And if curl is still too much overhead for the camera plugin, maybe we should consider removing completely and relying on e.g. MAVLink FTP.

Maybe in the future we can do that.

julianoes:
Closing in favour of #865.

PR #839, opened 2019-08-12
816 version generated files

### LINKED ISSUE #816: Move protoc generation out of cmake and version generated files
@julianoes, @thomasetter: I'd like to have your opinion on that :blush:.

We use protobuf + gRPC in `mavsdk_server`, and therefore generate code with the proto compiler (`protoc`), at build time. I believe this should change, mainly for the following reasons:

* We will soon want to autogenerate more code, using our own protoc plugin. It will make the autogeneration a bit more complex (will require to build a python module, etc), and therefore I don't feel like keeping that in our CMakeLists. We already have separate scripts for that [in Swift](https://github.com/mavlink/MAVSDK-Swift/blob/master/tools/generate_from_protos.bash) and [in Python](https://github.com/mavlink/MAVSDK-Python/blob/master/other/tools/run_protoc.sh).
* We will autogenerate the C++ header files, and it would make sense to version them, so that we can link to the github files in discussions. We already do that [in Swift](https://github.com/mavlink/MAVSDK-Swift/tree/master/Sources/MAVSDK-Swift/Generated) and [in Python](https://github.com/mavlink/MAVSDK-Python/tree/master/mavsdk/generated).
* When cross-compiling, we currently need to build for the host first, so that we get `protoc`, and then we can cross-compile. It takes time (e.g. on the CI), and sometimes it doesn't work. For instance when using dockcross on my machine, my host-built `protoc` relies on a version of glibc that is not available in the dockcross container, and therefore `protoc` doesn't run there.

For those reasons, I believe it makes sense to generate the files separately, and version them.

The new workflow would look a bit like this:

1. If you don't need to generate the files (typically when cross-compiling), you just run `cmake` (configure) and  `cmake --build` (build), and rely on the versioned generated files.
2. If you need to generate the files, you:

    a. run `cmake` (configure) so that you get `protoc`
    b. run `generate_protos.sh`
    c. run `cmake --build` (build)

@thomas.etter: Note that this would break the changes you did for automatically re-generating when the proto files change.

JonasVautherin:
@thomasetter: FYI

* Generate protobuf and gRPC source files from a script (instead of doing it in the cmake configure step)
* Version those generated files

Resolves #816.


JonasVautherin:
hmm not passing the style checks. @julianoes: do you know if there is a way to ignore a directory in `fix_style.sh`? I see some `style_ignore` logic but I'm not sure I completely get it :confused:.

julianoes:
@JonasVautherin it looks like the `style-ignore.txt` file was removed at some point. I've added it back and it seems to work.

JonasVautherin:
Ubuntu 14.04 with gcc6 is failing for an obscure reason :sweat_smile:.

julianoes:
Yer I've seen this in other places. It's the `coveralls` thing, maybe because coveralls was down or being maintained: https://status.coveralls.io/.

PR #841, opened 2019-08-13
workaround for the problem with E10T where the MAVLink the MAVLink pa…

byuarus:
Extended the workaround for the problem with E10T
> Comparison type mismatch between uint8_t and uint32_t (mavlink_parameters.h:407)


julianoes:
@byuarus thanks :smile: . I think you'll have to run `tools/fix_style.sh .`

darioxz:
Could it be that the datatype of the camera MAVLink parameter in the definition file is wrong? Which parameter causes a problem?

julianoes:
@darioxz it's a mismatch between the MAVLink param and the type documented in the xml.

julianoes:
Thanks @byuarus. I'll take care of the style fix in the other branch.

Issue #736, opened 2019-04-30
Health not clearing cache


douglaswsilva:
Telemetry Health Subscription does not clear cache when aircraft is disconnected, so if you turn aircraft off and on again, it will say it has home position and global position even though it has not locked GPS yet. This will prevent vehicle form taking off.

JonasVautherin:
@julianoes: how does a plugin know the vehicle is disconnected?

julianoes:
Good find @douglaswsilva.

@JonasVautherin we should reset in here:
https://github.com/Dronecode/DronecodeSDK/blob/0e6a1aeb6ce3132e04ee922720dfc875ea138bf1/plugins/telemetry/telemetry_impl.cpp#L126

irsdkv:
I have a same problem with param module.
[Cache used](https://github.com/mavlink/MAVSDK/blob/45d908df6a03e5826199cbea710707b4b7c7d1c2/src/core/mavlink_parameters.cpp#L94) during all backend runtime although UAV was rebooted or params values was changed from the other source (QGC or PX4 console, for example).

Also I think solution should have API for gRPC request  like "CacheResetRequest" for modules which use caches.

What are you think about this?

irsdkv:
Or may be it would be ideal if user will have ability to turn caching off/on for module

JonasVautherin:
> Also I think that solution should have API for gRPC request like "CacheResetRequest" for modules which use caches.

My opinion about caches is that it should be transparent. If the user needs to be aware of the fact there is a cache, then it is a bad cache :sweat_smile:.

Examples:

* Caching a "battery status" is fine, so that users receive the last position as soon as they subscribe to it instead of waiting until the next battery state (which may be a few seconds away).
* Caching parameters is fine if the autopilot broadcasts the new values whenever they change. It is not fine if MAVSDK has no way to know that the parameter was changed, and therefore we should just always request the param, with the caveat that sometimes this may time out.
* Caching the mission is not a good idea unless we have a way to know that the mission got updated on the drone.

If we can't have a real (transparent) cache, let's not have any cache, and push that as a MAVLink issue that needs to be fixed at the autopilot level (i.e. the autopilot should tell us when its state change).

What do you think?

irsdkv:
This is good clarification.
I dont know about specific use cases for Battery Status and Mission, but I think that Parameters module have no reasons to use cache:

- Seems that MAVSDK have no way to know that parameter was changed
[...]

PR #847, opened 2019-08-27
Remove caching in params module

irsdkv:
Due to [this](https://github.com/mavlink/MAVSDK/issues/736#issuecomment-522110195) conversation:

- Seems that MAVSDK has no way to know that parameter was changed on UAV after reboot or using console or QGC, for example
- Use cases of MAVSDK probably does not include real necessity of fast frequency parameters values polling (which may require cache using)

### DIFF
--- a/src/core/mavlink_parameters.cpp
+++ b/src/core/mavlink_parameters.cpp
@@ -342,7 +342,6 @@ void MAVLinkParameters::process_param_ext_value(const mavlink_message_t& message
                 // FIXME: workaround for mismatching type uint8_t which should be uint16_t.
                 ParamValue correct_type_value;
                 correct_type_value.set_uint16(static_cast<uint16_t>(value.get_uint8()));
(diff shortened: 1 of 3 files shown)


julianoes:
Closes #736.

JonasVautherin:
@julianoes: Does that really solve #736? The health cache issue comes from the telemetry plugin, right?

Also @hamishwillee mentioned there that the autopilot is supposed to broadcast updated params :sweat_smile:.

julianoes:
> The health cache issue comes from the telemetry plugin, right?

Right, we need to verify this one. I'm not sure which values are not up-to-date.

irsdkv:
Somebody, can you please merge this PR if all ok?
I have not enough permissions for that :)

JonasVautherin:
It's not completely clear to me from #736 if this is compliant or not :sweat_smile:. Asking there.

PR #860, opened 2019-09-23
Check existing libraries

caleridas:
Check for presence of required libraries instead of pulling them unconditionally
ourselves.

HACK: The cmake use is not 100% clean due to having to support cmake 3.1 for the
time being.


JonasVautherin:
Closing in favor of #865. Thanks @caleridas for the help!

PR #862, opened 2019-09-24
Target system ID check fix

irsdkv:
Some messages like [ATT_POS_MOCAP](https://mavlink.io/en/messages/common.html#ATT_POS_MOCAP) don't have `target_system` field.
For these messages [`mavlink_get_msg_entry()`](https://github.com/mavlink/c_library_v2/blob/a66f06dd300338aef7da306b403941bd9c0dd3a8/mavlink_helpers.h#L491) returned struct [`mavlink_msg_entry`](https://github.com/mavlink/c_library_v2/blob/a66f06dd300338aef7da306b403941bd9c0dd3a8/mavlink_types.h#L289) with field `target_system_ofs` setted to `0` (see [comment](https://github.com/mavlink/c_library_v2/blob/a66f06dd300338aef7da306b403941bd9c0dd3a8/mavlink_types.h#L287) and for example  [`MAVLINK_MESSAGE_CRCS`](https://github.com/mavlink/c_library_v2/blob/a66f06dd300338aef7da306b403941bd9c0dd3a8/common/common.h#L27) entry for msg `#138`).

So we need to check is this offset equals to `0` and allow message to be sent on each iface.

### DIFF
--- a/src/core/udp_connection.cpp
+++ b/src/core/udp_connection.cpp
@@ -138,7 +138,9 @@ bool UdpConnection::send_message(const mavlink_message_t& message)
     // on the matching link.
     const mavlink_msg_entry_t* entry = mavlink_get_msg_entry(message.msgid);
     const uint8_t target_system_id =
-        (entry ? reinterpret_cast<const uint8_t*>(message.payload64)[entry->target_system_ofs] : 0);
+        (entry && (entry->flags & MAV_MSG_ENTRY_FLAG_HAVE_TARGET_SYSTEM) ?
+             reinterpret_cast<const uint8_t*>(message.payload64)[entry->target_system_ofs] :
+             0);
 
     bool send_successful = true;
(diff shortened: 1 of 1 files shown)


julianoes:
Oh, thanks for the fix! I think we should also check for `MAV_MSG_ENTRY_FLAG_HAVE_TARGET_SYSTEM` as here:
https://github.com/PX4/Firmware/blob/07d656e971a72d1202651dfd3b4642736fb078d7/src/modules/mavlink/mavlink_main.cpp#L455

irsdkv:
You are right.
Message [MISSION_REQUEST_LIST](https://mavlink.io/en/messages/common.html#MISSION_REQUEST_LIST) have `target_sys_id` offset eq `0`, but have it:
`{43, 132, 2, 3, 3, 0, 1}`

PR #864, opened 2019-09-25
Move from json11 to jsoncpp

julianoes:
This way we can make use of system-wide installed jsoncpp.

FYI @caleridas.


third_party/jsoncpp/CMakeLists.txt
```diff
@@ -22,16 +22,14 @@ elseif(IOS)
...
 endforeach()
 
 ExternalProject_Add(
-    json11
-    GIT_REPOSITORY https://github.com/dropbox/json11
-    GIT_TAG v1.0.0
-    PREFIX json11
-    PATCH_COMMAND git apply ${PROJECT_SOURCE_DIR}/export_config.patch
+    jsoncpp
+    URL https://github.com/open-source-parsers/jsoncpp/archive/1.9.1.zip
```
julianoes: Nice!

JonasVautherin:
What's with `0.10.1`? It seems like it is a 4-years old release? And after a quick look it seems like it doesn't export the tinyxml target.

What was the issue with the latest version on Ubuntu 16?

caleridas:
> "we" might but for the community we have to support 16.04 for some time, at least until 20.04 is out.

*Probably* it is best then to just not build it yourselves on Ubuntu at all: it has libjsoncpp-1.7.2 packaged already, and security team is taking care of it. Picking ancient version just for convenience of your build setup is probably not a good idea.

(And this gets me to the other important reason to strongly push people to using *maintained* software -- I doubt you want to become security maintainer for third party software).

julianoes:
Let's keep everything together in #865, closing.

PR #868, opened 2019-10-01
core: add geometry file for global/local frames

julianoes:
This adds functionality to convert from local coordinates in meters to global latitude/longitude and back.

This will be used to easily create tests missions without hard-coded latitude/longitude coordinates.


julianoes:
> Nice! We could also use that to expose mission items with local coordinates, couldn't we?

Good idea, that would make the interface easier.

PR #874, opened 2019-10-11
zlib: use nicer patch to install cmake config

julianoes:
This cmake patch seems nicer because it also installs the cmake config.


julianoes:
Appveyor runs through fine and the unit tests pass and then it exits with not-null :man_shrugging:?

JonasVautherin:
> Appveyor runs through fine and the unit tests pass and then it exits with not-null man_shrugging?

Seems to fail consistently :thinking:

PR #877, opened 2019-10-14
Update grpc to 1.24.0

JonasVautherin:
It requires to recompile the proto files, and also to checkout with git, because they added a new submodule...


JonasVautherin:
Oh sorry, I could have mentioned that. It fixes the build with glibc 2.30, so we will need it eventually (e.g. I need it now) :blush:.

PR #878, opened 2019-10-15
Add shell plugin

irsdkv:
My attempt to implement MAVLink shell.

Implementation may contain some ineffective or hazardous code. But testing (With [MAVSDK-Python PR](https://github.com/mavlink/MAVSDK-Python/pull/120)) looks ok.


src/plugins/shell/shell_impl.cpp
```diff
@@ -0,0 +1,145 @@
...
+        }
+    }
+    if (shell_message.data.length() > 70 && _result_subscription) {
+        auto callback = _result_subscription;
+        Shell::Result arg{Shell::Result::ResultCode::DATA_TOO_LONG, {}};
+        _parent->call_user_callback([callback, arg]() { callback(arg); });
+        return;
+    }
+
+    _shell_message_mutex.lock();
```
julianoes: I suggest to always use `std::lock_guard` for anything that we add now, and also remove the old `lock()`, `unlock()`s at some point.

```
{
    std::lock_guard<std::mutex> lock(_shell_message_mutex);
    _shell_message = shell_message;
}
```

src/plugins/shell/shell_impl.cpp
```diff
@@ -0,0 +1,145 @@
...
+{
+    _parent->unregister_plugin(this);
+}
+
+void ShellImpl::shell_message_response_async(Shell::result_callback_t callback)
+{
+    _result_subscription = callback;
+}
+
+void ShellImpl::send_shell_message(const Shell::ShellMessage& shell_message)
```
julianoes: What about calling this function `shell_command()`?

src/plugins/shell/shell_impl.cpp
```diff
@@ -0,0 +1,145 @@
...
+{
+    if (!_parent->is_connected()) {
+        if (_result_subscription) {
```
julianoes: I think you need to create locks for all of the member variables to make the plugin thread-safe. You can probably just grab the lock for most of the action, and you probably need to have something to track if we are currently doing a shell roundtrip, or if we can already send the next thing.

irsdkv:
Im sorry for force-push, but it turned out that I created PR with the old version of the code.

src/plugins/shell/shell_impl.cpp
```diff
@@ -0,0 +1,145 @@
...
+            auto callback = _result_subscription;
+            Shell::Result arg{Shell::Result::ResultCode::NO_SYSTEM, {}};
+            _parent->call_user_callback([callback, arg]() { callback(arg); });
+            return;
+        }
+    }
+    if (shell_message.data.length() > 70 && _result_subscription) {
```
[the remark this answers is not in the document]
irsdkv: Now used [MAVLINK_MSG_SERIAL_CONTROL_FIELD_DATA_LEN](https://github.com/mavlink/MAVSDK/pull/878/files#diff-f4c5701958a5b01cfeef470e26440de5R106)

irsdkv:
Looks good except `ESC` trailing symbol.
image

```python
b'param show SYS_AUTOSTART\nSymbols: x = used, + = saved, * = unsaved\nx + SYS_AUTOSTART [732,1060] : 4011\n\n 767/1307 parameters used.\nnsh> \x1b[K'
```

src/plugins/shell/shell_impl.cpp
```diff
@@ -0,0 +1,192 @@
...
+                &_shell_message_timeout_cookie);
+        }
+    }
+}
+
+void ShellImpl::receive_shell_message_timeout()
+{
+    std::lock_guard<std::mutex> lock_result(_result_mutex);
+    std::lock_guard<std::mutex> lock_subscription(_result_subscription_mutex);
+    std::lock_guard<std::mutex> lock_cookie(_shell_message_timeout_cookie_mutex);
```
julianoes: It's quite dangerous to acquire multiple locks in this way. If you try to acquire them in another order in another method you have a deadlock. I think easiest is to only have one lock for all member variables together, or alternatively you only ever acquire one lock at a time.

src/plugins/shell/shell_impl.h
```diff
@@ -0,0 +1,51 @@
+#pragma once
+
+#include <atomic>
```
julianoes: Looks like this include is not used.

src/plugins/shell/shell_impl.cpp
```diff
@@ -0,0 +1,192 @@
...
+void ShellImpl::init()
+{
+    using namespace std::placeholders; // for `_1`
+    _parent->register_mavlink_message_handler(
+        MAVLINK_MSG_ID_SERIAL_CONTROL,
+        std::bind(&ShellImpl::process_shell_message, this, _1),
+        this);
+}
+
+void ShellImpl::deinit() {}
```
julianoes: You should call `unregister_all_mavlink_message_handlers()` here I think.

tools/generate_from_protos.sh
```diff
@@ -24,7 +24,7 @@ command -v ${protoc_binary} && command -v ${protoc_grpc_binary} || {
     exit 1
 }
 
-plugin_list="action calibration geofence gimbal camera core info mission offboard param telemetry"
+plugin_list="action calibration gimbal camera core info mission offboard param shell telemetry"
```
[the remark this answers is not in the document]
irsdkv: Oops
Sorry, my bad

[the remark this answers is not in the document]
irsdkv: Oops
Sorry, my bad

src/plugins/shell/shell_impl.cpp
irsdkv: I`m tried to implement some `promise`-`future` scheme.
It is not ideal I think, but it worked ok.

src/plugins/shell/shell_impl.cpp
```diff
@@ -18,60 +18,86 @@ void ShellImpl::init()
...
-    std::lock_guard<std::mutex> lock_subscription(_result_subscription_mutex);
-    if (!_parent->is_connected()) {
-        if (_result_subscription) {
-            auto callback = _result_subscription;
-            Shell::Result arg{Shell::Result::ResultCode::NO_SYSTEM, {}};
-            _parent->call_user_callback([callback, arg]() { callback(arg); });
+    {
+        std::lock_guard<std::mutex> lock(_future_mutex);
+        if (is_transfer_in_progress()) {
             return;
```
irsdkv: It is not good to just `return` if current transfer not completed yet.
May be need to return some `bool` value? 
What do you think?

src/plugins/shell/shell_impl.cpp
```diff
@@ -0,0 +1,213 @@
+#include "shell_impl.h"
+#include "system.h"
+#include "global_include.h"
+#include "px4_custom_mode.h"
+#include <cmath>
+#include <functional>
+#include <array>
```
julianoes: Same here, check what is needed.

src/plugins/shell/shell_impl.cpp
julianoes: Usually I would return a `Shell::Result` and in this case it would say `Shell::Result::BUSY`.

src/plugins/shell/include/plugins/shell/shell.h
```diff
@@ -0,0 +1,124 @@
...
+    struct ShellMessage {
+        bool need_response; /**< @brief Set if wants the receiver to send a response. */
+        uint16_t timeout; /**< @brief Timeout for reply data in ms. */
+        std::string data; /**< @brief Serial data. */
+    };
+
+    /**
+     * @brief Shell requests results type.
+     */
+    struct Result {
```
julianoes: `Result` is usually an `enum` so this is bit different then other plugins.

What about using this callback to return:
```
typedef std::function<void(Result result, ShellMessage response)> result_callback_t;
```

I'm not sure if it's nicer but it's more consistent :thinking:?
@JonasVautherin do you have an opinion?

src/plugins/shell/shell_impl.h
```diff
@@ -0,0 +1,55 @@
...
+    bool send_shell_message_mavlink();
+
+    void process_shell_message(const mavlink_message_t& message);
+
+    void receive_shell_message_timeout();
+
+    bool is_transfer_in_progress();
+
+    std::promise<void> _transfer_finished_promise{};
+    std::future<void> _transfer_finished_future{};
```
julianoes: It's an interesting idea to use futures in this way. Haven't thought about that.

src/plugins/shell/include/plugins/shell/shell.h
irsdkv: This does not allow [this feature](https://github.com/mavlink/MAVSDK/pull/878#discussion_r335893631). I should change it corresponding other modules.

irsdkv: [Done](https://github.com/mavlink/MAVSDK/pull/878/commits/d985378c1e4bb4ef93d9fa02d7f6352916d8158e#diff-f4c5701958a5b01cfeef470e26440de5R49)

src/plugins/shell/shell_impl.cpp
irsdkv: [Done](https://github.com/mavlink/MAVSDK/pull/878/commits/d985378c1e4bb4ef93d9fa02d7f6352916d8158e#diff-f4c5701958a5b01cfeef470e26440de5R49)

src/plugins/shell/include/plugins/shell/shell.h
irsdkv: [Done](https://github.com/mavlink/MAVSDK/pull/878/commits/d985378c1e4bb4ef93d9fa02d7f6352916d8158e#diff-8ca1b5f6cd83dbe50248a5028f19845eR78)

src/plugins/shell/shell_impl.cpp
julianoes: We can leave it for now and try it for a bit!

irsdkv:
image

```shell
[100%] Linking CXX executable unit_tests_backend.exe
```

So close... :smile:

src/plugins/shell/shell_impl.cpp
```diff
@@ -0,0 +1,215 @@
...
+{
+    _parent->unregister_plugin(this);
+}
+
+bool ShellImpl::is_transfer_in_progress()
+{
+    if (!_transfer_finished_future.valid()) {
+        return true;
+    }
+    return _transfer_finished_future.wait_for(std::chrono::seconds(0)) != std::future_status::ready;
```
julianoes: Could you use `valid()` here?
https://en.cppreference.com/w/cpp/thread/future/valid

irsdkv: Actually this is not what we want.
`valid` returns `true` until `get()` is called on the future. So in our case it will return `true` every time.
And in this case 
```cpp
if (!_transfer_finished_future.valid()) {
    return true;
}
```
seems like useless check. Is should fix it :)

julianoes: Oh, I see. Nevermind then.

JonasVautherin:
Thanks a lot @irsdkv! :rocket:

PR #883, opened 2019-10-19
Remove unused dependencies

JonasVautherin:
LibreSSL has been working well for a while (and removed our dependency on Golang at build time, by the way :raised_hands:). Time to remove OpenSSL and BoringSSL.


julianoes:
Jenkins is broken.

PR #884, opened 2019-10-20
Move from C++11 to C++17

julianoes:
This is work in progress to see if we could potentially move to C++17.

This will need a newer GCC on Ubuntu 16.04 which can be installed as described here:https://gist.github.com/jlblancoc/99521194aba975286c80f93e47966dc5


JonasVautherin:
Note: we'll have to test manylinux, android and ios

docker/Dockerfile.manylinux2010-x64-custom
```diff
@@ -0,0 +1,10 @@
+FROM dockcross/manylinux2010-x64
+
+ENV DEFAULT_DOCKCROSS_IMAGE manylinux2010-x64-custom
+
+ADD https://www.cpan.org/src/5.0/perl-5.28.1.tar.gz .
```
JonasVautherin: So we still need to add perl manually with 2010? Did you try without it?

julianoes: No, I did not.

julianoes: You're right. It's not needed anymore. So I'll remove our custom stuff.

julianoes:
Closing this as #1196 has been merged.

PR #887, opened 2019-10-23
cmake: remove -Werror

### LINKED ISSUE #611: Remove `-Werror` except for CI
We should probably disable to error on warnings (`-Werror`) except for CI runs.

Currently, `-Werror` is included here:
https://github.com/Dronecode/DronecodeSDK/blob/ef45e4ddb7a8c5dcd0ac20df9a330afdaa9a7be6/cmake/compiler_flags.cmake#L13

but it would be better if this was not set by default and added in CI.

julianoes:
I think it's a better idea to only use -Werror in CI because:
- This way the build is more robust when built for unknown environments with unknown compilers and compiler versions.
- Many developers get confused by warnings and are blocked assuming they are actual errors.

Fixes #611.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -23,7 +23,7 @@ jobs:
       - name: Install lcov
         run: sudo apt-get install -y lcov
       - name: configure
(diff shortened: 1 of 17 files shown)


julianoes:
9 months later, and I'm giving this another try :smile:.

julianoes:
Got it in within one year, come on!

PR #889, opened 2019-10-23
Remove debug build from appveyor

JonasVautherin:
Debug times out, for some reason (I would have expected Release to be slower). Maybe linking time makes the difference?

Also I don't think anybody cares about Win32, so let's not waste a machine for that.


julianoes:
I'm fine removing the debug builds.

But remember we needed it for the mavsdk_server as part of the Python package? Because Python 32 bit seems to be the default on Windows.

JonasVautherin:
You're right!

julianoes:
I've removed appveyor completely in the meantime.

PR #890, opened 2019-10-25
Add simple time synchronization with Autopilot

irsdkv:
Hi all
At this time MAVSDK does not allow to get FCU time.
I did a little research and think that this method of time sync will be the simplest.
In common idea it based on [mavros plugin](https://github.com/mavlink/mavros/blob/master/mavros/src/plugins/sys_time.cpp) but without filter.


src/core/global_include.cpp
```diff
@@ -158,4 +164,18 @@ bool are_equal(double one, double two)
     return (std::fabs(one - two) < std::numeric_limits<double>::epsilon());
 }
 
+FCUTime::FCUTime() {}
```
julianoes: What's `FCU`? Can we have a name that's easier to understand? Maybe `Autopilot`?

src/core/timesync.h
```diff
@@ -0,0 +1,41 @@
...
+
+#include <atomic>
+#include <mutex>
+
+#include "global_include.h"
+#include "mavlink_include.h"
+
+// Since not all vehicles support/require level calibration, this
+// is disabled for now.
+//#define LEVEL_CALIBRATION
```
julianoes: Hm?

src/plugins/mocap/mocap_impl.cpp
```diff
@@ -73,13 +74,23 @@ Mocap::Result MocapImpl::set_odometry(const Mocap::Odometry& odometry)
...
-            static_cast<uint64_t>(_parent->get_time().elapsed_s() * 1e3);
+        fcu_time_usec = std::chrono::duration_cast<std::chrono::microseconds>(
+                            _parent->get_fcu_time().now().time_since_epoch())
+                            .count();
+    } else {
+        fcu_time_usec =
+            std::chrono::duration_cast<std::chrono::microseconds>(
+                _parent->get_fcu_time()
+                    .time_in(std::chrono::microseconds(vision_position_estimate.time_usec))
+                    .time_since_epoch())
```
julianoes: Should we have a helper in `SystemImpl` for this or is this ever only used here?

src/plugins/mocap/mocap_impl.cpp
```diff
@@ -103,14 +114,25 @@ bool MocapImpl::send_vision_position_estimate()
...
     _attitude_position_mocap_mutex.lock();
     auto attitude_position_mocap = _attitude_position_mocap;
     _attitude_position_mocap_mutex.unlock();
 
     if (!attitude_position_mocap.time_usec) {
-        attitude_position_mocap.time_usec =
-            static_cast<uint64_t>(_parent->get_time().elapsed_s() * 1e3);
+        fcu_time_usec = std::chrono::duration_cast<std::chrono::microseconds>(
+                            _parent->get_fcu_time().now().time_since_epoch())
+                            .count();
```
julianoes: Nevermind, it looks like it's used more than once :smile:.

src/plugins/mocap/mocap_impl.cpp
```diff
@@ -135,13 +157,24 @@ bool MocapImpl::send_attitude_position_mocap()
...
 {
+    uint64_t fcu_time_usec{};
+
     _odometry_mutex.lock();
     auto odometry = _odometry;
     _odometry_mutex.unlock();
 
     if (!odometry.time_usec) {
-        odometry.time_usec = static_cast<uint64_t>(_parent->get_time().elapsed_s() * 1e3);
+        fcu_time_usec = std::chrono::duration_cast<std::chrono::microseconds>(
```
julianoes: How do you like this?

```

const uint64_t fcu_time_usec =
    (!odometry.time_usec)
    ? fcu_time_usec = std::chrono::duration_cast<std::chrono::microseconds>(_parent->get_fcu_time().now().time_since_epoch()).count()
    : std::chrono::duration_cast<std::chrono::microseconds>(_parent->get_fcu_time().time_in(std::chrono::microseconds(odometry.time_usec)).time_since_epoch()).count();
```

src/core/global_include.h
```diff
@@ -63,6 +67,32 @@ class FakeTime : public Time {
...
+        std::lock_guard<std::mutex> lock(_fcu_system_time_offset_mutex);
+        return dl_fcu_time_t(local_system_time_point + std::chrono::duration_cast<T>(_fcu_time_offset));
+    };
+
+private:
+    mutable std::mutex _fcu_system_time_offset_mutex{};
+    std::chrono::nanoseconds _fcu_time_offset{};
+
+    virtual dl_stime_t system_time();
+};
```
julianoes: Does this all need to be globally available? I guess so for the templates?

src/core/global_include.cpp
irsdkv: That's make sense :)

src/core/global_include.h
irsdkv: Refactored

irsdkv:
This is for EKF2 in MoCap, for example:
> Technically this can be set to 0 if there is correct timestamping (not just arrival time) and timesync (e.g NTP) between MoCap and (for example) ROS computers. [Link](https://dev.px4.io/v1.9.0/en/ros/external_position_estimation.html#tuning-EKF2_EV_DELAY)

> [Link] Clocks are synchronized when using MAVROS(https://github.com/PX4/Firmware/blob/549fb0d5de8a8eaebaa54d6e4d458876b53b60d2/src/modules/ekf2/ekf2_main.cpp#L1192)

Autopilot expect that time is synchronized.

Also may be it will be useful for some other purposes.

julianoes:
Somehow there is a build error with macOS:
http://ci.px4.io:8080/blue/organizations/jenkins/mavlink%2FMAVSDK/detail/PR-890/13/pipeline/29

irsdkv:
Checks passed now

julianoes:
@irsdkv sorry I had the review done this morning but forgot to send it. I didn't want to delay it yet again :see_no_evil:.

julianoes:
@irsdkv sorry for being so slow here. This can go in, right?

irsdkv:
Hi! All ok :)
I had some tests and seems that this can be merged.
Anyway if I will find issues in this module in future I will create fix PR :)

julianoes:
Ok thanks for the quick follow-up.

PR #891, opened 2019-10-29
Publish Windows artefacts

julianoes:
Let's give this a test.

Presumably this should be possible now:
https://changelog.travis-ci.com/windows-jobs-can-now-deploy-76366


JonasVautherin:
That's interesting, my last experience was that it was failing because of the secrets. And their fix was to not hide the secrets (i.e. publish the credentials required for releasing the artifact). Their thread is [here](https://travis-ci.community/t/current-known-issues-please-read-this-before-posting-a-new-topic/264/21).

.travis.yml
```diff
@@ -119,7 +119,8 @@ script:
     make -Cbuild/ios_simulator -j4;
   fi
 - if [[ "${BUILD_TARGET}" = "windows" ]]; then
-    cmake -G "Visual Studio 15 2017" -DCMAKE_INSTALL_PREFIX=${TRAVIS_BUILD_DIR}/build/debug/install -DBUILD_BACKEND=ON -DBUILD_SHARED_LIBS=OFF -Bbuild/debug -H${TRAVIS_BUILD_DIR};
+    cmake -G "Visual Studio 15 2017" -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=${TRAVIS_BUILD_DIR}/build/release/install -DBUILD_BACKEND=ON -DBUILD_SHARED_LIBS=OFF -Bbuild/release -H${TRAVIS_BUILD_DIR};
+    cmake --build build/release -j4;
```
JonasVautherin: > 'RuntimeLibrary': value 'MD_DynamicRelease' doesn't match value 'MDd_DynamicDebug'

For Windows, you need to set the configuration at build time (and we need it a configure time, because we build the dependencies there).

In appveyor we do:

```
cmake --build build --target install --config %configuration% -j 2
```

So here I would think something like `cmake --build build/release --config Release -j4;`

julianoes: Yes thanks

JonasVautherin:
BTW note that the issue (and contact with Travis CI support team) I had was in May 2019, and your link above is "a year old". So I'm really not convinced that this is fixed :confused:.

julianoes:
Merging because the Jenkins CI error is not related.

Issue #885, opened 2019-10-22
Landed State is jumping between modes


shrit:
Hello, 

I am doing simulation for three quadcopters in SITL Gazebo, I need to make the quadcopter fly at the same time and reach high take-off altitude. 
To do this, I use three threads and a while sleep inside the take-off function until the quadcopters finish their taking off procedure.
The problem arises when the first while loops in the takeoff function blocks on unwanted states. After many observations, I have noticed that at least in one quadcopter LandedState jump from ON_GROUND to IN_AIR directly without passing by TAKING_OFF.
These are the while loop from the takeoff function:

```

/* Wait until the landed state changes the mode */
while (true) {
std::this_thread::sleep_for(std::chrono::milliseconds(5));
if (landed_state_ == Telemetry::LandedState::TAKING_OFF) {
LogInfo() << "Landed State : "<< telemetry_->landed_state_str(landed_state_);
break;
}
LogInfo() << "Landed State : "<< telemetry_->landed_state_str(landed_state_);
}
while (landed_state_ == Telemetry::LandedState::TAKING_OFF) {
LogInfo() << "Taking off..." ;
std::this_thread::sleep_for(std::chrono::milliseconds(300));
}
LogInfo() << "Taking off has finished successfully...";
```

Knowing that is landed state is recovered asynchronous from the SDK
Do you have any idea about this issue??

JonasVautherin:
That's interesting... did you check the message coming from PX4? Does PX4 say `IN_AIR`, or does it say `TAKING_OFF` and MAVSDK somehow reports the wrong state?

shrit:
Nope, I have no idea how to check the PX4 messages, In fact, the SDK report rarely `TAKING_OFF` but most of the time, I receive `IN_AIR` instead.
The only simple hack I had to do is to replace `landed_state()` function by `flight_mode()` since there is a TAKEOFF flight mode.

julianoes:
I think this state is only sent at 1Hz. Is it possible that you're just missing the update between `LANDED` and `IN_AIR`?

shrit:
@julianoes, Ok I understand, I have to test it without the 50 milliseconds sleep.

shrit:
I am not sure if I understand correctly, the above code might be able to miss out on the transition between `ON_GROUND` and `TAKEOFF` _since the state is sent only once each second_.
In this case, it might be true that I am missing it as the above code sleep for 5 milliseconds, 

[...]

PR #892, opened 2019-10-31
Use landed state to improve taking off

shrit:
This pull request is made to allow easier debug to the problem described in the issue https://github.com/mavlink/MAVSDK/issues/885 
Please do not merge this pull request.
It might require several taking off with landed state to see the problem (block into infinite while loop)
On the other hand, flight mode will never block in the while loop.

Even if you receive landed state or flight mode asynchronously it will be the same problem

To use the landed sate:
`./offboard udp://:14540 landed_state `

To use the flight mode
`./offboard udp://:14540 flight_mode`


shrit:
Ok, Would you like to keep both of them? the landed state and the flight mode as a use case for the users?

julianoes:
Oh I'm stupid, this was only to debug! Sorry!

shrit:
@julianoes, However, we can replace the sleep by using the flight mode to wait for the takeoff process to be completed.
What do you think of the last commit, I think this can be merged

julianoes:
But this is then just a workaround for the fact that PX4 is not picking up the landed state correctly. However, that will be fixed with https://github.com/PX4/Firmware/pull/13075/commits/f913cc24aec47fab3c55aad35841d41e30a4199d.

shrit:
Ok, So I will replace it with the flight mode with landed state, then you can merge when it is fixed at PX4 level.

PR #895, opened 2019-11-04
Revert zlib patch for cmake config

julianoes:
This reverts commit 04cab3de1647a07946338cb325233c8bd66de95a.

This broke linking libz statically.


julianoes:
Merging without CI because we know it worked before.

PR #899, opened 2019-11-07
travis: use ccache on Linux and macOS

julianoes:
I think it is worth a try to speed up the build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: cpp
+cache: ccache
 
 matrix:
   fast_finish: true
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -80,6 +81,10 @@ install:
     docker run --rm dockcross/linux-armv7 > ./dockcross-linux-armv7;
     chmod +x ./dockcross-linux-armv7;
   fi
+- if [[ "${BUILD_TARGET}" = "osx_build" ]] || [[ "${BUILD_TARGET}" = "ios_build" ]] || [[ "${BUILD_TARGET}" = "ios_simulator_build" ]]; then
+    brew install ccache;
+    export PATH="/usr/local/opt/ccache/libexec:$PATH";
```
JonasVautherin: so that's a travis path that stays between builds?

julianoes: According to the docs, yes. I doubt it though.
https://docs.travis-ci.com/user/caching/#ccache-cache

CMakeLists.txt
```diff
@@ -12,6 +12,23 @@ endforeach()
 
 project(mavsdk_superbuild)
 
+# ccache
```
JonasVautherin: Why are we moving that here? Does it affect the superbuild?

julianoes: I was hoping that it would enable ccache for the superbuild dependencies.

JonasVautherin: I don't understand completely how this cache config works in cmake, but it seems to be just about setting a few variables. Those would have to be propagated to the dependencies (I would say that they are not in the current setup)

julianoes: You're right 🤔 .

julianoes:
We moved from travis-ci to GitHub workflows now, closing.

PR #901, opened 2019-11-12
Fix camera settings deadlocks

julianoes:
This fixes a few potential deadlocks for users of the API trying to do another API call within the callback as well as one deadlock reported in https://github.com/mavlink/MAVSDK-Swift/issues/160.

Fixes https://github.com/mavlink/MAVSDK-Swift/issues/160.

Could you test this one @douglaswsilva?


douglaswsilva:
Thanks, will test this soon and will let you know. 👍🏼

PR #910, opened 2019-11-20
clang-format: prevent style changes

julianoes:
This seems to be necessary with Clang 9.


julianoes:
It's only the macOS build that fails.

PR #913, opened 2019-11-25
add flight information to Info plugin

DanielePettenuzzo:
The flight information mavlink message was so far available only in the camera plugin. In this plugin the user can extract the flight_uuid as part of the folder name that the SDK creates for your photos. This will only work if you have a camera hooked up. 

This pr adds the flight information to the info plugin. The plugin automatically starts and periodically sends the mavlink command to get the flight information. The use can then pull out this info with the get_flight_information()  method. 

@julianoes if you confirm that this makes sense I will add comments everywhere to keep the same structure as the other oomments in the header files.

### DIFF
--- a/src/plugins/info/include/plugins/info/info.h
+++ b/src/plugins/info/include/plugins/info/info.h
@@ -83,6 +83,15 @@ public:
     };
 
     /**
+     * @brief Type containing system flight information.
+     */
+    struct FlightInfo {
+        uint32_t time_boot_ms; /**< @brief Time since system boot. */
+        uint64_t flight_uid; /**< @brief Flight counter. Starts from zero, is incremented by 1
+                                  at every disarm and is never reset even after reboots.*/
+    };
+
+    /**
      * @brief Type containing identification.
      */
     struct Identification {
@@ -115,6 +124,14 @@ public:
      * the product information about the system.
      */
     std::pair<Result, Product> get_product() const;
+
+    /**
+     * @brief Get system flight information.
+     *
+     * @return a pair containing the result of the request and if successful,
+     * the flight information about the system.
+     */
+    std::pair<Result, FlightInfo> get_flight_information() const;
 
     /**
      * @brief Returns a human-readable English string for an Result.
--- a/src/plugins/info/info_impl.cpp
+++ b/src/plugins/info/info_impl.cpp
@@ -24,6 +24,11 @@ void InfoImpl::init()
         MAVLINK_MSG_ID_AUTOPILOT_VERSION,
         std::bind(&InfoImpl::process_autopilot_version, this, _1),
         this);
+
+    _parent->register_mavlink_message_handler(
+        MAVLINK_MSG_ID_FLIGHT_INFORMATION,
+        std::bind(&InfoImpl::process_flight_information, this, _1),
+        this);
 }
 
 void InfoImpl::deinit()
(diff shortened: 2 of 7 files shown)


src/plugins/info/include/plugins/info/info.h
```diff
@@ -82,6 +82,13 @@ class Info : public PluginBase {
         char product_name[32]; /**< @brief Name of product. */
     };
 
+    struct FlightInfo {
+        uint32_t time_boot_ms;
+        uint64_t arming_time_utc;
+        uint64_t takeoff_time_utc;
+        uint64_t flight_uuid;
```
thomasetter: This has been unfortunately named in PX4 and MAVLink, as [UUID](https://en.wikipedia.org/wiki/UUID) is a concept which is defined to be 128 bits and it is confusing to see something completely different with the same name.

Can we choose a better name or do we have to be consistent with PX4 naming?

@julianoes could this still be changed in MAVLink 2?

julianoes: I agree it should probably not be called UUID if it isn't one. Now the question is what sort of type it is. If it's based on the param that is counted up it would be an `int32_t`.

@DanielePettenuzzo you thought about this, right?

src/plugins/info/info_impl.cpp
```diff
@@ -36,19 +41,26 @@ void InfoImpl::enable()
...
     _parent->send_autopilot_version_request();
+    _parent->send_flight_information_request();
 
     // We're going to retry until we have the version.
     _parent->add_call_every(
         std::bind(&InfoImpl::request_version_again, this), 1.0f, &_call_every_cookie);
+
+    // We're going periodically ask for the flight information
+    _parent->add_call_every(
+        std::bind(&InfoImpl::request_flight_information, this), 1.0f, &_flight_info_call_every_cookie);
```
julianoes: Do we really have to check every second? We could also check if armed/disarmed has changed and then request it again.

src/plugins/info/include/plugins/info/info.h
DanielePettenuzzo: The type and name of the variable are the same as in the mavlink message. I could rename it in the SDK to have some name that makes more sense. Something like flight_counter or flight_number?

src/plugins/info/info_impl.cpp
DanielePettenuzzo: @julianoes I have a fix ready for this. Basically check at every disarm. The issue now is that if we decide to do it this way we need to remove arming_time_utc and takeoff_time_utc because they will always be zero. Do you prefer check at every disarm and keep only boot_time and flight_uuid or keep it like it is now and keep the whole FlightInfo structure.

julianoes: So generally something should work when it is merged to MAVSDK. Hence, we either need to implement it on the PX4 side and get it merged, or you need to remove it from the MAVSDK API.

src/plugins/info/include/plugins/info/info.h
julianoes: Or `flight_uid`.

src/plugins/info/info_impl.cpp
DanielePettenuzzo: @julianoes I'm not sure I understand. What I meant is that they are 0 because you are disarmed when you read those. When you arm they are refreshed but we never ask for the message then because we check only at disarm since that's when the flight_uuid is incremented.

julianoes: Oh got it. Maybe just update it for both changes then.

DanielePettenuzzo: as discussed with @julianoes offline we will check for this message only at disarm and won't read arming_time_utc and takeoff_time_utc. This is maybe something more for the telemetry plugin. I will push it in the next few minutes

DanielePettenuzzo:
@julianoes I pushed the following changes:
- check for the flight information message only when connecting and then at every disarm (not periodic anymore)
- add comments to info.h similar to all other data structures
- remove arming_time_utc and takeoff_time_utc fields from the message because we would need to read them periodically and this would maybe be more something for the telemetry plugin
- renamed flight_uuid into flight_uid

DanielePettenuzzo:
@julianoes fixed style and added integration test

PR #916, opened 2019-12-02
telemetry_impl: Add process_attitude call

patrickelectric:
It's only possible to get attitude information via quaternion messages with the actual implementation.
There is vehicles that the default attitude message is ATTITUDE and not ATTITUDE_QUATERNION.
This patch allow both messages to be used. 

Signed-off-by: Patrick José Pereira <patrickelectric@gmail.com>

### DIFF
--- a/src/plugins/telemetry/telemetry_impl.cpp
+++ b/src/plugins/telemetry/telemetry_impl.cpp
@@ -39,6 +39,11 @@ void TelemetryImpl::init()
         this);
 
     _parent->register_mavlink_message_handler(
+        MAVLINK_MSG_ID_ATTITUDE,
+        std::bind(&TelemetryImpl::process_attitude, this, _1),
+        this);
+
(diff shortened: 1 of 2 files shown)


patrickelectric:
Hi Jonas,

Both messages can be used and the information _should be the same_.
Quaternions are much more robust if someone wants to avoid the euler limitations and euler is pretty easy for any interface to deal with.
Right now, AFAIK, ArduPilot vehicles by default only provide ATTITUDE messages and not ATTITUDE_QUATERNION messages, and probably PX4 only provides ATTITUDE_QUATERNION messages and ATTITUDE right ?

patrickelectric:
Just want to say that MAVSDK is one of the best modern C++ libraries around, great software and architecture, thanks! I hope to see it using C++14 and 17 soon :)

julianoes:
Thx! And see #884!

PR #919, opened 2019-12-03
Add missing const around core and integration_tests

patrickelectric:


julianoes:
macOS CI failure is unrelated.

PR #923, opened 2019-12-09
Set listening port of mavsdk_server in order to run multiple instances in parallel

JonasVautherin:
The new default is to leave `mavsdk_server` choose a free port for the gRPC server.

The port can be set with `-p <port>`, for instance:

```
./mavsdk_server -p 50051 udp://:14540
```

The next step would be to send the port over the callback of `runBackend(...)`, so that bindings like Swift or Python can actually directly know it and easily start multiple instances of `mavsdk_server`.

### DIFF
--- a/src/backend/src/backend_api.cpp
+++ b/src/backend/src/backend_api.cpp
@@ -2,10 +2,16 @@
 #include "backend.h"
 #include <string>
 
-void runBackend(const char* connection_url, void (*onServerStarted)(void*), void* context)
+void runBackend(const char* connection_url, const int mavsdk_server_port, void (*onServerStarted)(void*), void* context)
 {
     mavsdk::backend::MavsdkBackend backend;
-    backend.startGRPCServer();
+
+    auto grpc_port = backend.startGRPCServer(mavsdk_server_port);
+    if (grpc_port == 0) {
+        // Server failed to start
+        return;
+    }
+
     backend.connect(std::string(connection_url));
 
     if (onServerStarted != nullptr) {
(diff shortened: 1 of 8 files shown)


src/backend/src/backend_api.cpp
```diff
@@ -2,10 +2,16 @@
...
 
-void runBackend(const char* connection_url, void (*onServerStarted)(void*), void* context)
+void runBackend(const char* connection_url, const int mavsdk_server_port, void (*onServerStarted)(void*), void* context)
 {
     mavsdk::backend::MavsdkBackend backend;
-    backend.startGRPCServer();
+
+    auto grpc_port = backend.startGRPCServer(mavsdk_server_port);
+    if (grpc_port == 0) {
+        // Server failed to start
```
julianoes: Does it need a `LogErr()` here?

julianoes: Oh it does it inside, ok.

src/backend/src/mavsdk_server.cpp
```diff
@@ -2,33 +2,50 @@
...
+                return 1;
+            }
+
+            const std::string port(argv[i + 1]);
+            i++;
+
+            if (!is_integer(port)) {
+                usage();
+                return 1;
+            } else {
```
julianoes: I prefer no `else` but that's just a nit.

src/backend/src/mavsdk_server.cpp
```diff
@@ -38,5 +55,16 @@ void usage()
...
+}
+
+bool is_integer(const std::string& tested_integer) {
+    for (const auto& digit : tested_integer) {
+        if (!std::isdigit(digit)) {
+            return false;
+        }
+    }
+
+    return true;
```
julianoes: :+1:

src/backend/src/mavsdk_server.cpp
JonasVautherin: Done

PR #924, opened 2019-12-09
Adding DISTANCE_SENSOR and possibly other raw sensor messages.

physicsman:
**Moved to Issues**
-------------------
For offboard work I feel it would be useful to have access to the the raw sensor data. For example I am using Attitude+Lidar Rangefinder+Thermal Sensor  to create a heat map during a survey. 

**Suggestions/Ideas:**

A telemetry subclass ie `telemetry.raw()` that gathers sensor information and current readings available through mavlink. Also, IMO the sensor reading timestamp "time_boot_ms" (ex: https://mavlink.io/en/messages/common.html#DISTANCE_SENSOR) is especially useful from my perspective for sensor fusion and filtering to approximate vehicle state.

### DIFF
--- a/src/core/CMakeLists.txt
+++ b/src/core/CMakeLists.txt
@@ -29,6 +29,7 @@ add_library(mavsdk
     cli_arg.cpp
     thread_pool.cpp
     geometry.cpp
+    timesync.cpp
 )
 
 target_link_libraries(mavsdk
--- a/src/core/global_include.cpp
+++ b/src/core/global_include.cpp
@@ -4,10 +4,12 @@
 #include <cstdint>
 #include <limits>
 #include <cmath>
+#include <chrono>
 
 namespace mavsdk {
 
 using std::chrono::steady_clock;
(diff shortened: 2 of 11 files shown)


JonasVautherin:
Thanks for the PR!

However, it seems like it is not up-to-date with `develop`, and therefore in the "Files changed" tab, you can see changes that are not yours.

Ideally, you should make sure that your branch is up-to-date with `develop` (and only have your new commits on top), so that we can really see what changes you propose.

Also, is this meant to be a pull request (you have code you want to contribute), or is it more of a feature request (you want to discuss adding the distance sensor)? In the latter case, it should be an issue (and not a PR). In the former case, please make sure you are up-to-date with `develop` so we clearly see what code you want to merge!

Let me know if you have questions, and thanks for the contribution! :blush:

JonasVautherin:
Closing in favor of https://github.com/mavlink/MAVSDK/issues/925.

Thanks!

Issue #925, opened 2019-12-09
Adding DISTANCE_SENSOR and possibly other raw sensor messages.


physicsman:
For offboard work I feel it would be useful to have access to the the raw sensor data. For example I am using Attitude+Lidar Rangefinder+Thermal Sensor  to create a heat map during a survey. 

**Suggestions/Ideas:**

A telemetry subclass ie `telemetry.raw()` that gathers sensor information and current readings available through mavlink. Also, IMO the sensor reading timestamp "time_boot_ms" (ex: https://mavlink.io/en/messages/common.html#DISTANCE_SENSOR) is especially useful from my perspective for sensor fusion and filtering to approximate vehicle state.

irsdkv:
@JonasVautherin I think I can participate in part of implement receiving and handling `DISTANCE_SENSOR` message.
In which plugin this should implemented on your opinion?
In `Telemetry` (whenever it already large)?

JonasVautherin:
Maybe in telemetry... I'm just wondering what happens when you don't have a distance sensor at all. What if you subscribe to that stream? You just get no event?

@julianoes: do you have an opinion?

irsdkv:
> You just get no event?

It seems so:
https://github.com/irsdkv/Firmware/blob/master/src/modules/mavlink/mavlink_messages.cpp#L4277

JonasVautherin:
Ok! Let's put that in the telemetry plugin then! Thanks @irsdkv!

physicsman:
I don't know the implementation however I think multiple distance sensor support will be coming for object avoidance (i.e. a face-forward and face-downward sensor).

julianoes:
DISTANCE_SENSOR has been implemented in #1222.

PR #927, opened 2019-12-10
Add distance sensor message handling

irsdkv:
Related to #925 

I'm not sure that `DistanceSensor` typename is a good choice.
@JonasVautherin , @julianoes what are you think about naming for this type?


irsdkv:
And also may be need to keep all possible orientations (see [this commit](https://github.com/mavlink/MAVSDK/pull/927/commits/feb411f251a1b7ca246ef9b508631d0e730770d6))?
Or not?

julianoes:
CI should be fine. I manually stopped the iOS builds.

src/plugins/telemetry/include/plugins/telemetry/telemetry.h
```diff
@@ -363,6 +363,64 @@ class Telemetry : public PluginBase {
...
+            ROTATION_PITCH_270 = 25, /**< @brief Roll: 0, Pitch: 270, Yaw: 0 */
+            ROTATION_CUSTOM = 100 /**< @brief Custom orientation */
+        };
+
+        uint32_t time_boot_ms; /**< @brief  Timestamp (time since system boot). */
+        uint16_t min_distance; /**< @brief  Minimum distance the sensor can measure. */
+        uint16_t max_distance; /**< @brief  Maximum distance the sensor can measure. */
+        uint16_t current_distance; /**< @brief Current distance reading. */
+        SensorType type; /**< @brief Type of distance sensor. */
+        uint8_t id; /**< @brief Onboard ID of the sensor. */
```
julianoes: I don't like the fact that we are just 1 to 1 exposing the MAVLink message at this point. In my opinion the point of MAVSDK is to strip the information down to its essentials and present it in simplified way.

For instance here we have an `enum` for the orientation as well as a quaternion. What about translating the enum to quaternion inside the plugin and only exposing the quaternion?

Or what about using covariance as a float in meters with `NAN` for unknown instead of relying on the `uint8_t` type.

Another idea: should we return a vector of distance sensors where each ID is an entry instead of sending them one by one and having the client code to filter it?

irsdkv: > Or what about using covariance as a float in meters with NAN for unknown instead of relying on the uint8_t type.

It is a good idea but I'm little confused because of in this case seems that Covariance is not the same as `Odometry` Covariance:

> Measurement variance. Max standard deviation is 6cm.

Anyway what about smth like  that?
```cpp
float covariance_m_2; /**< @brief Measurement variance in square meters. NNan if unknown. 
```
```cpp
uint8_t covariance_cm_2 = mavlink_msg_distance_sensor_get_covariance(&message);
if (covariance_cm_2 != 255) {
    distance_sensor.covariance = covariance_cm_2 * covariance_cm_2 / 1e4;
} else {
    distance_sensor.covariance = std::numeric_limits<float>::quiet_NaN
}
```

irsdkv: > What about translating the enum to quaternion inside the plugin and only exposing the quaternion?

Is that does not make handling more complicated?
If users prefers to use standard orientations then handling `Quaternion` looks like a little bit overhead.

@physicsman are you have any idea for that? What way seems more useful for your purposes for example?

irsdkv: > should we return a vector of distance sensors where each ID is an entry instead of sending them one by one and having the client code to filter it?

Hm. Interesting idea. But now I don't know how exactly we should handle received messages.
Return vector on each received message? If yes, so user will should check `time_boot_ms` on each vector member to check is it recently received message or not.
If we will drop old received messages so we don't need a vector anymore. 
But in C++17 we can use `std::map<uint8_t /*sensor_id*/, std::option<DistanceSensor>>`. I think that this is interesting solution.
If I understand you correctly.

irsdkv: Hm. @julianoes , what are you think about some around individual subscription for each sensor?

```cpp
    std::map<uint8_t/*senor_id*/, Telemetry::distance_sensor_callback_t> _distance_sensors_subscriptions{};
```

JonasVautherin: > What about translating the enum to quaternion inside the plugin and only exposing the quaternion?

Not sure if it helps the discussion, but I don't understand why MAVLink does what it does instead of just exposing a quaternion :thinking:. Is it legacy, like the enum was there first? Just having the quaternion would seem enough to me.

JonasVautherin: > what are you think about some around individual subscription for each sensor?

So now I'm trying to think about how it would look like in the wrappers. We can easily auto-generate a wrapper for the version with the ID, and I believe it is easy to filter on the client side (be it in asyncio or in reactive extensions).

Now if we have a map of callbacks, that's more complicated. I guess we would need to expose a list of available sensor ids, and a way to subscribe with the id. Like `SubscribeAvailableDistanceSensors` or `FetchAvailableDistanceSensors`, and then `SubscribeDistanceSensor(id)`.

Either way, there is some logic required on the client side, because you need to select the sensor you are interested in, right?

irsdkv: > Just having the quaternion would seem enough to me.

I'm not worked with Distance Sensors yet. So I worried that quaternion will overcomplicated for use.
Ok, I will implement Quaternion only now

irsdkv: > Just having the quaternion would seem enough to me.

I'm not worked with Distance Sensors yet. So I worried that quaternion will overcomplicated for use.
Ok, I will implement Quaternion only now

irsdkv: > Now if we have a map of callbacks, that's more complicated. 

Yes, in gRPC that's true. Ok, let it go :)

src/plugins/telemetry/telemetry_impl.cpp
```diff
@@ -865,6 +882,65 @@ void TelemetryImpl::process_odometry(const mavlink_message_t& message)
...
+        {
+            std::array<float, 4> q{};
+            mavlink_msg_distance_sensor_get_quaternion(&message, q.data());
+            quaternion.w = q[0];
+            quaternion.x = q[1];
+            quaternion.y = q[2];
+            quaternion.z = q[3];
+            break;
+        }
+    }
```
irsdkv: Should I add some other orientations?

src/plugins/telemetry/include/plugins/telemetry/telemetry.h
JonasVautherin: > Ok, let it go :)

So which one do we go for? We keep the id like in MAVLink, or we go for another interface? I would be fine for both solutions, as long as we can auto-generate them. As a user, I don't think one of them is necessarily more complicated to deal with

irsdkv: > So which one do we go for?

I mean that leave ID's as in MAVLink, sorry

src/plugins/telemetry/telemetry_impl.cpp
julianoes: I don't know. Which orientations would PX4 ever send?

physicsman: If I'm reading this correctly there are two separate issues. One: vehicle orientation and as it relates to the quaternion and Two: returning distance sensor measurements?

 I'm looking more into Quaternions this evening. I was under the impression that the representation was a normalize vector representing a vehicle pointing and a rotation about that vector to add orientation. I was also under the impression that as long as I had the quaternion I could calculate vehicle attitude and sensor attitude in a fixed euclidian reference frame with a custom function or some off the shelf python package. 

The most useful thing to me would some sort of tree that contains the vehicle sensors and their values.

irsdkv: > Which orientations would PX4 ever send?

Looks like this depends on real sensor orientation and we can't expect that it will be some constrained list. And in case of this orientation is not `CUSTOM` quaternion does not set.
The only way to implement safe message processing is check all orientations.

irsdkv: > The most useful thing to me would some sort of tree that contains the vehicle sensors and their values.

@julianoes , @JonasVautherin what are you think about it? Can be implemented some around this in future?
This can be useful for me and I can help in implementation if it will start.

JonasVautherin: I'm not sure I understand the "tree" idea there. You have multiple distance sensors on your drone, and for each of them you want to get the value they read. Correct? Why would that be a tree instead of a list?

The current proposition (subscribing to a stream of events that contains something like <sensor_id, sensor_value>) contains all the information I can imagine one may need. The only thing is that on the client side, you guys have to filter by id if you want only the events coming from one of the sensors. But that's trivial to do, right?

physicsman: @JonasVautherin I meant tree as in a tree of raw sensor readings. They could be any sensor type and distance sensor would be a subtype. Subtypes of distance sensor could be acoustic, Lidar, etc..

When you subscribe to a stream is that automatic or can MAVSDK return a list of what we can subscribe to and then decide? Say if we wanted to conserve bandwidth and only subscribe to streams we need.

JonasVautherin: > When you subscribe to a stream is that automatic or can MAVSDK return a list of what we can subscribe to and then decide?

I would return the question to MAVLink: because MAVSDK uses MAVLink, it depends on how that works there. I am not sure if you can selectively register to the distance sensor you want using MAVLink. Can you?

hamishwillee: > I would return the question to MAVLink: because MAVSDK uses MAVLink, it depends on how that works there. I am not sure if you can selectively register to the distance sensor you want using MAVLink. Can you?

From a MAVLink perspective, in theory you can call [MAV_CMD_SET_MESSAGE_INTERVAL](https://mavlink.io/en/messages/common.html#MAV_CMD_SET_MESSAGE_INTERVAL) to control the rate of message emission. I'm not sure how widely this is supported in flight stacks, but on PX4 it is common to set the [mode](https://docs.px4.io/master/en/peripherals/mavlink_peripherals.html) to minimal and then specify the set of messages you want. 

FWIW I would think it reasonable for MAVSDK API to provide convenient filters on types/sensor ids you could register against. But I don't think it should be used for controlling sending rates as suggested here. That's more of an integration question.

JonasVautherin: > FWIW I would think it reasonable for MAVSDK API to provide convenient filters on types/sensor ids you could register against.

I am not against the idea, but my concern is auto-generation of that kind of code. Because C++ is (still) not auto-generated, C++ implementers have more freedom now, but they will (hopefully soon) be constrained by the auto-generation. So when possible, I'm trying to make sure that it will be possible to auto-generate the APIs we add now. Once everything is auto-generated, it will be time to think about improving them.

Does that make sense?

physicsman:
If I'm reading this right the only thing blocking a merge is a failed iOS build? Im going to try and build it on a Raspberry Pi Zero.

JonasVautherin:
I think there was still the discussion about the covariance, as mentioned by @julianoes:

> Or what about using covariance as a float in meters with NAN for unknown instead of relying on the uint8_t type?

src/plugins/telemetry/include/plugins/telemetry/telemetry.h
```diff
@@ -363,6 +363,44 @@ class Telemetry : public PluginBase {
...
+        enum class SensorType {
+            LASER = 0, /**< @brief Laser rangefinder, e.g. LightWare SF02/F or PulsedLight units. */
+            ULTRASOUND = 1, /**< @brief Ultrasound rangefinder, e.g. MaxBotix units. */
+            INFRARED = 2, /**< @brief Infrared rangefinder, e.g. Sharp units. */
+            RADAR = 3, /**< @brief Radar type, e.g. uLanding units. */
+            UNKNOWN = 4, /**< @brief Broken or unknown type, e.g. analog units. */
+        };
+
+        uint32_t time_boot_ms; /**< @brief  Timestamp (time since system boot). */
+        uint16_t min_distance; /**< @brief  Minimum distance the sensor can measure. */
```
JonasVautherin: Should we add the unit there? Is it `min_distance_m`?

julianoes:
Let me know when I should review this again.

Katawann:
Hi @irsdkv !

any updates on that feature ? I'm looking for to see it merged :)

physicsman:
For my use case have access to altitude_terrain would be sufficient.  I can get amsl and relative to takeoff from telemetry.positon

hrnbot (approved):

julianoes:
Replaced by #1222.

Issue #912, opened 2019-11-22
FlightMode options for MAVSDK limited to Auto mode and Offboard


wsolichin-bto:
So currently, with MAVSDK, if the PX4 is either in Position / Altitude / Stabilized / Manual Mode, it will be displayed as Unknown in MAVSDK using Telemetry Flight Mode from API Reference here: https://mavsdk.mavlink.io/develop/en/api_reference/classmavsdk_1_1_telemetry.html#classmavsdk_1_1_telemetry_1aae86fcb399b0368209fcbd35ecdd5fad. it will be very useful for monitoring the state of the drone if it can for example go to Altitude Mode or Position Mode to test if there's either height source sensors or position source sensors. Is there any other way to get the rest of the flight modes mentioned?

julianoes:
I agree that it can be useful to be able to know the flight mode in general, however, I don't think it should be used to check the state of the position estimation (e.g. if position control is possible or only altitude control. This information should be provided by status flags. That status information is a known problem and something that @MaEtUgR is working on.

bozkurthan:
@wsolichin-bto can you check now?
Other modes can be received by using telemetry plugin.

julianoes:
The modes have been added in #945.

PR #942, opened 2019-12-19
Additional Flight Modes

bozkurthan:
Hello,
Based on this #912 , there is a gap for PX4 modes to get flight modes from telemetry plugin. I added on PR but there is also problem about this PR. I'm using very old version (about 5 months ago), so there are some changes for implementation in current version. 

For example;

https://github.com/bozkurthan/MAVSDK/blob/782022d72be38ed30e77308e0336de511245733a/src/plugins/telemetry/telemetry_impl.cpp#L885-L908

I didn't change this function. I can also add these lines and test this on SiTL. But currently it's working for me.

Screenshot from 2019-12-19 14-41-05

### DIFF
--- a/src/plugins/telemetry/include/plugins/telemetry/telemetry.h
+++ b/src/plugins/telemetry/include/plugins/telemetry/telemetry.h
@@ -223,6 +223,7 @@ public:
      * https://docs.px4.io/en/config/flight_mode.html.
      */
     enum class FlightMode {
+        UNKNOWN, /**< @brief Mode not known. */
         READY, /**< @brief Armed and ready to take off. */
         TAKEOFF, /**< @brief Taking off. */
         HOLD, /**< @brief Hold mode (hovering in place (or circling for fixed-wing vehicles). */
@@ -231,13 +232,12 @@ public:
         LAND, /**< @brief Landing. */
         OFFBOARD, /**< @brief Offboard mode. */
         FOLLOW_ME, /**< @brief FollowMe mode. */
-        UNKNOWN, /**< @brief Mode not known. */
         MANUAL, /**< @brief Manual mode. */
-        POSCTL, /**< @brief Position mode. */
         ALTCTL, /**< @brief Altitude mode. */
-        RATTITUDE, /**< @brief Mode not known. */
+        POSCTL, /**< @brief Position mode. */
         ACRO, /**< @brief Acro mode. */
-        STABILIZED /**< @brief Stabilize mode. */
+        STABILIZED, /**< @brief Stabilize mode. */
+        RATTITUDE /**< @brief Rattitude mode. */
     };
 
     /**
--- a/src/plugins/telemetry/include/plugins/telemetry/telemetry.h
+++ b/src/plugins/telemetry/include/plugins/telemetry/telemetry.h
@@ -231,7 +231,13 @@ public:
         LAND, /**< @brief Landing. */
         OFFBOARD, /**< @brief Offboard mode. */
         FOLLOW_ME, /**< @brief FollowMe mode. */
-        UNKNOWN /**< @brief Mode not known. */
+        UNKNOWN, /**< @brief Mode not known. */
+        MANUAL, /**< @brief Manual mode. */
+        POSCTL, /**< @brief Position mode. */
+        ALTCTL, /**< @brief Altitude mode. */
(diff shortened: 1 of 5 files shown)


julianoes:
Thanks for that.

It doesn't seem to build yet. Can you check that?
https://travis-ci.org/mavlink/MAVSDK/jobs/627186819#L1942

bozkurthan:
Sure my fault. It has to be without `telemetry` class . Actually I will double check tomorrow with new mavlink clone and test them. I just added codes from website. So this miserable faults come out

src/plugins/telemetry/include/plugins/telemetry/telemetry.h
```diff
@@ -231,7 +231,13 @@ class Telemetry : public PluginBase {
         LAND, /**< @brief Landing. */
         OFFBOARD, /**< @brief Offboard mode. */
         FOLLOW_ME, /**< @brief FollowMe mode. */
-        UNKNOWN /**< @brief Mode not known. */
+        UNKNOWN, /**< @brief Mode not known. */
```
JonasVautherin: Wait... `UNKNOWN` must be the first one, as per [the proto files](https://github.com/mavlink/MAVSDK-Proto/blob/master/protos/telemetry/telemetry.proto#L238-L246) :thinking:. They must be in the same order in the proto file and in this file. Would you mind moving `UNKNOWN` at the top here, and make sure the others are in the right order?

bozkurthan: Sure but previous version wasn't like that. UNKNOWN is on the top, Copy that :). 

For others, should I sort like here?

https://github.com/mavlink/MAVSDK/blob/6665fee1ef803188bc7c7552174dfca43460919e/src/core/px4_custom_mode.h#L50-L70

Should they like this?
```
        UNKNOWN, /**< @brief Mode not known. */
        READY, /**< @brief Armed and ready to take off. */
        TAKEOFF, /**< @brief Taking off. */
        HOLD, /**< @brief Hold mode (hovering in place (or circling for fixed-wing vehicles). */
        MISSION, /**< @brief Mission mode. */
        RETURN_TO_LAUNCH, /**< @brief Returning to launch position (then landing). */
        LAND, /**< @brief Landing. */
        OFFBOARD, /**< @brief Offboard mode. */
        FOLLOW_ME, /**< @brief FollowMe mode. */
        MANUAL, /**< @brief Manual mode. */
        ALTCTL, /**< @brief Altitude mode. */
        POSCTL, /**< @brief Position mode. */
        ACRO, /**< @brief Acro mode. */
        STABILIZED, /**< @brief Stabilize mode. */
        RATTITUDE /**< @brief Rattitude mode. */
```

JonasVautherin: They should be in the same order as the proto file (linked above). The reason for that is that we generate code from those proto files. In the future, the C++ headers will be auto-generated, too. But in the meantime we have to take care of that manually =)

bozkurthan:
Yes I checked and I got it, but there is no declaration of new modes that I added (Position control, altitude control, manual, acro, rattitude, stabilized). So which order is correct for these? Also so then we have to add these modes on proto files right?

JonasVautherin:
> So which order is correct for these?

You can choose, since you are the one adding them :wink:.

> Also so then we have to add these modes on proto files right?

Right. There are currently 3 places where new features need to be added:

* C++ "core" (what you did in this PR)
* Corresponding proto file(s) (here [telemetry.proto](https://github.com/mavlink/MAVSDK-Proto/blob/master/protos/telemetry/telemetry.proto#L237-L247))
* Corresponding mavsdk_server implementation (that would be [here](https://github.com/mavlink/MAVSDK/blob/develop/src/backend/src/plugins/telemetry/telemetry_service_impl.h#L303-L327))

In the future, the mavsdk_server implementation will be auto-generated, as mentioned earlier.

If you have time for it, we would be glad if you could update the proto file and mavsdk_server implementation in new PRs (one for each) :blush:.

src/plugins/telemetry/include/plugins/telemetry/telemetry.h
JonasVautherin: > Should they like this?

That's correct! Please update that here, and then it's good to be merged :+1: :rocket:

PR #943, opened 2019-12-19
Add mavshell example

JonasVautherin:
Add example for the shell plugin.

Also, somehow the examples/ CMakeLists was not a root project, so I fixed that.

### DIFF
--- a/example/mavshell/mavshell.cpp
+++ b/example/mavshell/mavshell.cpp
@@ -1,0 +1,107 @@
+#include <future>
+#include <mavsdk/mavsdk.h>
+#include <mavsdk/plugins/shell/shell.h>
+#include <iostream>
+
+using namespace mavsdk;
+
+#define ERROR_CONSOLE_TEXT "\033[31m" // Turn text on console red
+#define TELEMETRY_CONSOLE_TEXT "\033[34m" // Turn text on console blue
+#define NORMAL_CONSOLE_TEXT "\033[0m" // Restore normal console colour
+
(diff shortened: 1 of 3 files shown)


example/mavshell/mavshell.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+int main(int argc, char** argv)
+{
+    if (!are_arguments_valid(argc, argv)) {
+        const auto binary_name = argv[0];
+        print_usage(binary_name);
+        return 1;
+    }
+
+    Mavsdk dc;
```
julianoes: Haha, just keep copying `dc`. It's not `DroneCore` anymore.

```suggestion
    Mavsdk mavsdk;
```

example/mavshell/mavshell.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+    while (true) {
+        std::string command;
+        getline(std::cin, command);
+
+        if (command == "exit") {
+            break;
+        }
+
+        Shell::ShellMessage msg{true, timeout_ms, command + "\n"};
```
julianoes: I would probably add the newline inside the plugin, no?

irsdkv: @JonasVautherin Recently I was create [PR](https://github.com/mavlink/MAVSDK/pull/944) with that, so no more need to append `\n` :)

JonasVautherin: Agreed, I'll change that once the change is merged :slightly_smiling_face:

Issue #940, opened 2019-12-18
Trouble using shell plugin


JonasVautherin:
I was trying to play with the shell plugin. So I connected a pixhawk to my computer (available on `serial:///dev/ttyACM0`), and ran the following code, which is essentially a modified `takeoff_and_land.cpp` example ([gist here](https://gist.github.com/JonasVautherin/8172654cb0b1de3b9b9046032b4ddf8c)):

```cpp
#include <chrono>
#include <cstdint>
#include <mavsdk/mavsdk.h>
#include <mavsdk/plugins/shell/shell.h>
#include <iostream>
#include <thread>

using namespace mavsdk;
using namespace std::this_thread;
using namespace std::chrono;

#define ERROR_CONSOLE_TEXT "\033[31m" // Turn text on console red
#define TELEMETRY_CONSOLE_TEXT "\033[34m" // Turn text on console blue
#define NORMAL_CONSOLE_TEXT "\033[0m" // Restore normal console colour

void usage(std::string bin_name)
{
    std::cout << NORMAL_CONSOLE_TEXT << "Usage : " << bin_name << " <connection_url>" << std::endl
              << "Connection URL format should be :" << std::endl
              << " For TCP : tcp://[server_host][:server_port]" << std::endl
              << " For UDP : udp://[bind_host][:bind_port]" << std::endl
              << " For Serial : serial:///path/to/serial/dev[:baudrate]" << std::endl
              << "For example, to connect to the simulator use URL: udp://:14540" << std::endl;
}

int main(int argc, char** argv)
{
    Mavsdk dc;
    std::string connection_url;
    ConnectionResult connection_result;

    bool discovered_system = false;
    if (argc == 2) {
        connection_url = argv[1];
        connection_result = dc.add_any_connection(connection_url);
    } else {
        usage(argv[0]);
        return 1;
    }

    if (connection_result != ConnectionResult::SUCCESS) {
        std::cout << ERROR_CONSOLE_TEXT
                  << "Connection failed: " << connection_result_str(connection_result)
                  << NORMAL_CONSOLE_TEXT << std::endl;
        return 1;
    }

    // We don't need to specify the UUID if it's only one system anyway.
    // If there were multiple, we could specify it with:
    // dc.system(uint64_t uuid);
    System& system = dc.system();

    std::cout << "Waiting to discover system..." << std::endl;
    dc.register_on_discover([&discovered_system](uint64_t uuid) {
[...]

PR #944, opened 2019-12-19
shell: add automatic line feed

irsdkv:
Add line feed due to https://github.com/mavlink/MAVSDK/issues/940

### DIFF
--- a/src/plugins/shell/include/plugins/shell/shell.h
+++ b/src/plugins/shell/include/plugins/shell/shell.h
@@ -68,6 +68,8 @@ public:
     /**
      * @brief Send the shell message.
      *
+     * If shell_message.data string have not trailing '\n' symbol - it will be added.
+     *
      * @param shell_message Shell `struct`.
      */
     Shell::Result shell_command(ShellMessage shell_message);
--- a/src/plugins/shell/include/plugins/shell/shell.h
+++ b/src/plugins/shell/include/plugins/shell/shell.h
@@ -70,6 +70,8 @@ public:
      *
      * If shell_message.data string have not trailing '\n' symbol - it will be added.
      *
+     * If response data looks like not completed try to increase timeout value in request.
+     *
      * @param shell_message Shell `struct`.
      */
     Shell::Result shell_command(ShellMessage shell_message);
--- a/src/plugins/shell/shell_impl.cpp
+++ b/src/plugins/shell/shell_impl.cpp
@@ -86,6 +86,10 @@ Shell::Result ShellImpl::shell_command(const Shell::ShellMessage& shell_message)
 
     _shell_message = shell_message;
 
+    if (_shell_message.data.back() != '\n') {
+        _shell_message.data.append(1, '\n');
+    }
+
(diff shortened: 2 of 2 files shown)


src/plugins/shell/include/plugins/shell/shell.h
```diff
@@ -54,6 +54,8 @@ class Shell : public PluginBase {
         NO_SYSTEM, /**< @brief No system connected. */
         CONNECTION_ERROR, /**< @brief %Connection error. */
         NO_RESPONSE, /**< @brief Response does not received */
+        NOT_COMPLETED, /**< @brief Response transfer potentially does not completed
+                                  (termination symbol does not received) */
```
julianoes:
```suggestion
                                  (termination symbol not received) */
```

src/plugins/shell/include/plugins/shell/shell.h
```diff
@@ -54,6 +54,8 @@ class Shell : public PluginBase {
         NO_SYSTEM, /**< @brief No system connected. */
         CONNECTION_ERROR, /**< @brief %Connection error. */
         NO_RESPONSE, /**< @brief Response does not received */
+        NOT_COMPLETED, /**< @brief Response transfer potentially does not completed
```
julianoes:
```suggestion
        NOT_COMPLETED, /**< @brief Response transfer potentially not completed
```

src/plugins/shell/include/plugins/shell/shell.h
```diff
@@ -54,6 +54,8 @@ class Shell : public PluginBase {
         NO_SYSTEM, /**< @brief No system connected. */
         CONNECTION_ERROR, /**< @brief %Connection error. */
         NO_RESPONSE, /**< @brief Response does not received */
+        NOT_COMPLETED, /**< @brief Response transfer potentially not completed
```
JonasVautherin: Isn't it equivalent to `BUSY`, like we had in other plugins? E.g. if you try to `upload_mission()` while it's already uploading/downloading, it will say `BUSY`. It may be more consistent.

Do you guys have an opinion on that?

src/plugins/shell/include/plugins/shell/shell.h
```diff
@@ -68,6 +70,8 @@ class Shell : public PluginBase {
     /**
      * @brief Send the shell message.
      *
+     * If shell_message.data string have not trailing '\n' symbol - it will be added.
```
JonasVautherin:
```suggestion
     * If shell_message.data has no trailing '\n' symbol, it will be added.
```

src/plugins/shell/include/plugins/shell/shell.h
irsdkv: from [this](https://github.com/mavlink/MAVSDK-Proto/pull/112#discussion_r359965790):

Yes, this really can confuse.

BUSY - shell is busy now
NOT_COMPLETED - in response of current request was not received ESC, so we suppose that transfer was interrupted or last slices of response lost.
May be need to use more clear name (or may be refuse this enum value)?

PR #951, opened 2020-01-03
Update dependencies

JonasVautherin:
Let's hope the newer versions build correctly :see_no_evil:.


JonasVautherin:
Seems like it works! :tada:

PR #961, opened 2020-01-23
mission: retry to clear the mission

julianoes:
The timeout was not implemented for clearing the mission. This meant that if the autopilot did not ack mission clear we would be stuck in the MISSION_CLEAR activity stay and be busy for the rest.

Reported by @douglaswsilva.


julianoes:
I discovered a crash because the wrong callback was called on clearing timeout, see: dbbadd3f.

PR #962, opened 2020-01-24
offboard: fix crash on plugin destruction

julianoes:
If we don't stop the call_every calls before destruction, we can run into segfaults where the method is destructed but still called.


julianoes:
Ignored git clone error on travis-ci macOS.

PR #963, opened 2020-01-24
mission: fix timeout notification

julianoes:
This was broken in the previous fix. The activity state would always be NONE if it is changed before it is used/checked.

Thanks @douglaswsilva!


julianoes:
Ignored git clone error on travis-ci macOS.

PR #964, opened 2020-01-27
Refactor mission transfer

julianoes:
This moves mission/geofence/rally point transfer into core which means the work that has to be done in the plugins can be simplified

### DIFF
--- a/src/plugins/geofence/geofence_impl.cpp
+++ b/src/plugins/geofence/geofence_impl.cpp
@@ -5,9 +5,7 @@
 
 namespace mavsdk {
 
-GeofenceImpl::GeofenceImpl(System& system) :
-    PluginImplBase(system),
-    _mavlink_geofence_item_messages()
+GeofenceImpl::GeofenceImpl(System& system) : PluginImplBase(system)
 {
     _parent->register_plugin(this);
 }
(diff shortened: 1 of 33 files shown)


julianoes:
Try again tomorrow, now libressl download is down:
```
 -- Using src='https://ftp.openbsd.org/pub/OpenBSD/LibreSSL/libressl-3.0.2.tar.gz'
CMake Error at libressl-stamp/download-libressl.cmake:159 (message):
```

julianoes:
@JonasVautherin ready for review :smile:

src/plugins/geofence/geofence_impl.cpp
```diff
@@ -183,64 +55,70 @@ void GeofenceImpl::assemble_mavlink_messages(
...
+Geofence::Result GeofenceImpl::convert_result(MAVLinkMissionTransfer::Result result)
 {
-    if (callback == nullptr) {
-        LogDebug() << "Callback is not set";
-        return;
+    switch (result) {
+        case MAVLinkMissionTransfer::Result::Success:
+            return Geofence::Result::SUCCESS;
+        case MAVLinkMissionTransfer::Result::ConnectionError:
+            return Geofence::Result::ERROR; // FIXME
```
JonasVautherin: How should those be fixed? By adding Geofence results?

julianoes: I planned to consolidate the different results in a future PR. I think the result type will be duplicated / or similar between the plugins but we will add some more error types that make sense. My goal was basically to not have any change in API for this PR but only swap out internals.

julianoes:
Thanks for merging. I didn't have the courage :smile:.

julianoes:
FYI @hamishwillee we're on the way with the plan we decided on a while ago :tada:

itrigitech:
Oh perfect. Thanks for that work! In March then I will be able to work in the needed plugins

hamishwillee:
Well done!

PR #966, opened 2020-01-27
Add takeoff, land vtol mission items to mission plugin

itrigitech:
I've moddified the mission plugin to accept takeoff, land, and takeoff_and_transition, transition_and_land mission items when uploading/downloading a mission. I've found this necessary as f vtol without this items the drone is not transitioning so needed for sure for vtols.

I've also created a new test only adding this two items on the mission.

For my tests everything works ok, if you ee something weird tell me.


julianoes:
@itrigitech thanks a lot for the contribution, appreciated. Unfortunately, I have some concerns with adding the command in this way.

I've designed the mission API of the mission plugin to be simple but as easy to use as possible, trying to prevent pitfalls that the mission items could create and making sure the exposed API is supported by PX4.
Now, this design has also drawbacks, the biggest one usually the lack of a specific feature. In your case you were missing VTOL transitions.

With your pull request we are essentially creating a backdoor enabling basically everything and my fear is that users using this API will run into problems and/or be confused.
As an example they might be confused about what command to use. Or, they might choose a command which is not compatible with the rest of the properties set.

Now, I'm not saying VTOL transitions are too complicated and that we should not support them, however, I think, we need to find a cleaner and easier-to-use way.

My suggestion is to create a `mission_vtol` plugin specifically for capabilities around VTOLs. I'm already working in that direction by extracting the mission transfer logic from the mission plugin to the core (see https://github.com/mavlink/MAVSDK/pull/964) and that way re-use it in `mission`, `mission_raw` (for raw MAVLink items), `geofence`, as well as `rally points`.

I think what would be best is to define the API/functionality required for VTOLs, so that we can then implement it together for the `mission_vtol` plugin. I would really appreciate your help there.

Sorry about that but I hope you can understand my reasoning.

itrigitech:
Yes. I imagined something like that but I didn't know if you where in active development of this features. For now I will use my branch until we have a final solution.

Anyway I think that we should have the ability to raw upload missions. Apart of VTOL transitions there are other commands like having loops,for example, not implemented. 

Also mention that in my personal use case I was confused in the begining because I was uploading missions generated with qgroundcontrol. If you use this, it changes the VTOL items for waypoints. The drone is not transitioning and it can crash easily. I think this is dangerous too. Or at least to leave it clear that the mission plugin it's only for non VTOL drones. Maybe changing the name or something.

julianoes:
> .. I didn't know if you where in active development of this features. 

Started on Monday because it had come up enough times :smile:.

> For now I will use my branch until we have a final solution.

Makes sense.

> Anyway I think that we should have the ability to raw upload missions. 

This will also come once the transfer is refactored.

> Or at least to leave it clear that the mission plugin it's only for non VTOL drones. Maybe changing the name or something.

Good point!

julianoes:
@itrigitech FYI #964 is merged.

julianoes:
@itrigitech I don't think we want to merge this as such, so I'm going to close this PR. The next step in my opinion would be to come up with a mission API specifically for VTOL. Also note the discussion here:
https://github.com/mavlink/MAVSDK-Proto/issues/168

PR #969, opened 2020-01-28
Expose Ground Truth SITL position

kamilritz:


src/plugins/telemetry/include/plugins/telemetry/telemetry.h
```diff
@@ -122,6 +122,17 @@ class Telemetry : public PluginBase {
         float yaw_rad_s; /**< @brief Yaw angular velocity */
     };
 
+    /**
+     * @brief Ground truth type.
+     *
+     * Ground truth position information available in simulation
```
julianoes:
```suggestion
     * Ground truth position information available in simulation.
```

src/integration_tests/telemetry_async.cpp
```diff
@@ -244,6 +252,14 @@ void print_fixedwing_metrics(Telemetry::FixedwingMetrics fixedwing_metrics)
     _received_fixedwing_metrics = true;
 }
 
+void print_ground_truth(Telemetry::GroundTruth ground_truth)
+{
+    std::cout << "Latitude: " << ground_truth.latitude_deg << " deg, "
+              << "Longitude: " << ground_truth.longitude_deg << " deg, "
+              << "Altitude: " << ground_truth.absolute_altitude_m << " m" << std::endl;
```
julianoes: You could actually just do `std::cout << ground_truth << std::endl;` since you implemented it.

src/integration_tests/telemetry_sync.cpp
```diff
@@ -26,7 +26,7 @@ TEST_F(SitlTest, TelemetrySync)
     EXPECT_EQ(telemetry->set_rate_gps_info(10.0), Telemetry::Result::SUCCESS);
     EXPECT_EQ(telemetry->set_rate_battery(10.0), Telemetry::Result::SUCCESS);
     EXPECT_EQ(telemetry->set_rate_actuator_control_target(10.0), Telemetry::Result::SUCCESS);
-    EXPECT_EQ(telemetry->set_rate_fixedwing_metrics(10.0), Telemetry::Result::SUCCESS);
+    EXPECT_EQ(telemetry->set_rate_ground_truth(10.0), Telemetry::Result::SUCCESS);
```
julianoes: Have you checked if this works if you don't enable the message in the startup script of PX4?

kamilritz: Yes.

src/integration_tests/telemetry_sync.cpp
```diff
@@ -26,7 +26,7 @@ TEST_F(SitlTest, TelemetrySync)
     EXPECT_EQ(telemetry->set_rate_gps_info(10.0), Telemetry::Result::SUCCESS);
     EXPECT_EQ(telemetry->set_rate_battery(10.0), Telemetry::Result::SUCCESS);
     EXPECT_EQ(telemetry->set_rate_actuator_control_target(10.0), Telemetry::Result::SUCCESS);
-    EXPECT_EQ(telemetry->set_rate_fixedwing_metrics(10.0), Telemetry::Result::SUCCESS);
```
JonasVautherin: Why was this removed? :thinking:

kamilritz: 🤔 there is no good reason, sry for that. Will add it back.

julianoes: Oops, missed it.

PR #971, opened 2020-01-28
workflows: build PRs from forks too

julianoes:
If we don't build on pull_request, then pull requests from forks are not built. By only building pushes to master and develop though we get rid of the duplicate builds otherwise.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,6 +1,14 @@
 name: C/C++ CI
 
-on: [push]
+on:
+  push:
+    branches:
+    - 'master'
+    - 'develop'
+  pull_request:
+    branches:
+    - '*'
+
 
 jobs:
   coverage:


julianoes:
I don't need to wait for CI, I didn't change anything other than CI itself :smile:.

JonasVautherin:
I thought they were being built on the forks themselves, and therefore only for PRs from forks which enabled github-actions (because it's still in beta). So that's a nice security measure, actually :slightly_smiling_face:. Isn't it how it works?

julianoes:
I don't think that works in practice because the checks don't show up here. And that's ultimately what counts. If CI can't check what is contributed from a fork it's rather useless in my opinion. I'm pretty sure all secrets are kept secret by GitHub. Otherwise, it's probably not just us that would be in trouble.

JonasVautherin:
What I meant was [this](https://github.community/t5/GitHub-Actions/Run-a-GitHub-action-on-pull-request-for-PR-opened-from-a-forked/td-p/15426):

> This is a mitigation against the possibility that a bad actor could open PRs against your repo and do things like list out secrets or just run up a large bill (once we start charging) on your account. The actions are in fact executed, but it happens against the _fork_, not against the base repo.

 > This does require that the forked repo also has actions enabled though. During the beta period that means that the owner of the forked repo must also be in the beta.

So I accepted the fact that github-actions would not be run on all PRs during the beta period. Is that wrong?

julianoes:
I see. Looking at this post I think the credentials are safe:
https://github.community/t5/GitHub-Actions/Run-a-GitHub-action-on-pull-request-for-PR-opened-from-a-forked/m-p/31835/highlight/true#M914

PR #976, opened 2020-01-29
Support for non-superbuild mavsdk_server build

julianoes:
This are some bits and pieces that could allow us to build mavsdk_server without superbuild.

### DIFF
--- a/src/backend/test/CMakeLists.txt
+++ b/src/backend/test/CMakeLists.txt
@@ -37,7 +37,7 @@ target_link_libraries(unit_tests_backend
     mavsdk_offboard
     mavsdk_telemetry
     gRPC::grpc++
-    protobuf
+    ${PROTOBUF_LIBRARIES}
     gtest
     gmock
     gmock_main
--- a/src/backend/test/CMakeLists.txt
+++ b/src/backend/test/CMakeLists.txt
@@ -37,6 +37,7 @@ target_link_libraries(unit_tests_backend
     mavsdk_offboard
     mavsdk_telemetry
     gRPC::grpc++
+    protobuf
     gtest
     gmock
     gmock_main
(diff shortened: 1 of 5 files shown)


tools/generate_from_protos.sh
```diff
@@ -4,11 +4,19 @@ set -e
 
 script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 
+
+if [ "$1" == "--no-superbuild" ]; then
+    protoc_binary=$(command -v protoc)
```
JonasVautherin: What about something like:

```
command -v protoc || { echo >&2 "Protobuf needs to be installed for this script to run!"; exit 1; }
command -v grpc_cpp_plugin || { echo >&2 "grpc_cpp_plugins needs to be installed for this script to run!"; exit 1; }
```

src/backend/test/CMakeLists.txt
```diff
@@ -37,6 +37,7 @@ target_link_libraries(unit_tests_backend
     mavsdk_offboard
     mavsdk_telemetry
     gRPC::grpc++
+    protobuf
```
JonasVautherin: I'm not sure if that's the right way to do it (mainly because the CI fails). Normally it comes in the `gRPC::grpc++` target, I would say. I wonder if `FindgRPC.cmake` should not take care of this (but I'm not sure how)

tools/generate_from_protos.sh
julianoes: Much better, thanks.

src/backend/test/CMakeLists.txt
julianoes: The original FindGRPC had something for protobuf but I did not take the time to understand what it did:
https://raw.githubusercontent.com/wastl/cmarmotta/master/cmake/FindgRPC.cmake

julianoes:
Trying again almost a year late.

julianoes:
No one needs this anyway, closing.

PR #983, opened 2020-02-11
Use OpenSSL instead of LibreSSL (for dTLS support)

JonasVautherin:
Checking what the CI thinks of OpenSSL


JonasVautherin:
Believe it or not, that's exactly what I was expecting the CI to think. So predictable :trollface:.

julianoes:
@JonasVautherin should we close this or do you plan on completing it?

PR #985, opened 2020-02-12
Move from travis-ci to GitHub workflows

julianoes:
- Move publication of artefacts over from travis-ci to GitHub workflows.
- Add publication of iOS framework.
- Move coveralls report to GitHub workflows.
- Disable travis-ci.


.github/workflows/main.yml
```diff
@@ -96,6 +98,42 @@ jobs:
         run: cmake --build build/release --target install -- -j2
       - name: package
         run: tools/create_packages.sh ./install .
+      - name: Publish artefacts
+        if: contains(github.ref, 'v')
+        uses: svenstaro/upload-release-action@v1-release
+        with:
+          repo_token: ${{ secrets.GITHUB_TOKEN }}
+          file_glob: true
+          file: '*.deb'
```
JonasVautherin: So according to [this](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet), it seems like we can't just `*.[deb|rpm]`, indeed :confused:. Too bad, it would prevent quite some duplication :sweat_smile:

julianoes: Right, I'll see if we can do that. We can also change one of those actions to fit what we need.

julianoes:
> Coverage increased (+8.5%) to 29.201% 

That's a bit odd.

julianoes:
But will accept that increase in coverage :smile:.

PR #986, opened 2020-02-13
workflows: only publish if actual version tag

julianoes:
Before this we would also try to publish the `develop` branch because the ref contains a `v`.


julianoes:
I'm not going to wait on iOS and merge this.

PR #989, opened 2020-02-14
Add lossy mission transfer test to CI

julianoes:
We need to include the passthrough plugin for the lossy mission transfer test.


julianoes:
Just iOS left, merging.

PR #1000, opened 2020-03-04
Emit the gRPC port on which mavsdk_server listens in onServerStarted

JonasVautherin:
For Android/iOS, we want `mavsdk_server` to start on a random gRPC port, but then we need to advertise it to the Java/Swift client.


JonasVautherin:
Hmm actually I think we want to do this differently :sweat_smile:.

PR #1001, opened 2020-03-05
Implement source system and component ID handling via Configuration class

### LINKED ISSUE #999: Organize & Cleanup Receiver System & Component ID Handling
Hello, I was looking into the source code and I found quite a few FIXME's in `mavsdk_impl.cpp` regarding how different system and component IDs are handled. I created this issue so we can try to fix these issues, and to offer my thoughts on how they should be implemented as well as hear other people's ideas.

1. [L256](https://github.com/mavlink/MAVSDK/blob/master/src/core/mavsdk_impl.cpp#L256): Agreed, there are a couple things wrong here. First of all, we should not be setting the SID to a known CID, that's bad usage of the definitions. However, I don't think we should simply hard code in another value. See point 3 for my thoughts on a better approach.
2. [L258](https://github.com/mavlink/MAVSDK/blob/master/src/core/mavsdk_impl.cpp#L258): Again, this time at least we are setting the CID to a value designated for a Component ID (but incrementing it by 1? :thinking: ). This is still a fairly arbitrary value, and furthermore is hardcoded.
3. Personally, I think the `Configuration` structure is rather limiting. It forces the user to use a specific SID and CID (which is automatically decided for them) which further has the implication of only being able to successfully run 1 instance of MAVSDK at a time (well, maybe 1 per `Configuration` but you get the idea). Furthermore, the hardcoded values are set to some weird values! :thinking: 

My suggestion is: as far as API goes, start with a `void set_mavlink_id(uint8_t sid, uint8_t cid)` or similar which allows for manual user setting of the SID and CID. The current `Configuration` idea can still remain as a feature but should be refactored to something like `set_auto_mavlink_id(MavType type)` which makes more sense to me - it allows the user to have MAVSDK automatically choose an *appropriate* SID and CID based on the type of system the user is implementing with the SDK. Also, I personally feel that setting either the SID and CID or the `MavType` should be mandatory at the constructor level, but I'm not sure how others feel about this?

Now for what those auto values should be:
* For `MavType GCS`: Quoted from MAVSDK source: "Usually QGC identifies itself with sysid 255" - we should just pick a standard ID that we want to be the default MAVSDK SID, and stick to it. According to the MAVLink documentation, it is good practice to count GCS SIDs down from 255. So I'm throwing the value `245` out there, as its reasonably close to the top but leaves some room for multiple QGC/other GCS instances to be run. In reality any SID would work, and we should probably do at least some basic searching before deciding on one, to minimize conflicts (at least with existing open source GCS). The CID should be set to `MAV_COMP_ID_MISSIONPLANNER` as this is most likely what the user needs (stated in docs as "Component that can generate/supply a mission flight plan (e.g. GCS or developer API).", which is most likely the usage of MAVSDK in the context of a GCS.)
* For `MavType CompanionComputer`: This is a bit more difficult, As the FIXME says it should be autodetected. The best way to do this would probably be to read incoming `MAV_HEARTBEAT` messages until one with the CID `MAV_COMP_ID_AUTOPILOT1` is discovered, and set our SID to the source SID of that message. As for CID, `MAV_COMP_ID_PATHPLANNER` would probably be the most idiomatic, considering the companion computer is usually involved in path planning, obstacle avoidance, etc.
* For `MavType Autopilot`: As far as default values go, the existing is fine - 1 as SID and and `MAV_COMP_ID_AUTOPILOT` as CID.

4. [L263](https://github.com/mavlink/MAVSDK/blob/master/src/core/mavsdk_impl.cpp#L263): See point 3.
5. [L78](https://github.com/mavlink/MAVSDK/blob/master/src/core/mavsdk_impl.cpp#L78): This is a tough one. Could someone explain to me why we are blanket ignoring all QGC messages? Is this a QGC specific thing, or do we just not want to handle messages from other GCS? If that's the case, we should again use heartbeat to get a list of GCS based on `MAV_COMP_ID_MISSIONPLANNER` or similar, and ignore anything in that list. This seems restrictive, however, as you can't communicate with other GCS... i.e if 2 MAVSDK instances were communicating with each other. Either way, checking a specific SID is a bad idea - I for one run multiple QGC instances with different SIDs, so this wouldn't work off the bat. Would something bad happen if we didn't ignore GCS messages? Instead of "blacklisting", maybe we could "whitelist" by deciding on a per-message basis if we are capable of handling it?

These are just my thoughts - please feel free to comment and share ideas. Once we have a good idea of what we want to do, I can open a PR and implement the changes!

coderkalyan:
close #999.

This PR implements a a new `Configuration` class, replacing the old `enum class` of the same name. It allows for more flexibility and full control over the source system and component IDs of `Mavsdk` objects, in addition to fixing certain defaults to be more precise.

The `Configuration` class has 2 constructors - 1 which takes in the system and component id, for full manual control. The second one takes in a `UsageType` - which is an `enum class` inside `Configuration` which reimplements the same functionality as the original `enum class Configuration`. It allows the SDK to automatically choose an appropriate system and component ID based on the usage of the Mavsdk, whether it be `Autopilot, GroundStation` or `CompanionComputer`. I have attached an example (`examples/custom_configuration`) which can configure itself as any one of these or to a manual ID based on CLI arguments.

One thing mentioned by the issue still needs to be sorted out: In a couple of places, there is a blanket filter on any message with the system ID `255` - this is intended to filter out message from QGC, but is neither robust (as QGC's system ID can change) nor a good idea/even needed? @JonasVautherin thought that this was a remnant from when we used to connect to the first system we saw, but should no longer be needed as we now have more control over it. Still waiting for @julianoes or someone else's thoughts on this matter. This filtering happens in https://github.com/mavlink/MAVSDK/blob/master/src/core/mavsdk_impl.cpp#L78 and https://github.com/mavlink/MAVSDK/blob/master/src/core/udp_connection.cpp#L246.


coderkalyan:
In addition, for autopilot configuration mode, the system ID is currently set to `1`. We might want to do some heartbeat packet inspection to find the system ID of an autopilot, then set our own system ID to that (of course this only works when there is only one autopilot connected to the MAVLink network). Not sure how to go about implementing this, tho, since heartbeat inspection is outside the scope of `set_configuration` method?

example/custom_configuration.cpp/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
+cmake_minimum_required(VERSION 2.8.12)
```
julianoes:
```suggestion
cmake_minimum_required(VERSION 3.1)
```

example/custom_configuration.cpp/custom_configuration.cpp
```diff
@@ -0,0 +1,184 @@
...
+                  << std::endl;
+    });
+
+    // Check if vehicle is ready to arm
+    while (telemetry->health_all_ok() != true) {
+        std::cout << "Vehicle is getting ready to arm" << std::endl;
+        sleep_for(seconds(1));
+    }
+
+    // Arm vehicle
```
julianoes: I don't think the rest of this example makes sense because if your configuration is `autopilot` you can't actually send arm, or at least you should not.

Also, it just confuses a reader because it doesn't really matter. Therefore, I would keep it short and remove all of the stuff down here. If it only does configuration and then nothing else, that's fine.

example/custom_configuration.cpp/custom_configuration.cpp
```diff
@@ -0,0 +1,184 @@
+//
+// Simple example to demonstrate custom configuration
+// (system and component id) in MAVSDK
```
julianoes: It's a good idea to add an example to show how this is used. However, it would probably be easier to understand if the example was specific (and not generic like this one). E.g. there would be an example for the companion computer case and the example would be to do something like avoidance, something that makes sense for a companion.

src/core/mavsdk.h
```diff
@@ -131,14 +131,50 @@ class Mavsdk {
...
 
-    /**
-     * @brief Possible configurations.
-     */
-    enum class Configuration {
-        Autopilot, /**< @brief SDK is used as an autopilot. */
-        GroundStation, /**< @brief SDK is used as a ground station. */
-        CompanionComputer /**< @brief SDK is used on a companion computer onboard the system (e.g.
-                             drone). */
+    class Configuration {
```
julianoes: I'm not sure if we really need this configuration class. 

I'm wondering if it would be enough to have these setters:
```
Mavsdk::set_system_id(uint8_t system_id);
Mavsdk::set_component_id(uint8_t component_id);
Mavsdk::set_system_id_and_component_id(Configuration configuration);
```

@JonasVautherin what's your feel?

coderkalyan: @julianoes Actually @JonasVautherin suggested I use the class based method in the first place, rather than manually setting the ID via methods...

JonasVautherin: I like the `Configuration` class better, but I'm coming from Java :see_no_evil:. In the end, it is a matter of preference, so I could live with both.

One thing is that I believe the `Configuration` class is more versatile: it could become a bigger thing (with more configuration options) without creating a mess. Whereas the 3 functions may have to completely change depending on how this all evolves (but it may not, that's a valid argument :laughing:).

julianoes: Ok convinced!

src/core/mavsdk.h
```diff
@@ -131,14 +131,50 @@ class Mavsdk {
...
+    public:
+        enum class UsageType;
+
+        Configuration(uint8_t system_id, uint8_t component_id);
+        Configuration(UsageType usage_type);
+        ~Configuration();
```
julianoes:
```suggestion
        ~Configuration() = default;
```

src/core/mavsdk.h
```diff
@@ -131,14 +131,50 @@ class Mavsdk {
...
+
+        //Configuration(const Configuration&) = delete;
+        //Configuration& operator=(const Configuration&) = delete;
```
julianoes: Remove these two lines.

src/core/mavsdk.h
```diff
@@ -131,14 +131,50 @@ class Mavsdk {
...
-        Autopilot, /**< @brief SDK is used as an autopilot. */
-        GroundStation, /**< @brief SDK is used as a ground station. */
-        CompanionComputer /**< @brief SDK is used on a companion computer onboard the system (e.g.
-                             drone). */
+    class Configuration {
+    public:
+        enum class UsageType;
+
+        Configuration(uint8_t system_id, uint8_t component_id);
+        Configuration(UsageType usage_type);
```
julianoes: Make sure the normal operator is deleted or creates a sane default.

src/core/mavsdk_impl.h
```diff
@@ -15,6 +15,19 @@ namespace mavsdk {
 
 class MavsdkImpl {
 public:
+    /** @brief Default System ID for GCS configuration type. */
```
julianoes: Docstrings are not required for impl headers. I think they are just destracting. Good variable naming is enough.

example/custom_configuration.cpp/custom_configuration.cpp
coderkalyan: @julianoes Okay, I will modify this to be a companion example which streams/logs telemetry, I'm working on avoidance but thats out of the scope of this example IMO. Once avoidance is merged, I could write another example which takes advantage of the custom configuration to make avoidance possible.

julianoes: Great :+1:.

example/companion_configuration/companion_configuration.cpp
```diff
@@ -0,0 +1,160 @@
+//
+// Simple example to demonstrate configuring MAVSDK for use
+// on a companion computer. This example simply streams telemetry.
```
julianoes: That's not true, it also tries to arm and all, right? :smile:

coderkalyan: @julianoes whoops - anyway thats fixed now

coderkalyan:
@julianoes Is this ready to be merged?

julianoes:
@coderkalyan oh yes that would be nice. Can you fix the conflicts please and push again? I'm going to review it now.

coderkalyan:
Yikes! I rebased, didn't go so well. Lemme try to fix

julianoes:
What did you do? Now it's a mess.

I would have done:
```
git fetch --all
git rebase origin/develop
git push -f
```

julianoes:
Let me try.

julianoes:
@coderkalyan nevermind, you didn't give me permission to push to your remote for this, so I can't push.

julianoes:
https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests

julianoes:
Ok, so there are 3 ways to get this merged:
1. You rebase or merge this so there are no conflicts.
2. You give me permission to your fork (see [instructions](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests)) and I force push it.
3. We close this PR and I push your changes to another branch and create a new PR.

Your choice.

coderkalyan:
@julianoes I have no problem with giving you access lol, I'm trying to follow that guide you posted but I don't see any checkbox for allowing edits from maintainers...

coderkalyan:
Could it be because it's an organization fork and not a personal one? I own the organization so...

coderkalyan:
@julianoes Seems to be good now, the CI is running currently. Can you start the review?

coderkalyan:
I ran fix_style.sh, so the style CI task should pass now @julianoes.

julianoes:
> __w/MAVSDK/MAVSDK/tools/docs/install/include/mavsdk/mavsdk.h:138: warning: Member Configuration(uint8_t system_id, uint8_t component_id) (function) of class mavsdk::Mavsdk::Configuration is not documented.
/__w/MAVSDK/MAVSDK/tools/docs/install/include/mavsdk/mavsdk.h:139: warning: Member Configuration(UsageType usage_type) (function) of class mavsdk::Mavsdk::Configuration is not documented.
/__w/MAVSDK/MAVSDK/tools/docs/install/include/mavsdk/mavsdk.h:149: warning: Member get_system_id() const (function) of class mavsdk::Mavsdk::Configuration is not documented.
/__w/MAVSDK/MAVSDK/tools/docs/install/include/mavsdk/mavsdk.h:155: warning: Member get_component_id() const (function) of class mavsdk::Mavsdk::Configuration is not documented.

src/core/mavsdk.h
```diff
@@ -135,23 +135,22 @@ class Mavsdk {
     public:
         enum class UsageType;
 
+        /**
+         * @brief Create new Configuration via manually configured
+         * system and component ID.
```
julianoes: You also need to document the `@param`s.

src/core/mavsdk.h
```diff
@@ -135,23 +135,22 @@ class Mavsdk {
...
         ~Configuration();
 
-        // Configuration(const Configuration&) = delete;
-        // Configuration& operator=(const Configuration&) = delete;
-
-        /*
-         * @brief Get the system id of this configuration
-         * @return `integer` system id, between 1 and 255
-         */
+        /** @brief Get the system id of this configuration */
```
julianoes: And the `@return`. And if I told you earlier to remove the comments, then I'm very sorry :confused:.

coderkalyan: No worries!

coderkalyan:
@julianoes Fingers crossed, the CI will fully pass this time and we'll be ready to merge :)

julianoes:
Arg, so close:
> Searching for documented defines.../__w/MAVSDK/MAVSDK/tools/docs/install/include/mavsdk/mavsdk.h:134: warning: Compound mavsdk::Mavsdk::Configuration is not documented.

julianoes:
Style and docs passed, cool!

coderkalyan:
No problem! Happy to help.

PR #1004, opened 2020-03-05
Add api to stop `mavsdk_server`

JonasVautherin:
This changes the C API from having only `runBackend()` to having `runBackend()`, `getPort()`, `attach()` and `stopBackend()`.

`runBackend()` is not blocking anymore, and that's replaced by `attach()`.

I had to remove the namespaces because that's not working with the C API. The alternative would be to have ifdefs in the API file (to differentiate between `struct MavsdkServer*` and `mavsdk::backend::MavsdkServer*`, and I am not sure what's worse. Here I find the `backend_api.h` file quite clear, at the cost of losing the namespaces only for the API files (`backend_api.*` and `backend.*`).

The limitation now is that if you register to an infinite stream and try to `stopBackend()`, then it will hang forever, waiting for the stream to finish (which won't happen since it is infinite). I fixed the telemetry plugin, so that won't happen here, but it could happen in other plugins. I'm not planning to fix that before the C++ auto-generation, so I would count it as "known issue" :innocent:.

I would also like to do some renaming (backend vs mavsdk_server is a bit confusing right now), but not in this PR.

### DIFF
--- a/src/backend/src/backend.cpp
+++ b/src/backend/src/backend.cpp
@@ -6,8 +6,7 @@
 #include "mavsdk.h"
 #include "grpc_server.h"
 
-namespace mavsdk {
-namespace backend {
+using namespace mavsdk::backend;
 
 class MavsdkBackend::Impl {
 public:
@@ -24,15 +23,21 @@ public:
     {
         _server = std::unique_ptr<GRPCServer>(new GRPCServer(_dc));
         _server->set_port(port);
-        return _server->run();
+        _grpc_port = _server->run();
+        return _grpc_port;
     }
 
     void wait() { _server->wait(); }
 
+    void stop() { _server->stop(); }
+
+    int getPort() { return _grpc_port; }
+
 private:
-    Mavsdk _dc;
+    mavsdk::Mavsdk _dc;
     ConnectionInitiator<mavsdk::Mavsdk> _connection_initiator;
     std::unique_ptr<GRPCServer> _server;
+    int _grpc_port;
 };
 
 MavsdkBackend::MavsdkBackend() : _impl(new Impl()) {}
@@ -50,6 +55,12 @@ void MavsdkBackend::wait()
 {
     _impl->wait();
 }
+void MavsdkBackend::stop()
+{
+    _impl->stop();
+}
 
-} // namespace backend
-} // namespace mavsdk
+int MavsdkBackend::getPort()
+{
+    return _impl->getPort();
+}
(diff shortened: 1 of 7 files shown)


src/backend/src/backend.cpp
```diff
@@ -6,8 +6,7 @@
 #include "mavsdk.h"
 #include "grpc_server.h"
 
-namespace mavsdk {
-namespace backend {
+using namespace mavsdk::backend;
```
julianoes: I don't understand this change. I can see how you can't have namespaces in C but you can't have the class either C?

src/backend/src/backend_api.h
```diff
@@ -10,12 +10,18 @@ extern "C" {
...
 #endif
 
-DLLExport void runBackend(
+DLLExport struct MavsdkBackend* runBackend(
     const char* system_address,
     const int mavsdk_server_port,
     void (*onServerStarted)(void*),
     void* context);
 
+DLLExport int getPort(struct MavsdkBackend* backend);
```
julianoes: I would prefer `get_port` especially since it's C, and in the other calls as well.

JonasVautherin: You're right. Back when I started I followed gRPC's naming, but that was a mistake. I'll change that, either in this PR or in a following one that would also fix the "dc" and "backend" everywhere.

src/backend/src/backend.cpp
JonasVautherin: If I keep the namespace here, it gets ambiguous when building in C++. Because the header file says `struct MavsdkBackend*` and the implementation calls it `mavsdk::backend::MavsdkBackend*`. An alternative I considered was to `#ifdef __cplusplus` and define all the functions once for C and once for C++. I've seen examples that have a more complicated structure to avoid the `#ifdef`, too. I haven't seen something simple and elegant, though :confused:.

I could also do something like `#define MAVSDK_BACKEND_TYPE "struct MavsdkBackend*"` for C, and `#define MAVSDK_BACKEND_TYPE "mavsdk::backend::MavsdkBackend"` for C++, so that the functions look like:

```
DLLExport MAVSDK_BACKEND_TYPE runBackend();
```

But I'm not sure what's best. Opinions? :sweat_smile:

julianoes: You clearly thought more about this than me :smile:.

I would probably put the `#ifdef __cplusplus` around the namespace, I think that's fairly common.

julianoes: I tried and I failed so let's merge this as is for now.

PR #1005, opened 2020-03-05
add pollin to serial connection read

DanielePettenuzzo:
On an intel nuc running ubuntu 18.04 I had issues where my sdk example
(mavlink_ftp_client) was returning 1 in the main but the serial connection
thread was blocked because of the blocking read call.

My setup was an intel nuc connected to the telem2 port of a Pixhawk 4.

Also the "close" of the file descriptor in the stop method was called before
joining the receive thread which could lead to undefined behavior.

### DIFF
--- a/src/core/serial_connection.cpp
+++ b/src/core/serial_connection.cpp
@@ -6,6 +6,7 @@
 #include <unistd.h>
 #include <fcntl.h>
 #include <termios.h>
+#include <sys/poll.h>
 #endif
 
 namespace mavsdk {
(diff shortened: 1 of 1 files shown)


src/core/serial_connection.cpp
```diff
@@ -1,6 +1,7 @@
 #include "serial_connection.h"
 #include "global_include.h"
 #include "log.h"
+#include <sys/poll.h>
```
julianoes: This will have to be ifdefd against Windows.

src/core/serial_connection.cpp
```diff
@@ -267,9 +269,20 @@ void SerialConnection::receive()
...
 
+    struct pollfd fds[1];
+    fds[0].fd = _fd;
+    fds[0].events = POLLIN ;
+
     while (!_should_exit) {
         int recv_len;
 #if defined(LINUX) || defined(APPLE)
+        int pollrc = poll( fds, 1, 1000);
+        if(pollrc == 0 || !(fds[0].revents & POLLIN)){
```
julianoes: You have to fix style:
```
tools/fix_style.sh .
```

Issue #1003, opened 2020-03-05
No result with set_current_mission_item_async


julianoes:
There was a report that `set_current_mission_item_async` does not call the callback at all.

This needs fixing and we also need an integration tests checking it.

```
    int current = 4;
    auto prom = std::make_shared<std::promise<Mission::Result>>();
    auto future_result = prom->get_future();
    std::cout << "Setting current mission item..." << std::endl;
    mission->set_current_mission_item_async(current,
    [prom](Mission::Result result) {
        prom->set_value(result);
    });    const Mission::Result result = future_result.get(); //Wait on result    if(result == Mission::Result::SUCCESS){
        std::cout << "OK" << std::endl;
    }else{
        std::cout << "NOK" << std::endl;
    }
```

rohan2734:
This bug doesnt seem it is for a beginner

coderkalyan:
https://discuss.px4.io/t/mavsdk-clear-mission-async-callback-never-gets-called/17686 seems to be a report that `clear_mission_async` doesn't call the callback either. Maybe related?

JonasVautherin:
> Maybe related?

Yes, maybe. Would you mind linking this issue from the discuss thread? :blush:

coderkalyan:
Sure thing, will do.

julianoes:
Fixed by #1006, and covered by an integration test in 1264.

PR #1006, opened 2020-03-06
Add mission set current item feature

Katawann:
this feature has been removed in commit c6f67067c160759b4a738aabf3a9443e0e48c055. It allows to set the current item in a mission (async)

related to issue #1003

### DIFF
--- a/src/core/mavlink_mission_transfer.cpp
+++ b/src/core/mavlink_mission_transfer.cpp
@@ -43,6 +43,14 @@ void MAVLinkMissionTransfer::clear_items_async(uint8_t type, ResultCallback call
     _work_queue.push_back(ptr);
 }
 
+void MAVLinkMissionTransfer::set_current_item_async(uint8_t type, int &current, ResultCallback callback)
+{
+    auto ptr = std::make_shared<SetCurrentWorkItem>(
+        _sender, _message_handler, _timeout_handler, type, current, callback);
+
+    _work_queue.push_back(ptr);
+}
+
 void MAVLinkMissionTransfer::do_work()
 {
     LockedQueue<WorkItem>::Guard work_queue_guard(_work_queue);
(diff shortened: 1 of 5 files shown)


src/core/mavlink_mission_transfer.cpp
```diff
@@ -767,4 +775,126 @@ void MAVLinkMissionTransfer::ClearWorkItem::callback_and_reset(Result result)
...
+}
+
+void MAVLinkMissionTransfer::SetCurrentWorkItem::start()
+{
+    std::lock_guard<std::mutex> lock(_mutex);
+
+    _started = true;
+
+    // If we coudln't find it, the requested item is out of range and probably an invalid argument.
+    if (_current < 0) {
```
julianoes: We should also check for greater than length of items. Although never mind, we don't have that information in this class, the autopilot will have to take care of it.

src/core/mavlink_mission_transfer.cpp
```diff
@@ -767,4 +775,126 @@ void MAVLinkMissionTransfer::ClearWorkItem::callback_and_reset(Result result)
...
+        callback_and_reset(Result::CurrentInvalid);
```
julianoes: This error was initially for something else but I guess it fits.

src/core/mavlink_mission_transfer.cpp
```diff
@@ -767,4 +775,126 @@ void MAVLinkMissionTransfer::ClearWorkItem::callback_and_reset(Result result)
...
+
+void MAVLinkMissionTransfer::SetCurrentWorkItem::process_mission_current(const mavlink_message_t& message)
+{
+    std::lock_guard<std::mutex> lock(_mutex);
+
+    mavlink_mission_current_t mission_current;
+    mavlink_msg_mission_current_decode(&message,&mission_current);
+
+    _timeout_handler.remove(_cookie);
+    _current = mission_current.seq;
```
julianoes: I think we should instead check if the returned current is correct and report success if so.

src/core/mavlink_mission_transfer.cpp
```diff
@@ -767,4 +775,126 @@ void MAVLinkMissionTransfer::ClearWorkItem::callback_and_reset(Result result)
...
+    std::lock_guard<std::mutex> lock(_mutex);
+
+    mavlink_statustext_t status;
+    mavlink_msg_statustext_decode(&message,&status);
+
+    _timeout_handler.remove(_cookie);
+    //TODO status.severity
+
+    callback_and_reset(Result::ProtocolError);
+}
```
julianoes: Why is this required? I don't think a status text output should concern us.

src/core/mavlink_mission_transfer.h
```diff
@@ -229,6 +262,8 @@ class MAVLinkMissionTransfer {
 
     void clear_items_async(uint8_t type, ResultCallback callback);
 
+    void set_current_item_async(uint8_t type, int &current, ResultCallback callback);
```
julianoes: I wouldn't use a reference here. Just pass it by value.

src/core/mavlink_mission_transfer.cpp
Katawann: From the docs they say:

> Drone (server) receives message and attempts to update the current mission sequence number.
> 
> - On success, the Drone should broadcast a MISSION_CURRENT message containing the current sequence number (seq).
> - On failure, the Drone should broadcast a STATUSTEXT with a MAV_SEVERITY and a string stating the problem. This may be displayed in the UI of receiving systems.

So I was not sure if that something we need to check and how to do it

src/core/mavlink_mission_transfer.cpp
Katawann: It was my thought but I did not find a way to check it unless we receive the item in parameters

Katawann:
> so either you or I should add unit tests 
Yep agree! I will not have the time in next days but it can be good practice for me to do it. I will commit here when it's done !

julianoes:
Oh, and we should also use this function in one of the integration tests, that way we're less likely to break it again.

Katawann:
Oups not a review I wanted to do...

Not working but I think I will need your help to perform the unit test. I'm not used to this function and I'm running out of time to do it properly. For now it fails because it blocks on the EXPECT_EQ. What I missing ?

julianoes:
@Katawann I've added some commits. Let's see if this passes now.

Katawann:
Great ! Thank you for your help. I will check what you've done and do it with the upload mission raw I will pull request in few times !

julianoes:
@Katawann the upload `mission_raw` feature should be straightforward, basically a copy from `mission` and doesn't have any tests. Although an integration test would be nice :).

julianoes:
I realized that two more unit tests for retransmission and timeout need to be added. I'm on that.

julianoes:
Unrelated CI error, merging.

PR #1011, opened 2020-03-10
Integration tests bugfixes

### LINKED ISSUE #1010: Passthrough integration test crashes
Sometimes `SitlTest.MavlinkPassthrough` fails, as seen in #1009. 

```
[10:34:32|Info ] HIGHRES_IMU.temperature [1] (100)11.8327 degrees C (mavlink_passthrough.cpp:61)
terminate called after throwing an instance of 'std::system_error'
  what():  Resource deadlock avoided
tools/run-sitl-tests.sh: line 15: 227053 Aborted                 (core dumped) PX4_SIM_SPEED_FACTOR=10 AUTOSTART_SITL=1 PX4_FIRMWARE_DIR=$PX4_FIRMWARE_DIR HEADLESS=1 build/debug/src/integration_tests/integration_tests_runner --gtest_filter="SitlTest.MavlinkPassthrough"
```

Reproduced with `stress -c 8` and `while tools/run-sitl-tests.sh /home/julianoes/src/Firmware ; do :; done`

julianoes:
This contains two small bugfixes and one major change in architecture. For more information check the commit messages.

Fixes #1010.

### DIFF
--- a/src/core/plugin_impl_base.cpp
+++ b/src/core/plugin_impl_base.cpp
@@ -4,6 +4,6 @@
 
 namespace mavsdk {
 
-PluginImplBase::PluginImplBase(System& system) : _parent(system.system_impl()) {}
+PluginImplBase::PluginImplBase(System& system) : _parent(system.system_impl_raw()) {}
 
 } // namespace mavsdk
--- a/src/core/plugin_impl_base.h
+++ b/src/core/plugin_impl_base.h
@@ -58,7 +58,7 @@ public:
     const PluginImplBase& operator=(const PluginImplBase&) = delete;
 
 protected:
-    std::shared_ptr<SystemImpl> _parent;
+    SystemImpl* _parent;
 };
 
 } // namespace mavsdk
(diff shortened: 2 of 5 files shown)


julianoes:
> Why is it that the plugins don't have to check if _parent is still there when it is a raw pointer?

Haha, right. They do have to check if they don't want to segfault. This assumes a user does not destruct the `Mavsdk` object like so:

```
Mavsdk* mavsdk = new Mavsdk();
mavsdk->add_udp_connection();
System& system = mavsdk->system();
auto action = std::make_shared<Action>(system);
delete mavsdk;action->arm(); // -> Segfault!
```

@bkueng made a valid comment that, and you @JonasVautherin have raised that before, that we should not call the same callback from multiple threads but always the same. So, each telemetry stream should pick one thread from the thread pool, so have a thread affinity.

julianoes:
cb13be59bd863e5682bb552699a68611172bd220 has been cherry-picked. For the rest #1143 should fix it.

PR #1012, opened 2020-03-18
Takeoff landed state

shrit:
This pull request addresses using the takeoff process using landed state instead of sleep.

### DIFF
--- a/examples/offboard_velocity/offboard_velocity.cpp
+++ b/examples/offboard_velocity/offboard_velocity.cpp
@@ -230,6 +230,30 @@ void usage(std::string bin_name)
               << "For example, to connect to the simulator use URL: udp://:14540" << std::endl;
 }
 
+Telemetry::landed_state_callback_t landed_state_callback(std::promise<void>& landed_promise)
+{
+    return [&landed_promise](Telemetry::LandedState landed) {
+        switch (landed) {
+            case Telemetry::LandedState::ON_GROUND:
+                std::cout << "On ground" << std::endl;
+                break;
+            case Telemetry::LandedState::TAKING_OFF:
+                std::cout << "Taking off..." << std::endl;
+                break;
+            case Telemetry::LandedState::LANDING:
+                std::cout << "Landing..." << std::endl;
+                break;
+            case Telemetry::LandedState::IN_AIR:
+                std::cout << "Taking off has finished." << std::endl;
+                landed_promise.set_value_at_thread_exit();
+                break;
+            case Telemetry::LandedState::UNKNOWN:
+                std::cout << "Unknown landed state." << std::endl;
+                break;
+        }
+    };
+}
+
 int main(int argc, char** argv)
 {
     Mavsdk dc;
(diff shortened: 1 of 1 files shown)


shrit:
@julianoes, if you find the principle is good, you can tell me so I will apply the same thing on other examples

examples/offboard_velocity/offboard_velocity.cpp
```diff
@@ -259,6 +259,17 @@ int main(int argc, char** argv)
     std::cout << "In Air..." << std::endl;
     sleep_for(seconds(5));
 
+    while (true) {
```
JonasVautherin: I would find it more elegant to use [landed_state_async](https://github.com/mavlink/MAVSDK/blob/develop/src/plugins/telemetry/include/plugins/telemetry/telemetry.h#L1196) with a future/promise to wait until the right state is reached.

That would get rid of the `while (true)` and active waiting (`sleep_for(milliseconds(500))`)...

shrit: Agreed,

julianoes: And since you added the wait, you can remove the `sleep_for(seconds(5))` above.

shrit:
@julianoes I am sorry, but I did not get what you mean here?

julianoes:
You are putting two completely different changes in to the same pull request. That is a bit confusing. It would be nicer to have two separate pull requests.

shrit:
Ah, you mean the example to examples. OK I will do it in a new pull request. Sorry I forgot that I have done that :smile:

shrit:
@julianoes It is strange since I am getting a run time error, with a future throwing excepting reclaiming that the promise has already been satisfied. I do not understand the reason behind this since the callback has been called once, but it seems that it being called twice for none clear reasons. :thinking:

julianoes:
@shrit `landed_state_async` is called every second, so that would explain why the promise is satisfied twice. You need to do `landed_state_async(nullptr)` inside the callback to unsubscribe again.

shrit:
@julianoes I have done as you told me, the promise does not throw an exception any longer. However, at the same time, the promise is still waiting after calling the `landed_state_async(nullptr)`

julianoes:
@shrit I've merged the example->examples PR. You can now rebase this PR and force-push.

example/offboard_velocity/offboard_velocity.cpp
```diff
@@ -211,6 +230,31 @@ void usage(std::string bin_name)
...
+                break;
+            case Telemetry::LandedState::TAKING_OFF:
+                std::cout << "Taking off..." << std::endl;
+                break;
+            case Telemetry::LandedState::LANDING:
+                std::cout << "Landing..." << std::endl;
+                break;
+            case Telemetry::LandedState::IN_AIR:
+                std::cout << "Taking off has finished." << std::endl;
+                landed_promise.set_value_at_thread_exit();
```
julianoes: What is this? I've never seen it.

Why don't you just do:
```
telemetry->landed_state_async(nullptr);
landed_promise.set_value();
```

shrit: Actually I had called `telemetry->landed_state_async(nullptr);` after the `set_value()` in line 249.
I will try to call async before setting the value

shrit:
OK, The problem I had to revert the `examples` to `example` in a previous commit here. I think I will open a new pull request, or I will drop the example commits

shrit:
@julianoes, Thanks, I think everything is fine now, it is working as magic.

PR #1016, opened 2020-03-20
Upload mission raw

Katawann:
This is a first draft and it just compiles yet but it bothers me that it is just copy/paste (but not exactly) from Mission....

I feel it would do more sense to do a `MissionRawItem` rather than `MissionRaw`. For me, `MissionRaw` is just a lower level than `MissionItem`. And `MissionItem` would just use `MissionRawItem`. If we want high level functionalities we'll use `MissionItem` and for more advanced settings like I need we can work with `MissionRawItem`.

also the `MavlinkMissionItemInt` is exactly a `MAVLinkMissionTransfer::ItemInt` so I don't get the purpose to create another struct for that

### DIFF
--- a/src/integration_tests/mission_raw_mission_changed.cpp
+++ b/src/integration_tests/mission_raw_mission_changed.cpp
@@ -1,6 +1,5 @@
 #include "integration_test_helper.h"
 #include "mavsdk.h"
-#include "plugins/mission/mission.h"
 #include "plugins/mission_raw/mission_raw.h"
 #include <cmath>
 #include <future>
@@ -30,7 +29,6 @@ TEST_F(SitlTest, MissionRawMissionChanged)
     System& system = dc.system();
     ASSERT_TRUE(system.has_autopilot());
 
-    auto mission = std::make_shared<Mission>(system);
     auto mission_raw = std::make_shared<MissionRaw>(system);
 
     std::promise<void> prom_changed{};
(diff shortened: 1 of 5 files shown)


Katawann:
Just updated to a working version (at least with the integration test), but my remarks still remain :)

julianoes:
>  And MissionItem would just use MissionRawItem.

I agree. There is always a trade-off between duplication and independence though. The idea with the plugins is that they do not depend on each other. That way you can only use a subset without problems. For now, I would argue this duplication is ok because the big part is now refactored away in core.

> also the MavlinkMissionItemInt is exactly a MAVLinkMissionTransfer::ItemInt so I don't get the purpose to create another struct for that

Right, it's suboptimal but I wouldn't worry too much about it. If we want to re-use the struct, we need to include an internal header of core when using MAVSDK. However, plugins should only include plugins and not core (except masvsdk, system, etc.). In the same way the core should depend on something defined in a plugin, otherwise they get coupled and you can't build core without a certain plugin.

src/integration_tests/mission_raw_mission_changed.cpp
```diff
@@ -49,24 +48,54 @@ TEST_F(SitlTest, MissionRawMissionChanged)
     // The mission change callback should not trigger yet.
     EXPECT_EQ(fut_changed.wait_for(std::chrono::milliseconds(500)), std::future_status::timeout);
 
-    std::vector<std::shared_ptr<MissionItem>> mission_items;
+    std::vector<std::shared_ptr<MissionRaw::MavlinkMissionItemInt>> mission_raw_items;
```
julianoes: You can also remove the `mission` include now.

src/plugins/mission_raw/include/plugins/mission_raw/mission_raw.h
```diff
@@ -113,6 +128,26 @@ class MissionRaw : public PluginBase {
...
+    void upload_mission_async(const std::vector<std::shared_ptr<MissionRaw::MavlinkMissionItemInt>>& mission_raw,
+            result_callback_t callback);
+
+    /**
+     * @brief Cancel a mission upload (asynchronous).
+     *
+     * This cancels an ongoing mission upload. The mission upload will fail
+     * with the result `Result::CANCELLED`.
+     */
+    //TODO void upload_mission_cancel();
```
julianoes: Please either implement it or remove it.

src/plugins/mission_raw/mission_raw.cpp
```diff
@@ -19,6 +19,12 @@ void MissionRaw::download_mission_cancel()
     _impl->download_mission_cancel();
 }
 
+void MissionRaw::upload_mission_async(
+    const std::vector<std::shared_ptr<MavlinkMissionItemInt> > &mission_raw, result_callback_t callback)
```
julianoes: You probably need to run `./tools/fix_style.sh .`

src/plugins/mission_raw/mission_raw_impl.cpp
```diff
@@ -146,6 +178,30 @@ void MissionRawImpl::download_mission_cancel()
     // TODO: Implement cancel.
 }
 
+void MissionRawImpl::upload_mission_async(const std::vector<std::shared_ptr<MissionRaw::MavlinkMissionItemInt> > &mission_raw, const MissionRaw::result_callback_t &callback)
+{
+    if (!_parent->does_support_mission_int()) {
+        LogWarn() << "Mission int messages not supported";
+        // report_mission_result(callback, Mission::Result::ERROR);
```
julianoes: This callback is required and should be implemented using `_parent->call_user_callback`.

src/plugins/mission_raw/include/plugins/mission_raw/mission_raw.h
Katawann: Not too lazy but kind in a rush with dev so I removed it :)

Katawann:
> > And MissionItem would just use MissionRawItem.
> 
> I agree. There is always a trade-off between duplication and independence though. The idea with the plugins is that they do not depend on each other. That way you can only use a subset without problems. For now, I would argue this duplication is ok because the big part is now refactored away in core.

OK I understand. And actually it makes sense to do this use of MissionItem/MissionRaw in the core of the companion computer I'm working on

julianoes:
Docs warnings:
```
 Gen/__w/MAVSDK/MAVSDK/tools/docs/install/include/mavsdk/plugins/mission_raw/mission_raw.h:130: warning: argument 'mission_items' of command @param is not found in the argument list of mavsdk::MissionRaw::upload_mission_async(const std::vector< std::shared_ptr< MissionRaw::MavlinkMissionItemInt >> &mission_raw, result_callback_t callback)
/__w/MAVSDK/MAVSDK/tools/docs/install/include/mavsdk/plugins/mission_raw/mission_raw.h:138: warning: The following parameters of mavsdk::MissionRaw::upload_mission_async(const std::vector< std::shared_ptr< MissionRaw::MavlinkMissionItemInt >> &mission_raw, result_callback_t callback) are not documented:
/__w/MAVSDK/MAVSDK/tools/docs/install/include/mavsdk/plugins/mission_raw/mission_raw.h:89: warning: Member operator==(const MavlinkMissionItemInt &other) const (function) of class mavsdk::MissionRaw::MavlinkMissionItemInt is not documented.
G/__w/MAVSDK/MAVSDK/tools/docs/install/include/mavsdk/plugins/mission_raw/mission_raw.h:130: warning: argument 'mission_items' of command @param is not found in the argument list of mavsdk::MissionRaw::upload_mission_async(const std::vector< std::shared_ptr< MissionRaw::MavlinkMissionItemInt >> &mission_raw, result_callback_t callback)
/__w/MAVSDK/MAVSDK/tools/docs/install/include/mavsdk/plugins/mission_raw/mission_raw.h:138: warning: The following parameters of mavsdk::MissionRaw::upload_mission_async(const std::vector< std::shared_ptr< MissionRaw::MavlinkMissionItemInt >> &mission_raw, result_callback_t callback) are not documented:
```

Katawann:
Hope it's good now :) Sorry for the multiple modifications though !

How do you do to generate this "Docs warning" ? Like this I can do proper pull request next time

Katawann:
> How do you do to generate this "Docs warning" ? Like this I can do proper pull request next time

Ok found it in the CI build. And corrected the conversion issue
```
D:\a\MAVSDK\MAVSDK\src\integration_tests\mission_raw_mission_changed.cpp(65,41): warning C4244: '=': conversion from 'double' to 'int32_t', possible loss of data [D:\a\MAVSDK\MAVSDK\build\release\src\integration_tests\integration_tests_runner.vcxproj]
D:\a\MAVSDK\MAVSDK\src\integration_tests\mission_raw_mission_changed.cpp(66,41): warning C4244: '=': conversion from 'double' to 'int32_t', possible loss of data [D:\a\MAVSDK\MAVSDK\build\release\src\integration_tests\integration_tests_runner.vcxproj]
```

Not sure if the iOS issue is related

julianoes:
Thanks! The CI error looks unrelated.

PR #1019, opened 2020-03-23
action.h: Fix broken link to Return mode docs

hamishwillee:


julianoes:
Thanks @hamishwillee. The action.h is now autogenerated, so it should be fixed here:
https://github.com/mavlink/MAVSDK-Proto/blob/c9ecfd46951ae25bc9792b6cdef1eb44b2850f9b/protos/action/action.proto#L63

Right @JonasVautherin ?

hamishwillee:
Cool. Note I did make one push to master by accident of one of these fixes - rest are PRs you can close. Sorry if that causes a problem :-(

JonasVautherin:
Right, sorry for the regression @hamishwillee! The good news is that autogeneration is centralizing the documentation in the proto files, to be reused in all the language bindings (currently the python code auto-generates documentation, too).

I'll fix the proto file while adding the header, today.

hamishwillee:
Great! Then just close this :-)

JonasVautherin:
FYI: fixed here: https://github.com/mavlink/MAVSDK-Proto/blob/master/protos/action/action.proto#L63

PR #1021, opened 2020-03-23
MissionRaw: add multiples features

Katawann:
I added multiple features for MissionRaw:
start/pause/clear/set_current/mission_finished/subscribe_progress/cancel upload/download

I still need to manage if a RTL command is sent. This is mostly copy/paste from Mission class.

Tests are still required

### DIFF
--- a/src/plugins/mission_raw/mission_raw_impl.cpp
+++ b/src/plugins/mission_raw/mission_raw_impl.cpp
@@ -22,16 +22,37 @@ void MissionRawImpl::init()
         MAVLINK_MSG_ID_MISSION_ACK,
         std::bind(&MissionRawImpl::process_mission_ack, this, _1),
         this);
-}
 
-void MissionRawImpl::deinit()
-{
-    _parent->unregister_all_mavlink_message_handlers(this);
+    _parent->register_mavlink_message_handler(
+        MAVLINK_MSG_ID_MISSION_CURRENT,
+        std::bind(&MissionRawImpl::process_mission_current, this, _1),
+        this);
+
+    _parent->register_mavlink_message_handler(
+        MAVLINK_MSG_ID_MISSION_ITEM_REACHED,
+        std::bind(&MissionRawImpl::process_mission_item_reached, this, _1),
+        this);
 }
 
 void MissionRawImpl::enable() {}
 
 void MissionRawImpl::disable() {}
+
+void MissionRawImpl::deinit()
+{
+    _parent->unregister_timeout_handler(_timeout_cookie);
+    _parent->unregister_all_mavlink_message_handlers(this);
+}
+
+void MissionRawImpl::reset_mission_progress()
+{
+    std::lock_guard<std::mutex> lock(_mission_data.mutex);
+    _mission_data.last_current_mavlink_mission_item = -1;
+    _mission_data.last_reached_mavlink_mission_item = -1;
+    _mission_data.last_total_mavlink_mission_item = -1;
+    _mission_data.last_current_reported_mavlink_mission_item = -1;
+    _mission_data.last_total_reported_mavlink_mission_item = -1;
+}
 
 void MissionRawImpl::process_mission_ack(const mavlink_message_t& message)
 {
@@ -53,16 +74,110 @@ void MissionRawImpl::process_mission_ack(const mavlink_message_t& message)
     }
 }
 
-std::vector<MAVLinkMissionTransfer::ItemInt> MissionRawImpl::convert_to_int_items(
-    const std::vector<std::shared_ptr<MissionRaw::MavlinkMissionItemInt>>& mission_raw)
+void MissionRawImpl::process_mission_current(const mavlink_message_t& message)
 {
-    std::vector<MAVLinkMissionTransfer::ItemInt> int_items;
+    mavlink_mission_current_t mission_current;
+    mavlink_msg_mission_current_decode(&message, &mission_current);
 
-    for (const auto& item : mission_raw) {
-        int_items.push_back(convert_mission_raw(item));
+    {
+        std::lock_guard<std::mutex> lock(_mission_data.mutex);
+        _mission_data.last_current_mavlink_mission_item = mission_current.seq;
     }
 
-    return int_items;
+    report_progress();
+}
+
+void MissionRawImpl::process_mission_item_reached(const mavlink_message_t& message)
+{
+    mavlink_mission_item_reached_t mission_item_reached;
+    mavlink_msg_mission_item_reached_decode(&message, &mission_item_reached);
+
+    {
+        std::lock_guard<std::mutex> lock(_mission_data.mutex);
+        _mission_data.last_reached_mavlink_mission_item = mission_item_reached.seq;
+    }
+
+    report_progress();
+}
+
+void MissionRawImpl::upload_mission_async(
+    const std::vector<std::shared_ptr<MissionRaw::MavlinkMissionItemInt>>& mission_raw,
+    const MissionRaw::result_callback_t& callback)
+{
+    if (_mission_data.last_upload.lock()) {
+        _parent->call_user_callback([callback]() {
+            if (callback) {
+                callback(MissionRaw::Result::BUSY);
+            }
+        });
+        return;
+    }
+
+    if (!_parent->does_support_mission_int()) {
+        _parent->call_user_callback([callback]() {
+            if (callback) {
+                callback(MissionRaw::Result::UNSUPPORTED);
+            }
+        });
+        return;
+    }
+
+    const auto int_items = convert_to_int_items(mission_raw);
+
+    _mission_data.last_upload = _parent->mission_transfer().upload_items_async(
+        MAV_MISSION_TYPE_MISSION,
+        int_items,
+        [this, callback, int_items](MAVLinkMissionTransfer::Result result) {
+            auto converted_result = convert_result(result);
+            auto converted_items = convert_items(int_items);
+            _parent->call_user_callback([callback, converted_result, converted_items]() {
+                if (callback) {
+                    callback(converted_result);
+                }
+            });
+        });
+}
+
+void MissionRawImpl::upload_mission_cancel()
+{
+    auto ptr = _mission_data.last_upload.lock();
+    if (ptr) {
+        ptr->cancel();
+    }
+}
+
+void MissionRawImpl::download_mission_async(
+    const MissionRaw::mission_items_and_result_callback_t& callback)
+{
+    if (_mission_data.last_download.lock()) {
+        _parent->call_user_callback([callback]() {
+            if (callback) {
+                std::vector<std::shared_ptr<MissionRaw::MavlinkMissionItemInt>> empty_items;
+                callback(MissionRaw::Result::BUSY, empty_items);
+            }
+        });
+        return;
+    }
+
+    _mission_data.last_download = _parent->mission_transfer().download_items_async(
+        MAV_MISSION_TYPE_MISSION,
+        [this, callback](
+            MAVLinkMissionTransfer::Result result,
+            std::vector<MAVLinkMissionTransfer::ItemInt> items) {
+            auto converted_result = convert_result(result);
+            auto converted_items = convert_items(items);
+            _parent->call_user_callback([callback, converted_result, converted_items]() {
+                callback(converted_result, converted_items);
+            });
+        });
+}
+
+void MissionRawImpl::download_mission_cancel()
+{
+    auto ptr = _mission_data.last_download.lock();
+    if (ptr) {
+        ptr->cancel();
+    }
 }
 
 MAVLinkMissionTransfer::ItemInt MissionRawImpl::convert_mission_raw(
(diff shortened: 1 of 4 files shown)


src/plugins/mission_raw/mission_raw_impl.cpp
```diff
@@ -22,16 +22,36 @@ void MissionRawImpl::init()
...
+    _parent->unregister_timeout_handler(_timeout_cookie);
     _parent->unregister_all_mavlink_message_handlers(this);
 }
 
-void MissionRawImpl::enable() {}
-
-void MissionRawImpl::disable() {}
+void MissionRawImpl::reset_mission_progress()
+{
+    std::lock_guard<std::recursive_mutex> lock(_mission_data.mutex);
```
julianoes: Let's not make the mistake of using the `std::recursive_mutex` again. I don't think this should be required.

src/plugins/mission_raw/mission_raw_impl.cpp
```diff
@@ -22,16 +22,36 @@ void MissionRawImpl::init()
...
+    _mission_data.last_current_mavlink_mission_item = -1;
+    _mission_data.last_reached_mavlink_mission_item = -1;
+    _mission_data.last_current_reported_mission_item = -1;
+    _mission_data.last_total_reported_mission_item = -1;
```
julianoes: Isn't `current_mavlink` and `current_reported` the same for `mission_raw`?

src/plugins/mission_raw/mission_raw_impl.cpp
```diff
@@ -87,40 +201,24 @@ MAVLinkMissionTransfer::ItemInt MissionRawImpl::convert_mission_raw(
...
-        default:
-            return MissionRaw::Result::UNKNOWN;
+    std::vector<MAVLinkMissionTransfer::ItemInt> int_items;
+
+    // FIXME item_i only increment when new NAV_WAYPOINT
+    int item_i = -1;
+
+    for (const auto& item : mission_raw) {
+        if (item->command == MAV_CMD_NAV_WAYPOINT) {
+            ++item_i;
```
julianoes: I think this logic is not needed. For current I would just use the actual MAVLink index. The map like for `mission` is not required here, right?

src/plugins/mission_raw/mission_raw_impl.cpp
Katawann: Changed. Note this is taken from mission_impl.cpp. Should we update in MissionImpl:reset_mission_progress() as well then ?

src/plugins/mission_raw/mission_raw_impl.cpp
Katawann: `current_mavlink` gives the number of MAV_CMD we sent and `current_reported` is mapping every command for a given waypoint right ?

IMO this is interesting to keep this mapping. My use case is that I want to generate a Mission but to customize like I want so I send `MissionRaw` rather than `Mission` but without to change the logic of following progress. To be correct we may add a `subscribe_malvink_progress` that returns the progress of the mavlink command (if my question above is correct)

src/plugins/mission_raw/mission_raw_impl.cpp
julianoes: We should but we need to be careful and test it.

src/plugins/mission_raw/mission_raw_impl.cpp
julianoes: Aha. I don't think the mapping makes sense because there might be more mission item types that we should count that we currently are not counting.

Since this is raw I would strongly suggest that also the progress is raw. Otherwise it gets confusing.

Katawann: ok yes I understand that it could not match all use cases. I'll do this modification and manage this on my side. So do we agree I have to remove all the `*_mission_item_*` related  and that progress/upload/download/indice/etc are defined by a `*_mavlink_message_item_*` or another name like this. Otherwise it would be a bit confusing to have "mission_item" name

julianoes: ok sounds good.

src/plugins/mission_raw/mission_raw_impl.cpp
```diff
@@ -22,16 +22,37 @@ void MissionRawImpl::init()
...
 }
 
-void MissionRawImpl::enable() {}
-
-void MissionRawImpl::disable() {}
+void MissionRawImpl::reset_mission_progress()
+{
+    std::lock_guard<std::mutex> lock(_mission_data.mutex);
+    _mission_data.last_current_mavlink_mission_item = -1;
+    _mission_data.last_reached_mavlink_mission_item = -1;
```
Katawann: I don't understand the difference between MISSION_CURRENT ( #42 ) and MISSION_ITEM_REACHED ( #46 ) so not sure if the logic to report progress and current progress ans so on is correct

julianoes: Good point. MISSION_REACHED would give you `n` while MISSION_CURRENT is `n+1`.

src/plugins/mission_raw/mission_raw_impl.h
```diff
@@ -43,8 +67,27 @@ class MissionRawImpl : public PluginImplBase {
...
+    struct MissionData {
+        mutable std::mutex mutex{};
+        int last_current_mavlink_mission_item{-1};
+        int last_reached_mavlink_mission_item{-1};
+        int last_total_mavlink_mission_item{-1};
+        int last_current_reported_mavlink_mission_item{-1};
+        int last_total_reported_mavlink_mission_item{-1};
+        int num_mission_items_to_download{-1};
+        int next_mission_item_to_download{-1};
+        int last_mission_item_to_upload{-1};
```
julianoes: These are not needed.

src/plugins/mission_raw/mission_raw_impl.cpp
```diff
@@ -156,63 +249,217 @@ MissionRawImpl::convert_items(const std::vector<MAVLinkMissionTransfer::ItemInt>
...
+    }
+
+    if (_mission_data.last_total_mavlink_mission_item < 0) {
+        return false;
+    }
+
+    return (
+        _mission_data.last_reached_mavlink_mission_item ==
+        _mission_data.last_total_mavlink_mission_item);
+}
```
Katawann: @julianoes Not sure how to manage the `mission_finished` and if we should manage it in the `MissionRaw`. Especially with the RTL, in mission_impl there is something like this:

```
    // It is not straightforward to look at "current" because it jumps to 0
    // once the last item has been done. Therefore we have to lo decide using
    // "reached" here.
    // It seems that we never receive a reached when RTL is initiated after
    // a mission, and we need to account for that.
    const unsigned rtl_correction = _enable_return_to_launch_after_mission ? 2 : 1;

    return (
        unsigned(_mission_data.last_reached_mavlink_mission_item + rtl_correction) ==
        _mission_data.mavlink_mission_item_to_mission_item_indices.size());
```

but I guess that not something we can do in `MissionRaw`. So we should rather return `reached_mavlink_mission_item` as well and let the user manage when the mission is finished ?

Note: it should be ```
_mission_data.last_reached_mavlink_mission_item + 1 ==
        _mission_data.last_total_mavlink_mission_item);
``` but I'll update it when I know what to do with the `mission_finished` function

**Comment by Katawann** [comment]
After a small headache, I could not find a generic way to implement the `is_mission_finished` because as you said:
> I don't think the mapping makes sense because there might be more mission item types that we should count that we currently are not counting.

The cleanest way and the "rawest" IMO is just to give update of what's happening and let the user choose what to do with that

File: src/plugins/mission_raw/mission_raw_impl.cpp (hunk 6 above)
Comment (julianoes) [reply src/plugins/mission_raw/mission_raw_impl.cpp]: I'd have to play with SITL and see how it works, I'm not sure about the edge cases from the top of my head, sorry.

Comment (Katawann) [reply src/plugins/mission_raw/mission_raw_impl.cpp]: I have something like this if I upload just 6 waypoints (without changing speed or anything and without RTL)

Start Mission
[04:13:45|Debug] mavlink_current: 1, mavlink_total: 6 (mission_raw_impl.cpp:370)
[04:13:45|Debug] mavlink_reached: 0, mavlink_total: 6 (mission_raw_impl.cpp:402)
[04:13:45|Debug] 1st waypoint reached
[04:13:50|Debug] mavlink_reached: 1, mavlink_total: 6 (mission_raw_impl.cpp:402)
[04:13:50|Debug] mavlink_current: 2, mavlink_total: 6 (mission_raw_impl.cpp:370)
[04:13:50|Debug] 2nd waypoint reached
[04:13:55|Debug] mavlink_reached: 2, mavlink_total: 6 (mission_raw_impl.cpp:402)
[04:13:55|Debug] mavlink_current: 3, mavlink_total: 6 (mission_raw_impl.cpp:370)
[04:13:55|Debug] 3rd waypoint reached
[04:14:01|Debug] mavlink_reached: 3, mavlink_total: 6 (mission_raw_impl.cpp:402)
[04:14:01|Debug] mavlink_current: 4, mavlink_total: 6 (mission_raw_impl.cpp:370)
[04:14:01|Debug] 4th waypoint reached
[04:14:06|Debug] mavlink_reached: 4, mavlink_total: 6 (mission_raw_impl.cpp:402)
[04:14:06|Debug] mavlink_current: 5, mavlink_total: 6 (mission_raw_impl.cpp:370)
[04:14:06|Debug] 5th waypoint reached
[04:14:11|Debug] MAVLink: info: Mission finished, loitering. (system_impl.cpp:257)
[04:14:11|Debug] mavlink_reached: 5, mavlink_total: 6 (mission_raw_impl.cpp:402)
[04:14:11|Debug] 6th waypoint reached (and last)

Ok let me know ! Not sure if the `mission_finished` would take all the possibilities that an user could generate...

Comment (Katawann) [reply src/plugins/mission_raw/mission_raw_impl.cpp]: and with RTL:
Start Mission
[05:14:20|Debug] mavlink_reached: 0, mavlink_total: 7 (mission_raw_impl.cpp:402)
[05:14:20|Debug] mavlink_current: 1, mavlink_total: 7 (mission_raw_impl.cpp:370)
[05:14:20|Debug] 1st waypoint reached
[05:14:24|Debug] mavlink_reached: 1, mavlink_total: 7 (mission_raw_impl.cpp:402)
[05:14:24|Debug] mavlink_current: 2, mavlink_total: 7 (mission_raw_impl.cpp:370)
[05:14:24|Debug] 2nd waypoint reached
[05:14:30|Debug] mavlink_reached: 2, mavlink_total: 7 (mission_raw_impl.cpp:402)
[05:14:30|Debug] mavlink_current: 3, mavlink_total: 7 (mission_raw_impl.cpp:370)
[05:14:30|Debug] 3rd waypoint reached
[05:14:35|Debug] mavlink_reached: 3, mavlink_total: 7 (mission_raw_impl.cpp:402)
[05:14:35|Debug] mavlink_current: 4, mavlink_total: 7 (mission_raw_impl.cpp:370)
[05:14:35|Debug] 4th waypoint reached
[05:14:40|Debug] mavlink_reached: 4, mavlink_total: 7 (mission_raw_impl.cpp:402)
[05:14:40|Debug] mavlink_current: 5, mavlink_total: 7 (mission_raw_impl.cpp:370)
[05:14:40|Debug] 5th waypoint reached
[05:14:45|Debug] MAVLink: info: Returning to launch (system_impl.cpp:257)
[05:14:45|Debug] mavlink_reached: 5, mavlink_total: 7 (mission_raw_impl.cpp:402)
[05:14:45|Debug] mavlink_current: 6, mavlink_total: 7 (mission_raw_impl.cpp:370)
[05:14:45|Debug] 6th waypoint reached (and last)
[05:14:45|Debug] MAVLink: info: RTL HOME activated (system_impl.cpp:257)
[05:14:45|Debug] MAVLink: info: RTL: climb to 518 m (31 m above home) (system_impl.cpp:257)
Flight mode: RTL

Comment (julianoes) [reply src/plugins/mission_raw/mission_raw_impl.cpp]: So you're saying it's all working? :smile: Or what am I reading?

Comment (Katawann) [reply src/plugins/mission_raw/mission_raw_impl.cpp]: The progress current and reached are working fine but I did not implement the `mission_finished`. The "waypoint reached" are generated by my companion computer and not `MissionRaw` so `MissionRaw` only gives me updates on reached/current

Comment (Katawann) [reply src/plugins/mission_raw/mission_raw_impl.cpp]: So I did not implement the function `is_mission_finished` below from `mission_impl` because to keep the Mission**Raw** I'm not sure if it makes sense to detect if there is a RTL command in mission uploaded.

```
bool MissionImpl::is_mission_finished() const
{
    std::lock_guard<std::recursive_mutex> lock(_mission_data.mutex);

    if (_mission_data.last_current_mavlink_mission_item < 0) {
        return false;
    }

    if (_mission_data.last_reached_mavlink_mission_item < 0) {
        return false;
    }

    if (_mission_data.mavlink_mission_item_to_mission_item_indices.size() == 0) {
        return false;
    }

    // It is not straightforward to look at "current" because it jumps to 0
    // once the last item has been done. Therefore we have to lo decide using
    // "reached" here.
    // It seems that we never receive a reached when RTL is initiated after
    // a mission, and we need to account for that.
    const unsigned rtl_correction = _enable_return_to_launch_after_mission ? 2 : 1;

    return (
        unsigned(_mission_data.last_reached_mavlink_mission_item + rtl_correction) ==
        _mission_data.mavlink_mission_item_to_mission_item_indices.size());
}
```

Which mean I do not have `current_mission_item` from `Mission` but I replaced it by `current_mavlink_mission_item` and `reached_mavlink_mission_item` to let the developer using `MissionRaw` in a "raw" way

Katawann: So for me it's all good ! Now this the test part that I never have the time to do 😅

Katawann:
Great ! For integration tests I will not have the time in the next days because I have to work on the developments for my company but I'll try to write something if I have free time !

julianoes:
We're about to auto-generate this plugin, so I need to merge this now and then we can add more tests later.

Katawann:
Yep sorry I could not find the time to create the tests. But I'm using this PR since one month and I had no surprise. Of course are still required to avoid to break it later

julianoes:
@Katawann that is very good to know! :smile:

PR #1022, opened 2020-03-26
Implement gimbal manager and gimbal device APIs to reflect updated MAVLink Gimbal protocol

coderkalyan:
This PR deprecates the existing Gimbal class in the gimbal plugin. In its place are 2 new classes, GimbalManager and GimbalDevice. Each of these implement a "server" and "client" API to interact with MAVLink enabled gimbals using the new API. 

https://github.com/mavlink/mavlink/pull/1174


coderkalyan:
Note to myself: Gimbal manager has to be refactored to connect to only 1 gimbal, which better respects the paradigm talked about in the design document. Multiple gimbal managers should be instantiated to connect to multiple gimbals. @julianoes @hamishwillee do you agree?

In addition, @julianoes I'm going to consider refactoring `Gimbal` into `GimbalLegacy` and reusing the `Gimbal` class as a "high level" wrapper, which can automatically create multiple managers and has a more friendly API. What are your thoughts on this?

hamishwillee:
> Note to myself: Gimbal manager has to be refactored to connect to only 1 gimbal, which better respects the paradigm talked about in the design document. Multiple gimbal managers should be instantiated to connect to multiple gimbals. @julianoes @hamishwillee do you agree?

Yes. Gimbal manager > Device relationship is 1:1.
The component that is going to own the manager queries for attached gimbal devices and instantiates however many it needs. 

> In addition, @julianoes I'm going to consider refactoring Gimbal into GimbalLegacy and reusing the Gimbal class as a "high level" wrapper, which can automatically create multiple managers and has a more friendly API (the current manager API is a pretty thin wrapper around the MAVLink commands). What are your thoughts on this?

This is IMO the approach suggested in the doc - a gimbal manager can abstract old-style gimbal behaviour.

src/plugins/gimbal/CMakeLists.txt
```diff
@@ -1,6 +1,10 @@
 add_library(mavsdk_gimbal
     gimbal.cpp
     gimbal_impl.cpp
+    gimbal_manager.cpp
+    gimbal_manager_impl.cpp
+    gimbal_device.cpp
+    gimbal_device_impl.cpp
```
julianoes: I think these should be new plugins in a separate folder to follow the general plugin structure.

Also, from the naming, I would suggest this as the naming to avoid confusion:
- gimbal: includes the old gimbal protocol and the new gimbal protocol and uses the new one if the autopilot supports it).
- gimbal_device_tester: just for testing, not for normal usage.

src/plugins/gimbal/gimbal_device.cpp
```diff
@@ -0,0 +1,54 @@
...
+#include "gimbal_device_impl.h"
+
+namespace mavsdk {
+
+GimbalDevice::GimbalDevice(System& system) : PluginBase(), _impl{new GimbalDeviceImpl(system)} {}
+
+GimbalDevice::~GimbalDevice() {}
+
+GimbalDevice::Result GimbalDevice::set_attitude(int operation_flags,
+        float quat_w, float quat_x, float quat_y, float quat_z,
```
julianoes: Check how other plugins use `struct`s to pass data like a quaternion in. That many parameters to a function are confusing and dangerous, in my opinion.

src/plugins/gimbal/gimbal_manager.cpp
```diff
@@ -0,0 +1,69 @@
...
+}
+
+void GimbalManager::set_attitude_async(float tilt_rate, float pan_rate,
+        float tilt_deg, float pan_deg, int operation_flags,
+        uint8_t id, result_callback_t callback)
+{
+    _impl->set_attitude_async(tilt_rate, pan_rate, tilt_deg, pan_deg, operation_flags, id, callback);
+}
+
+GimbalManager::Result GimbalManager::track_point(float x, float y, uint8_t id)
```
julianoes: Let's hold off with this functionality. I would rather add features one by one as we can test them. We have no experience with tracking yet (and it would require a camera interacting with the gimbal).

src/plugins/gimbal/include/plugins/gimbal/gimbal_device.h
```diff
@@ -0,0 +1,177 @@
...
+        HAS_ROLL_AXIS = 4, /**< @brief Supports roll axis control. */
+        HAS_ROLL_FOLLOW = 8, /**< @brief Supports follow mode for roll axis. */
+        HAS_ROLL_LOCK = 16, /**< @brief Supports lock mode for roll axis. */
+        HAS_PITCH_AXIS = 32, /**< @brief Supports pitch axis control. */
+        HAS_PITCH_FOLLOW = 64, /**< @brief Supports follow mode for pitch axis. */
+        HAS_PITCH_LOCK = 128, /**< @brief Supports lock mode for pitch axis. */
+        HAS_YAW_AXIS = 256, /**< @brief Supports yaw axis control. */
+        HAS_YAW_FOLLOW = 512, /**< @brief Supports follow mode for yaw axis. */
+        HAS_YAW_LOCK = 1024, /**< @brief Supports lock mode for yaw axis. */
+        HAS_INFINITE_YAW = 2048 /**< @brief Supports infinite yaw. */
```
julianoes: I wonder if we should duplicate this, or if maybe the gimbal device tester could be implemented using an example which uses the passthrough plugin and uses just direct MAVLink.

src/plugins/gimbal/gimbal_device.cpp
coderkalyan: Okay, will do!

src/plugins/gimbal/CMakeLists.txt
coderkalyan: Got it, I'm going to do some refactoring and moving code around today

src/plugins/gimbal/gimbal_manager.cpp
coderkalyan: I'll just comment it out for now

coderkalyan:
Progress update: I've moved gimbal manager into its own plugin. The "dummy" `gimbal_manager_test` and `gimbal_device_test` apps are in progress, @bbworld1 is helping me on that too. I will push code in some time. I'm holding off on the `GimbalDevice` API for now, not sure if its necessary or not. Once GimbalManager is done/working to some extent I will refactor `Gimbal` to `GimbalLegacy` and reuse that class for the high level API. This API should automatically manage multiple gimbal managers, and also allow fallback to internally use the legacy class.

Note to self: Found some function in SystemImpl that checks the component ID against MAV_COMP_ID_GIMBAL, this should be investigated/updated/removed.

examples/gimbal_device_test/gimbal_device_test.cpp
```diff
@@ -0,0 +1,81 @@
...
+                  << std::endl;
+        return 1;
+    }
+
+    bool received_information = false;
+    MavlinkPassthrough pass(system);
+    pass.subscribe_message_async(MAVLINK_MSG_ID_GIMBAL_DEVICE_INFORMATION, 
+    //pass.subscribe_message_async(MAVLINK_MSG_ID_GLOBAL_POSITION_INT,
+            [&received_information, &gimbal_id](const mavlink_message_t message) {
+        std::cout << "Received" << sizeof(message) << gimbal_id << std::endl;
```
julianoes: Nice, this goes in the right direction.

examples/gimbal_manager_test/gimbal_manager_test.cpp
```diff
@@ -0,0 +1,129 @@
+//
+// Dummy gimbal manager implementation written using MavlinkPassthrough.
```
julianoes: Aha, so this is faking to be a gimbal manager? Interesting :thinking:

examples/gimbal_raw/gimbal_raw.cpp
```diff
@@ -0,0 +1,114 @@
+//
+// Connect to and control a gimbal manager, using the "raw" GimbalManager API
```
julianoes: I don't get this one. What do you mean with `raw`?

src/plugins/gimbal/include/plugins/gimbal/gimbal_device.h
```diff
@@ -0,0 +1,177 @@
...
+/**
+ * @brief the GimbalDevice class provides callbacks for listening to gimbal device
+ * commands. It also provides the ability to command a gimbal device, i. e. from
+ * a custom gimbal manager.
+ *
+ * Warning: This is used for low level gimbal device control. this API should
+ * only be used if you are 1) testing a gimbal manager or 2) writing a
+ * custom gimbal manager, i. e. for a ground station.
+ */
+class GimbalDevice : public PluginBase {
```
julianoes: I'm confused. Is this to test a gimbal device or to fake a gimbal device?
And I think it's wrong to have a plugin so something inheriting from `PluginBase` inside of the `gimbal` plugin.

coderkalyan: @julianoes It was supposed to test a gimbal device, but I'm holding off on it for now, thats why I haven't moved it to its own plugin/worked on it.

examples/gimbal_raw/gimbal_raw.cpp
coderkalyan: @julianoes I mean `raw` in the same way as `MissionRaw`, in the sense that the GimbalManager API is low level control, versus the (to be written) higher level `Gimbal` API.

examples/gimbal_device_test/gimbal_device_test.py
```diff
@@ -0,0 +1,167 @@
...
+gimbal device component with the MAVLinkPassthrough module.
+This example uses the libpilot MAVLink component library (https://gitlab.avhs.best/amadoruavs/pilot/pilot).
+
+Originally authored by: Vincent Wang (2020-03-30).
+"""
+import argparse
+import time
+from pilot import Component
+from pilot.util import MavlinkId
+from pymavlink import mavutil
```
julianoes, line 13: What? I don't think we want to depend on pymavlink with MAVSDK.

coderkalyan: @julianoes Definitely not, that was not the point. Sorry! We were simply trying to use pymavlink for the dummy gimbal manager/device as it was easier than MavlinkPassthrough in MAVSDK as well as we encountered some system ID issues. I will remove this before merge, or if you prefer I take it out of tree now, I'm happy to do so!

@bbworld1 might be better to just keep this out of tree

examples/gimbal_raw/gimbal_raw.cpp
julianoes: Ok, and what is that needed for?

examples/gimbal_device_test/gimbal_device_test.py
julianoes: I see.

coderkalyan:
@JonasVautherin I've paused work on this to fix a few annoying bugs related to routing in MAVSDK which were hindering the development of this plugin due to its communication between different MAVLnk nodes/components: see https://github.com/mavlink/MAVSDK/pull/1053

coderkalyan:
This has been implemented (mostly by @julianoes) in a collection of smaller PRs.

julianoes:
Sorry for duplicating some of the work @coderkalyan. This came up at work internally and I had to get some of the parts done quickly.

PR #1024, opened 2020-03-30
Update examples based on recent Action class renaming

coderkalyan:
When Action got changed with the autogeneration of gRPC bindings, etc, Action::Result enum values changed from uppercase to title case - this PR reflects those changes in the examples. It also fixes the mavlink_ftp_server example, which was broken when I implemented the new Configuration system a short while back.


coderkalyan:
I believe the conflict is actually due to the `example` directory being renamed to `examples` in develop vs master, or something like that?

JonasVautherin:
Thanks for the changes! Your branch is not up-to-date with `develop`, hence the conflics (probably). Please try to update your branch and add your changes on top of the latest state :blush:

coderkalyan:
@JonasVautherin should be good now!

PR #1029, opened 2020-04-04
mission: flatten plugin, remove MissionItem class

julianoes:
In order to prepare for the auto-generation of the C++ part of the mission plugin, it probably makes sense to flatten the mission plugin and make the API less complex. Instead of having this MissionItem class we now just use a simple struct.

The disadvantage is that we can't check the "settings" of a mission item at construction but only when it is used as part of a mission upload.

Also, it would be nice to generate the current NAN defaults for floats in a mission item.


julianoes:
@JonasVautherin this is WIP, it's missing the `mavsdk_server` part.

julianoes:
@JonasVautherin ok this should be ready for review. Let me know if you agree that it is the right direction to flatten the mission plugin like that.

JonasVautherin:
> Also, it would be nice to generate the current NAN defaults for floats in a mission item.

Yes! I had been looking at a way to "mark" some fields as optional in the proto file, while giving them a default value. I would need to spend some more time on it but I think there was room for improvement there!

src/integration_tests/mission_transfer_lossy.cpp
```diff
@@ -142,6 +139,7 @@ void compare_mission(
     }
 
     for (unsigned i = 0; i < a.size(); ++i) {
-        EXPECT_EQ(*a.at(i), *b.at(i));
+        // FIXME: enable again
```
JonasVautherin, line 142: Another FIXME, in case they need to be fixed in this PR

src/backend/test/mission_service_impl_test.cpp
```diff
@@ -267,7 +279,8 @@ void MissionServiceImplUploadTest::checkItemsAreUploadedCorrectly(
     ASSERT_EQ(mission_items.size(), _uploaded_mission.size());
 
     for (size_t i = 0; i < mission_items.size(); i++) {
-        EXPECT_EQ(*mission_items.at(i), *_uploaded_mission.at(i));
+        // FIXME: add this again
```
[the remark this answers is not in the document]
julianoes: I thought I would wait for the auto-generated `operator==` but then realized we had it already in the removed code, so I added it back.

src/plugins/mission/include/plugins/mission/mission.h
```diff
@@ -1,11 +1,11 @@
 #pragma once
 
+#include <cmath>
 #include <vector>
 #include <memory>
 #include <functional>
```
JonasVautherin, line 6: Regarding the Windows error, does it require this?

```suggestion
#include <functional>
#include <string>
```

PR #1032, opened 2020-04-08
[BUG] - Result::Success -> Result::SUCCESS

TheLegendaryJedi:
Variable name typo

### DIFF
--- a/examples/takeoff_land/takeoff_and_land.cpp
+++ b/examples/takeoff_land/takeoff_and_land.cpp
@@ -112,7 +112,7 @@ int main(int argc, char** argv)
     std::cout << "Arming..." << std::endl;
     const Action::Result arm_result = action->arm();
 
-    if (arm_result != Action::Result::Success) {
+    if (arm_result != Action::Result::SUCCESS) {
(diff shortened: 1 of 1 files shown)


TheLegendaryJedi:
I was just following the tutorial and got this error, I didn't check other branches, sorry.

JonasVautherin:
`develop` is the default because it is the main development branch, so that's where pull requests should be directed to :blush:.

I understand that it is not always intuitive that you should go check the examples that correspond to the version you are using, but I'd like to note that we don't always break the API (actually we very rarely do so). It just that these days, we are really making a lot of (small) changes :innocent:.

TheLegendaryJedi:
> `develop` is the default because it is the main development branch, so that's where pull requests should be directed to .
> 
> I understand that it is not always intuitive that you should go check the examples that correspond to the version you are using, but I'd like to note that we don't always break the API (actually we very rarely do so). It just that these days, we are really making a lot of (small) changes .

@JonasVautherin thanks a lot for the explanation :) I'm kind of new to this environment.

PR #1034, opened 2020-04-09
Don't respond to timesync until time known

### LINKED ISSUE #1033: Timesync adjusting outgoing message timestamps
The usual method for timesync with mavlink is that outgoing messages are sent with the sender's clock source, and the receiver should translate it into their own time frame. However, we've seen now with MAVSDK that after an initial timesync exchange it adjusts the clock source for _outgoing_ messages. This causes massive confusion on the flight controller side, since it was trying to adjust timestamps on incoming messages to what the clock source on MAVSDK was _before_ it did the timesync. This might be the root cause of #960 ? It can also cause instability in both timesync algorithms since they try to estimate each other when the offset is constantly changing, instead of estimating a single offset which should be mostly constant.

To fix this, outgoing messages should be timestamped with a consistent clock source (preferably CLOCK_MONOTONIC_RAW, ie something that isn't affected by time jumps or temporary clock slewing on the host system), and only incoming messages should be adjusted via the timesync message.

FYI @irsdkv , @JonasVautherin suggested I tag you since you contributed this. Do you see any problems here? Would you have time to fix this?

Thanks @DanielePettenuzzo for the investigation.

jkflying:
The timesync protocol expects clock offsets to be constant (or at least have constant skew), however the MAVSDK has an initial big change when syncing with the flight controller. This can confuse other systems on the mavlink network.

This prevents the MAVSDK from responding to other timesync messages until its own timesync with the autopilot time has been determined.

This is a ~~fix~~ cover up for #1033 . The real fix is to allow the flight controller, MAVROS, etc to have multiple different external systems timesynced at the same time.

FYI @irsdkv  @DanielePettenuzzo @JonasVautherin 

Fixes #1033.

### DIFF
--- a/src/core/timesync.cpp
+++ b/src/core/timesync.cpp
@@ -42,7 +42,7 @@ void Timesync::process_timesync(const mavlink_message_t& message)
                          _parent.get_autopilot_time().now().time_since_epoch())
                          .count();
 
-    if (timesync.tc1 == 0) {
+    if (timesync.tc1 == 0 && _autopilot_timesync_acquired) {
         // Send synced time to remote system
         send_timesync(now_ns, timesync.ts1);
     } else if (timesync.tc1 > 0) {
@@ -75,6 +75,7 @@ void Timesync::set_timesync_offset(int64_t offset_ns, uint64_t start_transfer_lo
 
         // Save time offset for other components to use
         _parent.get_autopilot_time().shift_time_by(std::chrono::nanoseconds(offset_ns));
+        _autopilot_timesync_acquired = true;
 
         // Reset high RTT count after filter update
         _high_rtt_count = 0;
--- a/src/core/timesync.cpp
+++ b/src/core/timesync.cpp
@@ -27,6 +27,8 @@ void Timesync::do_work()
                                   _parent.get_autopilot_time().now().time_since_epoch())
                                   .count();
             send_timesync(0, now_ns);
+        } else {
+            _autopilot_timesync_acquired = false;
         }
         _last_time = _parent.get_time().steady_time();
     }
--- a/src/core/timesync.h
+++ b/src/core/timesync.h
@@ -30,5 +30,6 @@ private:
     static constexpr uint64_t _MAX_CONS_HIGH_RTT = 5;
     static constexpr uint64_t _MAX_RTT_SAMPLE_MS = 10;
     uint64_t _high_rtt_count{};
+    bool _autopilot_timesync_acquired = false;
 };
 } // namespace mavsdk


DanielePettenuzzo:
Cool this makes sense. I was having a time jump almost on every boot since px4 sends timesync messages at 10Hz and the MAVSDK at much slower rate. Therefore the first PX4 timesyncs have the system time from the sdk and then it jumped to the px4 time.

I will test this later today

src/core/timesync.cpp
```diff
@@ -42,7 +42,7 @@ void Timesync::process_timesync(const mavlink_message_t& message)
                          _parent.get_autopilot_time().now().time_since_epoch())
                          .count();
 
-    if (timesync.tc1 == 0) {
+    if (timesync.tc1 == 0 && _timesync_locked) {
```
irsdkv, line 45: May be
```suggestion
    if (timesync.tc1 == 0 && !_timesync_locked) {
```
?

src/core/timesync.cpp
```diff
@@ -75,6 +75,7 @@ void Timesync::set_timesync_offset(int64_t offset_ns, uint64_t start_transfer_lo
 
         // Save time offset for other components to use
         _parent.get_autopilot_time().shift_time_by(std::chrono::nanoseconds(offset_ns));
+        _timesync_locked = true;
```
irsdkv, line 78: May be need to unlock timesync if we have no response?

```
if (_high_rtt_count > _MAX_CONS_HIGH_RTT) {
    // Issue a warning to the user if the RTT is constantly high
    LogWarn() << "RTT too high for timesync: " << rtt_ns / 1000000.0 << " ms.";
    // Reset counter
     _high_rtt_count = 0;
     _timesync_locked = false;
}
```
https://github.com/mavlink/MAVSDK/pull/1034/files#diff-ec4205b6bc1adbc844894d870f234adaL85-L91

src/core/timesync.cpp
DanielePettenuzzo: This looks right to me. We don't want to respond to timesyncs until our time is not updated with the autopilot time. Otherwise get_autopilottime will return system time at first.

irsdkv: Okay, I understand now, thank you

irsdkv:
Should MAVSDK also stop to send timesync requests if this requests received from the other part?

https://github.com/mavlink/MAVSDK/blob/899ff7ab830cbcb404a14f21164ae991393b1a78/src/core/timesync.cpp#L22-L33

src/core/timesync.cpp
jkflying: No, this is correct. We don't want the remote system to do a timesync until our own clock is stable.

jkflying: Ah, I think it is the naming of the variable that isn't clear. I will make it more obvious.

jkflying:
> Should MAVSDK also stop to send timesync requests if this requests received from the other part?

No, this part is what we need so that MAVSDK can align with the autopilot. The part that is causing a problem is responding to the autopilot timesync before we have competed our own timesync, because now our clock source isn't constant.

src/core/timesync.cpp
jkflying: If we only have high RTT, I'm not sure what the best case is. I'd prefer to not confuse the flight controller if the clock on MAVSDK isn't stable...

irsdkv: I worry about case of MAVSDK will reconnect to FC during one session. Sometimes it can be a different FCs or FC with changed behaviour.

irsdkv: So may be in this place we should reset a flag:
```C++
 if (_parent.is_connected()) {
    uint64_t now_ns = std::chrono::duration_cast<std::chrono::nanoseconds>(
                          _parent.get_autopilot_time().now().time_since_epoch())
                          .count();
    send_timesync(0, now_ns);
} else {
    _autopilot_timesync_acquired = false;
}
```
https://github.com/mavlink/MAVSDK/blob/899ff7ab830cbcb404a14f21164ae991393b1a78/src/core/timesync.cpp#L22-L33

jkflying: Ok yes, good idea. I've added that.

JonasVautherin:
Thanks!

PR #1036, opened 2020-04-13
Autogenerate mission plugin

julianoes:
This converts the mission plugin to be auto-generated from the proto files.

Depends on https://github.com/mavlink/MAVSDK-Proto/pull/123


src/plugins/telemetry/include/plugins/telemetry/telemetry.h
```diff
@@ -458,9 +471,9 @@ class Telemetry : public PluginBase {
      * @brief Velocity type, represented in the Body (X Y Z) frame and in metres/second.
      */
     struct VelocityBody {
-        float x_m_s; /**< @brief Velocity in X in metres/second */
-        float y_m_s; /**< @brief Velocity in Y in metres/second */
-        float z_m_s; /**< @brief Velocity in Z in metres/second */
+        float x_m_s{}; /**< @brief Velocity in X in metres/second */
```
JonasVautherin, line 474: Seems like we're missing a few defaults there.

julianoes:
Something with the integration test is not matching yet:
```
/__w/MAVSDK/MAVSDK/src/integration_tests/mission.cpp:305: Failure
Value of: original == downloaded
  Actual: false
Expected: true
```

PR #1038, opened 2020-04-14
Added example to debug if all params arrive

julianoes:
This was just used for debugging. It's probably not worth merging.


julianoes:
@JonasVautherin yes we should add it internally at some point, or in the param plugin.

JonasVautherin:
I think in the param plugin. Whether we do some caching or not is an implementation detail, but in terms of API I believe this belongs to `param`. Or do you think it is needed internally? Or is it the case that other plugins would need to access the param plugin? We don't have any plugin depending on another plugin yet, but I don't see why that would be an issue...

PR #1042, opened 2020-04-16
Auto-generation of camera plugin

julianoes:
This adds the auto-generation of the camera plugin.


src/backend/src/plugins/telemetry/telemetry_service_impl.h
```diff
@@ -1115,7 +1115,7 @@ class TelemetryServiceImpl final : public rpc::telemetry::TelemetryService::Serv
 
         _telemetry.home_async(
             [this, &writer, &stream_closed_promise, is_finished, &subscribe_mutex](
-                const mavsdk::Telemetry::Position& home) {
+                const mavsdk::Telemetry::Position home) {
```
JonasVautherin, line 1118: Do we want to remove that `&`, or is that a mistake in the template?

julianoes: Hm, this probably happened by mistake. In general I would move to pass by value / copy as much as possible and then profile and optimize if needed.

julianoes:
The CI error is the usual dead lock on destruction as addressed in #1011.

src/backend/src/plugins/telemetry/telemetry_service_impl.h
JonasVautherin: Ok

JonasVautherin:
Haha not waiting for the iOS test? :sweat_smile:

PR #1053, opened 2020-04-23
Routing Step 1: Node Class

coderkalyan:
This is step one of my plans to clean up node handling, routing, UDP, etc in MAVSDK. This PR deprecates the System class in favor of the Node class, which is similar in implementation but designed for connecting to single MAVLink enabled nodes, not entire systems. This allows for more flexibility with more advanced uses of MAVSDK, connecting to multiple systems/components, etc.

Currently a draft, but @julianoes feel free to take a look whenever you want. It successfully detects other MAVLink nodes, but I'm working on integrating the system functions required to make it possible to port the plugins to Node.

I'm still thinking of the cleanest way to maintain backwards compatibility with System, @julianoes @JonasVautherin if you have any ideas let me know!


examples/node_test/node_test.cpp
```diff
@@ -0,0 +1,162 @@
+//
+// Simple example to test out multiple nodes
```
julianoes, line 2: For tests I suggest to use integration_tests instead of examples.

src/core/mavsdk.h
```diff
@@ -233,7 +233,7 @@ class Mavsdk {
      *
      * @param uuid UUID of system (or MAVLink system ID for systems that don't have a UUID).
      */
-    typedef std::function<void(uint64_t uuid)> event_callback_t;
+    typedef std::function<void(uint8_t system_id, uint8_t component_id)> event_callback_t;
```
julianoes, line 236: That sort of change is not backwards compatible. We would like to add new types, APIs but not change existing ones.

src/core/node.h
```diff
@@ -0,0 +1,77 @@
...
+
+    /**
+     * MavsdkImpl and PluginImplBase need access to NodeImpl class.
+     * This is not pretty but it's not easy to hide the methods from library
+     * users if not like that (or with an ugly reinterpret_cast).
+     */
+    friend MavsdkImpl;
+    friend PluginImplBase;
+
+    std::shared_ptr<NodeImpl> _node_impl;
```
julianoes, line 74: This should be a `std::unique_ptr`, I'd say.

src/core/mavsdk_impl.h
```diff
@@ -85,6 +96,17 @@ class MavsdkImpl {
...
 
     std::atomic<bool> _should_exit = {false};
+
+    void system_thread();
+    void send_heartbeat();
+
+    std::thread* _system_thread{nullptr};
+
+    /** @brief Interval in milliseconds to send heartbeat. */
+    // TODO kalyan - make this configurable??
```
julianoes, line 106: My answer is usually: no, let's leave it until someone asks.

src/core/mavsdk_impl.cpp
```diff
@@ -394,45 +453,84 @@ bool MavsdkImpl::does_system_exist(uint8_t system_id)
...
+        }
+        last_time = _time.steady_time();
+
+        std::this_thread::sleep_for(std::chrono::milliseconds(50)); // in case we want to do anything else in this thread
+    }
+}
+
+void MavsdkImpl::send_heartbeat()
+{
+    std::cout << "Sending heartbeat" << std::endl;
```
julianoes, line 521: You should always use `LogDebug()` instead of `std::cout`.

src/core/node_impl.h
```diff
@@ -0,0 +1,135 @@
...
+    void unregister_all_mavlink_message_handlers(const void* cookie);
+
+    void register_timeout_handler(std::function<void()> callback, double duration_s, void** cookie);
+    void refresh_timeout_handler(const void* cookie);
+    void unregister_timeout_handler(const void* cookie);
+
+    void add_call_every(std::function<void()> callback, float interval_s, void** cookie);
+    void change_call_every(float interval_s, const void* cookie);
+    void reset_call_every(const void* cookie);
+    void remove_call_every(const void* cookie);
```
julianoes, line 55: I'm wondering if all these API calls should be exposed by each node or where they fit it in :thinking:.

examples/node_test/node_test.cpp
coderkalyan: @julianoes Will do eventually, right now node_test is just my quick way of testing new code, I'll port it over before merge.

src/core/mavsdk_impl.cpp
coderkalyan: Done!

examples/node_test/node_test.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+void component_discovered(ComponentType component_type)
+{
+    std::cout << NORMAL_CONSOLE_TEXT << "Discovered a component with type "
+              << unsigned(component_type) << std::endl;
+}
+
+int main(int argc, char** argv)
+{
+    Mavsdk dc;
```
JonasVautherin, line 40: We should start calling this `mavsdk_server` instead of `dc` :sweat_smile:. `dc` is legacy from when the project was called "DroneCore" :laughing:

src/core/mavsdk_impl.cpp
```diff
@@ -73,49 +80,63 @@ std::string MavsdkImpl::version() const
...
         return;
     }
 
-    if (_systems.find(message.sysid) != _systems.end()) {
-        _systems.at(message.sysid)->system_impl()->process_mavlink_message(message);
+    //if (_systems.find(message.sysid) != _systems.end()) {
+        //_systems.at(message.sysid)->system_impl()->process_mavlink_message(message);
+    //}
+
+    uint16_t node_id = (message.sysid << 8) | message.compid;
```
JonasVautherin, line 137: Suggestion: could we put that in a function called something like `extract_node_id(message)`? This way it makes it easier to read, as the function hides the implementation detail (i.e. these `<<` and `|`).

src/core/mavsdk_impl.cpp
```diff
@@ -133,6 +154,19 @@ bool MavsdkImpl::send_message(mavlink_message_t& message)
     return true;
 }
 
+void MavsdkImpl::process_heartbeat(const mavlink_message_t& message)
+{
+    std::lock_guard<std::mutex> lock(_connections_mutex);
+
+    uint16_t node_id = (message.sysid << 8) | message.compid;
```
JonasVautherin, line 161: Oh, and this would be `extract_node_id(message)` again :blush:

src/core/mavsdk_impl.cpp
```diff
@@ -133,6 +154,19 @@ bool MavsdkImpl::send_message(mavlink_message_t& message)
...
+    std::cout << "HB Sysid: " << int(message.sysid) << " compid: " << int(message.compid) << std::endl;
```
JonasVautherin, line 162: We don't `cout`, we log at some level. `LogErr() <<`,  `LogInfo() <<`, `LogDebug() <<`, ...

src/core/mavsdk_impl.cpp
```diff
@@ -73,49 +80,63 @@ std::string MavsdkImpl::version() const
...
     }
 
-    if (_systems.find(message.sysid) != _systems.end()) {
-        _systems.at(message.sysid)->system_impl()->process_mavlink_message(message);
+    //if (_systems.find(message.sysid) != _systems.end()) {
+        //_systems.at(message.sysid)->system_impl()->process_mavlink_message(message);
+    //}
+
+    uint16_t node_id = (message.sysid << 8) | message.compid;
+    if (_nodes.find(node_id) != _nodes.end()) {
```
JonasVautherin, line 138: I would make this `if (node_exists(node_id)) {`, because I find that more readable than the `!= _nodes.end()` test. Later in the code, it is used again, and it could be `if (!node_exists(node_id))`. Again, just a suggestion.

src/core/mavsdk_impl.h
JonasVautherin: And that could go in a smaller, separate PR. So I'd argue that it is not in the scope of this PR anyway.

src/core/node.h
```diff
@@ -0,0 +1,77 @@
...
+public:
+    /** @private Constructor, used internally
+     *
+     * This constructor is not (and should not be) directly called by application code.
+     *
+     * @param parent `MavsdkImpl` dependency.
+     * @param system_id System id.
+     * @param comp_id Component id.
+     */
+    explicit Node(MavsdkImpl& parent, uint8_t system_id, uint8_t comp_id);
```
JonasVautherin, line 30: I'm not completely clear about the `node_id` used elsewhere in the code. Why is there a concept of `node_id`, but this concept does not exist inside the `Node` class? :thinking:

julianoes:
@JonasVautherin some of the docs for this is in https://docs.google.com/document/d/1UKXxG3Ph_9Nqws4hybiA0PPDo6L2PUYLB7mp_EwvI54/edit#.

JonasVautherin:
Nice document! @coderkalyan: I added comments directly there: https://docs.google.com/document/d/1UKXxG3Ph_9Nqws4hybiA0PPDo6L2PUYLB7mp_EwvI54/edit#. I would be interested in having your opinion.

julianoes:
@coderkalyan I don't know what you checked in but there is 4000 files in the diff now which makes it impossible to review or see what's going on. Please clean up.

JonasVautherin:
> +1,216,455 −114

:sweat_smile:

src/core/mavsdk_impl.cpp
coderkalyan: Sure, will do!

examples/node_test/node_test.cpp
coderkalyan: Oh, I was curious where that came from! I'll change it :)

src/core/mavsdk_impl.cpp
coderkalyan: Got it

coderkalyan:
@JonasVautherin No, I don't use a GUI tool for git. However I am very clumsy when it comes to contributing since my git workflow on private projects is usually more relaxed :) It is also possible that my friend @bbworld1 accidently messed something up, I'll take a look at the history!

src/core/node.h
coderkalyan: @JonasVautherin node_id is a single 16 bit integer, which is easier to manipulate than 2x8bit (sid & cid) since I just bitshift and bitwise OR it. It doesn't make sense to me that the user would ever want to use this, however, so I haven't exposed it.

coderkalyan: For instance in a map, it's easy to do map<uint16_t, whatever> as opposed to map<pair<uint8_t, uint8_t>, whatever>

julianoes:
@coderkalyan let's close this PR and work on smaller iterations in the future, ok?

PR #1054, opened 2020-04-23
shell: autogenerate plugin from proto

julianoes:
This changes the API because the existing API with sync and async command requests did not work for commands with continuous outputs like top. Therefore we changed this to a simple async stream of output plus a sync request to send a command line.

Depends on https://github.com/mavlink/MAVSDK-Proto/pull/136.


julianoes:
CI errors must be unrelated.

julianoes:
@irsdkv 

> It is hard to me to understand all changes in context of all MAVSDK improvements

Right, the changes are pretty hidden within all the noise. Sorry about that.
The best would probably be if you just check the new C++ API and whether it makes sense.

I did test it using the example against a Pixhawk, so I'm not too worried.

JonasVautherin:
I commented the API in https://github.com/mavlink/MAVSDK-Proto/pull/136, interested in having your opinion there :blush:

julianoes:
I'm gonna skip iOS and Windows CI and go ahead with merging.

PR #1057, opened 2020-04-23
Autogenerate MissionRaw from proto

julianoes:
This also cleans up the internals of mission raw quite a bit as it had recently been changed and some things were not quite as I wanted them to be.

@Katawann it would be good if you could review my changes in `mavsdk_raw_impl.cpp` if they look somewhat sane to you. I haven't tested this yet because there are no tests, so it is what it is :smile:.

Depends on https://github.com/mavlink/MAVSDK-Proto/pull/137.


tools/generate_from_protos.sh
```diff
@@ -10,6 +10,11 @@ third_party_dir="${script_dir}/../build/default/third_party"
 protoc_binary="${third_party_dir}/install/bin/protoc"
 protoc_grpc_binary="${third_party_dir}/install/bin/grpc_cpp_plugin"
 
+function snake_case_to_camel_case {
```
JonasVautherin, line 13: I thought I had changed `followme` to `follow_me`, but then I would have had to add something like this :thinking:. Weird :laughing:

julianoes: I'm confused because `follow_me` is not used in this script yet.

JonasVautherin: Yes, maybe I did that in the `proto` submodule actually, for Java. So I'm confused :laughing:

Katawann:
@julianoes I don't know mission reached/current logic good enough to say if the `_mission_progress.last.current = mission_item_reached.seq + 1` of the `process_mission_item_reached` function is okay or if user could run into some limitations according to the mission they could generate (with takeoff, RTL and so on). For the rest it looks good to me as well

julianoes:
@Katawann right, we need to check that. Best to create an integration test and specify the needed functionality and then check it.

PR #1059, opened 2020-04-24
mavlink_ftp: remove crc calls, add compare call

julianoes:
In my view it makes more sense to have a function to compare a local to a remote file instead of exposing the internal crc32 checking.

@MatejFranceskin let me know what you think.


julianoes:
@JonasVautherin odd mavsdk_server unit test fail.

JonasVautherin:
Seems like there is a race somewhere in the test... to be investigated

julianoes:
Now the mavsdk_server unit test passed but something else was flaky, I'm gonna ignore it :smile:.

MatejFranceskin:
I agree that to check that file was correctly transferred we can compare files and don't have to do CRCs. but I think that most of this was introduced also for the purpose of testing MavlinkFTP CRC method.

PR #1062, opened 2020-04-27
Autogen info

JonasVautherin:


src/plugins/info/info_impl.cpp
```diff
@@ -167,63 +157,63 @@ void InfoImpl::process_flight_information(const mavlink_message_t& message)
     _flight_information_received = true;
 }
 
-void InfoImpl::translate_binary_to_str(
-    uint8_t* binary, unsigned binary_len, char* str, unsigned str_len)
+std::string InfoImpl::translate_binary_to_str(uint8_t* binary, unsigned binary_len)
 {
+    std::string str(binary_len * 2, '0');
```
JonasVautherin, line 162: @julianoes: Does this length make sense to you? Seems to be working, but... :laughing:

src/plugins/info/info_impl.cpp
```diff
@@ -167,63 +157,63 @@ void InfoImpl::process_flight_information(const mavlink_message_t& message)
...
         _flight_info);
 }
 
-const char* InfoImpl::vendor_id_str(uint16_t vendor_id)
+const std::string InfoImpl::vendor_id_str(uint16_t vendor_id)
 {
     switch (vendor_id) {
         case 0x26ac:
-            return "Yuneec";
+            return "3D Robotics Inc.";
```
JonasVautherin, line 210: @julianoes: worth your attention, too. Note that I did not change "H520" below (line 219)

src/plugins/info/info_impl.cpp
julianoes: So `binary_len` is the number of bytes from `sizeof` but doesn't each byte need 2 hex chars?

src/plugins/info/info_impl.cpp
```diff
@@ -167,63 +157,63 @@ void InfoImpl::process_flight_information(const mavlink_message_t& message)
...
-    uint8_t* binary, unsigned binary_len, char* str, unsigned str_len)
+std::string InfoImpl::translate_binary_to_str(uint8_t* binary, unsigned binary_len)
 {
+    std::string str(binary_len * 2, '0');
+
     for (unsigned i = 0; i < binary_len; ++i) {
         // One hex number occupies 2 chars.
         // The binary is in little endian, therefore we need to swap the bytes for us to read.
-        snprintf(&str[i * 2], str_len - i * 2, "%02x", binary[binary_len - 1 - i]);
+        snprintf(&str[i * 2], str.length() - i * 2, "%02x", binary[binary_len - 1 - i]);
```
julianoes, line 167: Now that I think about it I'm not sure about little endian vs. big endian here. Because for the `uid2` it's basically just an array of bytes while for `uid` it might have had a certain endian type because it was a `uint64_t`.

JonasVautherin:
> I think you could remove the endian swap but not 100% sure.

With the swap, the git hashes are consistent with what QGC shows. However, the uid2 is not consistent with QGC. So I'm not sure which one to choose :sweat_smile:.

src/plugins/info/info_impl.cpp
JonasVautherin: The previous implementation was swapping for git hashes, but not for uid2. I thought it was not consistent, but I messed up probably. I'll revert that.

julianoes:
Coverage error seems unrelated.

julianoes:
Alright, not waiting for iOS.

PR #1064, opened 2020-04-29
Auto-generate the Ftp plugin from proto

julianoes:
This is all new. The Ftp plugin was not in proto at all before.


julianoes:
I think the CI error is unrelated.

PR #1066, opened 2020-04-29
Autogen tune

JonasVautherin:
Tested to work on a pixhawk with master (flashed today from QGC).


src/plugins/tune/tune_impl.cpp
```diff
@@ -113,38 +114,29 @@ void TuneImpl::play_tune_async(
...
-    std::string tune2_str;
-
-    if (tune_str.size() < 30) {
-        tune1_str = tune_str;
-        tune2_str = "";
-    } else {
-        tune1_str = tune_str.substr(0, 30);
-        tune2_str = tune_str.substr(30);
-    }
+    const uint32_t TUNE_FORMAT_QBASIC1_1 = 1;
```
JonasVautherin, line 122: @julianoes: Is that the right way to do it, or is there a mavlink-generated enum I missed?

julianoes: Yes, there is this one: https://github.com/mavlink/c_library_v2/blob/45d924df376c00e96d40a567c94f111ffc3b9bc1/common/common.h#L1889

src/plugins/tune/tune_impl.cpp
```diff
@@ -113,38 +114,29 @@ void TuneImpl::play_tune_async(
 
     LogDebug() << "About to send tune: " << tune_str;
 
-    if (tune_str.size() > 230) {
-        report_tune_result(callback, Tune::Result::TUNE_TOO_LONG);
+    if (tune_str.size() > 248) {
```
julianoes, line 117: This could be `MAVLINK_MSG_PLAY_TUNE_V2_FIELD_TUNE_LEN`

JonasVautherin: Thanks!

src/plugins/tune/tune_impl.cpp
```diff
@@ -113,38 +114,27 @@ void TuneImpl::play_tune_async(
 
     LogDebug() << "About to send tune: " << tune_str;
 
-    if (tune_str.size() > 230) {
-        report_tune_result(callback, Tune::Result::TUNE_TOO_LONG);
+    if (tune_str.size() > MAVLINK_MSG_PLAY_TUNE_V2_FIELD_TUNE_LEN) {
```
julianoes, line 117: Actually, the size can only be `MAVLINK_MSG_PLAY_TUNE_V2_FIELD_TUNE_LEN - 1` because we need one byte for 0 termination, right?

JonasVautherin: Haha I think we already had that discussion last October :rofl:. See here: https://github.com/mavlink/MAVSDK/pull/879#discussion_r336982971

julianoes: Ok but MAVSDK should adhere to the MAVLink spec which says there needs to be 0 termination within the 248 chars, right?
https://mavlink.io/en/messages/common.html#PLAY_TUNE_V2

JonasVautherin: Ok!

PR #1068, opened 2020-05-04
Improve cpp api

JonasVautherin:
C++ counterpart to https://github.com/mavlink/MAVSDK-Proto/pull/145


JonasVautherin:
Ignoring iOS simulator, which failed for some reason.

PR #1076, opened 2020-05-05
Change callback type name

julianoes:
This changes all callbacks from:

```
typedef std::function<void(int, double)> foo_callback_t;
```

to:
```
using FooCallback = std::function<void(int, double)>;
```
which is more in line with modern C++ rather than old style C typedefs.

I'm not sure it is worth the churn though. However, the time to do it is probably now because we changed most callback types already through the auto-generation.

### DIFF
--- a/examples/offboard_velocity/offboard_velocity.cpp
+++ b/examples/offboard_velocity/offboard_velocity.cpp
@@ -260,7 +260,7 @@ void usage(std::string bin_name)
               << "For example, to connect to the simulator use URL: udp://:14540" << std::endl;
 }
 
-Telemetry::landed_state_callback_t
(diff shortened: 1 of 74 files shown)


julianoes:
FYI @hamishwillee. I need to check the docs script to update it so it can handle the `using` instead of `typedef`.

hamishwillee:
> FYI @hamishwillee: I need to check the docs script to update it so it can handle the using instead of typedef.

OK. Lot of churn right now - when do you think I should do a develop docs update?

Note, full on for other product docs right now, so this is just trying to establish when I should commit time to the job.

julianoes:
CI error looks unrelated, merging.

PR #1077, opened 2020-05-05
fix to include the last byte in the uid2 in the hardware_uid string

DanielePettenuzzo:
I had this issue that when I update to the latest mavsdk the last character in my uid2 was not included in the hardware_uid in the info plugin. This is because snprintf doesn n-1 in the length argument. 

FYI @JonasVautherin

### DIFF
--- a/src/plugins/info/info_impl.cpp
+++ b/src/plugins/info/info_impl.cpp
@@ -172,12 +172,13 @@ std::string InfoImpl::swap_and_translate_binary_to_str(uint8_t* binary, unsigned
 
 std::string InfoImpl::translate_binary_to_str(uint8_t* binary, unsigned binary_len)
 {
-    std::string str(binary_len * 2, '0');
+    std::string str(binary_len * 2 + 1, '0');
 
     for (unsigned i = 0; i < binary_len; ++i) {
         // One hex number occupies 2 chars.
-        snprintf(&str[i * 2], str.length() - i * 2 + 2, "%02x", binary[i]);
+        snprintf(&str[i * 2], str.length() - i * 2, "%02x", binary[i]);
     }
+    snprintf(&str[binary_len * 2], str.length() - binary_len * 2 - 1, "%x", '\0');
 
     return str;
 }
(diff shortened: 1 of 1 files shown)


DanielePettenuzzo:
My autopilot uuid: 000100000000313238313138511400360023
What I saw in the sdk before this pr: 00010000000031323831313851140036002

DanielePettenuzzo:
I pushed a new commit. You mean something like this?

julianoes:
> You mean something like this?

I think so, yes.

DanielePettenuzzo:
I will test it again this morning

DanielePettenuzzo:
I tested this with a few different pixhawk's and I always get he right uuid. I was wondering, is it enough to increase the size by 1 and std::string will add the null terminator or do I need to add the null terminator like in this pr?

JonasVautherin:
> I was wondering, is it enough to increase the size by 1 and std::string will add the null terminator or do I need to add the null terminator like in this pr?

I would think that `std::string` doesn't require the `\0`. It will be added when you call `c_str()` or `data()`. But well, I'm the one who screwed up the `+ 1` here, so my feedback is to take with a grain of salt :see_no_evil:.

julianoes:
This looks correct now in my testing.

src/plugins/info/info_impl.cpp
```diff
@@ -172,12 +172,13 @@ std::string InfoImpl::swap_and_translate_binary_to_str(uint8_t* binary, unsigned
...
 std::string InfoImpl::translate_binary_to_str(uint8_t* binary, unsigned binary_len)
 {
-    std::string str(binary_len * 2, '0');
+    std::string str(binary_len * 2 + 1, '0');
 
     for (unsigned i = 0; i < binary_len; ++i) {
         // One hex number occupies 2 chars.
         snprintf(&str[i * 2], str.length() - i * 2, "%02x", binary[i]);
     }
+    snprintf(&str[binary_len * 2], str.length() - binary_len * 2 - 1, "%x", '\0');
```
JonasVautherin, line 181:
```suggestion
```

I'm pretty sure it's superfluous to add `\0` at the end of an `std::string` (which is what we return here). As mentioned, `\0` is not part of an `std::string`, and it is just added when doing the `c_str()` conversion.

PR #1078, opened 2020-05-06
Build examples in CI

### LINKED ISSUE #677: Build examples in CI
This got lost when I removed the Jenkinsfile.

### LINKED ISSUE #1027: Add examples to CI?
I was thinking, as there is often minor refactoring (such as @JonasVautherin 's recent Action gRPC port), we should check if all the official MAVSDK examples at least compile. A lot of people use them to get started, and as our documentation is pretty much just the docstrings + examples, it would be good to make sure that they actually build :). What do you guys think?

Even better would be to have integration tests which test that the examples do what they are supposed to do, but this would be harder.

julianoes:
This adds building the examples to CI. Fixes #1027, fixes #677.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -65,6 +65,10 @@ jobs:
         run: cmake --build build/release -j 2
       - name: install
         run: sudo cmake --build build/release --target install
+      - name: configure examples
+        run: (cd examples && cmake -DCMAKE_BUILD_TYPE=Release -j 2 -Bbuild -H.)
+      - name: build examples
+        run: (cd examples && cmake --build build -j 2)
       - name: test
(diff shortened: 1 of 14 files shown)


.github/workflows/main.yml
```diff
@@ -63,6 +63,12 @@ jobs:
         run: cmake -DCMAKE_BUILD_TYPE=Release -DSUPERBUILD=ON -DBUILD_BACKEND=ON -j 2 -Bbuild/release -H.
```
JonasVautherin, line 63: This will require the MAVLINK_PASSTHROUGH for building the examples

julianoes: Good point!

PR #1080, opened 2020-05-06
Add CI check for uncommitted proto changes, improved result descriptions

julianoes:
First we check if CI fails applying https://github.com/mavlink/MAVSDK-Proto/pull/148


julianoes:
The proto checks seems to work :).

julianoes:
> Seems like it's working, except for the docs that was changed in the proto files

I wanted it to fail to test, so now I can fix it.

PR #1083, opened 2020-05-07
Minor docs fixes

julianoes:
It doesn't look like I actually have to change anything for the "new" typedefs to work.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,5 @@ CMakeLists.txt.user*
 
 # windows swap files
 *~
+
+tools/docs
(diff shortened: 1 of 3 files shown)


hamishwillee:
This is not the bit that will fail :-). The bit that will cause problems is generating markdown from doxygen xml.

julianoes:
@hamishwillee I looked at the markdown output and it looked ok. Of course, we'll still have to actually check the links, etc.

hamishwillee:
Great! We can check links as a post process even. I'm OK for you to merge this if the check failed are OK

PR #1084, opened 2020-05-07
Several API cleanups after auto-generation

julianoes:
- I think we don't need the `result_str` API anymore and can clean that up.
- Merge `SpeedNed` into `VelocityNed`.

Depends on https://github.com/mavlink/MAVSDK-Proto/pull/149.


julianoes:
A couple more similar fixes are coming.

PR #1085, opened 2020-05-07
Improve output in enum operator<<

JonasVautherin:


JonasVautherin:
Hmm I really don't get why this would fail the "proto check" :thinking:

julianoes:
I think it's the same problem as in https://github.com/mavlink/MAVSDK/pull/1083/checks?check_run_id=651964551. I need to investigate it.

julianoes:
It might be because I updated grpc and thus protoc interleaved somewhere.

PR #1090, opened 2020-05-12
cmake don't use undocumented `-H` arg

julianoes:
`-H` is an undocumented cmake flag that we should not rely upon.

See: https://stackoverflow.com/a/54279862/8548472


JonasVautherin:
`-S` doesn't work on some platforms (like Ubuntu 16), I'm guessing because of the cmake version. Hence the use of `-H`. Curious to see what the CI says for iOS now, maybe that's not an issue anymore :slightly_smiling_face:

julianoes:
It does pass Ubuntu 16.04 in the test but that's only because I forgot to update the flag there. Ok nevermind, let's leave it at `-H`.

PR #1095, opened 2020-05-12
third_party: build absl separate as package

julianoes:
This is needed because otherwise gRPC doesn't install correctly otherwise.
The underlying issue is that gRPC dependencies don't work as "module" if the cmake version is lower than 3.13, so they need to be marked as "package" and added separately, as we already do it for a couple of other dependencies.


julianoes:
Well, I do not understand why absl doesn't build anymore, that's a mystery.

PR #1096, opened 2020-05-12
calibration: add generated from proto board level horizon calibration

TSC21:
Adds level horizon calibration to the backend.

### DIFF
--- a/src/plugins/calibration/calibration.cpp
+++ b/src/plugins/calibration/calibration.cpp
@@ -30,6 +30,11 @@ void Calibration::calibrate_magnetometer_async(CalibrateMagnetometerCallback cal
     _impl->calibrate_magnetometer_async(callback);
 }
 
+void Calibration::calibrate_level_horizon_async(CalibrateLevelHorizonCallback callback)
+{
+    _impl->calibrate_level_horizon_async(callback);
+}
+
 void Calibration::calibrate_gimbal_accelerometer_async(
     CalibrateGimbalAccelerometerCallback callback)
 {
--- a/src/plugins/calibration/calibration.cpp
+++ b/src/plugins/calibration/calibration.cpp
@@ -69,6 +69,8 @@ std::ostream& operator<<(std::ostream& str, Calibration::Result const& result)
             return str << "Timeout";
         case Calibration::Result::Cancelled:
             return str << "Cancelled";
+        case Calibration::Result::FailedArmed:
+            return str << "Failed Armed";
         default:
             return str << "Unknown";
     }
--- a/src/plugins/calibration/calibration_impl.cpp
+++ b/src/plugins/calibration/calibration_impl.cpp
@@ -41,7 +41,8 @@ void CalibrationImpl::calibrate_gyro_async(const CalibrationCallback& callback)
     std::lock_guard<std::mutex> lock(_calibration_mutex);
 
     if (_parent->is_armed()) {
-        report_failed("System is armed.");
+        Calibration::ProgressData progress_data;
+        call_user_callback(callback, Calibration::Result::FailedArmed, progress_data);
         return;
     }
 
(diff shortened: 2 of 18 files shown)


src/plugins/calibration/calibration.cpp
```diff
@@ -30,6 +30,11 @@ void Calibration::calibrate_magnetometer_async(CalibrateMagnetometerCallback cal
     _impl->calibrate_magnetometer_async(callback);
 }
 
+void Calibration::calibrate_levelhorizon_async(CalibrateLevelHorizonCallback callback)
```
JonasVautherin, line 33: Did that get generated like this? I think it should be `calibrate_level_horizon_async` :thinking:

src/plugins/calibration/include/plugins/calibration/calibration.h
```diff
@@ -135,6 +135,17 @@ class Calibration : public PluginBase {
...
+    /**
+     * @brief Callback type for calibrate_levelhorizon_async.
+     */
+
+    using CalibrateLevelHorizonCallback = std::function<void(Calibration::Result, ProgressData)>;
+
+    /**
+     * @brief Perform level horizon calibration.
+     */
+    void calibrate_levelhorizon_async(CalibrateLevelHorizonCallback callback);
```
julianoes, line 147: I'm confused why this is `calibrate_levelhorizon_async`. Given that it is also `LevelHorizon` it should be `calibrate_level_horizon_async`.

I don't see that it is wrong in proto. @JonasVautherin do you understand why that is?

src/plugins/calibration/calibration.cpp
julianoes: Ha, same comment :)

TSC21: Probably because I had that in place already. Let me check

src/plugins/calibration/include/plugins/calibration/calibration.h
JonasVautherin: You were 15 seconds faster on the comment :laughing:

src/plugins/calibration/calibration_impl.cpp
```diff
@@ -128,6 +128,33 @@ void CalibrationImpl::calibrate_magnetometer_async(const CalibrationCallback& ca
...
 }
 
+void CalibrationImpl::calibrate_levelhorizon_async(const CalibrationCallback& callback)
+{
+    std::lock_guard<std::mutex> lock(_calibration_mutex);
+
+    if (_parent->is_armed()) {
+        report_failed("System is armed.");
+        return;
+    }
```
julianoes, line 138: This is actually not done properly, also for the other calls. We should fix it for another time. We should properly answer using the callback (as below).

TSC21:
> > LogErr() << "Calibration error: " << Calibration::result_str(future_ret);
> 
> Oh, actually that does not exist anymore (the `result_str`) .

Let me fix that quickly.

src/plugins/calibration/calibration.cpp
julianoes: But then it wouldn't compile, right?

src/plugins/calibration/calibration_impl.cpp
TSC21: @julianoes addressed.

JonasVautherin:
I'm confused now. The CI error was not in the generated files, was it? Seems like you changed them again :confused:. Could it be? :sweat_smile:

TSC21:
> I'm confused now. The CI error was not in the generated files, was it? Seems like you changed them again . Could it be?

It was: https://github.com/mavlink/MAVSDK/runs/667306103#step:4:241. Still getting use to the autogeneration mechanism xD

PR #1098, opened 2020-05-13
Steps towards gimbal protocol v2

julianoes:
This is work in progress to add support for the new gimbal protocol.

See docs: https://mavlink.io/en/services/gimbal_v2.html

This will conflict with #1022 and we'll have to find a way to resolve that.


julianoes:
I think this should still be ready to go in as a first step.

julianoes:
(wrong button)

julianoes:
@JonasVautherin did you leave any review comments? Cause I can't find them. I might have overwritten them by force pushing.

julianoes:
Stopping the containers timed out. I'll ignore that for now.

PR #1100, opened 2020-05-14
telemetry: fix local position health flag

julianoes:
This flas was actually just fake before and a duplicate of the GPS available flag.

Hopefully this will help with https://github.com/PX4/Firmware/issues/14835.

FYI @kamilritz

### DIFF
--- a/src/plugins/telemetry/telemetry_impl.cpp
+++ b/src/plugins/telemetry/telemetry_impl.cpp
@@ -501,6 +501,8 @@ void TelemetryImpl::process_position_velocity_ned(const mavlink_message_t& messa
         auto arg = position_velocity_ned();
         _parent->call_user_callback([callback, arg]() { callback(arg); });
     }
+
+    set_health_local_position(true);
 }
 
 void TelemetryImpl::process_global_position_int(const mavlink_message_t& message)
@@ -732,8 +734,6 @@ void TelemetryImpl::process_gps_raw_int(const mavlink_message_t& message)
     const bool gps_ok = ((gps_raw_int.fix_type >= 3) && (gps_raw_int.satellites_visible >= 8));
 
     set_health_global_position(gps_ok);
-    // Local is not different from global for now until things like flow are in place.
-    set_health_local_position(gps_ok);
 
     if (_gps_info_subscription) {
         auto callback = _gps_info_subscription;


kamilritz:
I am not convinced that this will help. The local position message is sent no matter if the local_position is valid. By valid I mean if it is constrained by any measurment (GPS, flow, ...). The ekf2 is giving sending this information over vehicle_local_position.xy_valid field. But this field is not exposed over MAVLINK.

There is the [estimator status MAVLINK message](https://mavlink.io/en/messages/common.html#ESTIMATOR_STATUS), which reports the uncertainty of the position estimate. This could be used to check if it is safe to takeoff / local_position is valid.

julianoes:
> The local position message is sent no matter if the local_position is valid.

Are you sure?
https://github.com/PX4/Firmware/blob/76de299302c9a91a0e1ac316fa174ed001f15a2e/src/modules/mavlink/mavlink_messages.cpp#L2683-L2697

The question is when `vehicle_local_position` is started to be published via uORB.
In theory it would have a valid flag that we shoudl check in PX4 before sending the MAVLink message:
https://github.com/PX4/Firmware/blob/76de299302c9a91a0e1ac316fa174ed001f15a2e/msg/vehicle_local_position.msg#L5

kamilritz:
I will fix this true-false-true behavior, which is coming from a bug inside the estimator. But we need to think of a way that the MAVLINK user knows that he can takeoff in the offboard mode. One option would be:
> In theory it would have a valid flag that we should check in PX4 before sending the MAVLink message

Another option would be to send a valid flag to the user, or an uncertainty measurement.

julianoes:
@kamilritz right. Actually, we should use this flag, no?
https://mavlink.io/en/messages/common.html#MAV_SYS_STATUS_SENSOR_XY_POSITION_CONTROL

kamilritz:
This is a fantastic idea. We only want to start the offboard mission if the position controller is running.

julianoes:
Right but we need to check if that's sent out in PX4 yet, probably not.

julianoes:
I think this is an improvement for now until we can use https://mavlink.io/en/messages/common.html#MAV_SYS_STATUS_SENSOR_XY_POSITION_CONTROL.

Merging.

Issue #1103, opened 2020-05-20
Use std::unordered_map instead of std::map?


roangel:
Maybe there is another reason for it that I am not aware of, but wouldn't it be better (O(1) vs O(log(n)) lookup time) if we used `std::unordered_map` instead of `std::map` in these and similar cases?
https://github.com/mavlink/MAVSDK/blob/578c70c37cf7faf6603431dfba5d073c343b9db0/src/core/call_every_handler.h#L36
https://github.com/mavlink/MAVSDK/blob/578c70c37cf7faf6603431dfba5d073c343b9db0/src/core/system_impl.h#L320

julianoes:
Totally! I think at the time of writing I wasn't aware of `std::unordered_map` :man_facepalming:.

roangel:
Nice, I can take care of changing it. I was also thinking that this part:
https://github.com/mavlink/MAVSDK/blob/578c70c37cf7faf6603431dfba5d073c343b9db0/src/core/mavlink_message_handler.h#L29

Can also be an `std::unordered_map`, with the msg id as key? or even a custom hash of the msg id and the cookie? Would make the lookup time be so much faster (O(n) vs O(1))

julianoes:
Yes, I was looking at that today. The `msg_id`s can be duplicate but the `cookie` is unique, however, the unregistering won't happen very often.

Anecdotally, I've heard that just looping through a relatively short `std::vector` can be quicker in practice than having to do lookups in a map, so I would try to profile it before changing.

roangel:
> Anecdotally, I've heard that just looping through a relatively short `std::vector` can be quicker in practice than having to do lookups in a map, so I would try to profile it before changing.

Yes, I have also heard that it's because of the overhead of building the hash table behind `unordered_map`. Not sure at which quantities it starts paying off.

Issue #1031, opened 2020-04-07
mavlink_mission_item_to_mission_item_indices never cleared


Katawann:
_mission_data.mavlink_mission_item_to_mission_item_indices is never cleared which means we might have more mavlink_mission_item_to_mission_item_indices than we want if we download a list of items smaller than the previous one

https://github.com/mavlink/MAVSDK/blob/8afd4640cebf4fae242a9ff9650d354cebab2f48/src/plugins/mission/mission_impl.cpp#L241

julianoes:
Thanks for the report.

coderkalyan:
What's needed to fix this? Is it simply a matter of clear()ing the list before we start inserting into it?

Katawann:
yes only that hehe. I'm not working with the last version of MAVSDK and I don't have the time to upgrade for now with my job but yes if you clear it everything is fine !

coderkalyan:
Okay, if it's just clearing I can try to get a PR for this soon.

PR #1105, opened 2020-05-20
Changed map to unordered_map

roangel:
Fixes https://github.com/mavlink/MAVSDK/issues/1103

### DIFF
--- a/src/core/call_every_handler.h
+++ b/src/core/call_every_handler.h
@@ -3,7 +3,7 @@
 #include <mutex>
 #include <memory>
 #include <functional>
-#include <map>
+#include <unordered_map>
 #include "global_include.h"
 
 namespace mavsdk {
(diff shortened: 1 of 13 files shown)


roangel:
fixed style

roangel:
CI fails:
image

julianoes:
@roangel interesting, I wonder if that's related or not.

julianoes:
@roangel ok it looks unrelated and like a bug with the mission item count.

roangel:
Maybe it's this part?
image

julianoes:
@roangel I was just fixing the asan issue (due to threading) in the integration test but haven't looked for the thing producing the wrong number of mission items. I think you found it, thanks!

julianoes:
I think there might be a few more places even. And while looking at it I also stumbled across #1031.

julianoes:
Thanks @roangel. I'll follow up with the mission part in another PR.

PR #1107, opened 2020-05-25
Fix mission indices

### LINKED ISSUE #1031: mavlink_mission_item_to_mission_item_indices never cleared

julianoes:
This is the follow up to @roangel's PR #1105.

Also fixes #1031.

### DIFF
--- a/src/plugins/mission/mission_impl.cpp
+++ b/src/plugins/mission/mission_impl.cpp
@@ -476,7 +476,6 @@ MissionImpl::convert_to_int_items(const std::vector<MissionItem>& mission_items)
         _mission_data.mavlink_mission_item_to_mission_item_indices.push_back(item_i);
         int_items.push_back(next_item);
     }
-
     return int_items;
 }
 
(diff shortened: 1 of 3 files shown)


src/plugins/mission/mission_impl.cpp
```diff
@@ -477,10 +473,10 @@ MissionImpl::convert_to_int_items(const std::vector<MissionItem>& mission_items)
                                                   0,
                                                   MAV_MISSION_TYPE_MISSION};
 
-        _mission_data.mavlink_mission_item_to_mission_item_indices.insert(
-            std::pair<int, int>{static_cast<int>(int_items.size()), item_i});
+        _mission_data.mavlink_mission_item_to_mission_item_indices.push_back(item_i);
         int_items.push_back(next_item);
     }
+
```
roangel, line 479:
```suggestion
```

julianoes:
The error is another one of #882.

PR #1112, opened 2020-05-27
Bootstrapping of plugins from proto

julianoes:
The `generate_from_protos.sh` script can now bootstrap everything needed to add a new plugin. It does so by looking for new proto definitions in `proto/protos` and using templates for the `_impl.h` and `_impl.cpp` parts.

### DIFF
--- a/tools/generate_from_protos.sh
+++ b/tools/generate_from_protos.sh
@@ -14,8 +14,7 @@ function snake_case_to_camel_case {
     echo $1 | sed -r 's/(^|_)([a-z])/\U\2/g'
 }
 
-
-command -v ${protoc_binary} && command -v ${protoc_grpc_binary} || {
+command -v ${protoc_binary} > /dev/null && command -v ${protoc_grpc_binary} > /dev/null || {
     echo "-------------------------------"
     echo " Error"
     echo "-------------------------------"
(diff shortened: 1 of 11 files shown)


tools/generate_from_protos.sh
```diff
@@ -70,4 +67,52 @@ for plugin in ${plugin_list}; do
...
+    file_impl_h="${script_dir}/../src/plugins/${plugin}/${plugin}_impl.h"
+    if [[ ! -f "${file_impl_h}" ]]; then
+        ${protoc_binary} -I ${proto_dir} --custom_out=${tmp_output_dir} --plugin=protoc-gen-custom=${protoc_gen_dcsdk} --custom_opt="file_ext=h,template_path=${template_path_plugin_impl_h}" ${proto_dir}/${plugin}/${plugin}.proto
+        mv ${tmp_output_dir}/${plugin}/$(snake_case_to_camel_case ${plugin}).h ${file_impl_h}
+        echo "-> Creating ${file_impl_h}"
+    else
+        # Warn if file is not checked in yet.
+        if [[ ! $(git ls-files --error-unmatch ${file_impl_h} 2> /dev/null) ]]; then
+            echo "-> Not creating ${file_impl_h} because it already exists"
+        fi
```
JonasVautherin, line 80: Personal preference: I think this adds noise to the script, it will mostly never run (hence can easily get outdated) and if it happens while we are working on templates, then it will spam us every time we run the script :grin:.

This said, I could live with it :wink:.

tools/generate_from_protos.sh
```diff
@@ -70,4 +67,52 @@ for plugin in ${plugin_list}; do
...
+        if [[ ! $(git ls-files --error-unmatch ${file_cmake} 2> /dev/null) ]]; then
+            echo "-> Not creating ${file_cmake} because it already exists"
+        fi
+    fi
+
+    plugins_cmake_file="${script_dir}/../src/plugins/CMakeLists.txt"
+    if [[ ! $(grep ${plugin} ${plugins_cmake_file}) ]]; then
+        echo "-> Adding entry for '${plugin}' to ${plugins_cmake_file}"
+
+        # We want to append the plugin to the list but before the passthrough plugin.
```
JonasVautherin, line 111: We will lose the alphabetical order :scream: :scream: :scream: 

Just kidding :grin: :trollface:

julianoes: Yer, I thought about how to make that happen but gave up.

tools/generate_from_protos.sh
```diff
@@ -29,38 +28,36 @@ command -v ${protoc_binary} && command -v ${protoc_grpc_binary} || {
...
-echo ""
+echo "Found protoc ($(${protoc_binary} --version)): ${protoc_binary}"
+echo "Found grpc_cpp_plugin: ${protoc_grpc_binary}"
 
-mkdir -p ${backend_generated_dir}
-
-for plugin in ${plugin_list_and_core}; do
-    ${protoc_binary} -I ${proto_dir} --cpp_out=${backend_generated_dir} --grpc_out=${backend_generated_dir} --plugin=protoc-gen-grpc=${protoc_grpc_binary} ${proto_dir}/${plugin}/${plugin}.proto
-done
+plugin_list_and_core=$(cd ${script_dir}/../proto/protos && ls -d */ | sed 's:/*$::')
```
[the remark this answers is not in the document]
julianoes: Yes we have to, otherwise we get the path. Unless there is a flag for `ls` to omit the path.

tools/generate_from_protos.sh
julianoes: No, it shouldn't add noise. The output only appears if the files are there but have not been committed. Without this it's confusing if you change the proto but the changes are not reflected in the generated impl. That's because I don't want to overwrite any changed that were made.

PR #1113, opened 2020-05-28
Do not ignore QGC messages

JonasVautherin:
I think this was forgotten when removing the other connection hacks.


JonasVautherin:
For the record, already made some testing with QGC, and that was working.

julianoes:
Great, thanks!

PR #1125, opened 2020-06-11
Stop all plugins when stopping server

JonasVautherin:
Otherwise it hangs when `stopBackend()` is called.


julianoes:
This list is not auto-generated yet, right?

julianoes:
I'm ignoring the usual integration test failures.

PR #1126, opened 2020-06-11
serial: add support for flow control

bkueng:
Tested on Linux, not entirely sure about the Windows part.

Can be used as `serial_flowcontrol://<device>`

### DIFF
--- a/src/core/cli_arg.cpp
+++ b/src/core/cli_arg.cpp
@@ -46,6 +46,7 @@ bool CliArg::find_protocol(std::string& rest)
     const std::string udp = "udp";
     const std::string tcp = "tcp";
     const std::string serial = "serial";
+    const std::string serial_flowcontrol = "serial_flowcontrol";
     const std::string delimiter = "://";
 
     if (rest.find(udp + delimiter) == 0) {
@@ -58,7 +59,13 @@ bool CliArg::find_protocol(std::string& rest)
         return true;
     } else if (rest.find(serial + delimiter) == 0) {
         _protocol = Protocol::Serial;
+        _flow_control_enabled = false;
         rest.erase(0, serial.length() + delimiter.length());
+        return true;
+    } else if (rest.find(serial_flowcontrol + delimiter) == 0) {
+        _protocol = Protocol::Serial;
+        _flow_control_enabled = true;
+        rest.erase(0, serial_flowcontrol.length() + delimiter.length());
         return true;
     } else {
         LogWarn() << "Unknown protocol";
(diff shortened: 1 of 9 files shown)


src/core/cli_arg.cpp
```diff
@@ -58,8 +59,14 @@ bool CliArg::find_protocol(std::string& rest)
...
     } else if (rest.find(serial + delimiter) == 0) {
         _protocol = Protocol::Serial;
+        _flow_control = false;
         rest.erase(0, serial.length() + delimiter.length());
         return true;
+    } else if (rest.find(serial_flowcontrol + delimiter) == 0) {
+        _protocol = Protocol::Serial;
+        _flow_control = true;
+        rest.erase(0, serial_flowcontrol.length() + delimiter.length());
+        return true;
```
julianoes, line 69: I should have told you earlier, there is a test for this:
https://github.com/mavlink/MAVSDK/blob/develop/src/core/cli_arg_test.cpp

Would be great if you could add coverage for this.

src/core/cli_arg.h
```diff
@@ -29,6 +31,7 @@ class CliArg {
     std::string _path{};
     int _port{0};
     int _baudrate{0};
+    bool _flow_control{false};
```
julianoes, line 34:
```suggestion
    bool _flow_control_enabled{false};
```
or `_flow_control_enabled` so that it's clear what `true` means.

src/core/serial_connection.cpp
```diff
@@ -172,8 +180,14 @@ ConnectionResult SerialConnection::setup_port()
...
     dcb.StopBits = ONESTOPBIT;
-    dcb.fDtrControl = DTR_CONTROL_DISABLE;
-    dcb.fRtsControl = RTS_CONTROL_DISABLE;
+    if (_flow_control) {
+        dcb.fOutxCtsFlow = TRUE;
+        dcb.fDtrControl = DTR_CONTROL_HANDSHAKE;
+        dcb.fRtsControl = RTS_CONTROL_HANDSHAKE;
+    } else {
+        dcb.fDtrControl = DTR_CONTROL_DISABLE;
+        dcb.fRtsControl = RTS_CONTROL_DISABLE;
```
julianoes, line 189: Looks correct :smile:.

src/core/cli_arg.cpp
bkueng: Of course, I should have checked. Let's see if they pass.

julianoes:
@bkueng thanks! And @JonasVautherin is planning to do a release in the next days where it will be included.

PR #1127, opened 2020-06-11
Add support for statustext sequences

### LINKED ISSUE #1039: Support for multi-part statustext messages
We should add support for multipart statustext messages as specified in https://github.com/mavlink/mavlink/pull/1295.

Here https://github.com/mavlink/MAVSDK/blob/138ee4acebfbfa5875425048423014cb454ebee4/src/core/system_impl.cpp#L216-L258

krady21:
Fixes #1039


julianoes:
Sorry for not following up here @krady21. Let me merge it now and then we can improve it later.

PR #1133, opened 2020-06-17
Move to openssl

JonasVautherin:
We regularly have timeouts when checking out LibreSSL, and we would need some OpenSSL features at some point anyway (e.g. dTLS).


julianoes:
Closing until we need it again. :smile:

Issue #1137, opened 2020-07-06
Rebooting autopilot alone


shakeebbb:
I'm intending to reboot the PX4 autopilot using MAVSDK. I could find 'reboot' function inside the 'action' plugin which reboots the autopilot, camera, gimbal and the companion computer. To reboot the autopilot alone, I assume that 'reboot_async' function needs to be modified to turn off all parameters except the one corresponding to the autopilot. If that's correct and there is no other existing function to do that, can I send a pull request to make this function more generic by including flags with default arguments to this function?

julianoes:
Yes I think that would make sense, and see my review comments.

PR #1138, opened 2020-07-06
Added support to reboot/shutdown seperate components

shakeebbb:
Pull request for issue #1137

### DIFF
--- a/src/plugins/action/action_impl.cpp
+++ b/src/plugins/action/action_impl.cpp
@@ -78,22 +78,22 @@ Action::Result ActionImpl::kill() const
     return fut.get();
 }
 
-Action::Result ActionImpl::reboot() const
+Action::Result ActionImpl::reboot(const uint8_t component) const
 {
     auto prom = std::promise<Action::Result>();
     auto fut = prom.get_future();
 
-    reboot_async([&prom](Action::Result result) { prom.set_value(result); });
+    reboot_async([&prom](Action::Result result) { prom.set_value(result); }, component);
 
     return fut.get();
 }
 
-Action::Result ActionImpl::shutdown() const
+Action::Result ActionImpl::shutdown(const uint8_t component) const
 {
     auto prom = std::promise<Action::Result>();
     auto fut = prom.get_future();
 
-    shutdown_async([&prom](Action::Result result) { prom.set_value(result); });
+    shutdown_async([&prom](Action::Result result) { prom.set_value(result); }, component);
 
     return fut.get();
 }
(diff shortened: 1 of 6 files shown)


src/plugins/action/action_impl.cpp
```diff
@@ -234,31 +234,31 @@ void ActionImpl::kill_async(const Action::ResultCallback& callback) const
     });
 }
 
-void ActionImpl::reboot_async(const Action::ResultCallback& callback) const
+void ActionImpl::reboot_async(const Action::ResultCallback& callback, const uint8_t component) const
```
JonasVautherin, line 237: Shouldn't it be more something like an enum? e.g. `reboot(GIMBAL)`, `reboot(AUTOPILOT)`, `reboot(ALL)`?

julianoes: Yes, it could be an enum or the MAVLink component ID, I'd say.

JonasVautherin: Oh, right. There could be 2 cameras or 2 gimbals, again the same thing. But then shouldn't it be `drone.gimbal.reboot()` or `drone.camera.reboot()`? And in the future, when we support multiple cameras, it becomes something like `drone.camera[0].reboot()`.

julianoes: True, that would be nicer. However, @shakeebbb wants to reboot only the autopilot.

shakeebbb: Shouldn't MAV_COMPONENT enum be enough to handle multiple cameras and gimbals on the mavlink network?

shakeebbb:
Thanks for the heads up @julianoes! Is the development branch appropriate for this pull request or should I use the master branch?

julianoes:
> Is the development branch appropriate for this pull request or should I use the master branch?

`develop` is the one to work from, thx. Note to self: we should rename `master` to `release`.

src/plugins/action/action_impl.cpp
JonasVautherin: I see. What about `drone.action.reboot(bool autopilot_only)`? So that the action plugin can reboot only the autopilot OR everything. And then for individual components, this goes into their corresponding plugin. Would that work?

julianoes:
@shakeebbb are you still interested in getting this in? @JonasVautherin an idea what we should do?

JonasVautherin:
> @JonasVautherin an idea what we should do?

Still the same opinion on my side, as commented above :blush:.

shakeebbb:
@julianoes I would like to finish this, I am temporarily using my current fork to reboot my px4. My schedule is a bit packed this semester. I might take some time to get back to this if that's ok.

shakeebbb:
@julianoes I would be happy to finish this, I am temporarily using my current fork to reboot my px4. My schedule is a bit packed this semester. I might take some time to get back to this if that's ok.

julianoes:
@shakeebbb ping :)

julianoes:
Closing until @shakeebbb comes back.

Issue #1045, opened 2020-04-20
Asan issue in SITL v1.9 CI test


julianoes:
The address sanitizer found something in CI:

https://github.com/mavlink/MAVSDK/pull/1042/checks?check_run_id=601088428

```
2020-04-20T08:25:45.9611517Z waiting for system to disarm
2020-04-20T08:25:46.0028165Z [FlightMode: Flight Mode Land] Vehicle is at Lat: 47.3977 deg, Lon: 8.54564 deg.
2020-04-20T08:25:46.1263658Z [FlightMode: Flight Mode Land] Vehicle is at Lat: 47.3977 deg, Lon: 8.54564 deg.
2020-04-20T08:25:46.1504615Z [32m[08:25:46|Debug] [0mMAVLink: info: Landing detected (system_impl.cpp:257)
2020-04-20T08:25:46.1656642Z [32m[08:25:46|Debug] [0mMAVLink: info: DISARMED by Auto disarm on land (system_impl.cpp:257)
2020-04-20T08:25:46.2535044Z [FlightMode: Flight Mode Land] Vehicle is at Lat: 47.3977 deg, Lon: 8.54564 deg.
2020-04-20T08:25:46.3775860Z [FlightMode: Flight Mode Land] Vehicle is at Lat: 47.3977 deg, Lon: 8.54564 deg.
2020-04-20T08:25:46.5008294Z [FlightMode: Flight Mode Land] Vehicle is at Lat: 47.3977 deg, Lon: 8.54564 deg.
2020-04-20T08:25:46.6212400Z [FlightMode: Flight Mode Land] Vehicle is at Lat: 47.3977 deg, Lon: 8.54564 deg.
2020-04-20T08:25:46.7410665Z [FlightMode: Flight Mode Land] Vehicle is at Lat: 47.3977 deg, Lon: 8.54564 deg.
2020-04-20T08:25:46.9617132Z [0m[1;33m[0m=================================================================
2020-04-20T08:25:46.9622136Z ==8705==ERROR: AddressSanitizer: heap-use-after-free on address 0x603000283fd0 at pc 0x5557657b4891 bp 0x7f1c829fc6f0 sp 0x7f1c829fc6e0
2020-04-20T08:25:46.9622355Z READ of size 8 at 0x603000283fd0 thread T40
[...]

PR #1143, opened 2020-07-14
Remove thread pool

julianoes:
Before introducing the callback queue, we used a thread pool of 3 threads to call user callbacks.
The thread pool was written to avoid everything from stalling if a user blocked in one of the callbacks.
It generally worked provided the user did not block in more than 3 callbacks at any given time.

However, it had also some disadvantages:

- Due to the thread pool multiple threads could call a function at any given time.
  This means that one callback can be called from at least two threads at the same time which is quite likely if the callback is called at a high rate.
  This can create race conditions on the user side unless a mutex is used to prevent this.
  However, it's not very intuitive that a mutex is needed for that, and mutexes in general should be avoided, if possible, for performance reasons.
  Such race conditions also showed up as CI test failures, see: [MAVSDK#1010](https://github.com/mavlink/MAVSDK/issues/1010), [MAVSDK#1045](https://github.com/mavlink/MAVSDK/issues/1045).
- For developers the old thread pool behaviour was unclear and unpredictable.
  Basically, without code inspection, it was difficult to know if blocking in a callback is safe or not.
  When trying with one blocking callback, it would work, however, this can create a false sense of security as with 3 blocking callbacks it would suddenly lock up.
- Previously, when this happened, it was difficult to debug because it would either lead to a crash on destruction, or lock up at which point the backtrace of all threads need to be looked at using gdb.


julianoes:
Alright, let's get this in.

PR #1144, opened 2020-07-14
telemetry: handle lat = 0 lon = 0 correctly

julianoes:
For the case where the altitude is known but the global position is not, the message is sent with lat = 0 and lon = 0.

Still depends on the decision in https://github.com/PX4/Firmware/pull/15308.


JonasVautherin:
Oh... should it be `std::NAN` (looking at the CI failures)?

julianoes:
Ok, I'll try `std::nan` with `#include <cmath>`.

PR #1146, opened 2020-07-17
workflows: use checkout action v2

julianoes:
This should fix the coveralls error that we have been seeing:

```
Run coverallsapp/github-action@master
  with:
    github-token: ***
    path-to-lcov: ./lcov.info
    coveralls-endpoint: https://coveralls.io
Using lcov file: ./lcov.info
[error] "2020-07-16T13:40:56.682Z"  'error from getOptions'
fatal: Not a valid object name 3ce4b7b924ef507acc7570dce933938da28d9be9
```

See: https://github.com/coverallsapp/github-action/issues/55

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -129,13 +129,27 @@ jobs:
       - name: test (mavsdk_server)
         run: ./build/release/src/backend/test/unit_tests_backend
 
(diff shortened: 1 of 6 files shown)


julianoes:
The SITL tests timing out are unrelated, merging.

PR #1149, opened 2020-07-22
Plugin for manual control

julianoes:
This uses a joystick using SDL2.

Proto work by @AldoCF96


JonasVautherin:
Hmm, docs is apparently still failing :confused:

hamishwillee:
The generator script for markdown only exits if the XML docs directory is missing - otherwise it just prints warnings. Can you tell where the error originates? FYI I get the same output locally. 

Note:

1. There is this warning:
   > ...  manual_control/manual_control.h:27: warning: Compound mavsdk::ManualControl is not documented.
2. There is also a namespace file that is not being parsed that is undocumented: namespacemavsdk_1_1geometry.xml. This should be documented, but is not likely the cause of this problem. 
3. The parser does not know about friend methods so I also see this warning`UNHANDLED section: XXfriendYY`. I will add a section to handle that. at some point. Could this be causing CI to fail?

julianoes:
Thanks @hamishwillee, it's this one:
> ... manual_control/manual_control.h:27: warning: Compound mavsdk::ManualControl is not documented.

It greps for "warning" and fails if it finds any, see: https://github.com/mavlink/MAVSDK/blob/28084e4bbcf01ea7ff7d23082588c0b0c9543b04/tools/generate_docs.sh#L48

I'll fix it...

PR #1152, opened 2020-07-27
docker: use PX4 master for now

julianoes:
This is a workaround to prevent CI from being stuck for hours. Instead this should fail immediately if gzserver is not running.

To include https://github.com/PX4/Firmware/pull/15399.


julianoes:
I think this is better anyway, merging.

PR #1156, opened 2020-07-29
core: added ping capability

julianoes:


julianoes:
> Is there a reason for this being a draft PR?

Sorry, yes. There is still a debug printf in there. I'll update this soon.

julianoes:
Ready for review and merging.

JonasVautherin (approved):
LGTM

JonasVautherin (approved):

hshakula:
What was the reason for this feature?

julianoes:
Good question @hshakula, I don't remember, I guess just implementing the MAVLink ping protocol, which allows to get an estimate of the return latency.

However, according to [PING](https://mavlink.io/en/messages/common.html#PING) we should use [SYSTEM_TIME](https://mavlink.io/en/messages/common.html#SYSTEM_TIME) instead.

Why do you ask?

hshakula:
No particular reason. I was just analyzing MAVLink traffic in my network. And with 6 MAVSDK-based components per one system, when each of them does broadcast pings, I could not leave it unnoticed. So I decided to figure out why is it even needed. Considering it's there just for the sake of being there, I intend to disable this.

julianoes:
That's a good point. We could make it opt-in, or at least opt-out with an env variable. Would that help?

Issue #1116, opened 2020-05-30
Creating/Uploading a Survey Mission


petergerten:
Is there any way to create (or even just upload) a survey mission via MAVSDK.
If I have a QGC plan file, it seems that the importer won't work as TransectStyleComplexItem is not supported.

JonasVautherin:
[This example](https://github.com/mavlink/MAVSDK/blob/develop/examples/fly_mission/fly_mission.cpp) shows how to use the `Mission` plugin. Would that help?

petergerten:
@JonasVautherin I was having trouble to import a survey QGC plan and was confused on what exactly is supported and what not. Now I am stuck with a different issue (https://github.com/mavlink/MAVSDK-Python/issues/201), so I did not get further with this.

Is there an overview of what is supported or what is not supported when importing an QGC plan file via MAVSDK?

Jonsen94:
I checked the c-code. If I understand it correctly u can basically only use Mission-Plans from QGC which are build up from SimpleItems.
Just take a look into the plan file. It contains an "items"-Array. Everything there with a COMMAND_ID field can be used.

As a quick fix u can simple go into a plan file with a survey item. The survey item (inside the "items"-array) contains an "Items" array. Copy paste the content from the inner array into the outer one (don't forget to add correct commas) and remove the survey item. You can't change it from QGC so nicely afterwards but it will at least work with mavsdk.

petergerten:
@Jonsen94 yes, I checked the code as well - and it seems to be quite a big effort to implement support for survey missions as in TransectStyleComplexItem :-(

Jonsen94:
Are you sure? I mean, basically you could simple check if the "command" exists, if not, check if "TransectStyleComplexItem" exists. If so, check if "Items" array exists (I guess then we know we are inside a survey item?) take the "Items" array and parse it with the same for loop which parses the other "items" array. You do not have to re-implement the routings or so because they are already stored in there.

petergerten:
@Jonsen94 so the items within a TransectStyleComplexItem would be sufficient to parse the mission - and all other information is only for QGC to be show the survey configuration but not required for the internal mission representation - do I understand that correctly?

Jonsen94:
afaik yes.
If I understand the mission plan correctly the rest of the item is only for QGC so it can be loaded as a survey item.

Jonsen94:
@petergerten : PR is merged and should solve the issue now.
[...]

PR #1163, opened 2020-08-12
Added QGC complex mission item importing

Jonsen94:
Currently the import_qgroundcontrol_mission function can only import simple items (waypoints).

QGC also supports surveys, corridor scans and structure scans.

This pull should fix the first two and allow importing surveys and corridor scans.

See https://github.com/mavlink/MAVSDK/issues/1116

It's not tested as I have problems connecting the server with my python-mavsdk scripts. (Timeout issues).
But it compiles at least.

### DIFF
--- a/src/plugins/mission/mission_impl.cpp
+++ b/src/plugins/mission/mission_impl.cpp
@@ -1056,6 +1056,11 @@ Mission::Result MissionImpl::import_complex_mission_item(
         return Mission::Result::UnsupportedMissionCmd;
     }
 
+    // QGC supports more complex mission items than simple waypoints.
+    // Surveys and coridor scans (NOT structure scans) are stored in a so called "TransectStyleComplexItem" item inside the mission_items array.
+    // These ComplexItems also contain an array ("Items") which contains waypoints.
+    // It is used by GQC to keep survey parameters so one can edit it as a survey after importing.
+    // Structure scans are not supported as thes do not contain simple mission items.
     Json::Value complex_item = json_complex_mission_item["TransectStyleComplexItem"];
     for (auto& json_mission_item : complex_item["Items"]) {
         Mission::Result result =
--- a/src/plugins/mission/mission_impl.cpp
+++ b/src/plugins/mission/mission_impl.cpp
@@ -1024,6 +1024,45 @@ Mission::Result MissionImpl::build_mission_items(
     return result;
 }
 
+Mission::Result MissionImpl::import_simple_mission_item(
+        std::vector<Mission::MissionItem>& all_mission_items, const Json::Value& json_mission_item, MissionItem& new_mission_item)
+{
+    // Parameters of Mission item & MAV command of it.
+    MAV_CMD command = static_cast<MAV_CMD>(json_mission_item["command"].asInt());
+
+    // Extract parameters of each mission item
+    std::vector<double> params;
+    for (auto& p : json_mission_item["params"]) {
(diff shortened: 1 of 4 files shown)


Jonsen94:
Okay what am I missing.
I don't see where the added code fails in the test ?

src/plugins/mission/mission_impl.cpp
```diff
@@ -1024,6 +1024,49 @@ Mission::Result MissionImpl::build_mission_items(
...
+    const Json::Value& json_complex_mission_item,
+    MissionItem& new_mission_item)
+{
+    if (json_complex_mission_item["TransectStyleComplexItem"].isNull()) {
+        LogWarn() << "Unknown complex item type (" << json_complex_mission_item["complexItemType"]
+                  << ")";
+        return Mission::Result::UnsupportedMissionCmd;
+    }
+
+    Json::Value complex_item = json_complex_mission_item["TransectStyleComplexItem"];
```
julianoes, line 1059: Could you add a code comment for the reader on what this complex "TransectStyleComplexItem" actually is?

julianoes (changes requested):
Thanks, I think the change is nice and clean.

However, it would be great if you could add a unit test for such a complex mission file.
See https://github.com/mavlink/MAVSDK/blob/develop/src/plugins/mission/mission_import_qgc_test.cpp

Jonsen94:
Btw for documentation reasons if someone finds this pr:
Taken from the GQC git (https://github.com/mavlink/qgroundcontrol):
Complex Mission Item (which is a super class of Surveys, Corridor Scans and Structure Scans):  [ComplexMissionItem.h](https://github.com/mavlink/qgroundcontrol/blob/master/src/MissionManager/ComplexMissionItem.h)
Transect Mission Item (which is a subclass of the Complex Mission Item and a super class of Surveys and Corridor Scans, *note*: not of the structure scans): [TransectStyleComplexItem.h](https://github.com/mavlink/qgroundcontrol/blob/master/src/MissionManager/TransectStyleComplexItem.h)
The json key "Items" for the array I used for getting the simple mission items: [TransectStyleComplexItem.cc#L38](https://github.com/mavlink/qgroundcontrol/blob/master/src/MissionManager/TransectStyleComplexItem.cc#L38)

Other complex items:
[StructureScanComplexItem.h](https://github.com/mavlink/qgroundcontrol/blob/master/src/MissionManager/StructureScanComplexItem.h)
[CorridorScanComplexItem.h](https://github.com/mavlink/qgroundcontrol/blob/master/src/MissionManager/CorridorScanComplexItem.h)
[SurveyComplexItem.h](https://github.com/mavlink/qgroundcontrol/blob/master/src/MissionManager/SurveyComplexItem.h)

src/plugins/mission/mission_impl.cpp
Jonsen94: I added a comment and a test case.
The tests output lots of warnings (unsupported mission items) for 206 (MAV_CMD_DO_SET_CAM_TRIGG_DIST) and 20 (MAV_CMD_NAV_RETURN_TO_LAUNCH ) which are used in the test plan file.

julianoes: Ok thanks. I guess that's expected because they are not understood by the simple MAVSDK mission API.

julianoes (approved):
Thanks for the PR and follow up.

Issue #1158, opened 2020-08-05
set own usage_type?


ericwemhoff:
I don't see a way to see the Mavsdk's usage_type. I can create a Configuration with a desired usage type, but Mavsdk::set_configuration seems to ignore it:

mavsdk_impl.cpp:272:
void MavsdkImpl::set_configuration(Mavsdk::Configuration configuration)
{
    own_address.system_id = configuration.get_system_id();
    own_address.component_id = configuration.get_component_id();
}

And there's not a Mavsdk constructor that takes a usage_type or a Configuration. The default constructor sets it up to be GroundStation. My issue stems from having a gimbal I want to talk to, that seems to want to see heartbeats from a system whose usage type is CompanionComputer, or at least not one whose usage_type is GroundStation.

Am I missing something, or not understanding the intended usage?

MAVSDK$ git remote -v
origin	https://github.com/mavlink/MAVSDK.git (fetch)
origin	https://github.com/mavlink/MAVSDK.git (push)

MAVSDK$ git show
commit a1e83140a6a090ed77f3a79f2dada8237aedaedb (HEAD -> master, tag: v0.29.0, origin/master, origin/develop, origin/HEAD, develop)
Merge: dc529f87 0bf4fc5e
Author: Julian Oes <julian@oes.ch>
Date:   Tue Jul 28 17:36:53 2020 +0200

thanks!
Eric

julianoes:
Hi Eric,

> but Mavsdk::set_configuration seems to ignore it:

Why do you think that? Did you test it?

ericwemhoff:
Hi, yes I've done some testing.

BTW I should have been clear that I'm compiling MAVSDK from source, on the commit I indicated in my first message.

I tried to set the usage_type by doing this:
   mavsdk::Mavsdk::Configuration mavsdkcfg(mavsdk::Mavsdk::Configuration::UsageType::CompanionComputer);
   dc.set_configuration(mavsdkcfg);

To check whether it had an effect, I've instrumented the code with print statements to print the current configuration usage_type and output of MavsdkImpl::get_mav_type(), and I've looked at the content of the heartbeat messages that are being sent (looking at the 'type' field). The above lines don't seem to have an effect, based on those observations.

[...]

PR #1168, opened 2020-08-15
Update Configuration for indep. control of sysid, compid, usage_type.

ericwemhoff:
This regards issue #1158.

I think there are maybe two things here, closely related.

One is that the user ultimately doesnt have full control over the Mavsdk::Configuration object. This object basically exists to contain three parameters: system_id, component_id, and usage_type .. and, as far as I can tell, to help suggest good settings for them. Eg if the user provides a usage_type it'll set system_id and component_id automatically based on best guesses, and vice versa. The problem is that the user can't independently set them to arbitrary values, and I think this should be an option, for users with custom situations/requirements (as I seem to have).

The second issue is that even if a user is able to construct the right Configuration object, as it is you can only use it to set a MavsdkImpl object's internal system_id and component_id (stored in its "own_address MAVLinkAddress member, which is just a struct with those two vars), and not its usage_type. This happens by calling MavsdkImpl::set_configuration (defined in mavsdk_impl.cpp). In fact, MavsdkImpl doesnt have a usage_type member. I could only find one place where a system's usage type is looked up directly, and that is in MavsdkImpl::get_mav_type() (which is in turn used when creating heartbeat messages) and there it uses the usage_type from its member variable '_configuration'.

I kinda want to replace the current approach which uses the Configuration class, with a more flexible functionality that would manipulate MavsdkImpl member settings directly, via its constructor and some setters. But that would break the current API with the Configuration mechanism, and I might not fully understand the original intent..

This pull request only adds to the existing API, so existing code will still work, and tries to keep the changes minimal.

First, add setters to the Configuration class, making its system_id, component_id, and usage_type settings independently configurable.

Then, update MavsdkImpl::set_configuration so that in addition to updating the system_id and component_id, it'll update the "_configuration" member var as well. I looked thru the code in core/, and seems like that won't break anything. It didn't for my use case anyway (which at the moment is using MAVSDK to communicate with a gimbal from a linux laptop over a serial connection).

In reference to my use case that I described in issue #1158, I can now do the following to configure mavsdk how I need:

   mavsdk::Mavsdk dc;
   mavsdk::Mavsdk::Configuration mavsdkcfg(mavsdk::Mavsdk::Configuration::UsageType::CompanionComputer);
   mavsdkcfg.set_system_id(58);
   dc.set_configuration(mavsdkcfg);

And then the heartbeat messages have the right system_id and usage_type, or at least it's enough so that my gimbal will work.

Note, I haven't run any other tests..

@julianoes 
#1158

### DIFF
--- a/src/core/mavsdk.h
+++ b/src/core/mavsdk.h
@@ -162,11 +162,21 @@ public:
          */
         uint8_t get_system_id() const;
 
+        /** 
+         * @brief Set the system id of this configuration.
+         */
+        void set_system_id(uint8_t system_id);
+
         /**
          * @brief Get the component id of this configuration
          * @return `uint8_t` the component id stored in this configuration,from 1-255
          */
         uint8_t get_component_id() const;
+
+        /**
+         * @brief Set the component id of this configuration.
+         */
+        void set_component_id(uint8_t component_id);
 
         /**
          * @brief UsageTypes of configurations, used for automatic ID setting
@@ -181,6 +191,11 @@ public:
 
         /** @brief Usage type of this configuration, used for automatic ID set */
         UsageType get_usage_type() const;
+
+        /**
+         * @brief Set the usage type of this configuration.
+         */
+        void set_usage_type(UsageType usage_type);
 
     private:
         uint8_t _system_id;
--- a/src/core/mavsdk_impl.cpp
+++ b/src/core/mavsdk_impl.cpp
@@ -273,6 +273,7 @@ void MavsdkImpl::set_configuration(Mavsdk::Configuration configuration)
 {
     own_address.system_id = configuration.get_system_id();
     own_address.component_id = configuration.get_component_id();
+    _configuration.set_usage_type(configuration.get_usage_type());
 }
 
 std::vector<uint64_t> MavsdkImpl::get_system_uuids() const
--- a/src/core/mavsdk.cpp
+++ b/src/core/mavsdk.cpp
@@ -124,14 +124,30 @@ uint8_t Mavsdk::Configuration::get_system_id() const
     return _system_id;
 }
 
+void Mavsdk::Configuration::set_system_id(uint8_t system_id)
+{
+    _system_id = system_id;
+}
+
 uint8_t Mavsdk::Configuration::get_component_id() const
 {
     return _component_id;
+}
+
+void Mavsdk::Configuration::set_component_id(uint8_t component_id)
+{
+    _component_id = component_id;
 }
 
 Mavsdk::Configuration::UsageType Mavsdk::Configuration::get_usage_type() const
 {
     return _usage_type;
 }
+
+void Mavsdk::Configuration::set_usage_type(Mavsdk::Configuration::UsageType usage_type)
+{
+    _usage_type = usage_type;
+}
+
 
 } // namespace mavsdk


coderkalyan:
I personally think we should the UsageType enum into 2 enums, because it seems to be used for 2 different things: 1) autoselecting configuration options (what it was intended for) and 2) determining the HEARTBEAT get_mav_type().

As a side note, it would probably make sense to have a separate method of configuration for get_mav_type() because I've been finding in my work on the Avoidance plugin that other parameters in the HEARTBEAT message sent by the SDK might need to be changed (right now most of them are static).

Would like to have @julianoes opinion though.

src/core/mavsdk.h
```diff
@@ -182,6 +192,11 @@ class Mavsdk {
         /** @brief Usage type of this configuration, used for automatic ID set */
         UsageType get_usage_type() const;
 
+        /**
+         * @brief Set the usage type of this configuration.
```
julianoes, line 196: Maybe add a comment here:

> Note that this will set `system_id` and `component_id`.

src/core/mavsdk_impl.cpp
```diff
@@ -273,6 +273,7 @@ void MavsdkImpl::set_configuration(Mavsdk::Configuration configuration)
 {
     own_address.system_id = configuration.get_system_id();
     own_address.component_id = configuration.get_component_id();
+    _configuration.set_usage_type(configuration.get_usage_type());
```
julianoes, line 276: I don't think this is required, right? Or am I confused?

julianoes (commented):
Thanks for the PR. Check the inline comments.

julianoes:
@ericwemhoff did you see my review?

src/core/mavsdk.h
ericwemhoff: You mean a comment on set_usage_type()? That comment would not be accurate, there are separate setter methods for system_id and component_id. You may be confusing this function with MavsdkImpl::set_configuration()?

src/core/mavsdk_impl.cpp
ericwemhoff: I think it is required, in order to copy the usage type from the Configuration object into the MavsdkImpl object, where it is needed by get_mav_type().

src/core/mavsdk.h
julianoes: Oh, I might be.

src/core/mavsdk_impl.cpp
julianoes: Ok.

julianoes:
@JonasVautherin objections or ok to merge this?

coderkalyan:
@julianoes FWIW, I'm still voting for splitting the UsageType enum from the logic that returns the MAV_TYPE, I think it would be useful to take the opportunity to add more configurability to the sdk's HEARTBEAT and keep UsageType purely as a convenience function for automatically setting system and component IDs. I need to be able to configure the system_status of the SDK's heartbeat for my avoidance work, and I'd rather not publish another "hacky" PR to solve this.

JonasVautherin (approved):

julianoes:
@coderkalyan I'm happy to evolve and improve this more later. For now, this seems like a fair addition to unblock the use case of @ericwemhoff.

ericwemhoff:
Ok thanks much. One last comment, while I'm thinking about it..

I don't disagree with @coderkalyan, and the configurability of the heartbeat msg could be generalized more. And yeah, using the Configuration mechanism is maybe not the best way to do it. Eg, it seems the heartbeat msg fields like "base_mode" and "system_status" would naturally change over time, and so those don't feel like configuration-type parameters.. which would be an argument against using the Configuration mechanism for those. And yes the mav type setting could be kept in a separate enum field somewhere.

At the moment the other HB fields aren't settable at all... heartbeat msgs are generated here:

void SystemImpl::send_heartbeat()
{
    mavlink_message_t message;
    // GCSClient is not autopilot!; hence MAV_AUTOPILOT_INVALID.
    mavlink_msg_heartbeat_pack(
        get_own_system_id(),
        get_own_component_id(),
        &message,
        get_own_mav_type(),  <<< "type" (the setting I needed access to)
        MAV_AUTOPILOT_INVALID,  <<< "autopilot"
        0,  <<< "base_mode"
        0,  <<< "custom_mode"
        0);  <<< "system_status"
    send_message(message);
}

Here's the prototype for mavlink_msg_heartbeat_pack():
/**
 * @brief Pack a heartbeat message
 * @param system_id ID of this system
 * @param component_id ID of this component (e.g. 200 for IMU)
 * @param msg The MAVLink message to compress the data into
 *
 * @param type  Vehicle or component type. For a flight controller component the vehicle type (quadrotor, helicopter, etc.). For other components the component type (e.g. camera, gimbal, etc.). This should be used in preference to component id for identifying the component type.
 * @param autopilot  Autopilot type / class. Use MAV_AUTOPILOT_INVALID for components that are not flight controllers.
 * @param base_mode  System mode bitmap.
 * @param custom_mode  A bitfield for use for autopilot-specific flags
 * @param system_status  System status flag.
 * @return length of the message in bytes (excluding serial stream start sign)
 */
static inline uint16_t mavlink_msg_heartbeat_pack(uint8_t system_id, uint8_t component_id, mavlink_message_t* msg,
                               uint8_t type, uint8_t autopilot, uint8_t base_mode, uint32_t custom_mode, uint8_t system_status);

Ok cheers.

PR #1169, opened 2020-08-19
Disable timesync by default

JonasVautherin:
It's not completely clear to me when timesync should be enabled: it seems like it is a broadcast, so I'm not sure if all the components should send it or only the autopilot.

I disabled it by default, so those who use it (e.g. @irsdkv) should enable it with `system.enable_timesync()`.


julianoes (approved):
Makes sense, thanks!

irsdkv (approved):

PR #1174, opened 2020-08-25
Fix serial connection on windows

NicolasM0:
on windows, for serial port > 9, "\\\\.\\" is needed before "COM"
These characters are also compatible with port <= 9

more details here: https://stackoverflow.com/questions/27909666/createfile-serial-communication-issue

### DIFF
--- a/src/core/cli_arg.cpp
+++ b/src/core/cli_arg.cpp
@@ -111,6 +111,7 @@ bool CliArg::find_path(std::string& rest)
                     _path = "";
                     return false;
                 }
+            _path = "\\\\.\\" + _path;
             }
         } else {
             LogWarn() << "Invalid serial path";


JonasVautherin:
Thanks for the contribution!

CI failures (at least some) are unrelated. Let's hope #1175 fixes this, too. Please rebase when #1175 is merged :+1:.

julianoes (approved):
Thanks @NicolasM0 but the syntax is very strange.

julianoes:
I've tested this locally. Works for me on COM3. Merging!

NicolasM0:
> Thanks @NicolasM0 but the syntax is very strange.

Yes that is strange... after checking microsoft docs (https://docs.microsoft.com/fr-fr/windows/win32/api/fileapi/nf-fileapi-createfilea?redirectedfrom=MSDN#communications-resources), the correct syntax seems to be '\\.\\COMX'  (with one less \ ) but it work with double \

YogeshHC:
@julianoes @NicolasM0  `.\battery.exe Serial://COM16:57600` is this the correct way to connect to pixhawk(telemetry) over serial. 
When I try to run `Battery.exe` from MAVSDK examples I get `Unhandled exception at 0x00007FFC15F93B29 in battery.exe: Microsoft C++ exception: std::bad_alloc at memory location 0x00000065741BF020`.

julianoes:
@YogeshHC oh, that doesn't sound good.
Could you please create a new issue with this. And please also add information of what version you built and steps for someone to reproduce this.

PR #1177, opened 2020-08-26
Remove sysid from UDP remotes, which induced duplicated messages

JonasVautherin:
Until this PR, a UDP remote is identified by an `<ip, port, sysid>`. This means that if two systems exist on the same `UDPConnection` (which typically happens when you have an onboard component connecting to both the FCU and a ground station), two remotes will be created. As a consequence, broadcast messages (such as `HEARTBEAT` or `TIMESYNC`) will be sent twice on that `<ip, port>`, because of the different `sysid`s. This should not happen.

This PR removes the `sysid` from the remotes, so that they are now identified with `<ip, port>`. I believe that the only consequence of that (apart from not sending duplicated messages) is that directed messages will be sent to all remotes instead of being filtered by `target_system_id`. I believe this is fine, because that's anyway the case for all the broadcast messages.

Note: it seems to me that there was duplicated code there, too, that I removed. Not sure if that has consequences, but I don't think so.

### DIFF
--- a/src/core/udp_connection.cpp
+++ b/src/core/udp_connection.cpp
@@ -142,9 +142,14 @@ bool UdpConnection::send_message(const mavlink_message_t& message)
              reinterpret_cast<const uint8_t*>(message.payload64)[entry->target_system_ofs] :
              0);
 
+    // Send the message to all the remotes. A remote is a UDP endpoint
+    // identified by its <ip, port>. This means that if we have two
+    // systems on two different endpoints, then messages directed towards
+    // only one system will be sent to both remotes. The systems are
+    // then expected to ignore messages that are not directed to them.
     bool send_successful = true;
     for (auto& remote : _remotes) {
-        if (target_system_id != 0 && remote.system_id != target_system_id) {
+        if (target_system_id != 0) {
             continue;
         }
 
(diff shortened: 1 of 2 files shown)


julianoes:
> is that directed messages will be sent to all remotes instead

I think I disagree with that, they should still be filtered. However, the broadcast ones should not be filtered but also not broadcast to all remotes but only once. Could we do that?

JonasVautherin:
> the broadcast ones should not be filtered but also not broadcast to all remotes but only once

Why is that? A broadcast should be broadcasted, to me that means to all remotes... or doesn't it? :thinking:

julianoes:
Yes but what about the messages that are not broadcast?

JonasVautherin:
I don't know... nothing says that a sysid is linked to one UDP <dst_ip, dst_port>, right? Say you do multipath, you could have two connections coming from two different <dst_ip, dst_port> but linking to the same `UdpConnection` with the same sysid, couldn't you? Filtering them now is an optimization assuming that nobody wants to do multipath.

But it's your call, I could add a list per remote of `sysid` that have been observed (by receiving heartbeats, or by sending heartbeats), and directed messages would only be sent over the remotes that know the `sysid`. Should I do that?

julianoes:
If you do multipath then you could have two remotes with that sysid and you would use both paths again, no?

JonasVautherin:
Yes you're right. And because most messages are broadcasts, you will always somehow discover the `sysid`s linked to the remotes. I'll change that.

julianoes (approved):

julianoes:
@JonasVautherin can you check CI?
https://github.com/mavlink/MAVSDK/pull/1177/checks?check_run_id=1035794228#step:8:18

It's probably another mavlink extension that was added...

JonasVautherin:
Yes I reported it in https://github.com/mavlink/MAVSDK/issues/1178. Let me fix it quickly.

PR #1180, opened 2020-08-27
log_files: fix chunks_received size in download_log_file

NicolasM0:
I have always an error at the end of downloading a log file :
Sans titre

I use a mavsdk_server compiled on develop branch on Windows.

After digging a bit, I found that if log part size in not a multiple of MAVLINK_MSG_LOG_DATA_FIELD_DATA_LEN 
(it seems that there is no reason to be of a particular size),
chunks_received size must be "part_size / MAVLINK_MSG_LOG_DATA_FIELD_DATA_LEN +1"

### DIFF
--- a/src/core/mavsdk.cpp
+++ b/src/core/mavsdk.cpp
@@ -149,5 +149,4 @@ void Mavsdk::Configuration::set_usage_type(Mavsdk::Configuration::UsageType usag
     _usage_type = usage_type;
 }
 
-
 } // namespace mavsdk
--- a/src/core/mavsdk.h
+++ b/src/core/mavsdk.h
@@ -162,7 +162,7 @@ public:
          */
         uint8_t get_system_id() const;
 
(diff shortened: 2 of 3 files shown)


JonasVautherin (commented):
Interesting... Why could that be? Some null termination `\0`?

julianoes (approved):
I think that makes sense. Nice catch @NicolasM0.

JonasVautherin (changes requested):
Please update the branch (rebase `develop`) and fix the style (`./tools/fix_style.sh .`). Then it will be good to go :+1: :rocket:

NicolasM0:
> Please update the branch (rebase `develop`) and fix the style (`./tools/fix_style.sh .`). Then it will be good to go 👍 🚀

done !

JonasVautherin (approved):
Awesome! Thanks!

PR #1181, opened 2020-08-27
Fix examples following mavlink update

### LINKED ISSUE #1178: Fix examples following MAVLink update
As observed in https://github.com/mavlink/MAVSDK/pull/1102, it seems like #1176 broke the examples.

JonasVautherin:
Fixes examples, which were broken by an extension in the recent MAVLink update.

Resolves #1178.


julianoes (approved):
Ah, the examples! Thx.

PR #1184, opened 2020-09-02
Fixed loiter time (hold time) being ignored on QGC mission import.

### LINKED ISSUE #1183: Hold time is ignored for waypoints when importing with import_qgroundcontrol_mission
On waypoint import the hold time for a waypoint is ignored.
See https://github.com/mavlink/MAVSDK/blob/37be38dad47d575f29d76f886e7539e4f419ed67/src/plugins/mission/mission_impl.cpp `Mission::Result MissionImpl::build_mission_items`

All current code to parse a waypoint:
```
            if (command == MAV_CMD_NAV_WAYPOINT) {
                auto is_fly_through = !(int(params[0]) > 0);
                new_mission_item.is_fly_through = is_fly_through;
            }
            auto lat = params[4], lon = params[5];
            new_mission_item.latitude_deg = lat;
            new_mission_item.longitude_deg = lon;

            auto rel_alt = float(params[6]);
            new_mission_item.relative_altitude_m = rel_alt;
```
The hold time in params[0] is already used to determine if it is a 'fly-though' waypoint.
But then it is not used to set the hold time in seconds.

petergerten:
Resolves #1183.

### DIFF
--- a/src/plugins/mission/mission_impl.cpp
+++ b/src/plugins/mission/mission_impl.cpp
@@ -964,6 +964,7 @@ Mission::Result MissionImpl::build_mission_items(
                 auto is_fly_through = !(int(params[0]) > 0);
                 new_mission_item.is_fly_through = is_fly_through;
             }
+	    new_mission_item.loiter_time_s = params[0];
             auto lat = params[4], lon = params[5];
             new_mission_item.latitude_deg = lat;
             new_mission_item.longitude_deg = lon;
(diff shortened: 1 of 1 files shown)


JonasVautherin (changes requested):
Looks good!

Please fix style, though (regarding the CI failures): `./tools/fix_style.sh .`

JonasVautherin (approved):
Thanks a lot! :rocket:

PR #1185, opened 2020-09-03
workflows: enable Passthrough plugin for deb/rpm packages

julianoes:
I think it's mostly confusing if that's not available, so I suggest to add it by default.


MaEtUgR:
Also docs seem outdated: https://github.com/mavlink/MAVSDK-docs/pull/220

JonasVautherin (approved):

PR #1189, opened 2020-09-16
mission: fix potential segfault

julianoes:
I have seen a segfault happen in current_mission_item() and this is the only thing that potentially could have caused it, I think.


JonasVautherin (approved):

jtroeth1:
Hello,
Not sure if related...

I am seeing a segfault which has to be in these three lines:

System& uav = mavlink.system();
auto telemetry = std::make_shared<Telemetry>(uav);
auto action = std::make_shared<Action>(uav);

It has to be in these lines because i have no other mavlink lines in my code. I'm not sure whats going on in the background to cause this.
Any help appreciated.

Thanks,

JonasVautherin:
I think this is unrelated. Can you open a new issue with that, and specify which version of MAVSDK you are using?

PR #1192, opened 2020-09-24
telemetry: protect callbacks with mutex

julianoes:
Early on I forgot to protect the callbacks from access by different threads. This could lead to segfaults when a callback is subscribed or unsubscribed while being executed.


JonasVautherin (approved):
Good catch!

PR #1196, opened 2020-09-25
Move from C++11 to C++17 (second try)

julianoes:
Previous try was #884.


JonasVautherin:
Almost there, just missing iOS. I see a lot of errors of the sort: "only available on iOS 11 or newer":

```
error: aligned allocation function of type 'void *(std::size_t, std::align_val_t)' is only available on iOS 11 or newer
```

JonasVautherin (dismissed):

julianoes:
Rebased on top of develop/496795c1.

src/core/mavlink_parameters.h
```diff
@@ -238,25 +222,25 @@ class MAVLinkParameters {
 
         bool set_as_same_type(const std::string& value_str)
         {
-            if (_value.is<uint8_t>()) {
+            if (std::get_if<uint8_t>(&_value)) {
```
JonasVautherin, line 225: I suppose `_value.is` does not work with a variant, hence this change?

JonasVautherin (approved):
Nice! :rocket:

src/core/mavlink_parameters.h
julianoes: I think so. There is also `std::get` which throws instead: https://en.cppreference.com/w/cpp/utility/variant/get

julianoes:
:tada:

PR #1200, opened 2020-10-01
tools: update to clang-format-10

julianoes:
This is because we can build on Ubuntu 20.04 now, not 18.04 anymore.


JonasVautherin (approved):
Nice, I'll go back to `./tools/fix_style.sh .` instead of going through Docker, then :laughing:

PR #1205, opened 2020-10-06
MavlinkPassthrough: add API to send commands

julianoes:
This makes it possible to re-use the command retransmission in the passthrough plugin.

In the future we could also consider adding a plugin to send commands.


JonasVautherin (approved):
Nice!

PR #1210, opened 2020-10-06
Mission: use NAV_DELAY instead of LOITER_TIME

### LINKED ISSUE #1101: Mission generation should use NAV_DELAY instead of sending two waypoints on the same location
From @julianoes:
> The reason MAVSDK sends the same lat/lon twice in a row is because it adds a LOITER_TIME right after a NAV_WAYPOINT. The right approach would be to use NAV_DELAY [...].

Which is implemented in PX4 for quite some time (I verified). The reason PX4 rejects now waypoints that are on top of each other is because the path between those waypoints is not defined (zero distance). Often we however need to know where the vehicle is coming from geometrically and allowing N waypoints on the same location would force us to traverse the mission multiple times and quite a bit forward and backward, increasing the complexity of the autopilot in terms of memory, execution time and also in terms of our ability to QA and certify the code.

So in short: This is undesirable in principle and navigation commands should not exist in the same physical location one after the other.

coderkalyan:
Closes #1101, replacing MAV_CMD_NAV_LOITER_TIME with MAV_CMD_NAV_DELAY. Additionally, added checks for whether or not speed actually changed when inserting MAV_CMD_DO_CHANGE_SPEED mission items.

### DIFF
--- a/src/plugins/mission/mission_impl.cpp
+++ b/src/plugins/mission/mission_impl.cpp
@@ -281,10 +281,6 @@ MissionImpl::convert_to_int_items(const std::vector<MissionItem>& mission_items)
     std::vector<MAVLinkMissionTransfer::ItemInt> int_items;
 
     bool last_position_valid = false; // This flag is to protect us from using an invalid x/y.
-    MAV_FRAME last_frame;
-    int32_t last_x;
-    int32_t last_y;
-    float last_z;
 
     unsigned item_i = 0;
     _mission_data.mavlink_mission_item_to_mission_item_indices.clear();
@@ -314,10 +310,6 @@ MissionImpl::convert_to_int_items(const std::vector<MissionItem>& mission_items)
                 MAV_MISSION_TYPE_MISSION};
 
             last_position_valid = true; // because we checked has_valid_position
-            last_x = x;
-            last_y = y;
-            last_z = z;
-            last_frame = MAV_FRAME_GLOBAL_RELATIVE_ALT_INT;
 
             _mission_data.mavlink_mission_item_to_mission_item_indices.push_back(item_i);
             int_items.push_back(next_item);
@@ -369,9 +361,6 @@ MissionImpl::convert_to_int_items(const std::vector<MissionItem>& mission_items)
             }
         }
 
-        // FIXME: It is a bit of a hack to set a LOITER_TIME waypoint to add a delay.
-        //        A better solution would be to properly use NAV_DELAY instead. This
-        //        would not require us to keep the last lat/lon.
         // A loiter time of NAN is ignored but also a loiter time of 0 doesn't
         // make any sense and should be discarded.
         if (std::isfinite(item.loiter_time_s) && item.loiter_time_s > 0.0f) {
@@ -388,17 +377,17 @@ MissionImpl::convert_to_int_items(const std::vector<MissionItem>& mission_items)
 
                 MAVLinkMissionTransfer::ItemInt next_item{
                     static_cast<uint16_t>(int_items.size()),
-                    static_cast<uint8_t>(last_frame),
-                    MAV_CMD_NAV_LOITER_TIME,
+                    MAV_FRAME_MISSION,
+                    MAV_CMD_NAV_DELAY,
                     current,
                     autocontinue,
                     item.loiter_time_s, // loiter time in seconds
-                    NAN, // empty
-                    0.0f, // radius around waypoint in meters ?
-                    NAN, // don't change yaw
-                    last_x,
-                    last_y,
-                    last_z,
+                    -1, // no specified hour
+                    -1, // no specified minute
+                    -1, // no specified second
+                    0,
+                    0,
+                    0,
                     MAV_MISSION_TYPE_MISSION};
 
                 _mission_data.mavlink_mission_item_to_mission_item_indices.push_back(item_i);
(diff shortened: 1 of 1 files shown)


coderkalyan:
Just realized that some integration tests are failing. Lemme fix those before we merge :smile:

src/plugins/mission/mission_impl.cpp
```diff
@@ -371,14 +371,14 @@ MissionImpl::convert_to_int_items(const std::vector<MissionItem>& mission_items)
...
                     item.loiter_time_s, // loiter time in seconds
-                    NAN, // empty
-                    0.0f, // radius around waypoint in meters ?
-                    NAN, // don't change yaw
+                    -1, // no specified hour
+                    -1, // no specified minute
+                    -1, // no specified second
                     last_x,
                     last_y,
                     last_z,
```
JonasVautherin, line 384: I don't understand those changes. Where do they come from? Did the MAVLink message change?

JonasVautherin (commented):
Thanks for the changes! Did you find out why CI is failing?

src/plugins/mission/mission_impl.cpp
coderkalyan: @JonasVautherin see https://github.com/mavlink/MAVSDK/issues/1101, the idea was to switch over to MAV_CMD_NAV_DELAY.

coderkalyan:
@JonasVautherin Well I thought I figured it out and fixed it, but it still seems to be broken. The issue is regarding the logic to check if a speed change is needed - when the mission is uploaded, and then downloaded and reconstructed, something is going wrong on the reconstruction part and thats why the test is failing. I'll try to run the tests locally so I can solve the issue.

coderkalyan:
@JonasVautherin Fixed the error, the SITL test is passing now so we should be good to go :smiley:

JonasVautherin (approved):
LGTM

src/plugins/mission/mission_impl.cpp
```diff
@@ -269,15 +269,14 @@ MissionImpl::convert_to_int_items(const std::vector<MissionItem>& mission_items)
...
             _mission_data.mavlink_mission_item_to_mission_item_indices.push_back(item_i);
             int_items.push_back(next_item);
         }
 
-        if (std::isfinite(item.speed_m_s)) {
-            // The speed has changed, we need to add a speed command.
-
+        // only add a speed command if the speed value is valid,
+        // the speed has changed from the previous waypoint
+        if (std::isfinite(item.speed_m_s) && (item.speed_m_s != last_speed)) {
```
julianoes, line 279: The question is how the API should be interpreted.

There are two ways:
1. NAN means no change, so don't insert the item.
2. A value means change it to that value, if it has changed in the meantime or not.

The [docstring](https://github.com/mavlink/MAVSDK/blob/1f784fec29c0a04877c6fa9b7d1213d3ac6a0a1f/src/plugins/mission/include/plugins/mission/mission.h#L92-L93) says:

> Speed to use after this mission item (in metres/second)

I can see how this is a nice optimization for someone that always sets the speed of a mission item, however, I think that's not how the API was intended to be used in the first place. You should only set it if you actually want to set it. In previous versions this was a bit clearer because a mission item used to be a class with setters/getters and so by default speed was not set, unless you would do a `set_speed(speed)`.

src/plugins/mission/mission_impl.cpp
```diff
@@ -376,14 +374,14 @@ MissionImpl::convert_to_int_items(const std::vector<MissionItem>& mission_items)
...
                     item.loiter_time_s, // loiter time in seconds
-                    NAN, // empty
-                    0.0f, // radius around waypoint in meters ?
-                    NAN, // don't change yaw
+                    -1, // no specified hour
+                    -1, // no specified minute
+                    -1, // no specified second
                     last_x,
                     last_y,
                     last_z,
```
julianoes, line 387: I think these 3 can be set to 0, as they are marked reserved.

src/plugins/mission/mission_impl.cpp
```diff
@@ -507,6 +505,7 @@ std::pair<Mission::Result, Mission::MissionPlan> MissionImpl::convert_to_result_
 
         MissionItem new_mission_item{};
         bool have_set_position = false;
+        float cached_speed = 0.0;
```
julianoes, line 508:
```suggestion
        float cached_speed = 0.0f;
```

src/plugins/mission/mission_impl.cpp
```diff
@@ -531,6 +530,10 @@ std::pair<Mission::Result, Mission::MissionPlan> MissionImpl::convert_to_result_
 
                 new_mission_item.is_fly_through = !(int_item.param1 > 0);
 
+                // we may or may not get a change speed command, so assume the speed is the
+                // cached_speed for now
+                new_mission_item.speed_m_s = cached_speed;
```
julianoes, line 535: Shouldn't it be NAN if it's never been set? Now it would be set to 0, right?

src/plugins/mission/mission_impl.cpp
```diff
@@ -581,13 +584,19 @@ std::pair<Mission::Result, Mission::MissionPlan> MissionImpl::convert_to_result_
...
-                    new_mission_item.speed_m_s = int_item.param2;
+                    cached_speed = int_item.param2;
+                    new_mission_item.speed_m_s = cached_speed;
                 } else {
                     LogErr() << "Mission item DO_CHANGE_SPEED params unsupported";
                     result_pair.first = Mission::Result::Unsupported;
                 }
 
             } else if (int_item.command == MAV_CMD_NAV_LOITER_TIME) {
+                // MAVSDK doesn't use LOITER_TIME anymore, but it is possible a mission still uses
```
julianoes, line 595: Nice!

src/plugins/mission/mission_impl.cpp
```diff
@@ -581,13 +584,19 @@ std::pair<Mission::Result, Mission::MissionPlan> MissionImpl::convert_to_result_
...
+                // it
+                new_mission_item.loiter_time_s = int_item.param1;
```
julianoes, line 597: We should either support minutes, hours, etc. as well, or at least print an error if they are set and we are ignoring them.

julianoes (changes requested):

src/plugins/mission/mission_impl.cpp
coderkalyan: Understood, I can get rid of this then.

coderkalyan:
@julianoes @JonasVautherin I got rid of the "speed optimization" commit and cleaned up a little as per your suggestions. Please let me know if anything else needs to be done!

src/plugins/mission/mission_impl.cpp
```diff
@@ -5,6 +5,7 @@
 #include <fstream> // for `std::ifstream`
 #include <sstream> // for `std::stringstream`
 #include <cmath>
+#include <ctime>
```
julianoes, line 8: Is this required?

coderkalyan: @julianoes good catch, actually it's not - I added it when considering implementing the wait until hour/min/sec params but then decided to not implement that for now.

coderkalyan: Ok, fixed :)

julianoes (approved):
I think this looks good now, thanks!

JonasVautherin (approved):
CI failure is unrelated

PR #1218, opened 2020-10-08
Added Mavlink command reception support

yhabib29:
I have added Mavlink command reception for my personal use as it does not seem to be implemented yet.
I share it here in case someone is interested.

### DIFF
--- a/src/core/mavlink_commands.cpp
+++ b/src/core/mavlink_commands.cpp
@@ -20,10 +20,22 @@ MAVLinkCommands::MAVLinkCommands(SystemImpl& parent) : _parent(parent)
         MAVLINK_MSG_ID_COMMAND_ACK,
         std::bind(&MAVLinkCommands::receive_command_ack, this, std::placeholders::_1),
         this);
+
+    _parent.register_mavlink_message_handler(
+        MAVLINK_MSG_ID_COMMAND_LONG,
+        std::bind(&MAVLinkCommands::receive_command_long, this, std::placeholders::_1),
+        this);
+
+    _parent.register_mavlink_message_handler(
+        MAVLINK_MSG_ID_COMMAND_INT,
+        std::bind(&MAVLinkCommands::receive_command_int, this, std::placeholders::_1),
+        this);
 }
 
 MAVLinkCommands::~MAVLinkCommands()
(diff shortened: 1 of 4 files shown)


julianoes (commented):
Interesting, thanks for the pull request!

Could you describe your use case? I assume your plugin needs to be able to respond to a command but you don't want to subscribe two both long and int in every plugin?

I wonder if this should be added to a separate class e.g. `MavlinkCommandReceiver` because it doesn't really have anything to do with the existing `MavlinkCommands`, right? We could then also rename `MavlinkCommands` to `MavlinkCommandSender`.

yhabib29:
Yes, as you could guess, I am receiving commands on multiple plugins, I filter the target system and component ID, handle the command and send back a CMD_ACK. For instance, non-mavlink cameras directly connected to a companion computer with a running MAVSDK application, each camera having a different component ID and a separate instance of my custom plugin. When the plugin receives the command, it filters the system and component ID, trigger the user callback and send a CMD_ACK.

One thing I should have done here, is to filter the target system ID before calling the plugin callback, i.e. at the `MavlinkCommands` level.
Moreover, the CMD_ACK could also be sent from the `MavlinkCommands` class, according to the plugin callback result, ensuring we send an ack for each command received.

As you said, my changes have nothing to do with the existing `MavlinkCommands`, and it could be split in 2 classes. I cannot tell you which choice is the best, both suit me well.

JonasVautherin:
> As you said, my changes have nothing to do with the existing MavlinkCommands, and it could be split in 2 classes. I cannot tell you which choice is the best, both suit me well.

I like @julianoes' idea of having `MavlinkCommandReceiver` and `MavlinkCommandSender`, if that's easily feasible :+1: :blush:.

TSC21:
@yhabib29 do you have time to continue and finish this in the mean time? Any time window you can provide? Thanks!

TSC21:
Continued in https://github.com/mavlink/MAVSDK/pull/1236.

julianoes:
Thanks @yhabib29, it will come in with #1236.

yhabib29:
Thanks @TSC21 for your work, I just saw your message today ...

PR #1220, opened 2020-10-09
Added API for Distance sensor in telemetry plugin 

hrnbot:
Added API for Distance sensor in telemetry plugin .


JonasVautherin:
May I ask why you end up opening and closing so many PRs on this repo? It is just flooding my notifications, meaning that I am less prone to answer at all (I don't have infinite resources to always monitor all of them).

PR #1223, opened 2020-10-11
backend: build statically with musl

julianoes:
@JonasVautherin here is your work according to https://stackoverflow.com/questions/61799147/link-libc-statically.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -124,7 +124,26 @@ jobs:
         with:
           submodules: recursive
       - name: configure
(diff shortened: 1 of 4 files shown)


JonasVautherin:
Nice! Now I wonder how we should enable that. Somehow test if libc is musl, and use the `CMAKE_BUILD_SHARED=OFF` flag to enable that? And then make some dockcross images for the archs we want?

julianoes:
@JonasVautherin right, I was wondering too. Something like that.

PR #1228, opened 2020-10-16
Add meson build

coderkalyan:
This PR adds support for the meson build system. Currently, only the standard cpp-only library is built, and tests are not yet enabled. I will add gRPC building in another PR, as well as tests - but I'm not sure how exactly tests work yet so that will take some more time.

To try out:
> cd MAVSDK
> meson build
> cd build
> meson compile
> sudo meson install

Then, take a look at the takeoff_and_land example. There is a meson.build there that builds fine, or you can also use CMakeLists.txt by uncommenting the parts labeled "for meson". In the future I hope to make the usage process the same, but this requires generating a cmake config file with meson, which I don't yet know how to do.


julianoes:
@coderkalyan thanks for doing this.

One question that I wonder: Would it not be easier to just use meson to build it as a cmake project?

Given it's easier to add support for meson like you did, I'm not opposed, however, a meson build needs to be added to CI. Otherwise we are inevitably going to break it.

src/plugins/action/meson.build
```diff
@@ -0,0 +1,20 @@
+action_include = include_directories(['include'])
+
+# TODO: target properties
```
JonasVautherin, line 3: Out of curiosity: what's that? :blush:

JonasVautherin (commented):
Why is it that we duplicate the third party CMakeLists.txt into `subprojects` now?

And would you mind commenting a bit on those `*.wrap`? I don't know meson, I would love to get some insights :slightly_smiling_face:.

src/plugins/action/meson.build
coderkalyan: > One question that I wonder: Would it not be easier to just use meson to build it as a cmake project?
@julianoes It might be, but for one, meson <-> cmake interop is not 100% bug free, and I thought it would be nicer to have a "pure" meson build system.

> Given it's easier to add support for meson like you did, I'm not opposed, however, a meson build needs to be added to CI. Otherwise we are inevitably going to break it.
If you can point me to what needs to be done for adding it to CI (I've never used Travis), I'd be happy to add a meson workflow.

coderkalyan:
> One question that I wonder: Would it not be easier to just use meson to build it as a cmake project?

@julianoes It might be, but for one, meson <-> cmake interop is not 100% bug free, and I thought it would be nicer to have a "pure" meson build system.

> Given it's easier to add support for meson like you did, I'm not opposed, however, a meson build needs to be added to CI. Otherwise we are inevitably going to break it.

If you can point me to what needs to be done for adding it to CI (I've never used Travis), I'd be happy to add a meson workflow.

src/plugins/action/meson.build
coderkalyan: @JonasVautherin https://github.com/mavlink/MAVSDK/blob/develop/src/plugins/action/CMakeLists.txt#L10 Just some flags, nNot crucial to the build I think but I should add them in anyway :smiley:

coderkalyan:
> Why is it that we duplicate the third party CMakeLists.txt into subprojects now?
@JonasVautherin A side effect of how meson handles dependencies is that all subprojects need to be in the `/subprojects` directory. I know its not ideal, but meson's point of view is that it makes dependency handling cleaner for both the build system and us if all subprojects are in a specific directory.

coderkalyan:
> And would you mind commenting a bit on those *.wrap? I don't know meson, I would love to get some insights slightly_smiling_face.

@JonasVautherin This is my first time using `.wrap`s, but they are basically meson's dependency management solution, so basically you write a `.wrap` file containing respository/version information and any patches needed, and meson clones or downloads them into `subprojects/` automatically at buildtime. I actually didn't write these wraps, I got them from a public repository: https://wrapdb.mesonbuild.com/. Since most of MAVSDK's dependencies are popular C/C++ libraries, I found the majority of them on the wrapdb. For the ones that didn't have wraps, I just cloned them into `subprojects/` and added them as cmake subprojects. In the future I might write wraps by hand for the remaining projects and upload them to the WrapDB so we can use them.
For more information: https://mesonbuild.com/Wrap-dependency-system-manual.html

julianoes:
@coderkalyan for a CI check, please add a runner like this one https://github.com/mavlink/MAVSDK/blob/1f784fec29c0a04877c6fa9b7d1213d3ac6a0a1f/.github/workflows/main.yml#L55-L77

.github/workflows/main.yml
```diff
@@ -76,6 +76,21 @@ jobs:
...
 
+  ubuntu20-meson-non-superbuild:
+    name: ubuntu-20.04 (meson, non-backend, non-superbuild)
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: recursive
+      - name: install meson
+        run: sudo -H pip3 install meson
```
julianoes, line 87:
```suggestion
        run: pip3 install --user meson
```

julianoes:
@coderkalyan what's the status here? Are you still planning to add the backend here as well? Or are you ok if we just add docs to make sure that MAVSDK can be compiled as a meson cmake subproject (whatever it's called)? I hope you understand that having to maintain two parallel build systems is not really what maintainers are thrilled about.

julianoes:
@coderkalyan I'm gonna close this. However, I would very much appreciate a pull request to the docs explaining how to build it as part of a meson project.

PR #1234, opened 2020-10-21
Fix camera definition loading

JonasVautherin:
* Remove hack that seems obsolete and creates unwanted warnings
* First try to download the camera definition file, then fallback to stored files
* Only load camera definition once (it was doing it repeatedly before)

### DIFF
--- a/src/plugins/camera/camera_definition.cpp
+++ b/src/plugins/camera/camera_definition.cpp
@@ -529,20 +529,6 @@ bool CameraDefinition::set_setting(
         return false;
     }
 
-    // FIXME: this is a hack because some params have the wrong type
-    //        when they come from the camera.
-    MAVLinkParameters::ParamValue changed_value = value;
-    if (name.compare("CAM_MODE") == 0 || name.compare("CAM_COLORMODE") == 0 ||
-        name.compare("CAM_EXPMODE") == 0 || name.compare("CAM_METERING") == 0 ||
-        name.compare("CAM_PHOTOFMT") == 0 || name.compare("CAM_PHOTOQUAL") == 0 ||
-        name.compare("CAM_VIDRES") == 0 || name.compare("CAM_WBMODE") == 0 ||
-        name.compare("CAM_COLORENCODE") == 0 || name.compare("CAM_FLICKER") == 0) {
-        if (changed_value.is_uint8()) {
-            uint8_t temp = changed_value.get_uint8();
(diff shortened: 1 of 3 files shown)


src/plugins/camera/camera_definition.cpp
```diff
@@ -529,20 +529,6 @@ bool CameraDefinition::set_setting(
...
-    //        when they come from the camera.
-    MAVLinkParameters::ParamValue changed_value = value;
-    if (name.compare("CAM_MODE") == 0 || name.compare("CAM_COLORMODE") == 0 ||
-        name.compare("CAM_EXPMODE") == 0 || name.compare("CAM_METERING") == 0 ||
-        name.compare("CAM_PHOTOFMT") == 0 || name.compare("CAM_PHOTOQUAL") == 0 ||
-        name.compare("CAM_VIDRES") == 0 || name.compare("CAM_WBMODE") == 0 ||
-        name.compare("CAM_COLORENCODE") == 0 || name.compare("CAM_FLICKER") == 0) {
-        if (changed_value.is_uint8()) {
-            uint8_t temp = changed_value.get_uint8();
-            changed_value.set_uint32(uint32_t(temp));
```
julianoes, line 542: But removing this will cause segfaults with H520, right?

julianoes (commented):
One comment, the rest looks fine.

src/plugins/camera/camera_definition.cpp
julianoes: It might not segfault anymore because of these checks:
https://github.com/mavlink/MAVSDK/blob/1f784fec29c0a04877c6fa9b7d1213d3ac6a0a1f/src/core/mavlink_parameters.h#L416-L419

JonasVautherin: I haven't seen such warning on my H520 :thinking:. I'll do more testing.

JonasVautherin: I can set `CAM_EXPMODE` and `CAM_COLORMODE` without a problem on my E90. I'm improving my script to make it easier to test others, but I haven't seen segfaults until now.

julianoes: Sounds good!

JonasVautherin: Tested with `CAM_PHOTOQUAL`, `CAM_PHOTOFMT`, I think it works.

julianoes (approved):

PR #1235, opened 2020-10-21
Immediately call new_system_callback if a system is already present

JonasVautherin:
Otherwise code like this never triggers if the system is detected before the callback is set:

```python
    print("Waiting for drone to connect...")
    async for state in drone.core.connection_state():
        if state.is_connected:
            print(f"Drone discovered")
            break
```


julianoes (approved):
Oh, right! Thanks for finding this!

PR #1239, opened 2020-10-22
curl: patch file for libressl librt linking

dakejahl:
If trying to use CMAKE_USE_OPENSSL in curl, the build fails with this error
```
[100%] Linking C executable curl
/work/install/lib/libcrypto.a(getentropy_linux.c.o): In function `getentropy_fallback':
/work/build/linux-armv6/third_party/libressl/libressl/src/libressl/crypto/compat/getentropy_linux.c:357: undefined reference to `clock_gettime'
/work/build/linux-armv6/third_party/libressl/libressl/src/libressl/crypto/compat/getentropy_linux.c:427: undefined reference to `clock_gettime'
```
Added patch file for curl when using **CMAKE_USE_OPENSSL**, this forces linking against rt when building curl.

This now allows HTTPS URLs with curl -- tested on a raspberry pi successfully.

### DIFF
--- a/third_party/curl/CMakeLists.txt
+++ b/third_party/curl/CMakeLists.txt
@@ -37,5 +37,6 @@ ExternalProject_add(
     curl
     URL https://github.com/curl/curl/archive/curl-7_70_0.zip
     PREFIX curl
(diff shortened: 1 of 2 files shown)


JonasVautherin (approved):
Thanks a lot!

Issue #1213, opened 2020-10-08
Offboard disabled when processing other components heartbeat


yhabib29:
When using Offboard plugin, I have noticed that the plugin is processing heartbeats coming from other components on the same system (Camera, Gimbal, ...).
My workaround is to filter Heartbeats mavlink message based on the message component ID, using default Autopilot value MAV_COMP_ID_AUTOPILOT1.

I just added these lines at the beginning of the "process_heartbeat" function:
```
void OffboardImpl::process_heartbeat(const mavlink_message_t& message)
{
    // Process only Heartbeat coming from the autopilot
    if (message.compid != MAV_COMP_ID_AUTOPILOT1) {
        return;
    }
```

JonasVautherin:
I believe it makes sense. @julianoes: Is that correct?

julianoes:
Good point! We would appreciate a PR with the fix.

yhabib29:
No problem, I will try to do it during this week-end.

PR #1242, opened 2020-10-25
Filter heartbeat in Offboard plugin to consider only autopilot heartbeats

yhabib29:
Linked to issue #1213


JonasVautherin (changes requested):
Seems like you'll need to fix the style :see_no_evil:.

This should do it:

```
./tools/fix_style.sh .
```

yhabib29:
Yep thanks, it should be OK now ?

julianoes:
Thanks @yhabib29!

PR #1243, opened 2020-10-25
Ignore CAM_MODE when processing 'normal' parameters

JonasVautherin:
When we get `CAM_MODE`, we hardcode it as an `uint32_t`, without honoring the type set in the camera definition file. This is fine, because we do not treat `CAM_MODE` like a normal camera parameter: we get (save) and set it independently.

However, this mismatch between the type in the camera definition and the hardcoded one causes _a lot_ of warnings to appear when processing normal parameters. This PR fixes that by not processing `CAM_MODE` together with the normal parameters.

At the same time, I removed a few workarounds that tried to fix issues on the camera side (i.e. the camera definition says a param is one type, and the actual param sent over mavlink is another type), because that really confused me a lot while debugging. If a camera messes up with the types, I find it okay for MAVSDK to ignore that parameter and report it, such that the camera side can be fixed.

Tested against the E90 and the AuterionSim.

### DIFF
--- a/src/core/mavlink_parameters.cpp
+++ b/src/core/mavlink_parameters.cpp
@@ -35,12 +35,6 @@ void MAVLinkParameters::set_param_async(
     const void* cookie,
     bool extended)
 {
-    // if (value.is_float()) {
-    //     LogDebug() << "setting param " << name << " to " << value.get_float();
-    // } else {
-    //     LogDebug() << "setting param " << name << " to " << value.get_int();
-    // }
-
     if (name.size() > PARAM_ID_LEN) {
         LogErr() << "Error: param name too long";
         if (callback) {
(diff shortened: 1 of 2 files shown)


JonasVautherin:
Hmmm... it seems like not processing `CAM_MODE` there removes the filtering. So `possible_settings` will show all the settings, whereas before it was showing only the possible settings for the current `CAM_MODE` (which makes sense).

@julianoes: I'm a bit out of my depth there. Do you remember how this works? When processing `CAM_MODE`, does it add some params to this "exclusion" list, or something?

JonasVautherin:
Actually it doesn't work, because in order to extract the exclusions, we need to compare the value of the `CAM_MODE` options between the param we received (over MAVLink) and what we expect (from the camera definitions). Hardcoding `CAM_MODE` to `uint32_t` makes a mess, I believe (because at this point, the param value is an `Any`, and it requires a big ugly switch to extract and cast it to `uint32_t`). Not to say that this is super convoluted and opens opportunities to other nasty bugs like that.

So my next idea would be to not hardcode `CAM_MODE` as an `uint32_t`, but use its real type, defined in the camera definition. Question then: can we live with not knowing the `CAM_MODE` unless we get a camera definition? I'll investigate tomorrow.

PR #1246, opened 2020-10-26
Hardcode `CAM_MODE` to `uint32_t` to remove type warnings

JonasVautherin:
New try after failing on https://github.com/mavlink/MAVSDK/pull/1243.

This PR now hardcodes `CAM_MODE` everywhere as `uint32_t` and forces it to come from the `CAMERA_SETTINGS` message, and to be set through the `MAV_CMD_SET_CAMERA_MODE` message.

What makes it complex is that this parameter can be defined in the camera definition file, with exclusions and all, but the camera definition file is optional, and consequently `CAM_MODE` has a mavlink command message just for itself. In the absence of a camera definition file, we hardcode that enum to a `uint32_t`. But if later a camera definition file arrives, it may redefine it completely.

To be really compliant with the MAVLink implementations, MAVSDK should account for all that, with the camera definition taking precedence in case `CAM_MODE` is defined both in `CAMERA_SETTINGS` and in the camera definition. Which IMO makes it weird, because it could make the `CAMERA_SETTINGS` message wrong. Anyway, I don't need that right now, so that could be contributed later (good luck :see_no_evil:).

### DIFF
--- a/src/core/mavlink_parameters.cpp
+++ b/src/core/mavlink_parameters.cpp
@@ -35,11 +35,12 @@ void MAVLinkParameters::set_param_async(
     const void* cookie,
     bool extended)
 {
-    // if (value.is_float()) {
-    //     LogDebug() << "setting param " << name << " to " << value.get_float();
-    // } else {
-    //     LogDebug() << "setting param " << name << " to " << value.get_int();
-    // }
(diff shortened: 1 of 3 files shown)


JonasVautherin:
Grmbl, working for AuterionSim, but now not working for the Yuneec camera definition files.

PR #1247, opened 2020-10-29
Fix cam mode warnings

JonasVautherin:
* If camera definition defines `CAM_MODE`, save it with the right type when coming from `CAMERA_SETTINGS`
* Pass shared_ptr<promise> by value instead of reference, so that it behaves like one would expect
* Send correct result in the `InProgress` callback, so that it can properly be ignored (otherwise the promise is resolved twice when it happens)


julianoes (approved):
:heart_eyes:

JonasVautherin:
ActionHover is failing in integration tests, but I think it is unrelated.

julianoes:
Yes, it has been failing before.

PR #1250, opened 2020-11-05
fix examples to work with 0.34.0

yashom7:
In _examples/fly_multiple_drones/fly_multiple_drones.cpp_ and _examples/multiple_drones/multiple_drones.cpp_
`if (systems.size() < num_systems_discovered)` should be if `(systems.size() > num_systems_discovered)` otherwise 
``if (num_systems_discovered != total_ports_used)`` is always evaluated true and code aborts. 

```
    for (auto system : mavsdk.systems()) {
        std::thread t(&takeoff_and_land, std::ref(system));
        threads.push_back(std::move(t));
        sleep_for(seconds(1));
    }
```
If there is no time interval between thread creation it seems that the first vehicle rejects the initial command sent to it and the thread aborts resulting in only one vehicle executing the example.

In _examples/takeoff_land/takeoff_and_land.cpp_ access vehicle after discovering it.

In _examples/follow_me/follow_me.cpp_ target location is incorrectly reported as vehicle location. Fixed that and added a reporting vehicle location using telemetry

### DIFF
--- a/examples/fly_multiple_drones/fly_multiple_drones.cpp
+++ b/examples/fly_multiple_drones/fly_multiple_drones.cpp
@@ -121,7 +121,7 @@ int main(int argc, char* argv[])
     mavsdk.subscribe_on_new_system([&mavsdk, &num_systems_discovered]() {
         const auto systems = mavsdk.systems();
 
-        if (systems.size() < num_systems_discovered) {
+        if (systems.size() > num_systems_discovered) {
             std::cout << "Discovered system" << std::endl;
             num_systems_discovered = systems.size();
         }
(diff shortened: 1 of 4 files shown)


JonasVautherin (changes requested):
Thanks! CI is failing because of the style check. Mind running `./tools/fix_style.sh .`? :blush:

yashom7:
> Thanks! CI is failing because of the style check. Mind running `./tools/fix_style.sh .`? 😊

I just tried that. Brew only lets me install clang-format-11/8 and the docker has 9 while MAVSDK wants 10. How do I get around this?

JonasVautherin (approved):

Issue #1173, opened 2020-08-24
Error message if no manual control is available


julianoes:
It would be nice to have an error message if starting altitude or position control using manual_control fails.

This would be similar as it is done for offboard: https://github.com/mavlink/MAVSDK/blob/758a6437826aa1e6a1abda5f109a7bca21599d61/src/plugins/offboard/offboard_impl.cpp#L43

yashom7:
Hi!
I think I can get on with implementing this.

Just to clarify. This error message reporting will go in the Manual Control plugin and is used if some means of control i.e. a joystick is not available?

afwilkin:
From what I am understanding (and can confirm by testing), manual control needs a setpoint before entering the mode. This is an example for offboard `offboard->set_velocity_body({0,` 0, 0, 0}); `
So, in code it looks like this:
`auto offboard = std::make_shared<Offboard>(*system);
 offboard->set_velocity_body({0, 0, 0, 0}); /* Needed */
 Offboard::Result offboard_result = offboard->start();`

Without the `offboard->set_velocity_body({0, 0, 0, 0}); ` then we get the Result shown in Julian's comment above (Offboard::Result::NoSetpointSet;). 

**Manual Control**
Manual Control needs to be started in a very similar fashion

`manual_control->set_manual_control_input(0.f, 0.f, 0.f, 0.f); /*Needed*/
 auto manual_control_result = manual_control->start_position_control();`

However, if the `manual_control->set_manual_control_input(0.f, 0.f, 0.f, 0.f);` is NOT there, manual control will not be achieved. This should be an error. However, the code returns ManualControl::Result::Success instead of a result that would indicate that manual control is not going to work. (For example: ManualControl::Result::NoSetpointSet)

**TODO**
Implement something like ManualControl::Result::NoSetpointSet that will tell the developer they need to insert `manual_control->set_manual_control_input(0.f, 0.f, 0.f, 0.f); ` into their code

@julianoes is that a fair analysis?

julianoes:
Sorry for the late response here.

> Just to clarify. This error message reporting will go in the Manual Control plugin and is used if some means of control i.e. a joystick is not available?

Yes.

> @julianoes is that a fair analysis?

100%, thanks for writing that up @afwilkin.

PR #1251, opened 2020-11-06
No setpoint manual control

yashomdighe:
For #1173

Essentially just implements the same thing that offboard does with `Mode::NotActive` for this by adding` enum class Input { NotSet, Set } _input = Input::NotSet`. 
One thing that I am not sure about is that offboard has a `stop_sending_setpoints()` method to reset the `_mode` to `Mode::NotActive` after it has been set by some other method. I dont know if this is also needed for manual control. If it is need I will need some insights to implement it.

Tested this by calling the `start_position_control()` without setting any setpoints first with `set_manual_control_input()` with the following result 

```
[01:45:25|Info ] MAVSDK version: v0.34.0-6-g5c2031d9 (mavsdk_impl.cpp:20)
Waiting to discover system...
[01:45:25|Info ] New system on: 127.0.0.1:14580 (with sysid: 1) (udp_connection.cpp:190)
[01:45:25|Debug] New: System ID: 1 Comp ID: 1 (mavsdk_impl.cpp:417)
[01:45:25|Debug] Component Autopilot (1) added. (system_impl.cpp:314)
[01:45:25|Debug] Discovered 1 component(s) (UUID: 5283920058631409231) (system_impl.cpp:471)
Discovered system
Waiting for system to be ready
System is ready
Position control start failed: Unknown
```

I dont know if this is how @julianoes wanted it to be done so if its not the solution that is desired I can try to do what is required after discussing.

### DIFF
--- a/src/plugins/manual_control/manual_control_impl.cpp
+++ b/src/plugins/manual_control/manual_control_impl.cpp
@@ -27,9 +27,9 @@ void ManualControlImpl::enable() {}
 void ManualControlImpl::disable() {}
 
 void ManualControlImpl::start_position_control_async(const ManualControl::ResultCallback callback)
-{   
-    if (_input == Input::NotSet){
-        if(callback){
+{
+    if (_input == Input::NotSet) {
+        if (callback) {
             callback(ManualControl::Result::Unknown);
         }
         return;
@@ -44,10 +44,10 @@ void ManualControlImpl::start_position_control_async(const ManualControl::Result
 
 ManualControl::Result ManualControlImpl::start_position_control()
 {
-    if (_input == Input::NotSet){
+    if (_input == Input::NotSet) {
         return ManualControl::Result::Unknown;
     }
-    
+
     auto prom = std::promise<ManualControl::Result>();
     auto fut = prom.get_future();
 
@@ -58,8 +58,8 @@ ManualControl::Result ManualControlImpl::start_position_control()
 
 void ManualControlImpl::start_altitude_control_async(const ManualControl::ResultCallback callback)
 {
-    if (_input == Input::NotSet){
-        if(callback){
+    if (_input == Input::NotSet) {
+        if (callback) {
             callback(ManualControl::Result::Unknown);
         }
         return;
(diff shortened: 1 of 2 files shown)


src/plugins/manual_control/manual_control_impl.cpp
```diff
@@ -28,6 +28,13 @@ void ManualControlImpl::disable() {}
 
 void ManualControlImpl::start_position_control_async(const ManualControl::ResultCallback callback)
 {
+    if (_input == Input::NotSet) {
+        if (callback) {
+            callback(ManualControl::Result::Unknown);
+        }
+        return;
+    }
```
JonasVautherin, line 36: Hmm I think it should be called more [like this](https://github.com/mavlink/MAVSDK/blob/ba2d819a3f6b2c0ea0279f6b45082e844c547051/src/plugins/manual_control/manual_control_impl.cpp#L144-L145) (I haven't tested it, though):

```suggestion
    if (_input == Input::NotSet) {
        if (callback) {
            auto temp_callback = callback;
            _parent->call_user_callback(
                [temp_callback]() { temp_callback(ManualControl::Result::Unknown); });
        }
        return;
    }
```

@julianoes: does that seem correct?

JonasVautherin (changes requested):
Just a small remark regarding the callback, otherwise this seems good to me.

Note: CI failure is unrelated.

src/plugins/manual_control/manual_control_impl.cpp
yashom7: Should I make the changes or wait for Julian's review?

JonasVautherin: You can make the change (and test it, I did not :see_no_evil:  :innocent:).

yashomdighe: I tested it. Both styles of callback are working. If the method that you suggest is preferred due to any reason I will make the change and commit

JonasVautherin: Yes it is preferred, because then it goes through the event loop instead of keeping the thread for itself. Something like that :smile:.

JonasVautherin (approved):
Thanks!

JonasVautherin:
For some reason, gRPC is failing to build on manylinux2014-aarch64, I don't understand why this happens now... Anyway I'm trying to fix that by updating gRPC, sorry for the delay :innocent:.

PR #1254, opened 2020-11-13
Add PositionVelocity to offboard

JonasVautherin:
This adds a way to send a position and a velocity together, where the velocity is used as feed-forward.

@Jaeyoung-Lim: Does that seem correct?


Jaeyoung-Lim (dismissed):
Thanks!

I think it is a little implicit that yaw field in `VelocityNedYaw` will be ignored, but I think it is reasonable

JonasVautherin:
Agreed, I will look into updating the API to improve that :+1:.

julianoes (approved):

PR #1255, opened 2020-11-13
Update dependencies

JonasVautherin:
For some reason, manylinux2014-aarch64 stopped building gRPC. A fix was merged in gRPC (though I don't know why the bug appeared now but well...), and so I'm updating gRPC to get the fix.

At the same time, gRPC now depends on re2 :confused:, which I therefore added.

I took the opportunity to update curl, protobuf and libressl, too.


JonasVautherin:
I actually think that the change is in manylinux2014-aarch64 is here: https://github.com/dockcross/dockcross/commit/b4666459f32eaa226d4c1f446ed3fdbe60f47521. I don't get the `undefined reference` to `sleep_for`, though.

julianoes (dismissed):

julianoes (approved):
Awesome!

PR #1260, opened 2020-11-20
core: fix potential race on discovery

julianoes:
We should set the connected state correct before calling the callback.


JonasVautherin (approved):
Good catch! :rocket:

PR #1261, opened 2020-11-20
manual_control: use proper result if input not set

### LINKED ISSUE #1173: Error message if no manual control is available

julianoes:
A result/error of unknown is not a good user feedback. Instead we need
to communicate the actual reason.

Closes #1173.


JonasVautherin (approved):
Thanks!

PR #1262, opened 2020-11-20
CI: Run iOS jobs in parallel

JonasVautherin:
This splits the iOS jobs in order to run them in parallel. That requires a third job, that gets the frameworks built by each, package them into an xcframework, and deploys it (if it is a release).

If that works, it will divide the CI time by a factor of 2.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -377,17 +377,56 @@ jobs:
       - uses: actions/checkout@v2
         with:
           submodules: recursive
-      - name: configure (ios)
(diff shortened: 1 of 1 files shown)


julianoes (approved):
Nice! Why only 2 days retention?

JonasVautherin:
> Nice! Why only 2 days retention?

The default is 90, but I did not really see the point. We essentially just need it between the jobs, not really after that. Do you think it should be longer?

julianoes:
Oh got it. In that case it's of course enough.

julianoes (approved):
Awesome!

PR #1266, opened 2020-11-23
third_party: use git apply, not patch

julianoes:
Because we don't necessarily have patch on Win.


JonasVautherin (approved):
Oh right, I should have caught that earlier. Thanks!

PR #1267, opened 2020-11-24
Fix Windows build when MAVSDK with SUPERBUILD is external project

julianoes:
Does that make any sense? It seems to work.


julianoes:
I would assume that's unrelated:
```
 [12:37:29|Info ] MAVSDK version: be404ace (mavsdk_impl.cpp:20)
[12:37:29|Error] bind error: Address already in use (udp_connection.cpp:85)
/__w/MAVSDK/MAVSDK/src/integration_tests/mission_transfer_lossy.cpp:21: Failure
  mavsdk.add_udp_connection()
    Which is: Bind error
  ConnectionResult::Success
    Which is: Success
```

JonasVautherin (approved):

PR #1268, opened 2020-11-24
Add option to build macos framework

JonasVautherin:
For Python, we need `mavsdk_server` as a macOS (almost-)static binary (the equivalent of manylinux).
For Swift, the idea would be to add macOS into the xcframework, such that it would "just work" with SwiftPM. That is what this PR is about.

It needs some testing with MAVSDK-Swift, and then the artifact would be rename from `mavsdk_server_ios.zip` to maybe `mavsdk_server.xcframework.zip`, or something like that.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -445,7 +445,7 @@ jobs:
         run: bash ./src/backend/tools/package_backend_framework.bash
       - uses: actions/upload-artifact@v2
         with:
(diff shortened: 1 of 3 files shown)


JonasVautherin:
Note for self: in the macOS framework, the `Modules` symbolic link is missing (`Modules -> Versions/Current/Modules`).

julianoes (approved):
:man_shrugging:

PR #1271, opened 2020-11-27
Handle Result as parameter

JonasVautherin:
A `Result` needs to be treated differently than other types in the templates. This is fixing the templates when `Result` is used as a parameter to a function.

@TSC21: FYI


TSC21:
Cool! Will try it out ans let you know

PR #1275, opened 2020-12-02
core: attempt to fix connection race

julianoes:
This is an attempt to prevent the race between the _connected flag and calling the on_change subscription.


julianoes:
I'm gonna ignore the ios CI error and merge this.

PR #1276, opened 2020-12-03
Fix ios_simulator to xcode 12.0.1, because it seems to fail with 12.2

JonasVautherin:
Let's see if that helps, but locally it builds with 12.2 :thinking:.


JonasVautherin:
Does not fix the issue.

PR #1277, opened 2020-12-03
Move CI to macOS 11.0 Big Sur

JonasVautherin:
Running in parallel of #1276. I now wonder if the issue is not coming from having Xcode 12.2 with macOS 10.5. This moves up to macOS 11.0.


JonasVautherin:
Not solving it either.

PR #1278, opened 2020-12-03
Fix iOS simulator build following update to Xcode 12

JonasVautherin:
Using latest version from https://github.com/leetal/ios-cmake

### DIFF
--- a/tools/ios.toolchain.cmake
+++ b/tools/ios.toolchain.cmake
@@ -335,6 +335,9 @@ if(USED_CMAKE_GENERATOR MATCHES "Xcode")
   endif()
 endif()
 
+# Required from Xcode 12 on, because it now builds for arm macbooks and we don't support that.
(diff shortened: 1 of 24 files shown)


JonasVautherin:
Aaaand this breaks the libreSSL builds already. :disappointed:

tools/ios.toolchain.cmake
```diff
@@ -335,6 +335,9 @@ if(USED_CMAKE_GENERATOR MATCHES "Xcode")
   endif()
 endif()
 
+# Required from Xcode 12 on, because it now builds for arm macbooks and we don't support that.
+set(CMAKE_XCODE_ATTRIBUTE_EXCLUDED_ARCHS[sdk=iphonesimulator*] "arm64")
```
julianoes, line 339: I was looking for that cmake argument but did not find it, nice!

julianoes (approved):

tools/ios.toolchain.cmake
JonasVautherin: Yeah you were right on that from the beginning :100:

PR #1279, opened 2020-12-04
Add cache to github workflows (for dependencies)

JonasVautherin:
Trying to use the [cache action](https://docs.github.com/en/free-pro-team@latest/actions/guides/caching-dependencies-to-speed-up-workflows). This should do the following:

1. Try to find a cached directory matching key `${{ github.job }}-${{ hashFiles("./third_party/**") }}`, where `github.job` is the unique name of the job (e.g. 'iOS'), and `hashFiles("./third_party/**")` computes one hash representing all the files in `third_party/`, which are CMakeLists and patches for our dependencies. Therefore, this should be a unique key for the job and for that set of dependencies.
2. If there is a cache hit, it means that `./build/ios/third_party/install` has been saved for that job, which would be e.g. all the dependencies for the 'iOS' job. In that case, we want to disable the superbuild and use those dependencies, and we do it by adding e.g. `-DSUPERBUILD=OFF -DCMAKE_PREFIX_PATH=$(pwd)/build/ios/third_pary/install` to the configure step later on.
3. When the configure step runs, if `${{ deps_source }}` is defined, it will run without superbuild and with the cached deps. Otherwise it will run the superbuild and will cache the installed dependencies at the end, for the next run of that job (or a later job where the dependencies have not changed).

First time I try that, let's hope it works :crossed_fingers:.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -101,7 +101,7 @@ jobs:
           submodules: recursive
       - name: install clang-format
         run: sudo apt-get update && sudo apt-get install -y clang-format-10
-      - name: install dcsdkgen
+      - name: install protoc-gen-mavsdk
         run: |
              cd proto/pb_plugins
              pip3 install --user -r requirements.txt
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -352,8 +352,16 @@ jobs:
       - uses: actions/checkout@v2
         with:
           submodules: recursive
+      - uses: actions/cache@v2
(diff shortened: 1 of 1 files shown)


JonasVautherin:
Except for the alpine build (for which I opened an issue [here](https://github.com/actions/cache/issues/479)), I think that it is working pretty well. The iOS builds take ~7min instead of almost an hour before, and others like the dockcross builds are ~10min faster :rocket:.

julianoes (approved):
Ship it!

PR #1281, opened 2020-12-05
Add openssl

JonasVautherin:
Because it's so much fun 😅.

More seriously, the reason for moving to OpenSSL instead of keeping LibreSSL is that we would want to use dTLS if we were to encrypt the MAVLink communications, and LibreSSL does not support dTLS.


JonasVautherin:
For the env variables, probably worth trying this:

```
Error: The `set-env` command is disabled. Please upgrade to using Environment Files or opt into unsecure command execution by setting the `ACTIONS_ALLOW_UNSECURE_COMMANDS` environment variable to `true`. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/
```

julianoes:
@JonasVautherin what's special about these env vars? Can't you just do `export FOO=$(blacommand)`?

JonasVautherin:
@julianoes: About the env variables, it's a new security feature in the github workflows. Now environment variables need to be set through the file `$GITHUB_ENV`, to prevent some vulnerability. It is easy to use, we just have to know about it :+1:.

JonasVautherin:
Note for self: cache is broken in dockcross builds, and inexistant in some musl builds. But no need to fix before OpenSSL works.

.github/workflows/main.yml
```diff
@@ -232,17 +165,34 @@ jobs:
...
+        run: dpkg-buildpackage -us -uc -b
+
   rpm-package:
     name: ${{ matrix.container_name }} (package, non-mavsdk_server)
     runs-on: ubuntu-18.04
     container: mavsdk/mavsdk-${{ matrix.container_name }}
     strategy:
       matrix:
-        container_name: [fedora-32, fedora-33]
+        container_name: [fedora-33]
```
julianoes, line 189: Why only Fedora 33? My idea is to always have 2, so always the two versions that are supported.

third_party/grpc/grpc_openssl_ios.patch
```diff
@@ -0,0 +1,45 @@
+diff --git a/src/core/tsi/ssl_transport_security.cc b/src/core/tsi/ssl_transport_security.cc
```
julianoes, line 1: Tricky stuff! Thanks for figuring that out.

third_party/re2/CMakeLists.txt
```diff
@@ -1,23 +1,20 @@
 cmake_minimum_required(VERSION 3.1)
 
-project(external-curl)
+project(external-re2)
```
julianoes, line 3: :smile:

.github/workflows/main.yml
JonasVautherin: I did not manage to install perl-FindBin on Fedora-32, and eventually gave up :sweat_smile:

PR #1287, opened 2020-12-10
tools: support docker and podman

julianoes:
This was accidentally changed to podman beforehand. I'm now changing it back to docker but also supporting podman if available.


JonasVautherin (approved):
Ok!

PR #1295, opened 2021-01-28
Add android-x86 and android-x86_64 to CI

JonasVautherin:
It got merged into dockcross, we can now start using it :tada:.


JonasVautherin:
Windows CI error is unrelated, nasm.us is down at the moment.

PR #1298, opened 2021-01-30
Added Orbit Command.

DoppleGangster:


JonasVautherin (changes requested):
Looks good! You'll have to fix the style, though, by running the corresponding script:

```
./tools/fix_style.sh .
```

DoppleGangster:
Style Updated. Thanks!

DoppleGangster:
Is the docs failure something that should have been fixed with the fix_style.sh?

JonasVautherin:
Oh, we're missing some more documentation in `action.proto` :see_no_evil:, but I think that's the last missing piece.

[The enum here](https://github.com/mavlink/MAVSDK-Proto/blob/main/protos/action/action.proto#L188-L194) should be documented (similar to e.g. [this one](https://github.com/mavlink/MAVSDK-Proto/blob/main/protos/camera/camera.proto#L218-L222)).

Would you mind opening a PR to MAVSDK-Proto with that change? I think after that we will be fine here, too :innocent:

DoppleGangster:
👍 Sure not a problem.

DoppleGangster:
I believe it should be good to go now.

JonasVautherin (approved):
Nice! Thanks a lot!

I just re-ran the CI because SITL failed on 1.11, but it seems fine otherwise :tada:

DoppleGangster:
No problem! Thanks for helping me along the way!

PR #1310, opened 2021-02-04
Gimbal v2: implement take_control and release_control

JonasVautherin:


src/plugins/gimbal/gimbal_protocol_v1.cpp
```diff
@@ -156,4 +156,56 @@ void GimbalProtocolV1::set_roi_location_async(
...
+    std::thread([this, &callback]() {
+        std::unique_lock<std::mutex> lock(_control_thread_mutex);
+
+        while (_control_callback) {
+            _system_impl.call_user_callback(
+                [this, callback]() { callback(_current_control_status); });
+            _control_thread_cv.wait_for(lock, std::chrono::seconds(1));
+        }
+    });
+}
```
julianoes, line 209: I don't really get that. I assume it's overkill that we could just ignore. I guess we'll leave it in now and clean it up later.

julianoes (approved):
Works for me against https://github.com/PX4/PX4-SITL_gazebo/pull/703 :tada:

src/plugins/gimbal/gimbal_protocol_v1.cpp
JonasVautherin: I just want to allow an application to use the same logic for both v1 and v2. So you need to get some status updates if you expect them, right? :blush:

julianoes: Hm, fair enough. If it works you can leave it in :)

PR #1314, opened 2021-02-09
fixed library versioning

mattes-bru:
I recently tried to build the mavsdk library in a yocto Bitbake environment, where it failed with a qa issue. the problem where that the plugins 'failure' and 'manual_control' don't provide a versioned so file. This patch fixed the issue. 

Feel free to reject this PR if you left out the version in these two plugins by purpose.


julianoes (approved):
Oh, that was an oversight. Thanks a lot!

JonasVautherin:
Awesome, thanks for the fix!

PR #1324, opened 2021-02-18
Add support for list_photos()

JonasVautherin:
Add a way to list photos either since the moment MAVSDK connected to the drone, or all the photos on the sdcard.

The camera plugin caches the information on the pictures that are taken, meaning that for getting the images SINCE_CONNECTION, it should mostly be in the cache. However, if an image is missing, MAVSDK will retrieve it.

When fetching all the images on the sdcard, it may take longer depending on how many images are there. The way it works is that it requests the "captured" message for the next photo, and waits until a (not necessarily this one) "captured" message is received. This way it does not take all the bandwidth by e.g. sending 2000 requests at once.

There are two mechanisms to avoid deadlocks:
1. If a "captured" message is not received 1 second after the request message has been sent, try again. After 3 tries, time out the whole `list_photos()` function.
2. It could happen that the condition_variable is unlocked by a "captured" message that is not the one we are expecting. In that case, the same image is being queried again. After three such trials, the `safety_count_boundary` counter makes it fail. I believe it should not happen unless the vehicle is behaving in a wrong way, but still we should not crash.

Tested against the Auterion Cloud Simulator and locally from MAVSDK-Python.

### DIFF
--- a/src/plugins/camera/camera_impl.cpp
+++ b/src/plugins/camera/camera_impl.cpp
@@ -139,6 +139,7 @@ void CameraImpl::manual_enable()
 {
     refresh_params();
 
+    request_status();
     request_camera_information();
     request_flight_information();
 
@@ -285,6 +286,19 @@ MavlinkCommandSender::CommandLong CameraImpl::make_command_request_camera_captur
     cmd_req_camera_cap_stat.target_component_id = _camera_id + MAV_COMP_ID_CAMERA;
 
     return cmd_req_camera_cap_stat;
+}
+
+MavlinkCommandSender::CommandLong
+CameraImpl::make_command_request_camera_image_captured(const size_t photo_id)
+{
+    MavlinkCommandSender::CommandLong cmd_req_camera_image_captured{};
+
+    cmd_req_camera_image_captured.command = MAV_CMD_REQUEST_MESSAGE;
+    cmd_req_camera_image_captured.params.param1 = static_cast<float>(MAVLINK_MSG_ID_CAMERA_IMAGE_CAPTURED);
+    cmd_req_camera_image_captured.params.param2 = static_cast<float>(photo_id);
+    cmd_req_camera_image_captured.target_component_id = _camera_id + MAV_COMP_ID_CAMERA;
+
+    return cmd_req_camera_image_captured;
 }
 
 MavlinkCommandSender::CommandLong CameraImpl::make_command_request_storage_info()
@@ -726,6 +740,15 @@ void CameraImpl::process_camera_capture_status(const mavlink_message_t& message)
             (camera_capture_status.image_status == 2 || camera_capture_status.image_status == 3);
         _status.received_camera_capture_status = true;
         _status.data.recording_time_s = float(camera_capture_status.recording_time_ms) / 1e3f;
+        _status.image_count = camera_capture_status.image_count;
+
+        if (_status.image_count_at_connection == -1) {
+            _status.image_count_at_connection = camera_capture_status.image_count;
+        }
+
+        if (_status.last_advertised_image_index == -1) {
+            _status.last_advertised_image_index = camera_capture_status.image_count - 1;
+        }
     }
 
     check_status();
@@ -760,25 +783,31 @@ void CameraImpl::process_camera_image_captured(const mavlink_message_t& message)
     mavlink_msg_camera_image_captured_decode(&message, &image_captured);
 
     {
+        Camera::CaptureInfo capture_info = {};
+        capture_info.position.latitude_deg = image_captured.lat / 1e7;
+        capture_info.position.longitude_deg = image_captured.lon / 1e7;
+        capture_info.position.absolute_altitude_m = image_captured.alt / 1e3f;
+        capture_info.position.relative_altitude_m = image_captured.relative_alt / 1e3f;
+        capture_info.time_utc_us = image_captured.time_utc;
+        capture_info.attitude_quaternion.w = image_captured.q[0];
+        capture_info.attitude_quaternion.x = image_captured.q[1];
+        capture_info.attitude_quaternion.y = image_captured.q[2];
+        capture_info.attitude_quaternion.z = image_captured.q[3];
+        capture_info.attitude_euler_angle =
+            to_euler_angle_from_quaternion(capture_info.attitude_quaternion);
+        capture_info.file_url = std::string(image_captured.file_url);
+        capture_info.is_success = (image_captured.capture_result == 1);
+        capture_info.index = image_captured.image_index;
+
+        if (capture_info.is_success) {
+            _status.photo_list.insert(std::make_pair(image_captured.image_index, capture_info));
+        }
+
+        _captured_request_cv.notify_all();
+
         std::lock_guard<std::mutex> lock(_capture_info.mutex);
-
-        if (_capture_info.callback) {
-            Camera::CaptureInfo capture_info = {};
-            capture_info.position.latitude_deg = image_captured.lat / 1e7;
-            capture_info.position.longitude_deg = image_captured.lon / 1e7;
-            capture_info.position.absolute_altitude_m = image_captured.alt / 1e3f;
-            capture_info.position.relative_altitude_m = image_captured.relative_alt / 1e3f;
-            capture_info.time_utc_us = image_captured.time_utc;
-            capture_info.attitude_quaternion.w = image_captured.q[0];
-            capture_info.attitude_quaternion.x = image_captured.q[1];
-            capture_info.attitude_quaternion.y = image_captured.q[2];
-            capture_info.attitude_quaternion.z = image_captured.q[3];
-            capture_info.attitude_euler_angle =
-                to_euler_angle_from_quaternion(capture_info.attitude_quaternion);
-            capture_info.file_url = std::string(image_captured.file_url);
-            capture_info.is_success = (image_captured.capture_result == 1);
-            capture_info.index = image_captured.image_index;
-
+        if (_status.last_advertised_image_index < capture_info.index && _capture_info.callback) {
+            _status.last_advertised_image_index = capture_info.index;
             const auto temp_callback = _capture_info.callback;
             _parent->call_user_callback(
                 [temp_callback, capture_info]() { temp_callback(capture_info); });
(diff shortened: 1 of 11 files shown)


src/plugins/camera/camera_impl.cpp
```diff
@@ -1583,4 +1612,117 @@ void CameraImpl::format_storage_async(Camera::ResultCallback callback)
...
+                        _parent->call_user_callback([callback]() {
+                            callback(Camera::Result::Timeout, std::vector<Camera::CaptureInfo>{});
+                        });
+                        return;
+                    }
+
+                    _parent->send_command_async(
+                        make_command_request_camera_image_captured(i), nullptr);
+                    cv_status =
+                        _captured_request_cv.wait_for(lock, std::chrono::duration<double>(1));
```
JonasVautherin, line 1696: I timeout after 1 second here, that's completely arbitrary. @julianoes is it an okay-ish value?

JonasVautherin:
Note for self - Windows error: 

```
2021-02-18T19:00:47.5574732Z D:\a\MAVSDK\MAVSDK\src\plugins\camera\camera_impl.cpp(298,51): error C2220: the following warning is treated as an error [D:\a\MAVSDK\MAVSDK\build\release\src\plugins\camera\mavsdk_camera.vcxproj]
2021-02-18T19:00:47.7473789Z D:\a\MAVSDK\MAVSDK\src\plugins\camera\camera_impl.cpp(298,51): warning C4244: '=': conversion from 'const size_t' to 'float', possible loss of data [D:\a\MAVSDK\MAVSDK\build\release\src\plugins\camera\mavsdk_camera.vcxproj]
```

src/plugins/camera/camera_impl.cpp
```diff
@@ -287,6 +288,19 @@ MavlinkCommandSender::CommandLong CameraImpl::make_command_request_camera_captur
     return cmd_req_camera_cap_stat;
 }
 
+MavlinkCommandSender::CommandLong
+CameraImpl::make_command_request_camera_image_captured(const size_t photo_id)
+{
+    MavlinkCommandSender::CommandLong cmd_req_camera_image_captured{};
+
+    cmd_req_camera_image_captured.command = MAV_CMD_REQUEST_MESSAGE;
+    cmd_req_camera_image_captured.params.param1 = MAVLINK_MSG_ID_CAMERA_IMAGE_CAPTURED;
```
julianoes, line 297:
```suggestion
    cmd_req_camera_image_captured.params.param1 = static_cast<float>(MAVLINK_MSG_ID_CAMERA_IMAGE_CAPTURED);
```

src/plugins/camera/camera_impl.cpp
```diff
@@ -726,6 +740,15 @@ void CameraImpl::process_camera_capture_status(const mavlink_message_t& message)
...
         _status.received_camera_capture_status = true;
         _status.data.recording_time_s = float(camera_capture_status.recording_time_ms) / 1e3f;
+        _status.image_count = camera_capture_status.image_count;
+
+        if (_status.image_count_at_connection == -1) {
+            _status.image_count_at_connection = camera_capture_status.image_count;
+        }
+
+        if (_status.last_advertised_image_index == -1) {
+            _status.last_advertised_image_index = camera_capture_status.image_count - 1;
```
julianoes, line 750: Why -1?

src/plugins/camera/camera_impl.cpp
```diff
@@ -760,25 +783,31 @@ void CameraImpl::process_camera_image_captured(const mavlink_message_t& message)
...
-            capture_info.attitude_euler_angle =
-                to_euler_angle_from_quaternion(capture_info.attitude_quaternion);
-            capture_info.file_url = std::string(image_captured.file_url);
-            capture_info.is_success = (image_captured.capture_result == 1);
-            capture_info.index = image_captured.image_index;
+        _captured_request_cv.notify_all();
 
+        std::lock_guard<std::mutex> lock(_capture_info.mutex);
+        if (_status.last_advertised_image_index < capture_info.index && _capture_info.callback) {
+            _status.last_advertised_image_index = capture_info.index;
```
julianoes, line 810: What's the relation between `_capture_info` and `_status`? and would we need both mutexes here?

src/plugins/camera/camera_impl.cpp
```diff
@@ -1583,4 +1612,117 @@ void CameraImpl::format_storage_async(Camera::ResultCallback callback)
         cmd_format, std::bind(&CameraImpl::receive_command_result, this, _1, callback));
 }
 
+std::pair<Camera::Result, std::vector<Camera::CaptureInfo>>
+CameraImpl::list_photos(Camera::PhotosRange photos_range)
+{
+    auto prom = std::make_shared<
+        std::promise<std::pair<Camera::Result, std::vector<Camera::CaptureInfo>>>>();
```
julianoes, line 1619: `prom` doesn't need to be a `shared_ptr` here, right? Because we don't go out of scope until the prom has been set.

src/plugins/camera/camera_impl.cpp
```diff
@@ -1583,4 +1612,117 @@ void CameraImpl::format_storage_async(Camera::ResultCallback callback)
...
+}
+
+void CameraImpl::list_photos_async(
+    Camera::PhotosRange photos_range, const Camera::ListPhotosCallback callback)
+{
+    {
+        std::lock_guard<std::mutex> lock(_status.mutex);
+
+        if (_status.is_fetching_photos) {
+            _parent->call_user_callback([callback]() {
```
julianoes, line 1637: I think you need to check if `callback` is not `nullptr`?

src/plugins/camera/camera_impl.cpp
```diff
@@ -1583,4 +1612,117 @@ void CameraImpl::format_storage_async(Camera::ResultCallback callback)
...
+                callback(Camera::Result::Busy, std::vector<Camera::CaptureInfo>{});
+            });
+            return;
+        } else {
+            _status.is_fetching_photos = true;
+        }
+
+        if (_status.image_count == -1) {
+            LogErr() << "Cannot list photos: camera status has not been received yet!";
+            _parent->call_user_callback([callback]() {
```
julianoes, line 1647: Also here, right?

src/plugins/camera/camera_impl.cpp
```diff
@@ -1583,4 +1612,117 @@ void CameraImpl::format_storage_async(Camera::ResultCallback callback)
...
+
+    switch (photos_range) {
+        case Camera::PhotosRange::SinceConnection:
+            start_index = _status.image_count_at_connection;
+            break;
+        case Camera::PhotosRange::All:
+        default:
+            start_index = 0;
+            break;
+    }
```
julianoes, line 1664:
```suggestion
    const int start_index = [photos_range]() {
        switch (photos_range) {
            case Camera::PhotosRange::SinceConnection:
               return _status.image_count_at_connection;
            case Camera::PhotosRange::All:
            // FALLTHROUGH
            default:
                return 0;
    }();
```

src/plugins/camera/camera_impl.cpp
```diff
@@ -1583,4 +1612,117 @@ void CameraImpl::format_storage_async(Camera::ResultCallback callback)
...
+
+    std::thread([this, start_index, callback]() {
```
julianoes, line 1666: Have you considered doing this with `call_every` instead of a thread?

src/plugins/camera/camera_impl.cpp
```diff
@@ -1583,4 +1612,117 @@ void CameraImpl::format_storage_async(Camera::ResultCallback callback)
...
+
+                auto request_try_number = 0;
+                const auto request_try_limit =
+                    3; // Timeout if the request times out that many times
+                auto cv_status = std::cv_status::timeout;
+
+                while (cv_status == std::cv_status::timeout) {
+                    request_try_number++;
+                    if (request_try_number >= request_try_limit) {
+                        _parent->call_user_callback([callback]() {
```
julianoes, line 1687: Check if callback is nullptr?

src/plugins/camera/camera_impl.cpp
```diff
@@ -1583,4 +1612,117 @@ void CameraImpl::format_storage_async(Camera::ResultCallback callback)
...
+
+                    _parent->send_command_async(
+                        make_command_request_camera_image_captured(i), nullptr);
+                    cv_status =
+                        _captured_request_cv.wait_for(lock, std::chrono::duration<double>(1));
+                }
+            }
+
+            if (safety_count == safety_count_boundary) {
+                _parent->call_user_callback([callback]() {
```
julianoes, line 1701: Check callback?

src/plugins/camera/camera_impl.cpp
```diff
@@ -1583,4 +1612,117 @@ void CameraImpl::format_storage_async(Camera::ResultCallback callback)
...
+            for (auto capture_info : _status.photo_list) {
+                if (capture_info.first >= start_index) {
+                    photo_list.push_back(capture_info.second);
+                }
+            }
+
+            _status.is_fetching_photos = false;
+
+            const auto temp_callback = callback;
+            _parent->call_user_callback([temp_callback, photo_list]() {
```
julianoes, line 1721: Check callback?

src/plugins/camera/camera_impl.h
```diff
@@ -180,6 +188,11 @@ class CameraImpl : public PluginImplBase {
         Camera::Status data{};
         bool received_camera_capture_status{false};
         bool received_storage_information{false};
+        int image_count{-1};
+        int image_count_at_connection{-1};
+        int last_advertised_image_index{-1};
+        std::map<int, Camera::CaptureInfo> photo_list;
```
julianoes, line 194: Do you need an ordered map (`std::map`) or is an unordered map good enough (`std::unordered_map`)?

julianoes (changes requested):

src/plugins/camera/camera_impl.cpp
JonasVautherin: Because an image index is the image count - 1 (if there are 3 images, the last image is number 2). At least I think it works :sweat_smile:

src/plugins/camera/camera_impl.cpp
JonasVautherin: aha, yes you're right. Now I'm concerned about the risk of a deadlock if I use both :thinking:. We would need to be sure that we always acquire them in the same order, right?

src/plugins/camera/camera_impl.cpp
julianoes: Ok, got it.

src/plugins/camera/camera_impl.cpp
JonasVautherin: I was not comfortable waiting on a condition variable in `call_every`. That would block the event loop, wouldn't it?

src/plugins/camera/camera_impl.cpp
JonasVautherin: Actually, it should all be in `_capture_info`. I get the first part of your comment now :laughing:

src/plugins/camera/camera_impl.cpp
julianoes: Nevermind, leave it as is!

src/plugins/camera/camera_impl.h
JonasVautherin: I think that conceptually, the ordered map makes sense since I convert it to a vector for the user, and I want the images ordered there... not sure in terms of performance, though

julianoes (approved):

Issue #1326, opened 2021-02-19
Messages forwarding feature


yhabib29:
Hi,

For my projects needs, I had to implement messages forwarding.
An example use case is:
- 245.190 = Ground Control Station (GCS)
- 1.195 = Companion Computer (CC)
- 1.1 = PX4 Autopilot (PX4)

CC is connected to the GCS (UDP) and to the PX4 (serial).
We need the GCS to be able to send command/messages directly to the PX4.

Thus, I have added messages forwarding functionality based on PX4 source code.
The forward_message function was adapted from Firmware v1.11.3:
    https://github.com/PX4/PX4-Autopilot/blob/v1.11.3/src/modules/mavlink/mavlink_main.cpp#L460

See changes in: [add-messages-forwarding](/yhabib29/MAVSDK/tree/add-messages-forwarding)

I had to keep track of the origin connection from which the message was received to ignore it when forwarding.
There might be a better way to do this.

```diff --git a/src/core/connection.h b/src/core/connection.h
index 02ba17b4..5b977325 100644
--- a/src/core/connection.h
+++ b/src/core/connection.h
@@ -8,7 +8,7 @@ namespace mavsdk {
 
 class Connection {
 public:
-    typedef std::function<void(mavlink_message_t& message)> receiver_callback_t;
+    typedef std::function<void(mavlink_message_t& message, Connection *connection)> receiver_callback_t;
 
     Connection(receiver_callback_t receiver_callback);
     virtual ~Connection();
@@ -25,7 +25,7 @@ public:
 protected:
     bool start_mavlink_receiver();
     void stop_mavlink_receiver();
-    void receive_message(mavlink_message_t& message);
+    void receive_message(mavlink_message_t& message, Connection *connection);
 
     receiver_callback_t _receiver_callback{};
     std::unique_ptr<MAVLinkReceiver> _mavlink_receiver;
```

If this feature is useful to you, I can make a pull request.
However, i did not update the integration tests, etc ...

julianoes:
I think message forwarding would be a good feature. However, I wouldn't make it an option of the mavsdk configuration, so of the mavsdk instance but rather of connections. So you don't necessarily want forwarding between all connections but maybe only between some of them. So whenever you add a connection, you could specify if it should forward or not.

yhabib29:
Oh, good point, that is a better solution !
I will:
- Add a parameter `message_forwarding` in functions creating connection (default to `false`).
[...]

PR #1330, opened 2021-02-23
Add messages forwarding

yhabib29:
Adding message forwarding functionality based of PX4 firmware source code.

See #1326


yhabib29:
I had to keep `Connection` pointer in `receive_message` function to keep track of the connection which received current message.
Maybe you know a better way of doing this ?

```diff --git a/src/core/mavsdk_impl.cpp b/src/core/mavsdk_impl.cpp
index 986dee3d..78adc4c9 100644
--- a/src/core/mavsdk_impl.cpp
+++ b/src/core/mavsdk_impl.cpp

+void MavsdkImpl::receive_message(mavlink_message_t& message, Connection* connection)
{
+    /** @note: Forward message if option is enabled and multiple interfaces are connected.
+     *  Performs message forwarding checks for every messages if message forwarding
+     *  is enabled on at least one connection, and in case of a single forwarding connection,
+     *  we check that it is not the one which received the current message.
+     */
+    if (_connections.size() > 1 && connection->forwarding_connections_count() > 0 &&
+        (connection->forwarding_connections_count() && connection->do_forward_messages())) {
+        forward_message(message, connection);
+    }
+
     // Don't ever create a system with sysid 0.
     if (message.sysid == 0) {
         return;
```

src/core/mavsdk_impl.cpp
```diff
@@ -110,8 +110,71 @@ std::vector<std::shared_ptr<System>> MavsdkImpl::systems() const
...
+        (target_system_id == get_own_system_id() && target_component_id == get_own_component_id());
+
+    // We don't forward heartbeats unless it's specifically enabled.
+    const bool heartbeat_check_ok =
+        (message.msgid != MAVLINK_MSG_ID_HEARTBEAT || forward_heartbeats_enabled);
+
+    if (!targeted_only_at_us && heartbeat_check_ok) {
+        std::lock_guard<std::mutex> lock(_connections_mutex);
+
+        uint succesfull_emissions = 0;
```
julianoes, line 151: `uint` is a thing? :astonished:

src/core/connection.h
```diff
@@ -8,27 +8,34 @@ namespace mavsdk {
 
 class Connection {
 public:
-    typedef std::function<void(mavlink_message_t& message)> receiver_callback_t;
+    typedef std::function<void(mavlink_message_t& message, Connection* connection)>
+        receiver_callback_t;
 
-    Connection(receiver_callback_t receiver_callback);
+    Connection(receiver_callback_t receiver_callback, bool forward_messages = false);
```
julianoes, line 14: I suggest to use an enum for the forwarding option because that will allow to add more rules later and is better in terms of type-safety.

So what about:
```
enum class ForwardingOption {
    ForwardingOff = 0,
    ForwardingOn = 1,
};
```

julianoes (commented):
Looks very clean, thanks for that!

See my inline comments.

src/core/mavsdk_impl.cpp
yhabib29: Haha maybe you prefer uint8_t ?

src/core/connection.h
yhabib29: Yep, good point !

src/core/mavsdk_impl.cpp
julianoes: No, I think it's fine actually.

julianoes:
Sorry, I missed that you updated this. I'll review it again soon.

JonasVautherin:
The failing CI check has be fixed, so if you rebase your branch it will pass :+1:

yhabib29:
Just an additional remark, user should know what he is doing when enabling this option as it increases a lot the bandwidth usage when for example forwarding messages from the autopilot to a GroundStation through a radio.

Maybe adding message filtering could be interesting ...

src/core/mavsdk_impl.cpp
```diff
@@ -110,8 +110,71 @@ std::vector<std::shared_ptr<System>> MavsdkImpl::systems() const
...
+        }
+    }
+
+    // If it's a message only for us, we keep it, otherwise, we forward it.
+    const bool targeted_only_at_us =
+        (target_system_id == get_own_system_id() && target_component_id == get_own_component_id());
+
+    // We don't forward heartbeats unless it's specifically enabled.
+    const bool heartbeat_check_ok =
+        (message.msgid != MAVLINK_MSG_ID_HEARTBEAT || forward_heartbeats_enabled);
```
julianoes, line 146: I'm just wondering here. Should we forward heartbeats or not? The intuitive case would be to forward everything including heartbeats, right?

src/core/mavsdk_impl.cpp
```diff
@@ -110,8 +110,71 @@ std::vector<std::shared_ptr<System>> MavsdkImpl::systems() const
...
+
+    if (!targeted_only_at_us && heartbeat_check_ok) {
+        std::lock_guard<std::mutex> lock(_connections_mutex);
+
+        uint8_t successful_emissions = 0;
```
julianoes, line 151: Sorry, I might have confused you here. I didn't mean that we need to specifically try to use short types, just the normal `unsigned` is fine. I just didn't know that `uint` is a thing.

```suggestion
        unsigned successful_emissions = 0;
```

src/core/connection.h
```diff
@@ -8,27 +8,36 @@ namespace mavsdk {
...
     bool start_mavlink_receiver();
     void stop_mavlink_receiver();
-    void receive_message(mavlink_message_t& message);
+    void receive_message(mavlink_message_t& message, Connection* connection);
 
     receiver_callback_t _receiver_callback{};
     std::unique_ptr<MAVLinkReceiver> _mavlink_receiver;
+    ForwardingOption _forwarding_option;
+
+    static uint8_t _forwarding_connections_count;
```
julianoes, line 40: I think this can be a `unsigned`, sorry for confusing you earlier.

I also wonder whether this should be a `static std::atomic<unsigned>` because a new connection is added in different thread than where the receiver thread.

src/core/connection.cpp
```diff
@@ -38,9 +45,19 @@ void Connection::stop_mavlink_receiver()
     }
 }
 
-void Connection::receive_message(mavlink_message_t& message)
+void Connection::receive_message(mavlink_message_t& message, Connection* connection)
+{
+    _receiver_callback(message, connection);
+}
+
+bool Connection::do_forward_messages() const
```
julianoes, line 53: What about?
```suggestion
bool Connection::should_forward_messages() const
```

src/core/mavsdk_impl.cpp
```diff
@@ -110,8 +110,71 @@ std::vector<std::shared_ptr<System>> MavsdkImpl::systems() const
...
+        for (auto it = _connections.begin(); it != _connections.end(); ++it) {
+            if ((*it).get() == connection || !(**it).do_forward_messages()) {
+                continue;
+            }
+            if ((**it).send_message(message)) {
+                successful_emissions++;
+            }
+        }
+        if (successful_emissions == 0) {
+            LogErr() << "forward fail";
```
julianoes, line 161:
```suggestion
            LogErr() << "Message forwarding failed";
```

src/core/mavsdk_impl.cpp
```diff
@@ -110,8 +110,71 @@ std::vector<std::shared_ptr<System>> MavsdkImpl::systems() const
...
+
+void MavsdkImpl::receive_message(mavlink_message_t& message, Connection* connection)
 {
+    /** @note: Forward message if option is enabled and multiple interfaces are connected.
+     *  Performs message forwarding checks for every messages if message forwarding
+     *  is enabled on at least one connection, and in case of a single forwarding connection,
+     *  we check that it is not the one which received the current message.
+     */
+    if (_connections.size() > 1 && connection->forwarding_connections_count() > 0 &&
+        (connection->forwarding_connections_count() && connection->do_forward_messages())) {
```
julianoes, line 174: You check twice if `connection->forwarding_connections_count()` but once should be enough, and you don't need the parentheses in the second part since it's all `&&`.

So what about?
```suggestion
    if (_connections.size() > 1 && connection->forwarding_connections_count() > 0 &&
        connection->do_forward_messages()) {
```

src/core/mavsdk_impl.cpp
```diff
@@ -110,8 +110,71 @@ std::vector<std::shared_ptr<System>> MavsdkImpl::systems() const
...
+    // We don't forward heartbeats unless it's specifically enabled.
+    const bool heartbeat_check_ok =
+        (message.msgid != MAVLINK_MSG_ID_HEARTBEAT || forward_heartbeats_enabled);
+
+    if (!targeted_only_at_us && heartbeat_check_ok) {
+        std::lock_guard<std::mutex> lock(_connections_mutex);
+
+        uint8_t successful_emissions = 0;
+        for (auto it = _connections.begin(); it != _connections.end(); ++it) {
+            if ((*it).get() == connection || !(**it).do_forward_messages()) {
```
julianoes, line 153: So here we check if the connection is us, right?

Maybe a comment would be good to explain that to the reader.

src/core/connection.h
```diff
@@ -8,27 +8,36 @@ namespace mavsdk {
...
+        ForwardingOption forwarding_option = ForwardingOption::ForwardingOff);
     virtual ~Connection();
 
     virtual ConnectionResult start() = 0;
     virtual ConnectionResult stop() = 0;
 
     virtual bool send_message(const mavlink_message_t& message) = 0;
 
+    bool do_forward_messages() const;
+    uint8_t forwarding_connections_count() const;
```
julianoes, line 25:
```suggestion
    unsigned forwarding_connections_count() const;
```

Sorry for the confusion.

julianoes (commented):
Sorry, I made another few notes that we should work through. Other than that this looks good!

src/core/mavsdk_impl.cpp
yhabib29: No problem, but it was OK for me as they might not be more than 255 connections ?

src/core/connection.h
yhabib29: Yes, you are right, I missed that !

src/core/connection.cpp
yhabib29: I agree with that !

src/core/mavsdk_impl.cpp
julianoes: Oh, because the map is limited to 255 as well, right?

src/core/mavsdk_impl.cpp
yhabib29: Yes, exactly, I will add a comment.

src/core/mavsdk_impl.cpp
yhabib29: Yes you are right, I think I wanted to leave it as an option, but then I forgot.
But maybe we should just enable it.

src/core/mavsdk_impl.cpp
yhabib29: Hmm 🤔 I think I just wanted to keep it as small as possible, but it was arbitrary.
For the map, you meant the vector of Connection ? It does not seem to be limited to 255.

yhabib29: I will just change it to `unsigned` :+1:

src/core/mavsdk_impl.cpp
yhabib29: Well I think I made a mistake, the conditions should be:
1. At least 2 connections.
2. At least 1 forwarding connection.
3. [If only one forwarding connection, it should not be the current one] = [At least 2 forwarding connections or current connection is not forwarding].

src/core/mavsdk_impl.cpp
julianoes: I would enable it, and remove the specific code. And if we find out why it was there we can add it in :smile:.

src/core/mavsdk_impl.cpp
julianoes: I don't understand 3. :confused:

julianoes:
By the way, I've tested it and it works. Sadly forwarding via TCP does not work because TCP is only made to connect to a server, so far, and not to be a server. We probably need to add something like `setup_tcp_remote`, but with a better name :).

src/core/mavsdk_impl.cpp
yhabib29: Well if we have 3 connections: C1, C2 (forwarding) and C3.
If we receive a message from C2, we do not need to enter `forward_message` and performs all the checks as the all other connections do not forward messages.

Thus the new conditions are:
```
if (_connections.size() > 1 && 
    connection->forwarding_connections_count() > 0 &&
    (connection->forwarding_connections_count() > 1 || !connection->should_forward_messages())) 
{
    forward_message(message, connection);
}
```

Thus the 3rd condition checks that:
- Either we have at least 2 forwarding connections.
- Or (if there is only one forwarding connection) the connection from which we receive the message is not the the only forwarding connection.

Do you see what I mean ?

julianoes: Aha, so the rules for your example would be as follows:
C1 not forwarding
C2 forwarding
C3 not forwarding

Message arriving at C1 -> forwarded to C2
Message arriving at C2 -> not forwarded
Message arriving at C3 -> forwarded to C2

Is that correct?

yhabib29: Yep, that was the idea !

yhabib29:
Oh I see ! I do not have a lot of experience with TCP but as I understand, the only difference will be in the connection procedure, right ?

To sum up what you said: (correct me if I am wrong)
1. Currently, the TCP connection acts as a client, thus it tries to connect to a TCP server.
2. This new `setup_tcp_remote()` function should create a TCP server waiting for incoming connection(s).
    - Does this connection need to be blocking (until a first client connects) ? Or maybe use a timeout and let the user try again until it succeeds ?
    - Should it accept multiple clients ?

For the name, maybe `setup_tcp_connection` or `setup_tcp_server`, idk haha

src/core/mavsdk_impl.cpp
julianoes: Ok, I think that's fine but we'll have to document that in https://github.com/mavlink/MAVSDK-docs/.

julianoes (approved):
I would say this is good to go in, thanks a lot @yhabib29!

yhabib29:
All right, thank you for your help @julianoes !

For the documentation and TCP, let me now if you need some more help. 👍

Issue #1319, opened 2021-02-11
Using multiple UDP remotes


yhabib29:
Hello,
I am trying to use MAVSDK to build 2 programs:
- A "client" program which runs on the companion computer on each drone.
- A GroundStation (GS) which can connect to each client.

I am using 2 UDP connections to connect each client to the GS. However, I can create only one UDP remote at a time otherwise it overwrites the existing system (is_single_system in mavsdk_impl.cpp).

How should I connect my GS to the other clients where each program is using Mavsdk.

JonasVautherin:
I think you're right: right now MAVSDK can only add one remote. This would need to be updated in the implementation :+1:. Open for contributions, though :innocent:

yhabib29:
All right, I will try to implement it in few days.
Thank you for the confirmation 👍🏼

yhabib29:
To summarize the UDP connections system in Mavsdk.
- add_udp_connection("IP:PORT"):
    Create a new UdpConnection instance listening on IP:PORT.
    When the first message coming from a new endpoint (IP_ENDPOINT:PORT_ENDPOINT) is received, a new remote is created to send messages to IP_ENDPOINT:PORT_ENDPOINT.
    
- setup_udp_remote(IP, PORT):
    Create a new UdpConnection instance listening on 0.0.0.0:0 (port = 0 ??).
    Then a remote is created to send messages to IP:PORT

When sending a message, the mavsdk_impl iterates over all connections to send the message.
However, if one of the connections is a UDP connection and has no remote, the loop will break even if other connections could send the message.

see [add-multi-udp-remotes](habib29/MAVSDK/tree/add-multi-udp-remotes)

yhabib29:
In addition, could it be interesting to add a vector of available system IDs for each connection ?

When sending a message to a specific system ID, this will allow to ignore connections which have never see the corresponding system ID.

The reason is that for radio connections, it might be better to preserve the bandwidth from useless messages.

julianoes:
> However, if one of the connections is a UDP connection and has no remote, the loop will break even if other connections could send the message.

That sounds like a bug, right?

> When sending a message to a specific system ID, this will allow to ignore connections which have never see the corresponding system ID.

Yes, I think that would be a good optimization.

yhabib29:
> That sounds like a bug, right?

Well, I think yes but I did not know if it was done in purpose ...
Here is the quick changes I did, but there might be a better solution.
``` 
[...]

PR #1331, opened 2021-02-23
Add multi udp remotes

yhabib29:
Add multiple UDP remote connections functionality, and fix `send_message` function bug: it breaks when one of the connections cannot send the message even if other connections could sent it.

See #1319


yhabib29:
Added a vector of known system IDs in `Connection` class to avoid sending message through interfaces which have never seen the relevant system ID.

If set, get `target_system_id` from message (see [PX4 Firmware source code](https://github.com/PX4/PX4-Autopilot/blob/v1.11.3/src/modules/mavlink/mavlink_main.cpp#465))

src/core/mavsdk_impl.cpp
```diff
@@ -125,13 +125,6 @@ void MavsdkImpl::receive_message(mavlink_message_t& message)
         std::swap(_systems[message.sysid], it->second);
         _systems[message.sysid]->system_impl()->set_system_id(message.sysid);
         _systems.erase(it);
-    } else if (_is_single_system) {
-        auto it_begin = _systems.begin();
-        if (it_begin->first != message.sysid) {
-            std::swap(_systems[message.sysid], it_begin->second);
-            _systems[message.sysid]->system_impl()->set_system_id(message.sysid);
-            _systems.erase(it_begin);
-        }
```
julianoes, line 134: Hm, how come this is not required anymore?

julianoes (commented):
Thanks for this PR! I think it mostly good. I'm just a bit confused about one part, see inline comment.

julianoes:
@JonasVautherin would be good to get your review also since you looked at this more recently.

src/core/mavsdk_impl.cpp
yhabib29: Maybe I did not understand well the purpose of this part, can you explain what it does ?

src/core/mavsdk_impl.cpp
```diff
@@ -155,13 +148,30 @@ bool MavsdkImpl::send_message(mavlink_message_t& message)
...
+            }
+        }
+
+        if ((**it).send_message(message)) {
+            successful_emissions++;
         }
     }
 
+    if (successful_emissions == 0) {
+        LogErr() << "send fail";
```
julianoes, line 171: While we're at it:
```suggestion
        LogErr() << "Sending message failed";
```

src/core/mavsdk_impl.cpp
yhabib29: Ok I see, this what is done for the first condition, when it founds a system 0:
```
if (it != _systems.end()) {
    std::swap(_systems[message.sysid], it->second);
    _systems[message.sysid]->system_impl()->set_system_id(message.sysid);
    _systems.erase(it);
}
```

But this part I removed is used only when no system 0 was found in `_systems` and `_is_single_system` is true (when `setup_udp_remote` was used).
Thus, it seems to me to be swapping the first system in `_systems `(which cannot be 0), by the system ID of the current message.

src/core/mavsdk_impl.cpp
yhabib29: Thanks, I will change that !

julianoes:
@yhabib29 I tried this and it does not seem to work for the integration tests (also the integration tests in CI are not working).

I think the issue is the one I raised above where the default system(0) is not swapped out with a working one anymore.

```
cmake --build build -j8 && build/src/integration_tests/integration_tests_runner --gtest_filter="SitlTest.ActionHoverSyncHigher"
[  1%] Built target gtest
...
[100%] Built target integration_tests_runner
Note: Google Test filter = SitlTest.ActionHoverSyncHigher
[==========] Running 1 test from 1 test suite.
[ RUN      ] SitlTest.ActionHoverSyncHigher
[10:41:39|Debug] Model chosen: 'iris' (integration_test_helper.h:59)
SITL model: iris
[10:41:39|Info ] MAVSDK version: v0.37.0-9-g90ab84ad (mavsdk_impl.cpp:21)
[10:41:39|Info ] Waiting for system connect (action_hover_sync.cpp:37)
[10:41:39|Info ] New system on: 127.0.0.1:14580 (with sysid: 1) (udp_connection.cpp:190)
[10:41:39|Debug] New: System ID: 1 Comp ID: 1 (mavsdk_impl.cpp:428)
[10:41:39|Debug] Component Autopilot (1) added. (system_impl.cpp:324)
[10:41:39|Debug] Component Gimbal (154) added. (system_impl.cpp:324)
[10:41:39|Error] Sending message failed (mavsdk_impl.cpp:171)
[10:41:39|Error] Sending message failed (mavsdk_impl.cpp:171)
[10:41:39|Error] connection send error (520) (mavlink_commands.cpp:291)
[10:41:40|Error] Sending message failed (mavsdk_impl.cpp:171)
[10:41:40|Error] Sending message failed (mavsdk_impl.cpp:171)
[10:41:40|Error] connection send error (520) (mavlink_commands.cpp:291)
[10:41:41|Error] Sending message failed (mavsdk_impl.cpp:171)
[10:41:41|Error] Sending message failed (mavsdk_impl.cpp:171)
[10:41:41|Error] connection send error (520) (mavlink_commands.cpp:291)
[10:41:42|Warn ] No UUID received, using system ID instead. (system_impl.cpp:424)
[10:41:42|Debug] Discovered 2 component(s) (UUID: 1) (system_impl.cpp:481)
[10:41:42|Error] Sending message failed (mavsdk_impl.cpp:171)
[10:41:42|Info ] Waiting for system to be ready (action_hover_sync.cpp:59)
[10:41:42|Error] Sending message failed (mavsdk_impl.cpp:171)
[10:41:42|Error] Error: Send message failed (mavlink_parameters.cpp:300)
[10:41:42|Error] Error: Param for gyro cal failed. (telemetry_impl.cpp:1228)
[10:41:42|Error] Sending message failed (mavsdk_impl.cpp:171)
[10:41:42|Error] Error: Send message failed (mavlink_parameters.cpp:300)
[10:41:42|Error] Error: Param for accel cal failed. (telemetry_impl.cpp:1239)
[10:41:42|Error] Sending message failed (mavsdk_impl.cpp:171)
[10:41:42|Error] Error: Send message failed (mavlink_parameters.cpp:300)
[10:41:42|Error] Error: Param for mag cal failed. (telemetry_impl.cpp:1250)
[10:41:42|Error] Sending message failed (mavsdk_impl.cpp:171)
[10:41:42|Error] Error: Send message failed (mavlink_parameters.cpp:300)
[10:41:42|Error] Error: Param to determine hitl failed. (telemetry_impl.cpp:1274)
[10:41:43|Error] Sending message failed (mavsdk_impl.cpp:171)
[10:41:44|Error] Sending message failed (mavsdk_impl.cpp:171)
[10:41:45|Error] Sending message failed (mavsdk_impl.cpp:171)
```

src/core/mavsdk_impl.cpp
yhabib29: Ok I see, this what is done for the first condition, when it founds a system 0:
```
if (it != _systems.end()) {
    std::swap(_systems[message.sysid], it->second);
    _systems[message.sysid]->system_impl()->set_system_id(message.sysid);
    _systems.erase(it);
}
```

But this part I removed is used only when no system 0 was found in `_systems` and `_is_single_system` is true (when `setup_udp_remote` was used).
Thus, it seems to me to be swapping the first system in `_systems `(which cannot be 0), by the system ID of the current message.

yhabib29:
I have launched the same test, and the problem seems to come from these lines:
```
if (meta && meta->flags & MAV_MSG_ENTRY_FLAG_HAVE_TARGET_SYSTEM) {
    int target_system_id = (_MAV_PAYLOAD(&message))[meta->target_system_ofs];
    if (!(**it).has_system_id(target_system_id)) {
        continue;
    }
}
```

Here, I wanted to ignore connections that had never met the target system (if one was set).
When I comment this part, the test complete without any errors.

src/core/connection.cpp
```diff
@@ -43,7 +43,7 @@ void Connection::stop_mavlink_receiver()
 
 void Connection::receive_message(mavlink_message_t& message)
 {
-    if (_system_ids.find(message.sysid) != _system_ids.end()) {
+    if (_system_ids.find(message.sysid) == _system_ids.end()) {
         _system_ids.insert(message.sysid);
```
yhabib29, line 47: The problem was coming from here, I was not adding the new system_id when receiving a message from a new system_id.

yhabib29:
If you try now it should be OK. On my side, It still fails due to the error in relative altitude ...

But did you see my [reply](https://github.com/mavlink/MAVSDK/pull/1331/files#r588688521) ? 
What do you think ? Is this part still needed ? I think it is not needed anymore, but I might be wrong.

julianoes:
Since I've merged #1330 this needs rebasing or merging with `develop`. Then I'll test it again to see if it works, from the integration tests it already looks better.

julianoes:
Thanks for rebasing.

I tested this and it works fine with the integration tests and examples which have been ported to the new `std::vector<std::shared_ptr<System>> systems()` API.

However, I don't think it works with the `System& system()` API which is deprecated but is still used in `mavsdk_server`, here:
https://github.com/mavlink/MAVSDK/blob/8aa7b352212bc20be92fcc5e6dc9ec71f877bcc4/src/mavsdk_server/src/grpc_server.h#L56-L93, at least when I tried testing it, it would not connect.

yhabib29:
How can I reproduce this error ?
I have never used mavsdk_server ...

julianoes:
@yhabib29 if you apply the patch attached, you should see it.

You can apply the patch with:
```
git apply patch.txt
```

Then, you start the integration test:
```
cmake --build build/default -j8 && build/default/src/integration_tests/integration_tests_runner --gtest_filter="SitlTest.ActionHoverSyncDefault"
```

And then you start PX4 SITL
```
make px4_sitl gazebo_iris
```

[patch.txt](https://github.com/mavlink/MAVSDK/files/6118222/patch.txt)

yhabib29:
Okay, I have applied the patch, but I don't understand how it is supposed to exit from the following loop:

```diff --git a/src/integration_tests/action_hover_sync.cpp b/src/integration_tests/action_hover_sync.cpp
index 2d78bb57..96cf974c 100644
--- a/src/integration_tests/action_hover_sync.cpp
+++ b/src/integration_tests/action_hover_sync.cpp
@@ -33,34 +33,21 @@ void takeoff_and_hover_at_altitude(float altitude_m)

+    auto& system = mavsdk.system();
 
-    auto system = systems.at(0);
+    LogInfo() << "Waiting for system";
+    while (true) {
+        LogInfo() << "waiting";
+        std::this_thread::sleep_for(std::chrono::seconds(1));
+    }

```

I get stuck in this loop ...

julianoes:
Haha, I'm stupid. Let me fix this patch!

julianoes:
It looks like that with the correct patch it actually works, so I was wrong.
[patch2.txt](https://github.com/mavlink/MAVSDK/files/6123859/patch2.txt)

yhabib29:
> It looks like that with the correct patch it actually works, so I was wrong.
> [patch2.txt](https://github.com/mavlink/MAVSDK/files/6123859/patch2.txt)

Oh well, great ! :smile:  

> For use cases using a unique reference, `_is_single_system` is still needed but the code below will limit the usage of multiple UDP remotes.
> Maybe we can create a special System reference always pointing to the last System which received a message ?
> Or make `_is_single_system` flag independent of UDP remote creation, and disable multi UDP remotes when this flag is `true`.
> 
> ```
> } else if (_is_single_system) {
>     auto it_begin = _systems.begin();
>     if (it_begin->first != message.sysid) {
>         std::swap(_systems[message.sysid], it_begin->second);
>         _systems[message.sysid]->system_impl()->set_system_id(message.sysid);
>         _systems.erase(it_begin);
>     }
> ```

Nevertheless, as I removed this part in this PR, user might not be able  to use the single system functionality properly ...

julianoes:
`_is_single_system` I don't remember what this was for. I need to read through it again.

@JonasVautherin also wanted to review this, so I'll also wait for his comment.

src/core/mavsdk_impl.cpp
JonasVautherin: It was added for the `setup_udp_remote` indeed, [here](https://github.com/mavlink/MAVSDK/pull/774/files#diff-1dc7f324308bcafacd5b550b936eedb0818ac54d2b094e89a8d184a3b1759a12R94-R100)

src/core/mavsdk_impl.cpp
```diff
@@ -239,11 +232,27 @@ bool MavsdkImpl::send_message(mavlink_message_t& message)
...
+        // Checks whether connection knows target system ID by extracting target system if set.
+        // https://github.com/PX4/PX4-Autopilot/blob/v1.11.3/src/modules/mavlink/mavlink_main.cpp#L472
+        const mavlink_msg_entry_t* meta = mavlink_get_msg_entry(message.msgid);
+
+        if (meta && meta->flags & MAV_MSG_ENTRY_FLAG_HAVE_TARGET_SYSTEM) {
+            int target_system_id = (_MAV_PAYLOAD(&message))[meta->target_system_ofs];
+            if (!(**it).has_system_id(target_system_id)) {
+                continue;
+            }
         }
```
JonasVautherin, line 246: Just for my understanding: this is an optimization, right? It won't send a targeted message to a connection that has not been seen to have this sysid, but a broadcast will be sent. Correct?

Or is it required for some other reason? If it is just an optimization, I'm not sure if it is worth keeping, and actually for forwarding we may need to handle it differently to avoid loops, right?

src/core/mavsdk_impl.cpp
```diff
@@ -239,11 +232,27 @@ bool MavsdkImpl::send_message(mavlink_message_t& message)
...
+
+        if ((**it).send_message(message)) {
+            successful_emissions++;
+        }
+    }
+
+    if (successful_emissions == 0) {
```
JonasVautherin, line 253: That's an interesting one :smile:. Is a message successfully sent if it was sent on at least one connection, or if it was sent on all connections? Here we decide that it is successfully sent if it could be sent on one connection, correct?

src/core/connection.cpp
```diff
@@ -47,6 +50,10 @@ void Connection::stop_mavlink_receiver()
 
 void Connection::receive_message(mavlink_message_t& message, Connection* connection)
 {
+    // Register system ID when receiving a message from a new system.
+    if (_system_ids.find(message.sysid) == _system_ids.end()) {
```
JonasVautherin, line 54: Out of curiosity: is it more costly to `find` first and `insert` if it does not exist, or to always `insert`? Given that `_system_ids` only has a few elements generally. Does it even make sense to think about that, or is it a detail?

julianoes: It probably does the same check inside, right? But if it already exists, we don't want to overwrite it, would be my thought.

yhabib29: I am not really expert on this question, but what I understood is that finding over an `unordered_set` is very efficient, and as @julianoes said, `insert` might also calls to `find` function.

src/core/mavsdk_impl.cpp
yhabib29: Yes, but as you said it depends on how we want to interpret a successful emission.
But for me, the main thing was to let the loop iterate over all connections even if one fails to send the message.

src/core/mavsdk_impl.cpp
yhabib29: Yes, it is meant to be an optimization. A broadcast should be sent as I registered the "virtual" system ID 0 in the `Connection` constructor.
But I am not sure that the following line returns 0 when it is not a targeted message:
```
 int target_system_id = (_MAV_PAYLOAD(&message))[meta->target_system_ofs];

```

About its utility, let's consider an exemple, if we have an autopilot **(FC,1.1)** linked to a Companion Computer **(CC,1.195)** which is connected to a Ground Station **(GS,245.190)** and another device **(O,12.195)**, all using `mavsdk`:
- FC <--> CC = Serial
- CC <--> GS = Radio
- CC <--> O = Radio

Let's say the **GS** sends messages at high rate to the **FC** through the **CC**.
If we do not filter messages, the **CC** will forward messages targeted to the **FC** with system ID 1 to both the **FC** and **O**, even though **O** will ignore it. Indeed, the `Connection` linking **CC** to **O** will never receive any message from the sys ID 1, so it will not register it. Thus, the (radio) bandwidth will be loaded unnecessarily.

By the way, you are right for the loop problem.
When a `mavsdk` instance forward a message, it ignores the `Connection` from which it received the message.
However, if the network contains a loop, the message will be forwarded forever.
In the example above, if we add the following link:
- FC <--> GS = Radio (or whatever)
Then we create a loop: GS -> CC -> FC -> GS -> CC -> FC -> ...

src/core/connection.cpp
JonasVautherin: I don't think that [insert](https://www.cplusplus.com/reference/unordered_set/unordered_set/insert/) overwrites it:

> Inserts new elements in the unordered_set.
> 
> Each element is inserted only if it is not equivalent to any other element already in the container (elements in an unordered_set have unique values).

That's why I was wondering if that test was not redundant.

src/core/mavsdk_impl.cpp
JonasVautherin: ok :+1:

src/core/mavsdk_impl.cpp
JonasVautherin: Got it, good to know that it is an optimization. Other than that, I think that it doesn't hurt, so if it works, it can stay :+1:.

JonasVautherin (approved):
Looks good to me: passes the integration tests, and I tried it myself with `mavsdk_server` (using MAVSDK-Python) and it was working :+1:.

Thanks @yhabib29! :rocket:

src/core/mavsdk_impl.cpp
yhabib29: All right ! 😄 
Maybe a comment should be added to prevent user from creating such loop when activating forwarding option ...

src/core/connection.cpp
yhabib29: You are right, I got it know, indeed we could replace it if you prefer !

julianoes (approved):
Looks good now, thanks @yhabib29!

yhabib29:
Happy to help 👍🏼😁

PR #1335, opened 2021-03-01
Some static analysis

julianoes:
Motivated by weekend lecture.


CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.10.2)
+cmake_minimum_required(VERSION 3.10.2..3.19.6)
```
JonasVautherin, line 1: I'm not sure I get what this means, as a "minimum required".  Doesn't "minimum 3.10.2" say that 3.11, 3.12, 3.13, ... are supposedly fine already? :laughing:

JonasVautherin (dismissed):
Nice!

CMakeLists.txt
julianoes: Yes I don't get it either I have to say.

julianoes: It's about policy settings: https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html

julianoes: Actually, the maximum version was only added in 3.12, so this line doesn't make much sense :smile:.

julianoes: I'm gonna remove the change.

JonasVautherin: Yeah I think the feature is weird anyway. "If you are between 3.10.2 and 3.19.6, use the NEW policy behavior, but if you use 3.19.7, use the OLD behavior". Who wants that? :roll_eyes:

Also the minimum version is really just saying "hey CMake, if you are version 3.9 and this variable says 3.10, don't run and show an error". But if you set the variable to 3.10.2 and use CMake features that require a newer version (say 3.12), that won't prevent CMake 3.11 to run and fail. Like nothing will parse your files and check that you only use features available with 3.10.2 and older.

The only way is therefore to run your CI with CMake 3.10.2, and use that variable as an indication that this is the lowest version you are testing in CI. However, I am convinced that in most CMake projects out there, this variable (minimum_required) is meaningless: you just have to try with your CMake and you'll see if it fails, whether or not your CMake is older than this variable.

PR #1337, opened 2021-03-02
Added FOV and video stream light spectrum enum to Camera class

markusachtelik:
[old]
This PR adds support to subscribe to subscribe to VIDEO_STREAM_STATUS messages from the camera. https://mavlink.io/en/messages/common.html#VIDEO_STREAM_STATUS

[new]
after discussions, we decided to not duplicate the redundant mavlink messages VIDEO_STREAM_STATUS and VIDEO_STREAM_INFORMATION. Instead, we extend the current structure, and update it whenever data from either of these messages arrives.

The PR depends on https://github.com/mavlink/MAVSDK-Proto/pull/207. Once merged, I'll update the submodule here.

### DIFF
--- a/src/plugins/camera/camera_impl.cpp
+++ b/src/plugins/camera/camera_impl.cpp
@@ -60,6 +60,11 @@ void CameraImpl::init()
         this);
 
     _parent->register_mavlink_message_handler(
+        MAVLINK_MSG_ID_VIDEO_STREAM_STATUS,
+        std::bind(&CameraImpl::process_video_stream_status, this, _1),
+        this);
+
+    _parent->register_mavlink_message_handler(
         MAVLINK_MSG_ID_FLIGHT_INFORMATION,
         std::bind(&CameraImpl::process_flight_information, this, _1),
         this);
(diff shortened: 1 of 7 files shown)


JonasVautherin (commented):
Looks good! Just need to update the proto submodule once the corresponding proto PR is merged! 

Thanks! :rocket:

JonasVautherin (approved):

src/plugins/camera/camera_impl.cpp
```diff
@@ -345,6 +350,17 @@ MavlinkCommandSender::CommandLong CameraImpl::make_command_request_video_stream_
     return cmd_req_video_stream_info;
 }
 
+MavlinkCommandSender::CommandLong CameraImpl::make_command_request_video_stream_status()
+{
+    MavlinkCommandSender::CommandLong cmd_req_video_stream_status{};
+
+    cmd_req_video_stream_status.command = MAV_CMD_REQUEST_VIDEO_STREAM_STATUS;
```
julianoes, line 357: See: https://mavlink.io/en/messages/common.html#MAV_CMD_REQUEST_VIDEO_STREAM_STATUS

This command has been replaced by https://mavlink.io/en/messages/common.html#MAV_CMD_REQUEST_MESSAGE.

julianoes (commented):
I'd suggest one change, otherwise it's fine.

src/plugins/camera/camera_impl.cpp
julianoes: Like this: https://github.com/mavlink/MAVSDK/blob/261c3516e72fe007fbc285f2903679bc27aadceb/src/plugins/camera/camera_impl.cpp#L291-L303

julianoes: Ok, we'll leave it as is for now and transition later.

julianoes (approved):

PR #1339, opened 2021-03-02
Temporary fix for CI build

JonasVautherin:
On Ubuntu we need a newer version of python-protobuf than what comes with the system


JonasVautherin:
Closing in favor of https://github.com/mavlink/MAVSDK-Proto/pull/208

PR #1343, opened 2021-03-03
mission: move QGC plan import from mission to mission_raw

julianoes:
This moves the method to import QGroundControl .plan JSON mission files from the mission plugin to the mission_raw plugin.

The reasoning is as follows:
The imported missions did not really fit the Mission API, and I guess that the mission progress was often wrong. When the missions were imported into mission, many items/commands were unsupported and dropped with a warning. Such a warning is not visible in the language wrappers and could lead to surprises. The documentation said that the method would fail if any unsupported item was being imported, however, that was not true anymore in code.

Therefore, in my opinion, it makes a lot more sense to implement the import in mission_raw where all/any mission items and commands can be imported. It's should be more complete, without surprises, and therefore safer to use.

And in later steps we can also add geofence and rally items to this API.

This PR contains the proto change and the implementation in code. It also slightly improves the parsing by moving it to a separate file and adding a bit more unit test coverage.

Depends on https://github.com/mavlink/MAVSDK-Proto/pull/211.


examples/fly_qgc_mission/fly_qgc_mission.cpp
```diff
@@ -130,44 +124,26 @@ int main(int argc, char** argv)
...
-        const Mission::Result result = future_result.get();
-        handle_mission_err_exit(result, "Mission start failed: ");
-    }
+    result = mission_raw.start_mission();
+    handle_mission_err_exit(result, "Mission start failed: ");
 
-    while (!mission.is_mission_finished().second) {
+    while (telemetry.armed()) {
         sleep_for(seconds(1));
     }
```
JonasVautherin, line 148: I guess I'd rather use `telemetry.subscribe_armed()` and a fut/prom, but that's fine like this, too

src/plugins/mission/mission_impl.cpp
```diff
@@ -933,6 +931,7 @@ Mission::Result MissionImpl::convert_result(MAVLinkMissionTransfer::Result resul
     }
 }
 
+#if 0
```
JonasVautherin, line 934: Shouldn't that code be removed?

third_party/openssl/CMakeLists.txt
```diff
@@ -29,6 +29,7 @@ if(ANDROID)
         PATCH_COMMAND patch -p 0 < ${PROJECT_SOURCE_DIR}/dockcross-android.patch
         CONFIGURE_COMMAND ${CMAKE_COMMAND} -E env ANDROID_NDK=$ENV{CROSS_ROOT} <SOURCE_DIR>/Configure --sysroot=$ENV{CROSS_ROOT}/sysroot ${OPENSSL_PLATFORM} --prefix=${CMAKE_INSTALL_PREFIX} no-shared
         BUILD_COMMAND make
+        INSTALL_COMMAND make install_sw
```
JonasVautherin, line 32: what's that? :laughing:

JonasVautherin (dismissed):
Looks good except for an ifdef that I don't really get

examples/fly_qgc_mission/fly_qgc_mission.cpp
julianoes: Hehe, it's simple enough. And I don't care too much for this example. There is lots that could be improved in my opinion.

src/plugins/mission/mission_impl.cpp
julianoes: Oops, thanks.

third_party/openssl/CMakeLists.txt
julianoes: Oh, that was just a try to prevent it from installing tons of docs every time but it did not work. Let me remove it again.

julianoes:
@JonasVautherin fixed what you raised, squashed commits and force pushed.

JonasVautherin (approved):
Nice, thanks!

PR #1345, opened 2021-03-04
Gimbal improvements: added example, and v2 ROI

julianoes:
FYI @TSC21.


examples/CMakeLists.txt
```diff
@@ -14,6 +14,7 @@ add_subdirectory(fly_multiple_drones)
 add_subdirectory(fly_qgc_mission)
 add_subdirectory(follow_me)
 add_subdirectory(geofence_inclusion)
+add_subdirectory(gimbal)
```
TSC21, line 17: `gimbal_control` maybe?

TSC21:
Awesome!

examples/CMakeLists.txt
julianoes: Yer, I was wondering that too. Often we just name the example the same as the plugin, so from that perspective it's fine.

JonasVautherin (approved):

JonasVautherin (approved):

PR #1350, opened 2021-03-09
TrackingServer - Fixed is_command_sender_ok result checking

MatejFranceskin:
Typo in checking the result of is_command_sender_ok method prevented tracking commands from being executed.
Also removed duplicated log output.


JonasVautherin (approved):

julianoes (approved):
Oops, thanks!

PR #1353, opened 2021-03-11
Improve realtime factor estimate

julianoes:
It turns out that the timestamps vary quite a lot and our filtering did not give accurate results, we actually need to integrate the simulated and real time. I decided to use a circular buffer for the integration, although some online integrator would have probably been easier, a well.


JonasVautherin (approved):
Ok! :laughing:

PR #1356, opened 2021-03-12
Add catalyst builds to CI

JonasVautherin:
I have no clue if that works :see_no_evil:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -416,6 +416,9 @@ jobs:
           - name: ios_simulator
             platform: SIMULATOR64
             sdk: iphonesimulator
+          - name: iphoneos catalyst
+            platform: MAC_CATALYST
+            sdk: iphoneos
     steps:
       - uses: actions/checkout@v2
         with:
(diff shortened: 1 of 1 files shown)


JonasVautherin:
Two notes:

1. It fails, I don't really know why and I don't have time to look into this. Open for contributions: if someone can fix that, I can probably continue the integration.
2. I also tried to build for M1, but it seems like it cannot be cross-compiled, so it will have to wait until github actions provide M1 machines.

rligocki:
Based on this discussion (https://gitlab.kitware.com/cmake/cmake/-/issues/20132) I would suggest to change macOS SDK to iOS in catalyst build. 

In iOS configure, there is iOS SDK (Line 16 in configure part):
`-- Using SDK: /Applications/Xcode_12.4.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS14.4.sdk`
In catalyst build there is macOS SDK (Line 14 in configure part):
`-- Using SDK:  /Applications/Xcode_12.4.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.1.sdk`

rligocki:
In configure part of build on line 104 there is another clue. This could be proof of my previous idea and future collision. 
`-- Using minimum deployment version: 13.0 (SDK version: 11.1)`

julianoes:
@rligocki feel free to create a PR. If you happen to have an m1 to test that would be great.

PR #1358, opened 2021-03-16
Logfiles: test if target is an unexisting file before download

JonasVautherin:
This is checking if the target log file is not a directory and if it does not already exist. It errors with an InvalidArgument error if the argument is not appropriate.

This hopes to improve issues like https://github.com/mavlink/MAVSDK-Python/issues/323 and https://github.com/mavlink/MAVSDK-Python/issues/312 where it was not clear how to use the arguments.

### DIFF
--- a/src/plugins/log_files/log_files_impl.cpp
+++ b/src/plugins/log_files/log_files_impl.cpp
@@ -8,13 +8,13 @@
 #include <cstring>
 
 #if __has_include(<filesystem>)
-  #include <filesystem>
-  namespace fs = std::filesystem;
+#include <filesystem>
+namespace fs = std::filesystem;
 #elif __has_include(<experimental/filesystem>)
-  #include <experimental/filesystem>
-  namespace fs = std::experimental::filesystem;
+#include <experimental/filesystem>
+namespace fs = std::experimental::filesystem;
 #else
-  error "Missing the <filesystem> header."
+error "Missing the <filesystem> header."
 #endif
 
 namespace mavsdk {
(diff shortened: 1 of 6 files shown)


JonasVautherin:
Aha do I need to add `-lstdc++fs` somewhere? @julianoes would that be acceptable?

JonasVautherin:
Not sure if we want this, but just trying to do it for the exercise (would be good to know if we can benefit from `filesystem` later or not, also for e.g. the mavftp plugin)

julianoes:
`-lstdc++fs` no idea? If it builds for everything we support, yes.

src/plugins/log_files/log_files_impl.cpp
```diff
@@ -1,11 +1,22 @@
...
 #include <cstring>
 
+#if __has_include(<filesystem>)
+#include <filesystem>
+namespace fs = std::filesystem;
+#elif __has_include(<experimental/filesystem>)
+#include <experimental/filesystem>
+namespace fs = std::experimental::filesystem;
+#else
+error "Missing the <filesystem> header."
```
julianoes, line 17:
```suggestion
#error "Missing the <filesystem> header."
```

JonasVautherin:
I needed to link `-lstdc++fs` for some platforms but not all (not necessary on Windows, but it was failing with macOS/iOS and Android so I removed it there). It was needed for Ubuntu 18 at least (together with the `#ifdef` logic with `experimental/filesystem`, as described [here](https://askubuntu.com/questions/1256440/how-to-get-libstdc-with-c17-filesystem-headers-on-ubuntu-18-bionic)).

I also needed to bump the DEPLOYMENT_TARGET of iOS to 13.0 to use `path`, which is fine because almost all the iOS devices out there are at least on iOS 13.

julianoes (approved):

Issue #1316, opened 2021-02-10
Getting "timeout" issue while uploading mission on PixHawk (PX4) via MAVSDK


bb17pugc:
Hi,
I am using MAVSDK to upload mission from terminal to my PixHawk 4 (PX4) via telemetry.
Sometimes it works like a charm and sometime it give the "timout" error and stops.

Can someone please tell the fix for this issue? I tired to go into the file and updated the timeout value from "none" to "50" but it didn't worked. Same issue, sometimes it work and sometimes not work.

Complete code file is attached here. (due to editor restriction, changed the extension from ".py" to ".txt").
I am using Ubuntu 20.04

![image](https://user-images.githubusercontent.com/49359384/107490425-afec2300-6bab-11eb-907a-bc6f73d448e6.png)

[mission.txt](https://github.com/mavlink/MAVSDK/files/5957145/mission.txt)

julianoes:
Your code looks all fine to me.

I'm suspecting a problem with the connection. Are you using a radio or a direct FTDI cable?
And which PX4 version are you using?

Maybe we could resolve this by just increasing the timeout or making it configurable.

bb17pugc:
I am using the **FTDI Cable** to connect the Telemetry with my laptop.
Telemetry: **RFD900+**

PX4 Version: **1.11**
Where we can increase that time and what is the unit (sec/msec) of time to increase?

julianoes:
Ok, and is the upload working when the drone is close, and presumably the connection is good?

julianoes:
The timeout is set in https://github.com/mavlink/MAVSDK/blob/855a4b5d1e7c13a601781792f25dc5fe08ff5ae3/src/core/mavlink_mission_transfer.h#L248 so you would have to rebuild mavsdk_server yourself.

bb17pugc:
> Ok, and is the upload working when the drone is close, and presumably the connection is good?

Drone is at about 3-4 feet from the GCS telemetry. Even then the problem is coming.
Sometime it works with the 200+ mission items, and sometimes it didn't work even with the 5 mission items.

I will try the change in timeout and then let you know the feedback. Thank you.
What is the unit of time we are giving? Is this seconds or something else?

julianoes:
It's seconds there. Are you rebuilding mavsdk_server yourself already?

bb17pugc:
> It's seconds there. Are you rebuilding mavsdk_server yourself already?

[...]

PR #1362, opened 2021-03-18
core: make timeout configurable

julianoes:
This extracts the hard-coded timeouts of 0.5 seconds into MavsdkImpl and exposes it in Mavsdk for the user to adapt based on whether MAVSDK is communicating just over localhost or over a high latency radio link.

Should fix #617.

### DIFF
--- a/src/core/system_impl.cpp
+++ b/src/core/system_impl.cpp
@@ -25,7 +25,8 @@ SystemImpl::SystemImpl(MavsdkImpl& parent, uint8_t system_id, uint8_t comp_id, b
     _receive_commands(*this),
     _timesync(*this),
     _ping(*this),
-    _mission_transfer(*this, _message_handler, _parent.timeout_handler)
+    _mission_transfer(
+        *this, _message_handler, _parent.timeout_handler, [this]() { return timeout_s(); })
 {
     _target_address.system_id = system_id;
     // FIXME: for now use this as a default.
@@ -107,6 +108,11 @@ void SystemImpl::refresh_timeout_handler(const void* cookie)
 void SystemImpl::unregister_timeout_handler(const void* cookie)
 {
     _parent.timeout_handler.remove(cookie);
+}
+
+double SystemImpl::timeout_s() const
+{
+    return _parent.timeout_s();
 }
 
 void SystemImpl::enable_timesync()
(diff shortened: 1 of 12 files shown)


src/core/system_impl.cpp
```diff
@@ -25,7 +25,8 @@ SystemImpl::SystemImpl(MavsdkImpl& parent, uint8_t system_id, uint8_t comp_id, b
     _receive_commands(*this),
     _timesync(*this),
     _ping(*this),
-    _mission_transfer(*this, _message_handler, _parent.timeout_handler)
+    _mission_transfer(
+        *this, _message_handler, _parent.timeout_handler, [this]() { return timeout_s(); })
```
JonasVautherin, line 29: I am not sure I understand why there is this `TimeoutSCallback` callback at all, since it just returns the value of the `timeout_s` member? :thinking:

julianoes: I want to get the latest timeout set by the user by the time a transfer is initiated. As the MissionTransfer instance is created with the system we might miss the user updating the timeout. Therefore, I've added the callback to get the value later. The alternative would be to pass the timeout every time you do a transfer, so then the timeout value would have to be fed through via the plugin which is using the MissionTransfer.
I guess the problem there is that the plugin is not talking to the SystemImpl API but directly getting the MissionTransfer class here:
https://github.com/mavlink/MAVSDK/blob/develop/src/core/system_impl.h#L232

JonasVautherin (dismissed):

MuhammadBilal1:
@julianoes How can we use this new parameter of Timeout? I was getting this issue https://github.com/mavlink/MAVSDK/issues/1316 and still getting this issue while uploading the mission.

I was checking the documentation and don't find where I need to put this timeout parameter.
Can you please guide how to use and in which method this needs to be used?

JonasVautherin:
Here: https://github.com/mavlink/MAVSDK/blob/main/src/core/mavsdk.h#L271-L279

JonasVautherin:
Oh right, it's not exposed (yet), only in C++ :confused:

beenjamminnallday:
Hello, 
I am also getting the issue #1316, has this been exposed to MAVSDK-Python? If not, @JonasVautherin could you please guide me on how to do it? 

Any help would be appreciated.
BM

JonasVautherin:
It is exposed in the core plugin: https://github.com/mavlink/MAVSDK-Proto/blob/main/protos/core/core.proto#L22

Let me make a new release of MAVSDK-C++ today, then updating Python to use the latest version should bring it :+1:

beenjamminnallday:
Thank you very much, could you please write here when you will have updated the Python version? Thank you very much for all your help and constant support :) You guys are the best!!

PR #1366, opened 2021-03-19
core: only use UID from autopilot

julianoes:
This prevents the error "Error: UUID changed" that we saw when connecting to a system with multiple components if the heartbeat of one of the components other than the autopilot would arrive first.

We now leave the uid to 0 as long we have no autopilot discovered. We set it to the MAVLink system ID if the autopilot does not send the autopilot_version message or has autopilot_version.uid set to 0.


JonasVautherin (approved):
Oh, that makes sense!

PR #1367, opened 2021-03-21
Add dependency on get_entries to force its use before downloading a log

JonasVautherin:
Also generate temporary file names with `std::tmpnam(nullptr)` in integration tests instead of hardcoding them.

Should help for issues like https://github.com/mavlink/MAVSDK-Python/issues/323 and https://github.com/mavlink/MAVSDK-Python/issues/312.

Depends on https://github.com/mavlink/MAVSDK-Proto/pull/218.

### DIFF
--- a/src/integration_tests/log_files.cpp
+++ b/src/integration_tests/log_files.cpp
@@ -36,15 +36,14 @@ TEST(HardwareTest, LogFiles)
             LogInfo() << "Entry " << entry.id << ": "
                       << " at " << entry.date << ", " << size_mib
                       << " MiB, bytes: " << entry.size_bytes;
-            std::stringstream file_path_stream;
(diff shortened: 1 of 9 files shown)


src/integration_tests/log_files.cpp
```diff
@@ -36,15 +36,14 @@ TEST(HardwareTest, LogFiles)
             LogInfo() << "Entry " << entry.id << ": "
                       << " at " << entry.date << ", " << size_mib
                       << " MiB, bytes: " << entry.size_bytes;
-            std::stringstream file_path_stream;
-            file_path_stream << "/tmp/logfile_" << entry.id << ".ulog";
+            std::string file_path = std::tmpnam(nullptr);
```
julianoes, line 39: Nice, I didn't know about this.

julianoes (approved):

PR #1374, opened 2021-03-23
third_party: fix patch does not apply

julianoes:
Cmake from version 3.18 tries to apply the patch every time anything in cmake files changed and it triggers a reconfigure. This then leads to "patch does not apply" errors. By checking out cleanly we can avoid this.

### DIFF
--- a/third_party/absl/CMakeLists.txt
+++ b/third_party/absl/CMakeLists.txt
@@ -28,6 +28,6 @@ ExternalProject_add(
     GIT_REPOSITORY https://github.com/abseil/abseil-cpp.git
     GIT_TAG 20200923.3
     PREFIX absl
-    PATCH_COMMAND git apply ${PROJECT_SOURCE_DIR}/absl_c++11.patch
+    PATCH_COMMAND git checkout . && git apply ${PROJECT_SOURCE_DIR}/absl_c++11.patch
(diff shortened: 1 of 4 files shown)


third_party/absl/CMakeLists.txt
```diff
@@ -28,6 +28,6 @@ ExternalProject_add(
     GIT_REPOSITORY https://github.com/abseil/abseil-cpp.git
     GIT_TAG 20200923.3
     PREFIX absl
-    PATCH_COMMAND git apply ${PROJECT_SOURCE_DIR}/absl_c++11.patch
+    PATCH_COMMAND git checkout . stash && git apply ${PROJECT_SOURCE_DIR}/absl_c++11.patch
```
JonasVautherin, line 31:
```suggestion
    PATCH_COMMAND git checkout . && git apply ${PROJECT_SOURCE_DIR}/absl_c++11.patch
```

JonasVautherin (commented):
I'm all for it if it fixes that issue! :rocket:

third_party/absl/CMakeLists.txt
julianoes: Oops, I missed this one.

JonasVautherin (approved):

PR #1381, opened 2021-03-25
Remove tmpnam

julianoes:
This removes the warning about tmpnam that I get at linking, and it means I can actually see and check the files that I downloaded.


julianoes:
Self merging, I hope you're not mad @JonasVautherin :wink:.

JonasVautherin (commented):
I actually used the random name also because otherwise the integration tests would fail ("file already exists" with the hardcoded names).

But it seems like it passed the CI tests, so I guess that's fine.

PR #1392, opened 2021-03-31
Fix CI build for cmake 3.20 regarding -j 2

JonasVautherin:
Seems like `-j2` may be better? Also I'm not sure if it makes sense at all in the configure step, does it? :thinking: 

The reason for that change is that some CI checks (seems like ubuntu 18.04 updated to cmake 3.20 recently) give the following kind of error at configure time:

```
Run cmake -DCMAKE_BUILD_TYPE=Release -DSUPERBUILD=OFF -DWERROR=OFF -j 2 -Bbuild/release -H.
CMake Error: Unknown argument -j
CMake Error: Run 'cmake --help' for all supported options.
```

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -30,9 +30,9 @@ jobs:
       - name: Install lcov
         run: sudo apt-get update && sudo apt-get install -y lcov
       - name: configure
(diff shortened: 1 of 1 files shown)


julianoes (commented):
Funny, thanks.

Issue #1391, opened 2021-03-31
Structure Cmake for Cpp package managers


bazfp:
It would be useful to be able to use MAVSDK with cpp package managers (Hunter, Conan etc) for automatic dependency resolution. Instead of depending on a `SUPERBUILD` or local dependencies.

I've been implementing this library with [hunter](https://github.com/cpp-pm/hunter) and have found success by doing the following:

0. `SUPERBUILD=OFF`
1. Remove git commands from Cmakelists.txt, such as getting version from git tags
2. Don't use git submodules 
3. Remove Find modules provided by https://github.com/mavlink/MAVSDK/blob/main/src/cmake/Modules/
4. Use common lib names when linking: 
`JsonCpp::jsoncpp` -> `jsoncpp_lib`
`tinyxml2::tinyxml2` -> `tinyxml2`

This won't work for `mavsdk_server` but it is a start. There is likely to be more changes to this to get it to work (for Hunter I have included Hunter_gate but obviously this differs for each manager :) )

JonasVautherin:
How did you update `Findtinyxml2.cmake`? Does it still work for the other cases?

bazfp:
I removed lines 18-23.

It is unlikely to work in other cases - that why I have not submitted a PR.

JonasVautherin:
But why doesn't `find_package(TINYXML2)` work in the first place? Isn't the package manager (e.g. hunter) supposed to work with it? Like by setting some `CMAKE_PREFIX_PATH` or something?

We use pkgconfig as a fallback already...

bazfp:
Yes you're correct. It was not working due to the aliased lib names in the provided Find cmake modules. I no longer use the provided `src/cmake/Modules/Find*` modules now and lean on the package manager for finding packages. I have had to rename the libs that were linked as they were incorrect.

`Jsoncpp::jsoncpp` -> `jsoncpp_lib`
`tinyxml2::tinyxml2` -> `tinyxml2`

JonasVautherin:
FYI, I believe it should be fixed for jsoncpp using https://github.com/mavlink/MAVSDK/pull/1395. It's not a release yet, so I don't really know how it goes with Hunter, but I think jsoncpp now has the behavior you expect.

bazfp:
Aha, Good news :)

I will wait for fresh `jsoncpp` & `tinyxml2` releases with the namespaced library names, and push them to hunter.

[...]

PR #1393, opened 2021-04-02
Initial hunter support #1391

bazfp:
Added a simple hunter support for getting dependencies. Builds fine here

The lib variables are a bit jank, but not sure what to do when the default cmake aliases them to something different than the actual library!

Build with `SUPERBUILD=OFF`

### DIFF
--- a/src/plugins/camera/CMakeLists.txt
+++ b/src/plugins/camera/CMakeLists.txt
@@ -9,7 +9,7 @@ target_link_libraries(mavsdk_camera
     PUBLIC
     mavsdk
     PRIVATE
-    tinyxml2::tinyxml2
+    ${TINYXML2_LIB_NAME}
     CURL::libcurl
 )
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,11 @@
 cmake_minimum_required(VERSION 3.10.2)
 
+include("src/cmake/HunterGate.cmake")
+HunterGate(
+    URL "https://github.com/ruslo/hunter/archive/v0.23.214.tar.gz"
+    SHA1 "e14bc153a7f16d6a5eeec845fb0283c8fad8c358"
+)
+
 execute_process(
     COMMAND git describe --abbrev=8 --dirty --always --tags
     WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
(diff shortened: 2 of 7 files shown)


src/plugins/camera/CMakeLists.txt
```diff
@@ -9,7 +9,7 @@ target_link_libraries(mavsdk_camera
     PUBLIC
     mavsdk
     PRIVATE
-    tinyxml2::tinyxml2
+    ${TINYXML2_LIB_NAME}
```
JonasVautherin, line 12: Can't we use the targets with Hunter? I think that's the modern way...

JonasVautherin (changes requested):
I am not against supporting Hunter on the principle, but I think it should work with minimal changes. Dropping the targets syntax for instance seems like a regression. And what if somebody comes tomorrow and wants to support a package manager that requires the target syntax?

My second question would be about keeping it separate: if Hunter could use the target syntax (`tinyxml2::tinyxml2`), it seems like this could all be done out of the project (somehow at the superbuild level), because it seems like all Hunter does is provide `hunter_add_package` and then override `find_package`. So ideally anything lower than the root (e.g. `src/CMakeLists.txt`) should not even know that Hunter is being used...

Does that make sense?

JonasVautherin:
This said, we are applying a patch to jsoncpp 1.8.4, so it may not be the same as what Hunter provides. I looked into updating to 1.9.4 to be closer to upstream, and I think we would need [this PR](https://github.com/open-source-parsers/jsoncpp/pull/1271).

That's mainly FYI, I'll try to update jsoncpp as soon as possible.

julianoes:
@bazfp FYI tinyxml2 is updated now.

JonasVautherin:
For the record I think that we would like to have https://github.com/leethomason/tinyxml2/pull/866 and https://github.com/leethomason/tinyxml2/pull/865 merged, and released in a new version of tinyxml. The @bazfp could maybe get that new version into Hunter?

bazfp:
> @bazfp could maybe get that new version into Hunter?

Can do, still waiting on jsoncpp release

bazfp:
Update to this - seems upstream hunter has accepted @julianoes forked release of `jsoncpp`

I will revisit this to confirm it builds with hunter

PR #1394, opened 2021-04-02
Fix jsoncpp Werror with NDK r22b

JonasVautherin:
JsonCpp fails to build on Android with NDK r22b. This is not a problem for the CI yet, but it will come soon (the CI is on r21). This fixes it.

I am also looking into updating JsonCpp, which may allow us to remove all the patches. But that may take longer, so I fixed this issue first.

### DIFF
--- a/third_party/jsoncpp/CMakeLists.txt
+++ b/third_party/jsoncpp/CMakeLists.txt
@@ -31,6 +31,6 @@ ExternalProject_Add(
     GIT_REPOSITORY https://github.com/open-source-parsers/jsoncpp
     GIT_TAG 1.8.4
     PREFIX jsoncpp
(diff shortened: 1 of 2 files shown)


julianoes:
We could also just make sure we don't try to build with `-Werror` for everything downstream, right? Because these warnings will keep coming up for other compilers.

JonasVautherin:
Yeah... but I don't think we set it for downstream at all, so that is something to do for each dependency separately. In this case, it seemed to me that the default for jsoncpp was without werror, but for some reason it was failing here for Android. I did not look further into it because I think we should update to jsoncpp 1.9.4 as soon as possible, so this PR was just to not fail in the meantime.

julianoes (approved):

PR #1395, opened 2021-04-02
Update jsoncpp

JonasVautherin:
This is not using upstream, but a PR there (i.e. it will have to wait before being mergeable here).

Good news is that it removes the patches :+1:.


third_party/jsoncpp/CMakeLists.txt
```diff
@@ -28,9 +28,10 @@ endforeach()
 
 ExternalProject_Add(
     jsoncpp
-    GIT_REPOSITORY https://github.com/open-source-parsers/jsoncpp
-    GIT_TAG 1.8.4
+    GIT_REPOSITORY https://github.com/sergeyrachev/jsoncpp
+    GIT_TAG cmake-config-improvements
+    #GIT_REPOSITORY https://github.com/open-source-parsers/jsoncpp
+    #GIT_TAG 1.8.4
```
JonasVautherin, line 34: This is the part that will require a new release upstream (instead of using Sergey's PR)

julianoes:
Awesome!

cdunn2001:
FYI: I just merged open-source-parsers/jsoncpp#1271 . Please let us know if that causes a problem for you.

JonasVautherin:
I suggest we move to jsoncpp's HEAD until they make a new release, so that we already benefit from those CMake improvements :+1:

julianoes (approved):

PR #1398, opened 2021-04-06
Update README: link to build guide moved

potaito:
I believe this should be the link, correct? https://mavsdk.mavlink.io/main/en/cpp/guide/build.html

EDIT: Here's another link :thinking:  https://mavsdk.mavlink.io/develop/en/contributing/build.html


julianoes (approved):
Yes! Thx.

Issue #1400, opened 2021-04-08
subscribe_mission_progress() of class MissionRaw provides more output than expected if vehicle stops at waypoint


fas2037:

fas2037:
Steps to reproduce: Adapt fly_mission.cpp of examples to use MissionRaw and push back 1 waypoint. Start mission and subscribe to MissionProgress. 
Expected Results: As soon as the first and only waypoint is reached The MissionProgress Callback is called once when the waypoint is reached to confirm this.
Actual Result: MissionProgress is called multiple times with different ooutput messages, even though the vehicle stopsat the waypoin.
```
Arming...
[08:35:23|Debug] MAVLink: info: Armed by external command (system_impl.cpp:250)
Armed.
Mission status update: 0 / 1
[08:35:23|Debug] MAVLink: info: Takeoff to 8.9 meters above home. (system_impl.cpp:250)
[08:35:24|Debug] MAVLink: info: Takeoff detected (system_impl.cpp:250)
[08:35:45|Debug] MAVLink: info: Mission finished, loitering. (system_impl.cpp:250)
Mission status update: 1 / 1
Mission status update: 0 / 1
Mission status update: 1 / 1
Mission status update: 0 / 1
Mission status update: 1 / 1
Mission status update: 0 / 1
Mission status update: 1 / 1
Mission status update: 0 / 1
Mission status update: 1 / 1
Mission status update: 0 / 1
Mission status update: 1 / 1
Mission status update: 0 / 1
Mission status update: 1 / 1
Mission status update: 0 / 1
Mission status update: 1 / 1
Mission status update: 0 / 1
Mission status update: 1 / 1
Mission status update: 0 / 1
Mission status update: 1 / 1
Mission status update: 0 / 1
Mission status update: 1 / 1
Mission status update: 0 / 1
```

fas2037:
[mission.txt](https://github.com/mavlink/MAVSDK/files/6283640/mission.txt)
Make sure to use the MissionRaw class.

PR #1402, opened 2021-04-08
Add last reached mission item to mission raw progress (#1400)

fas2037:
Currently MissionProgress of class MissionRaw can't handle the case in which the vehicle stays at a mission item (either if this is the last mission item or the autocontinue flag is set to false). In this case the two messages CURRENT and REACHED conflict with each other. To differntiate between the current mission item and the last reached mission item a new property of MissionProgress `reached` is introduced . If a vehicle holds at a mission item `current` and `reached` will be identical. If the vehicle approaches another mission item `current` will be the sequence ID of the next mission item.
This fixes issue #1400.

### DIFF
--- a/third_party/jsoncpp/CMakeLists.txt
+++ b/third_party/jsoncpp/CMakeLists.txt
@@ -31,6 +31,6 @@ ExternalProject_Add(
     GIT_REPOSITORY https://github.com/open-source-parsers/jsoncpp
     GIT_TAG 1.8.4
     PREFIX jsoncpp
-    PATCH_COMMAND git checkout . && git apply ${PROJECT_SOURCE_DIR}/fixlibname.patch
+    PATCH_COMMAND git checkout . && git apply ${PROJECT_SOURCE_DIR}/fixlibname.patch && git apply ${PROJECT_SOURCE_DIR}/fix_werror.patch
     CMAKE_ARGS "${CMAKE_ARGS}"
     )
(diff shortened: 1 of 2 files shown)


JonasVautherin (changes requested):
CI is failing because of the style check. Fix the style by running the script:

```
./tools/fix_style .
```

julianoes (approved):
Nice, thanks for the fix. Does this work properly for you now?

fas2037:
@julianoes Yes, it works fine, thanks.

@JonasVautherin It seems like the test fails while running `git diff --exit-code`. How is it possible to fix this?

JonasVautherin:
> @JonasVautherin It seems like the test fails while running git diff --exit-code. How is it possible to fix this?

@fas2037: Fix the style by running the `fix_style` script :slightly_smiling_face: :

```
./tools/fix_style.sh .
```

JonasVautherin:
Whoops, I messed up, let me re-open it

julianoes:
@fas2037 we just figured it out. The problem is not that the style is wrong but rather that this field is defined by the proto file and auto-generated, so it can't changed that way.

I need to re-think why this happens in the first place.

fas2037:
@julianoes Okay, so from my side there aren't any changes required?

julianoes:
I guess not. The only thing you could help me with, is uploading a mission.plan file to the issue so that I can use that to reproduce the issue. You'll have to rename it to .txt so GitHub accepts it.

PR #1405, opened 2021-04-08
follow_me: fix double sending of location

julianoes:
We changed the behavior of add_call_every a while ago. It now triggeres immedately as well as later. This means that the manual immediate trigger is no longer necessary. Otherwise, this leads to weird double bursts.


JonasVautherin (dismissed):

potaito:
Now sometimes 1-2 messages go missing and never arrive :/

JonasVautherin (approved):
Oh, tricky

PR #1406, opened 2021-04-08
Fix race in System initialization

JonasVautherin:
A system should not be discovered before the object is created, otherwise we end up with race conditions. Instead a system should be constructed first, and then started

### DIFF
--- a/src/core/system.h
+++ b/src/core/system.h
@@ -36,16 +36,24 @@ public:
      * This constructor is not (and should not be) directly called by application code.
      *
      * @param parent `MavsdkImpl` dependency.
-     * @param system_id System id.
-     * @param comp_id Component id.
-     * @param connected If true then the system doesn't wait for heartbeat to go into connected
-     * state
      */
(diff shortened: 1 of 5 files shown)


src/core/system.h
```diff
@@ -36,17 +36,23 @@ class System {
...
 
+    /**
+     * @brief Initialize the system.
+     *
+     * @param system_id System id.
+     * @param component_id Component id.
+     * @param connected If true then the system doesn't wait for heartbeat to go into connected
+     * state
+     */
+    void init(uint8_t system_id, uint8_t component_id, bool connected) const;
```
julianoes, line 54: I think this should be private, right? Because users of the API should not be able to call this, right?

JonasVautherin: It used to be in the constructor, so it was exposed :thinking:. I don't think it can be private. The alternative is to use a factory, so that it's clear for the user how to instantiate it...

julianoes: Ok you're right. But please add a comment that this is to be used internally only.

julianoes (approved):

PR #1408, opened 2021-04-09
tools: don't display command not found

julianoes:
This just caught my eye.

### DIFF
--- a/tools/run-docker.sh
+++ b/tools/run-docker.sh
@@ -2,13 +2,13 @@
 
 dockerimage=mavsdk/mavsdk-ubuntu-20.04-px4-sitl-v1.11
 
-if type podman > /dev/null
+if type podman 2> /dev/null
 then
     podman run -it --rm -v $(pwd):/home/user/MAVSDK:z $dockerimage "$@"
     echo "sudo needed to repair file ownership after podman ran: sudo chown -R $USER:$USER ."
     sudo chown -R $USER:$USER .
 
-elif type docker > /dev/null
+elif type docker 2> /dev/null
 then
     docker run -it --rm -v $(pwd):/home/user/MAVSDK:z -e LOCAL_USER_ID=`id -u` $dockerimage "$@"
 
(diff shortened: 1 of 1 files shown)


JonasVautherin (changes requested):
Hmm which terminal did you test it with? For me with zsh and sh:

sh:
```
$ type podman > /dev/null
$ type podman 2> /dev/null
podman: not found
```

zsh:

```
% type podman > /dev/null
% type podman 2> /dev/null
podman not found
```

So it seems to me that `>` does what I expect, and `2>` does not :thinking:

julianoes:
odd, I tested it in zsh.

julianoes:
```
tools/run-docker.sh tools/fix_style.sh .
tools/run-docker.sh: line 5: type: podman: not found
Starting with UID: 1000
```

julianoes:
Bash:
```
type podman > /dev/null
bash: type: podman: not found
```

julianoes:
And since this is made for bash I think the PR is correct.

tools/run-docker.sh
```diff
@@ -2,13 +2,13 @@
 
 dockerimage=mavsdk/mavsdk-ubuntu-20.04-px4-sitl-v1.11
 
-if type podman > /dev/null
+if type podman 2> /dev/null
```
JonasVautherin, line 5:
```suggestion
if type podman > /dev/null 2> /dev/null
```

tools/run-docker.sh
```diff
@@ -2,13 +2,13 @@
...
 then
     podman run -it --rm -v $(pwd):/home/user/MAVSDK:z $dockerimage "$@"
     echo "sudo needed to repair file ownership after podman ran: sudo chown -R $USER:$USER ."
     sudo chown -R $USER:$USER .
 
-elif type docker > /dev/null
+elif type docker 2> /dev/null
```
JonasVautherin, line 11:
```suggestion
elif type docker > /dev/null 2> /dev/null
```

JonasVautherin (commented):
What about a tradeoff? Works for me on `sh`, `zsh` and `bash`.

julianoes:
Hm, you're right. :)

PR #1414, opened 2021-04-19
follow_me: send velocity array

potaito:
The velocity fields of the follow-me mavlink message is currently hard-coded to `NAN`. With this PR mavsdk will send the actual values instead.

@julianoes sorry it took me so long.

### DIFF
--- a/src/plugins/follow_me/follow_me_impl.cpp
+++ b/src/plugins/follow_me/follow_me_impl.cpp
@@ -279,7 +279,10 @@ void FollowMeImpl::send_target_location()
     const float alt = static_cast<float>(_target_location.absolute_altitude_m);
 
     const float pos_std_dev[] = {NAN, NAN, NAN};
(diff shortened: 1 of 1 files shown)


JonasVautherin (changes requested):
I think it's failing the style check. Can you run the following script?

```
./tools/fix_style.sh .
```

JonasVautherin (approved):
Looks good to me :+1:

Out of curiosity: could it be that we someday decide to use the others, too? E.g. `accel`, `attitude`, `rates`? Or is there a reason for never using them?

julianoes (commented):
Awesome, thanks @potaito :potato:

potaito:
> Looks good to me +1
> 
> Out of curiosity: could it be that we someday decide to use the others, too? E.g. `accel`, `attitude`, `rates`? Or is there a reason for never using them?

I suppose so, yes. But it wasn't part of my tests, so I did not include it :sweat_smile:

PR #1415, opened 2021-04-19
Do not freeze event loop while downloading camera definition

JonasVautherin:
Currently, the event loop freezes while the camera definition is being downloaded, resulting in a lot of timeouts. Moreover, it keeps trying indefinitely if the download fails.

This PR makes the download run in a separate thread, and gives up after it fails downloading 3 times.

### DIFF
--- a/src/plugins/camera/camera_impl.cpp
+++ b/src/plugins/camera/camera_impl.cpp
@@ -899,37 +899,57 @@ void CameraImpl::process_camera_information(const mavlink_message_t& message)
     mavlink_camera_information_t camera_information;
     mavlink_msg_camera_information_decode(&message, &camera_information);
 
-    {
-        std::lock_guard<std::mutex> lock(_information.mutex);
-        _information.data.vendor_name = (char*)(camera_information.vendor_name);
-        _information.data.model_name = (char*)(camera_information.model_name);
-        _information.data.focal_length_mm = camera_information.focal_length;
-        _information.data.horizontal_sensor_size_mm = camera_information.sensor_size_h;
-        _information.data.vertical_sensor_size_mm = camera_information.sensor_size_v;
-        _information.data.horizontal_resolution_px = camera_information.resolution_h;
-        _information.data.vertical_resolution_px = camera_information.resolution_v;
+    std::lock_guard<std::mutex> lock(_information.mutex);
 
-        if (_information.subscription_callback) {
-            const auto temp_callback = _information.subscription_callback;
-            const auto temp_information = _information.data;
-            _parent->call_user_callback(
-                [temp_callback, temp_information]() { temp_callback(temp_information); });
-        }
+    _information.data.vendor_name = (char*)(camera_information.vendor_name);
+    _information.data.model_name = (char*)(camera_information.model_name);
+    _information.data.focal_length_mm = camera_information.focal_length;
+    _information.data.horizontal_sensor_size_mm = camera_information.sensor_size_h;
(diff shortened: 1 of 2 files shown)


src/plugins/camera/camera_impl.cpp
```diff
@@ -899,39 +899,59 @@ void CameraImpl::process_camera_information(const mavlink_message_t& message)
...
-            LogDebug() << "Successfully loaded camera definition";
-        } else {
-            LogDebug() << "Failed to fetch camera definition!";
-        }
+    if (should_fetch_camera_definition()) {
+        _is_fetching_camera_definition = true;
+
+        std::thread([this, camera_information]() {
+            std::string content{};
+            auto succeeded = fetch_camera_definition(camera_information, content);
```
julianoes, line 924:
```suggestion
            const auto has_succeeded = fetch_camera_definition(camera_information, content);
```

src/plugins/camera/camera_impl.cpp
```diff
@@ -899,39 +899,59 @@ void CameraImpl::process_camera_information(const mavlink_message_t& message)
...
+
+            if (succeeded) {
+                LogDebug() << "Successfully loaded camera definition";
+
+                _camera_definition.reset(new CameraDefinition());
+                _camera_definition->load_string(content);
+                refresh_params();
+            } else {
+                LogDebug() << "Failed to fetch camera definition!";
```
julianoes, line 933:
```suggestion
                LogWarn() << "Failed to fetch camera definition!";
```

src/plugins/camera/camera_impl.cpp
```diff
@@ -899,39 +899,59 @@ void CameraImpl::process_camera_information(const mavlink_message_t& message)
...
+            std::lock_guard<std::mutex> thread_lock(_information.mutex);
+            _is_fetching_camera_definition = false;
+        }).detach();
     }
 }
 
+bool CameraImpl::should_fetch_camera_definition() const
+{
+    return !_camera_definition && !_is_fetching_camera_definition &&
+           !_has_camera_definition_timed_out;
```
julianoes, line 952: So if we ever timeout we won't try ever again? Shouldn't we try again after a longer timeout or so?

src/plugins/camera/camera_impl.cpp
julianoes: Oh, you have a warning below, I see, nevermind then.

src/plugins/camera/camera_impl.cpp
JonasVautherin: Not sure... what would be the use-case for that? It takes 3-5 seconds to timeout, so that's 15 before it gives up. Two questions:

1. When does it happen that it fails 3 times in a row when you connect the app, but works later?
2. If it retries every 30 or 60 seconds, my guess is that the user will learn how to restart the app faster than that.

My feeling is that if you need the camera controls (e.g. for a survey), you will never want to get them in the middle of the mission: you want them before you start the mission, as a preflight check.

But happy to make it retry every 30 or 60 seconds if you think there is a use-case for that :+1:

julianoes: Yer, you're probably right. I was just thinking of the case where the companion takes some time to boot up.

JonasVautherin: Is that an issue? The `CAMERA_INFORMATION` message arrives only when the camera component sends it, and at this point it should be ready to serve the camera definition, shouldn't it?

julianoes: Oh, you're right. All good then.

PR #1416, opened 2021-04-19
Fix race condition when stopping mavsdk_server

JonasVautherin:
The problem right now is that a callback can still be active after the grpc server is stopped and before Mavsdk is deleted.

Could be reproduced by adding some delay when stopping the server:

```c++
 void mavsdk_server_stop(MavsdkServer* mavsdk_server)
 {
     mavsdk_server->stop();
+    std::this_thread::sleep_for(std::chrono::milliseconds(500));
     delete mavsdk_server;
 }
```

### DIFF
--- a/src/mavsdk_server/src/plugins/calibration/calibration_service_impl.h
+++ b/src/mavsdk_server/src/plugins/calibration/calibration_service_impl.h
@@ -142,11 +142,10 @@ public:
         register_stream_stop_promise(stream_closed_promise);
 
         auto is_finished = std::make_shared<bool>(false);
-
-        std::mutex subscribe_mutex{};
+        auto subscribe_mutex = std::make_shared<std::mutex>();
 
         _calibration.calibrate_gyro_async(
-            [this, &writer, &stream_closed_promise, is_finished, &subscribe_mutex](
+            [this, &writer, &stream_closed_promise, is_finished, subscribe_mutex](
                 mavsdk::Calibration::Result result,
                 const mavsdk::Calibration::ProgressData calibrate_gyro) {
                 rpc::calibration::CalibrateGyroResponse rpc_response;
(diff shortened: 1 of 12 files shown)


src/mavsdk_server/src/plugins/calibration/calibration_service_impl.h
```diff
@@ -142,11 +142,10 @@ class CalibrationServiceImpl final : public rpc::calibration::CalibrationService
         register_stream_stop_promise(stream_closed_promise);
 
         auto is_finished = std::make_shared<bool>(false);
-
-        std::mutex subscribe_mutex{};
+        auto subscribe_mutex = std::make_shared<std::mutex>();
```
julianoes, line 145: I guess that works. Could you also use one `shared_ptr` around the mutex and bool together?

JonasVautherin: The problem is that if passed as a reference, the reference inside the callback gets invalid once the function returns.

> Could you also use one shared_ptr around the mutex and bool together?

Not sure what you mean :thinking:. An `std::shared_ptr<std::pair<bool, std::mutex>>`?

julianoes: For instance, yes.

JonasVautherin: Why is that? Not sure I get it :confused:

julianoes: Ah sorry, I just meant as an optimization, so you only need to do the reference counting once.

JonasVautherin: Oh, I understand now! This function is called exactly once per subscription, and in the current state, that usually means at most once (because MAVSDK-C++ does not support registering more than one callback yet). So I don't really worry about the `shared_ptr` overhead.

But of course I can change it, I'll do that tomorrow :+1:

julianoes (approved):

src/mavsdk_server/src/plugins/calibration/calibration_service_impl.h
julianoes: Since it's tested, let's leave it!

PR #1418, opened 2021-04-22
Initiate UDP connection from MAVSDK

JonasVautherin:
## The feature

This PR adds a way for MAVSDK to initiate a UDP connection, as follows:

* `udp://:14540` stays the same, it means that MAVSDK listens for heartbeats on 14540
* `udp://0.0.0.0:14540` explicitly states that MAVSDK should be have as a server (like above) and listen for heartbeats on 14540
* `udp://192.168.1.12:14540` says that MAVSDK should send heartbeats to `192.168.1.12:14540`, and therefore initiate the connection. __This is the new feature.__

It should not break any workflow, because if somebody used the format `udp://192.168.1.12:14540` before, that was probably not what they wanted to do (the behavior then was filtering the network interface, but nobody ever does that with MAVSDK as far as I know).

## The fixes

While implementing integration tests to test the feature, I encountered two bugs:

1. When using `set_udp_remote` (in order to initiate the MAVLink connection), we would create a new system with our system_id and comp_id. Which makes no sense: we are initiating the connection to a system that is not us, but we don't know its sysid/compid. So I set it to 0/0 now.
2. In order to consider a system as discovered, we had this `if (!_connected && _uuid_initialized)` test. But `_uuid_initialized` is only ever true when the system represents an autopilot. If MAVSDK itself represents an autopilot, this just fails and systems are never discovered.

Those are small changes, but I must say I have _no clue_ what that will imply more generally. I don't really get why we had the `_uuid_initialized` condition in the first place, and I am not completely sure how starting a system with 0/0 is handled. I am pretty sure it will fail in some use-cases [here](https://github.com/mavlink/MAVSDK/blob/main/src/core/mavsdk_impl.cpp#L206-L212) (e.g. what happens if you initiate the connection to two remote systems, since both will create a 0/0 system?), but fixing that seems to require more fundamental changes in the discovery mechanism, so I'm hoping this is not needed yet!

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -90,6 +90,8 @@ jobs:
         run: ./build/release/src/unit_tests_runner
       - name: test (mavsdk_server)
         run: ./build/release/src/mavsdk_server/test/unit_tests_mavsdk_server
+      - name: test FTP server
+        run: ./build/release/src/integration_tests/integration_tests_runner --gtest_filter="FtpTest.TestServer"
 
   ubuntu20-proto-check:
     name: ubuntu-20.04 (proto check)
--- a/src/core/mavsdk_impl.cpp
+++ b/src/core/mavsdk_impl.cpp
@@ -268,15 +268,15 @@ ConnectionResult MavsdkImpl::add_any_connection(
 
     switch (cli_arg.get_protocol()) {
         case CliArg::Protocol::Udp: {
-            std::string path = Mavsdk::DEFAULT_UDP_BIND_IP;
-            int port = Mavsdk::DEFAULT_UDP_PORT;
-            if (!cli_arg.get_path().empty()) {
-                path = cli_arg.get_path();
+            int port = cli_arg.get_port() ? cli_arg.get_port() : Mavsdk::DEFAULT_UDP_PORT;
+
+            if (cli_arg.get_path().empty() || cli_arg.get_path() == Mavsdk::DEFAULT_UDP_BIND_IP) {
+                std::string path = Mavsdk::DEFAULT_UDP_BIND_IP;
+                return add_udp_connection(path, port, forwarding_option);
+            } else {
+                std::string path = cli_arg.get_path();
+                return setup_udp_remote(path, port, forwarding_option);
             }
-            if (cli_arg.get_port()) {
-                port = cli_arg.get_port();
-            }
-            return add_udp_connection(path, port, forwarding_option);
         }
 
         case CliArg::Protocol::Tcp: {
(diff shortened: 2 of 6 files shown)


julianoes (approved):
This all looks correct to me. I guess the one thing to check is the Ftp integration tests.

Also: nice commit messages and PR description. This makes the review a breeze.

julianoes:
Also, we can then add the server mode to the TCP connection.

julianoes:
This is what I currently see for the FtpTest:
```
cmake --build build -j8 && build/src/integration_tests/integration_tests_runner --gtest_filter="FtpTest.*"
ninja: no work to do.
Note: Google Test filter = FtpTest.*
[==========] Running 4 tests from 1 test suite.
[----------] 4 tests from FtpTest
[ RUN      ] FtpTest.ListDirectory
[08:32:05|Info ] MAVSDK version: v0.39.0-15-g228ceb76 (mavsdk_impl.cpp:21)
unknown file: Failure
C++ exception with description "vector::_M_range_check: __n (which is 0) >= this->size() (which is 0)" thrown in the test body.
[  FAILED  ] FtpTest.ListDirectory (15 ms)
[ RUN      ] FtpTest.DownloadFile
[08:32:05|Info ] MAVSDK version: v0.39.0-15-g228ceb76 (mavsdk_impl.cpp:21)
[08:32:05|Error] bind error: Address already in use (udp_connection.cpp:86)
../src/integration_tests/ftp.cpp:188: Failure
  ret
    Which is: Bind error
  ConnectionResult::Success
    Which is: Success
[  FAILED  ] FtpTest.DownloadFile (10 ms)
[ RUN      ] FtpTest.UploadFiles
[08:32:05|Info ] MAVSDK version: v0.39.0-15-g228ceb76 (mavsdk_impl.cpp:21)
[08:32:05|Error] bind error: Address already in use (udp_connection.cpp:86)
../src/integration_tests/ftp.cpp:220: Failure
  ret
    Which is: Bind error
  ConnectionResult::Success
    Which is: Success
[  FAILED  ] FtpTest.UploadFiles (11 ms)
[ RUN      ] FtpTest.TestServer
[08:32:05|Info ] MAVSDK version: v0.39.0-15-g228ceb76 (mavsdk_impl.cpp:21)
unknown file: Failure
C++ exception with description "vector::_M_range_check: __n (which is 0) >= this->size() (which is 0)" thrown in the test body.
[  FAILED  ] FtpTest.TestServer (0 ms)
[----------] 4 tests from FtpTest (36 ms total)

[==========] 4 tests from 1 test suite ran. (36 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 4 tests, listed below:
[  FAILED  ] FtpTest.ListDirectory
[  FAILED  ] FtpTest.DownloadFile
[  FAILED  ] FtpTest.UploadFiles
[  FAILED  ] FtpTest.TestServer

```

TSC21:
@julianoes @JonasVautherin what are we missing here to bring it in?

julianoes:
@TSC21 fixing the Ftp test.

TSC21:
> @TSC21 fixing the Ftp test.

Is that something I can help with? @JonasVautherin

TSC21 (approved):

PR #1419, opened 2021-04-23
core: fix ever increasing heartbeats

julianoes:
When a system was lost and re-discovered we would yet again start to send heartbeats but never actually stop the call_every. This leads to more and more traffic, memory, and CPU used.

Manually tested in normal configuration where heartbeats stop, and onboard configuration where heartbeats are always sent.

### DIFF
--- a/src/core/mavsdk_impl.cpp
+++ b/src/core/mavsdk_impl.cpp
@@ -394,9 +394,7 @@ void MavsdkImpl::set_configuration(Mavsdk::Configuration configuration)
 {
     _configuration = configuration;
 
-    if (configuration.get_always_send_heartbeats()) {
-        start_sending_heartbeat();
-    }
+    start_stop_sending_heartbeats();
 }
 
 std::vector<uint64_t> MavsdkImpl::get_system_uuids() const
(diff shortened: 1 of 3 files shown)


src/core/mavsdk_impl.cpp
```diff
@@ -680,10 +678,22 @@ void MavsdkImpl::process_user_callbacks_thread()
     }
 }
 
-void MavsdkImpl::start_sending_heartbeat()
+void MavsdkImpl::start_stop_sending_heartbeats()
```
JonasVautherin, line 681: Disagreed :sweat_smile:. Those are clearly two functions: `start_sending_heartbeats()` and `stop_sending_heartbeats()`.

Whenever I will see `start_stop_blah()` in the code, it will tell me "there is an `if` statement that you MUST understand now, but for some reason it is hidden inside the function. So let's have that if statement to decide whether `start` or `stop` should be called directly :grin:.

Then I don't know why `stop_send_heartbeats` would ever be called from `set_connected()` (which always sets `_connected` to `true`).

What about something like this:

1. Whenever `_connected = false`, also do `call_every_handler.remove(_heartbeat_send_cookie); _heartbeat_send_cookie = nullptr;` => sounds a lot like a `set_disconnected()` function.
2. Because a `System` wants at most one heartbeat sending loop, in `start_sending_heartbeats`, check `if (_heartbeat_send_cookie)` (which means that it is already sending) and ignore the call if it is, otherwise call `call_every_handler.add`

JonasVautherin (changes requested):

JonasVautherin (approved):

PR #1422, opened 2021-04-28
core: add has_ground_station method to system_impl and system

QuentinVecchio:
This PR introduces the method `has_ground_station` in system_impl and system so we can detect if there is a ground station (mission planner) in the system.

### DIFF
--- a/src/core/system.cpp
+++ b/src/core/system.cpp
@@ -44,6 +44,11 @@ bool System::has_gimbal() const
     return _system_impl->has_gimbal();
 }
 
+bool System::has_ground_station() const
+{
+    return _system_impl->has_ground_station();
+}
+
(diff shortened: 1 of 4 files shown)


JonasVautherin:
Two questions:

1. Why do we need that? Are there components that specifically talk to ground stations? :thinking:
2. Are we sure that a ground station is a component of a system? Feels weird to me to have a System that "has" an autopilot and also "has" a ground station... what does that mean? Is it a drone or is it a ground control thingy? :sweat_smile:

julianoes:
@JonasVautherin I remember you are opposed to this change and had an alternative solution/suggestion. I guess we can close this PR then?

QuentinVecchio:
Correct, we can close this branch

PR #1423, opened 2021-04-29
tools: fix snake_case_to_camel_case for macOS

julianoes:
The problem is that BSD (non GNU) sed on macOS does not support case conversions. Hence we need to implement the snake_case to camelCase conversion in awk.


JonasVautherin (approved):
argh :sweat_smile:

PR #1424, opened 2021-04-29
New VehicleAction in MissionItem

rligocki:
based on:
- https://github.com/mavlink/MAVSDK-Proto/pull/226
- https://github.com/mavlink/MAVSDK-Swift/issues/198

### DIFF
--- a/src/integration_tests/mission_takeoff_land.cpp
+++ b/src/integration_tests/mission_takeoff_land.cpp
@@ -70,15 +70,21 @@ void do_mission_with_takeoff_and_land(float mission_altitude_m)
         Telemetry::LandedState::TakingOff,
         Telemetry::LandedState::InAir,
         Telemetry::LandedState::Landing,
+        Telemetry::LandedState::OnGround,
         Telemetry::LandedState::TakingOff,
         Telemetry::LandedState::InAir,
         Telemetry::LandedState::Landing,
         Telemetry::LandedState::OnGround};
+    
 
     std::vector<Telemetry::LandedState> landed_states;
-    telemetry->subscribe_landed_state([&landed_states](Telemetry::LandedState landed_state) {
-        LogInfo() << landed_state;
-        landed_states.push_back(landed_state);
+    Telemetry::LandedState landed_state_previous = Telemetry::LandedState::Unknown;
+    telemetry->subscribe_landed_state([&landed_states, &landed_state_previous](Telemetry::LandedState landed_state) {
+        if(landed_state_previous != landed_state){
+            LogInfo() << landed_state;
+            landed_state_previous = landed_state;
+            landed_states.push_back(landed_state);
+        }
     });
 
     while (!telemetry->health_all_ok()) {
--- a/src/integration_tests/mission_takeoff_land.cpp
+++ b/src/integration_tests/mission_takeoff_land.cpp
@@ -76,12 +76,10 @@ void do_mission_with_takeoff_and_land(float mission_altitude_m)
         Telemetry::LandedState::OnGround};
 
     std::vector<Telemetry::LandedState> landed_states;
-    int index = 0;
-    telemetry->subscribe_landed_state(
-        [&index, &landed_states](Telemetry::LandedState landed_state) {
-            LogInfo() << landed_state;
-            landed_states.push_back(landed_state);
-        });
+    telemetry->subscribe_landed_state([&landed_states](Telemetry::LandedState landed_state) {
+        LogInfo() << landed_state;
+        landed_states.push_back(landed_state);
+    });
 
     while (!telemetry->health_all_ok()) {
         LogInfo() << "Waiting for system to be ready";
--- a/src/integration_tests/mission_takeoff_land.cpp
+++ b/src/integration_tests/mission_takeoff_land.cpp
@@ -65,7 +65,26 @@ void do_mission_with_takeoff_and_land(float mission_altitude_m)
         pc.check_current_alitude(position.relative_altitude_m);
     });
 
-    while (!telemetry->health_all_ok()) {
+    Telemetry::LandedState landed_states_template[] = {
+        Telemetry::LandedState::OnGround,
+        Telemetry::LandedState::TakingOff,
+        Telemetry::LandedState::InAir,
+        Telemetry::LandedState::Landing,
+        Telemetry::LandedState::TakingOff,
+        Telemetry::LandedState::InAir,
+        Telemetry::LandedState::Landing,
+        Telemetry::LandedState::OnGroud};
+
+    Telemetry::LandedState landed_states[8];
+    int index = 0;
+    telemetry->subscribe_landed_state([](Telemetry::LandedState landed_state) {
+        LogInfo() << landed_state;
+        landed_states[index] = landed_state;
+        index++
+    })
+
+        while (!telemetry->health_all_ok())
+    {
         LogInfo() << "Waiting for system to be ready";
         LogDebug() << "Health: " << telemetry->health();
         std::this_thread::sleep_for(std::chrono::seconds(1));
@@ -75,15 +94,23 @@ void do_mission_with_takeoff_and_land(float mission_altitude_m)
     LogInfo() << "Creating and uploading mission";
 
     Mission::MissionItem new_item{};
+    Mission::MissionPlan mission_plan{};
 
     new_item.latitude_deg = 47.398170327054473;
     new_item.longitude_deg = 8.5456490218639658;
     new_item.relative_altitude_m = mission_altitude_m;
     new_item.vehicle_action = Mission::MissionItem::VehicleAction::Takeoff;
-
-    Mission::MissionPlan mission_plan{};
     mission_plan.mission_items.push_back(new_item);
 
+    new_item.latitude_deg = 47.398370327054473;
+    new_item.vehicle_action = Mission::MissionItem::VehicleAction::Land;
+    mission_plan.mission_items.push_back(new_item);
+
+    new_item.latitude_deg = 47.398570327054473;
+    new_item.vehicle_action = Mission::MissionItem::VehicleAction::Takeoff;
+    mission_plan.mission_items.push_back(new_item);
+
+    new_item.latitude_deg = 47.398770327054473;
     new_item.vehicle_action = Mission::MissionItem::VehicleAction::Land;
     mission_plan.mission_items.push_back(new_item);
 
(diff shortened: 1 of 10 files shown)


src/plugins/mission/mission_impl.cpp
```diff
@@ -463,6 +464,60 @@ MissionImpl::convert_to_int_items(const std::vector<MissionItem>& mission_items)
             int_items.push_back(next_item);
         }
 
+        if (item.vehicle_action != VehicleAction::None) {
+            // There is a camera action that we need to send.
+
```
JonasVautherin, line 469:
```suggestion
```

JonasVautherin (changes requested):
Except for a wrong comment, it looks good to me :blush:. Did you test it? Does it work? :rocket:

JonasVautherin:
Seeing the errors in CI (e.g. in `mission.h`), I come to wonder: did you auto-generate the code from the proto file, or did you write it manually? `mission.h` should be auto-generated, you should not have to write anything there yourself, right?

julianoes:
Thank @rligocki. Now it would be nice if there was a new integration test using it.

rligocki:
Yep I wrote it manually. Didn't know that it is possible to generate templates based on MAVSDK-Proto. I will have a look on generation of on C++ code based on Proto.

julianoes:
@rligocki there is a comment telling you that the file is auto-generated.
And please have a look at the docs: https://mavsdk.mavlink.io/main/en/cpp/contributing/plugins.html#generate-h-and-cpp-files

rligocki:
Now it looks like it might be ready for testing. What is best way to do it?

JonasVautherin:
> What is best way to do it?

You can write a small script that uses it, or directly go for an integration test (like one of the `mission_*.cpp` [here](https://github.com/mavlink/MAVSDK/tree/main/src/integration_tests)). Maybe `mission_action.cpp`?

Also the style is wrong, please run `./tools/fix_style.sh .`, and the "proto" CI test is failing because you did not update the proto submodule in this PR. Just make it point to your MAVSDK-Proto PR commit and it should be fine :+1:.

rligocki:
It looks like, there is some change in MAVSDK-Proto in core.proto file. Connection state is missing value 1. In my MAVSDK-Proto PR there are only additional code with VehicleAction. MAVSDK PR there is one removed line with that value.

JonasVautherin:
> Connection state is missing value 1.

Wasn't this resolved in https://github.com/mavlink/MAVSDK/pull/1420? Is your branch up to date?

Also you forgot to fix the style: `./tools/fix_style.sh .`

rligocki:
> > Connection state is missing value 1.
> 
> Wasn't this resolved in #1420? Is your branch up to date?
> 

https://github.com/mavlink/MAVSDK-Proto/blob/main/protos/core/core.proto - It looks like it is in main, so I was up to date.

> Also you forgot to fix the style: `./tools/fix_style.sh .`
> 

On my mac I was not able to compile these tools. So I need to do fix_style on different computer at home.

JonasVautherin:
> It looks like it is in main, so I was up to date.

There two of them: the proto submodules, and then the MAVSDK branch itself. Both need to be up to date :+1:. This said, the CI tests pass... where do you see the issue with the connection state? When building the integration tests? If yes, just update your branch to `mavsdk:main` :+1:

rligocki:
Still don't understand that problem with MAVSDK-Proto. On main branch, there is removed one connection state. Fixed problem with style. Should I create PR for connection state error?

JonasVautherin:
> On main branch, there is removed one connection state. Fixed problem with style. Should I create PR for connection state error?

You have a fork of `mavlink/MAVSDK`, say `~/mavsdk`, right? Inside of it, there is the proto submodule, `~/mavsdk/proto`. Those are two different git repositories, but they are related.

You need `~/mavsdk/proto` to point to the HEAD of `mavlink/MAVSDK-Proto`, and `~/mavsdk` to point to the HEAD of `mavlink/MAVSDK`. The connection changes are both in the proto submodule (where the interface removes it) and in the mavsdk repo (where it is removed in the implementation)

Does that make sense?

rligocki:
> You have a fork of `mavlink/MAVSDK`, say `~/mavsdk`, right? Inside of it, there is the proto submodule, `~/mavsdk/proto`. Those are two different git repositories, but they are related.
> 
This is what I understand. I see no problem there.

> You need `~/mavsdk/proto` to point to the HEAD of `mavlink/MAVSDK-Proto`, and `~/mavsdk` to point to the HEAD of `mavlink/MAVSDK`. The connection changes are both in the proto submodule (where the interface removes it) and in the mavsdk repo (where it is removed in the implementation)
But said that `~/mavsdk/proto` should point to MAVSDK-Proto PR. So you suggest to change proto git to point at mavlink/MAVSDK-Proto main branch?

rligocki:
I am really lost. I need help at this situation. Cannot find where exactly I made mistake and how to solve it. Tried to find solution, but unfortunately it is first time I contribute to public repository so I have no experience at this case.

JonasVautherin:
No worries, it can be tricky in the beginning :wink:. 

Should be fine now, both your branches were not up-to-date with upstream. Read about syncing with upstream e.g. [here](https://ardalis.com/syncing-a-fork-of-a-github-repository-with-upstream/) (it has to be done on both your mavsdk and mavsdk-proto forks until your proto is merged upstream).

JonasVautherin:
The docs CI check is failing, I commented on the proto PR accordingly: https://github.com/mavlink/MAVSDK-Proto/pull/226

Other than that, it would be nice if you could add an integration test. Note that the integration tests run on the iris SITL multicopter, so ideally use actions that work for a multicopter (i.e. takeoff and land), and maybe write a test that checks that the iris behaves correctly when you use VTOL actions (I don't know what PX4 is expected to do, but a test for it would be nice).

I won't be asking you to change the integration tests such that they can run a VTOL, but that would be a nice addition later :blush:.

rligocki:
Thanks a lot for your help. I though, that you would give me some advice, but this was faster way. 

Where are those integration tests located? I can check them and then create group of new tests for this purpose.

julianoes:
Does this answer it? https://mavsdk.mavlink.io/main/en/cpp/guide/test.html

JonasVautherin (commented):
Does the integration test prove that the takeoff action works? My feeling is that it would also takeoff without the takeoff action :thinking:.

Could we make it such that it lands and takes off in the middle of a mission? That I am pretty sure is not yet working :laughing:

src/integration_tests/mission_takeoff_land.cpp
```diff
@@ -73,18 +73,18 @@ void do_mission_with_takeoff_and_land(float mission_altitude_m)
...
 
     Telemetry::LandedState landed_states[8];
     int index = 0;
-    telemetry->subscribe_landed_state([](Telemetry::LandedState landed_state) {
-        LogInfo() << landed_state;
-        landed_states[index] = landed_state;
-        index++
-    })
+    telemetry->subscribe_landed_state(
+        [&index, &landed_states](Telemetry::LandedState landed_state) {
```
rligocki, line 81: I am worried about &index. As I understand C++, &index is address of int value. Is it problem, then to do index++?

JonasVautherin: It's correct :+1:

JonasVautherin: Note that instead of using arrays, you could have used vectors. But I believe this works.

src/integration_tests/mission_transition.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+    new_item.latitude_deg = 47.398170327054473;
+    new_item.longitude_deg = 8.5456490218639658;
+    new_item.relative_altitude_m = mission_altitude_m;
+    new_item.vehicle_action = Mission::MissionItem::VehicleAction::Takeoff;
+
+    Mission::MissionPlan mission_plan{};
+    mission_plan.mission_items.push_back(new_item);
+
+    new_item.vehicle_action = Mission::MissionItem::VehicleAction::TransitionToFw;
```
JonasVautherin, line 87: What's the expected behavior when flying a multicopter? I guess it is ignored, isn't it?

rligocki: I guess so. I was thinking, that if landing and takeoff would work, this action will provide that information.

src/integration_tests/mission_takeoff_land.cpp
rligocki: Modified PR to vectors. Thanks for your advices ;)

src/integration_tests/mission_takeoff_land.cpp
```diff
@@ -0,0 +1,171 @@
...
+        std::this_thread::sleep_for(std::chrono::seconds(1));
+    }
+
+    while (telemetry->armed()) {
+        // Wait until we're done.
+        std::this_thread::sleep_for(std::chrono::seconds(1));
+    }
+    for (int i = 0; i < 8; i++) {
+        ASSERT_EQ(landed_states_template, landed_states);
+    }
```
JonasVautherin, line 168: Either you want to do `landed_states_template.at(i)` and `landed_states.at(i)`, or remove the for loop (not sure if that works?), but something is wrong here :see_no_evil:

JonasVautherin (changes requested):

src/integration_tests/mission_takeoff_land.cpp
rligocki: I just wanted to make sure, that those two vector are same. So I did check 8 times. 🙈🙈 My mistake. already fixed.

JonasVautherin (approved):

JonasVautherin (changes requested):
LGTM, but will need to be merged after the acceptance_radius PR (because it requires an update of the proto submodule)

JonasVautherin:
@rligocki: seems like the integration tests still fail. Would you mind having a look?

rligocki:
I will try to look at it during today or tomorrow.

JonasVautherin:
@rligocki: I merge the proto PR and update the branch here (don't forget to pull the changes). Once the integration tests are fixed, that can go in :+1:

JonasVautherin:
@rligocki are you going to finish this or is it on hold for now? Asking because if it is on hold, I will just revert MAVSDK-Proto so that it does not block other features in the meantime :+1:

rligocki:
Now I was able to found some time for this problem. I will fix those tests so that they will pass.

julianoes:
@rligocki what's the status here? Would be nice to get this in soon.

rligocki:
Cannot upload those vehicle mission item on PX4. There is an error.

rligocki:
Currently there should be no error during runtime. One last thing, that needs to be done is transition mission validation. Now there is no measurement if transition happened. Also I cannot find way to run transition test with gazebo_standard_vtol plane.

PR #1426, opened 2021-05-03
core: use vector for systems

### LINKED ISSUE #1425: One drone arming and flying does not allow the other to arm.
Currently I am creating a code that uses multiple classes. In the zip file I shared I use two classes one for mavsdk and connections. and other for addressing individual drones.

the drones are drone0 and drone1
image

as You can see when trying to arm drone 1 it says it is already arm, but it is not drone 0 is armed. 

My setup is right now running two simulations without changing the location:

` ./Tools/sitl_multiple_run.sh 2
HEADLESS=1  ./Tools/jmavsim_run.sh -l
`
and the second one in a new terminal 
```
HEADLESS=1  ./Tools/jmavsim_run.sh -p 4561 -l
```

[server_sample.zip](https://github.com/mavlink/MAVSDK/files/6410384/server_sample.zip)

julianoes:
When a new system is added we expect it to be added to the end of the `systems()` vector. This can easily be accomplished by using a std::vector for the systems instead of a std::unordered_map.

For a small number of systems using a vector for that sort of lookup should be faster anyway than a map, according to common advice.

Fixes #1425.

### DIFF
--- a/src/core/mavsdk_impl.cpp
+++ b/src/core/mavsdk_impl.cpp
@@ -104,7 +104,7 @@ std::vector<std::shared_ptr<System>> MavsdkImpl::systems() const
         if (system.first == 0) {
             continue;
         }
-        systems_result.push_back(std::shared_ptr<System>(system.second));
+        systems_result.push_back(system.second);
     }
 
     return systems_result;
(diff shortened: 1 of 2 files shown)


JonasVautherin (approved):

AldoCF96:
Is the issue fixed?

julianoes:
Yes, it should be fixed if you build yourself from the main branch. Otherwise you have to wait a few days until we do the next release.

PR #1165, opened 2020-08-12
Add set_actuator

coderkalyan:
Adds support for the MAV_CMD_DO_SET_ACTUATOR mavlink command.

Related proto: https://github.com/mavlink/MAVSDK-Proto/pull/172

PR #1427, opened 2021-05-05
info: add legacy uid (uid1)

julianoes:
This is for backwards compatibility.

Depends on https://github.com/mavlink/MAVSDK-Proto/pull/227.

Contains #1165.


JonasVautherin (approved):
Thanks a lot @julianoes! :rocket:

PR #1430, opened 2021-05-10
In mission, pitch and yaw should be degrees between -180 and 180

JonasVautherin:
According to the specs for [GIMBAL_MANAGER_PITCHYAW](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW), pitch and yaw are degrees between -180 and 180.


JonasVautherin:
Still, when I upload a mission and download it in QGC, the "Waypoint" items don't have the gimbal part set at all. Feels like PX4 does not handle that command somehow?

As an integration test, I was planning to use the picture angle coming from the `capture_info` messages, if that makes sense :thinking:.

PR #1431, opened 2021-05-11
mavsdk_server: lazily initialize plugins

julianoes:
This changes mavsdk_server so that we don't initialize all the plugins in the very beginning but instead when a first request over gRPC is made. This is done by giving the services a reference to the mavsdk instance, so that they can get the first system when asked for something.
    
This has the benefits that:
 - We no longer need to use the deprecated null reference.
 - We can avoid some of the init storm where plugins get parameters, etc. (less parameters requested, less log noise).
 - We need less RAM and CPU by not starting plugins we don't use.
    
However, it might have the drawback that:
 - Some plugins will respond a bit slower the first time something is
      requested because they first need to get a param, etc.
    
Note this commit currently disables most of the mavsdk_server unit tests because I did not know how to pass in a mock mavsdk object instead of a mock plugin.

Depends on https://github.com/mavlink/MAVSDK-Proto/pull/228.


JonasVautherin (approved):

JonasVautherin:
Thanks!

PR #1432, opened 2021-05-11
mission: add gimbal v2 control/release

julianoes:
When using gimbal v2 commands in a mission we need to acquire control before setting the gimbal the first time, and release it again when the mission is done.


JonasVautherin (approved):
Thanks a lot @julianoes!

JonasVautherin:
Working!

PR #1436, opened 2021-05-17
Update tinyxml2 to 8.1.0

JonasVautherin:


JonasVautherin:
Whoops... some errors when building for Android, and... this is for Ubuntu 18.04:

```
CMake 3.15 or higher is required.  You are running version 3.10.2
```

julianoes (approved):

julianoes:
I'm confused that we don't have `fseek` on some platforms.

JonasVautherin:
> I'm confused that we don't have fseek on some platforms.

I think it's more that on `android-x86_64`, it goes into the `defined(__x86_64__)` condition and tries to use `fseeko64`, which apparently does not exist.

JonasVautherin:
I tried with a fork of tinyxml2 that sets cmake 3.10.2 (Ubuntu 18.04) and uses [this PR](https://github.com/leethomason/tinyxml2/pull/865) for Android.

I already see that tinyxml2 seems to require a version older than 3.10.2, let's see what Android says about my other patch.

julianoes (approved):

PR #1437, opened 2021-05-18
Fix legacy_uid 

JonasVautherin:
Looking at 31523ce4d04759a12b54dd1ff9f781e63a4b6888, we used to do it differently, and it turns out that the new way seems to return a "00000000000000" value with the H520

I'll try to get this tested on an H520 before merging it :smile:


JonasVautherin:
Oh, actually `uid` is already a `uint64_t` in the [AUTOPILOT_VERSION](https://mavlink.io/en/messages/common.html#AUTOPILOT_VERSION) message :thinking:.

PR #1439, opened 2021-05-20
offering the ability to set a component description 

frederictaillandier:
and adding new Companion Computer ids for configuration

the component description should allow to define in a more explicit way the component than the component id

accorfing @TSC21  recommendations

### DIFF
--- a/src/core/mavsdk.h
+++ b/src/core/mavsdk.h
@@ -1,13 +1,14 @@
 #pragma once
 
-#include <string>
-#include <memory>
-#include <vector>
 #include <functional>
+#include <memory>
+#include <optional>
+#include <string>
+#include <vector>
 
+#include "connection_result.h"
 #include "deprecated.h"
 #include "system.h"
-#include "connection_result.h"
 
 namespace mavsdk {
 
@@ -185,9 +186,12 @@ public:
         enum class UsageType {
             Autopilot, /**< @brief SDK is used as an autopilot. */
             GroundStation, /**< @brief SDK is used as a ground station. */
-            CompanionComputer, /**< @brief SDK is used as a companion computer on board the MAV. */
-            Custom /**< @brief the SDK is used in a custom configuration, no automatic ID will be
+            CompanionComputer, /**< @brief SDK is used as a companion computer element 1 on board the MAV. */
+            Custom, /**< @brief the SDK is used in a custom configuration, no automatic ID will be
                       provided */
+            CompanionComputer2, /**< @brief SDK is used as a companion computer element 2 on board the MAV. */
+            CompanionComputer3, /**< @brief SDK is used as a companion computer element 3 on board the MAV. */
+            CompanionComputer4 /**< @brief SDK is used as a companion computer element 4 on board the MAV. */
         };
 
         /**
@@ -250,11 +254,24 @@ public:
          */
         void set_usage_type(UsageType usage_type);
 
+        /**
+         * @brief Retrieve the component's description of this configuration if it has one
+         * UUseful to set a more custom name than the eventually too generic component_id
+         */
+        std::optional<std::string const> get_component_description() const;
+
+        /**
+         * @brief Set the component's description of this configuration
+         * Useful to set a more custom name than the eventually too generic component_id
+         */
+        void set_component_description(std::string const & component_description);
+
     private:
         uint8_t _system_id;
         uint8_t _component_id;
         bool _always_send_heartbeats;
         UsageType _usage_type;
+        std::optional<std::string> _component_description;
     };
 
     /**
--- a/src/core/mavsdk.cpp
+++ b/src/core/mavsdk.cpp
@@ -136,6 +136,21 @@ Mavsdk::Configuration::Configuration(UsageType usage_type) :
             _component_id = MavsdkImpl::DEFAULT_COMPONENT_ID_CC;
             _always_send_heartbeats = true;
             break;
+        case Mavsdk::Configuration::UsageType::CompanionComputer2:
+            _system_id = MavsdkImpl::DEFAULT_SYSTEM_ID_CC;
+            _component_id = MAV_COMPONENT::MAV_COMP_ID_ONBOARD_COMPUTER2;
+            _always_send_heartbeats = true;
+            break;
+        case Mavsdk::Configuration::UsageType::CompanionComputer3:
+            _system_id = MavsdkImpl::DEFAULT_SYSTEM_ID_CC;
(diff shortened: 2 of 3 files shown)


src/core/mavsdk.h
```diff
@@ -250,11 +254,24 @@ class Mavsdk {
...
 
+        /**
+         * @brief Retrieve the component's description of this configuration if it has one
+         * UUseful to set a more custom name than the eventually too generic component_id
+         */
+        std::optional<std::string const> get_component_description() const;
+
+        /**
+         * @brief Set the component's description of this configuration
+         * Useful to set a more custom name than the eventually too generic component_id
```
TSC21, line 265:
```suggestion
         * Useful to set the description of the a component that share the same type as other components
```

TSC21 (changes requested):
Some minor comments. Other than that, I think this approach is correct. Waiting for @julianoes and @JonasVautherin suggestions.

JonasVautherin:
My opinion is that the configuration should not be used to name components. Instead I would use the [manual interface](https://github.com/mavlink/MAVSDK/blob/main/src/core/mavsdk.h#L193-L201) and set the sysid/compid individually for each component.

I see the problem as this:

MAVLink sysids and compids are similar to setting up a network with static IPs. The sysid defines a subnet (at least that's how mavlink-router does the routing), so typically all the components running on a drone should have the same sysid (which is "1" by default).

This leaves us with the compids. 1 is usually the FMU, so we have the remaining ones for the components running onboard the drone. The drone maintainers should keep a list of such components and their assigned id ("component X is 1:42, component Y is 1:43, component Z is 1:44"). Because some drone has a RadioactivitySensor should not mean that the MAVLink standard keeps a hardcoded MAV_RADIOACTIVITY_SENSOR_COMPID. We should just say "the radioactivity sensor has sysid:compid 1:55 in our system".

More than that, other components should not actually hardcode 1:55 to talk to the sensor, they should detect it. Just like cameras. The reserved compids for cameras are legacy, I believe, they can be detected differently.

TSC21:
@JonasVautherin I would just argue how much abstraction we would want to add on this. For me as an high-level user, I would prefer not to care about defining component and sys ids at all and just use pre-configured configs. The description would just help me identify the component when the MAV_TYPEs are the same, or rather, when the component_ids have the same suffix. This is just to fix the problem where when we name a component ID, we are basically stating clear that that component is that thing, which when having something like an `ONBOARD_COMPUTER` id and then you actually need more than one id to identify services running on that system, you basically are obliged to add more comp ids with idx. That's why I believe that we are missing a better conjugation of MAV_TYPE with MAV_COMPONENT or even the existence of component instances (a camera is a component, multiple cameras can use the same component, but could be indexed with a component instance).

JonasVautherin:
Not sure I completely get it. So you are saying that it's better for a radioactivity sensor to be hardcoded as `COMPANION_COMPUTER3` than to be hardcoded as `1:43`?

My point is that sysid/compid are a network-level concept. Having a sysid:compid means that you can talk MAVLink, and it should not mean that you are a camera, a printer, or a connected lightbulb.

Especially since in that case, we are not even saying "all drones always have two connected lightbulbs, so we will reserved two compids for them and call them LIGHTBULB_1 and LIGHTBULB_2". Here we are saying "I have a lightbulb that needs a compid, and I find it more convenient to call it `COMPANION_COMPUTER_2` than giving it an actual id" :grin:

julianoes:
Ok, so I guess this PR triggered the bigger of discussion of how MAVLink sysid/compids should work vs. MAV_TYPE.
(In fact I had this discussion just recently with @hamishwillee when we talk about adding the COMPANION_COMPUTER_X compids.)

So the MAVLink spec makes this fairly clear: https://mavlink.io/en/services/heartbeat.html#component-identity

> Historically the component id was also used to determine the component type. New code must not make any assumption about the type from the id used (type is determined from HEARTBEAT.type).

And:

> MAVLink recommends that by default components use a type-appropriate component id from MAV_COMPONENT, and provide an interface to change the component id if needed. For example, a camera component might use any of the MAV_COMP_ID_CAMERAn ids, and should not use MAV_COMP_ID_GPS2.

So while it is good to have a component ID that "matches", it is not a strict requirement as the MAV_TYPE is what is ultimately deciding.

The reason is that it not possible to allocate 4-6 of every component type within the 255 available, so we have to make due with MAV_TYPE and potentially more arbitrary component MAV_COMPID_USERx. Because of course, someone is going to come and say they need 10 companion IDs, or 20 cameras, you name it.

---

> @JonasVautherin 
> Because some drone has a RadioactivitySensor should not mean that the MAVLink standard keeps a hardcoded MAV_RADIOACTIVITY_SENSOR_COMPID. We should just say "the radioactivity sensor has sysid:compid 1:55 in our system".

Exactly, and now for this case you would probably use a MAV_TYPE_SENSOR (which is missing, we should add it!).

---

> @TSC21 
> I would just argue how much abstraction we would want to add on this. For me as an high-level user, I would prefer not to care about defining component and sys ids at all and just use pre-configured configs.

Right, so how do we make this easy to deal with. There are two challenges I see:

1. MAV_TYPE/compid mapping: A component which does not just broadcast needs to know who else is there and what to send to whom. Therefore, each component needs to keep a map of all discovered components and their MAV_TYPE. That's not trivial but it's very doable, in my opinion. And note that some sort of mapping is already required as soon as you move from one camera to two cameras because at that point you need know how many components of type camera that you have. It could be CAMERA0 and CAMERA5, so you already can't infer CAMERA0 and CAMERA1.

2. Setting up compids: To me this is like local networking. In the same way as you set up IPs, an integrator needs to set up the compids and make sure they don't collide within a system. Now we could try to simplify this by having a central broker that assigns compids after an initial discovery (like DHCP) but we currently don't have that, and MAVLink does not rely on a broker. Alternatively, we could use auto-negotiation where a new participant to the network listens to all components and then tries to pick one that is still free. 

And so this brings us to this pull request. What IPs do we need to make this easiest for an integrator? Is it this PR which allows to set COMPANION_COMPUTERx or could we instead add the negotiation to prevent conflicts, and not have the user deal with this? The goal of MAVSDK was always to make MAVLink easier, and by exposing all of MAVLink, just through a different API we don't actually help with that.
I'm not saying we need to implement auto-negotiation now but I wonder if we should add multiple companions until then, or if this is a seldom enough problem that we can workaround by setting the compid manually without having to make the API more complex.

@frederictaillandier I feel bad for the amount of discussion your well intended PR caused, sorry :smile:.

frederictaillandier:
@julianoes  Haha don't worry, anyway, as  we discussed in private with @JonasVautherin , I plan to close this PR and get it properly tested in another fork before submitting it again. But please continue the conversation as I need to know what I should do.

PR #1442, opened 2021-05-24
Ability to customize logging output with a user-defined callback

levevg:
Problem:
- uncontrollable logging to stdout may be undesirable
- some applications may want to store logs in some kind of permanent storage for later investigations

Proposed solution:
- give users an ability to define custom callback which is called for every log message. The callback is able to discard default logging to stdout, or keep it and just add some another output

### DIFF
--- a/src/core/log_callback.h
+++ b/src/core/log_callback.h
@@ -1,0 +1,24 @@
+#pragma once
+
+#include <string>
+#include <functional>
+
+namespace mavsdk::log {
+
+// defined numeric values can be useful for filtering message by severity
+typedef enum {
+  Debug = 0,
+  Info = 1,
+  Warn = 2,
+  Err = 3
+} Level;
+
+/** @brief User-defined callback for logging. Returning true from this callback
+ * prevents default mavsdk`s logging to stdout. Returning false keeps it.
+ */
+typedef std::function<bool(Level level, const std::string &message,
+                           const std::string &file, int line)> Callback;
+
+extern Callback callback;
+
+}
--- a/src/core/CMakeLists.txt
+++ b/src/core/CMakeLists.txt
@@ -98,6 +98,7 @@ install(FILES
     deprecated.h
     system.h
     mavsdk.h
+    log_callback.h
     plugin_base.h
     geometry.h
     DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/mavsdk"
--- a/src/core/log.cpp
+++ b/src/core/log.cpp
@@ -20,6 +20,10 @@
 
 namespace mavsdk {
 
+namespace log {
+  Callback callback = nullptr;
+}
+
 void set_color(Color color)
 {
 #if defined(WINDOWS)
--- a/src/core/log.h
+++ b/src/core/log.h
@@ -2,6 +2,7 @@
 
 #include <sstream>
 #include "global_include.h"
+#include "log_callback.h"
 
 #if defined(ANDROID)
 #include <android/log.h>
@@ -37,18 +38,24 @@ public:
 
     virtual ~LogDetailed()
     {
+        if (log::callback &&
+            log::callback(_log_level, _s.str(), _caller_filename, _caller_filenumber))
+        {
+          return;
+        }
+
 #if ANDROID
         switch (_log_level) {
-            case LogLevel::Debug:
+            case log::Level::Debug:
                 __android_log_print(ANDROID_LOG_DEBUG, "Mavsdk", "%s", _s.str().c_str());
                 break;
-            case LogLevel::Info:
+            case log::Level::Info:
                 __android_log_print(ANDROID_LOG_INFO, "Mavsdk", "%s", _s.str().c_str());
                 break;
-            case LogLevel::Warn:
+            case log::Level::Warn:
                 __android_log_print(ANDROID_LOG_WARN, "Mavsdk", "%s", _s.str().c_str());
                 break;
-            case LogLevel::Err:
+            case log::Level::Err:
                 __android_log_print(ANDROID_LOG_ERROR, "Mavsdk", "%s", _s.str().c_str());
                 break;
         }
@@ -58,16 +65,16 @@ public:
 #else
 
         switch (_log_level) {
-            case LogLevel::Debug:
+            case log::Level::Debug:
                 set_color(Color::Green);
                 break;
-            case LogLevel::Info:
+            case log::Level::Info:
                 set_color(Color::Blue);
                 break;
-            case LogLevel::Warn:
+            case log::Level::Warn:
                 set_color(Color::Yellow);
                 break;
-            case LogLevel::Err:
+            case log::Level::Err:
                 set_color(Color::Red);
                 break;
         }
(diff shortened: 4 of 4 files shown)


src/core/log_callback.h
```diff
@@ -0,0 +1,24 @@
+#pragma once
+
+#include <string>
+#include <functional>
+
+namespace mavsdk::log {
+
+// defined numeric values can be useful for filtering message by severity
+typedef enum {
```
JonasVautherin, line 9: Since we are changing this, what about an `enum class`? Feels more modern :blush:

JonasVautherin (commented):
Sounds like a good idea!

src/core/log_callback.h
```diff
@@ -0,0 +1,24 @@
...
+  Err = 3
+} Level;
+
+/** @brief User-defined callback for logging. Returning true from this callback
+ * prevents default mavsdk`s logging to stdout. Returning false keeps it.
+ */
+typedef std::function<bool(Level level, const std::string &message,
+                           const std::string &file, int line)> Callback;
+
+extern Callback callback;
```
julianoes, line 22: Instead of `extern` I would prefer a mechanism with a clear API in `mavsdk.h`, e.g. `Mavsdk::subscribe_log(Callback)`

src/core/log_callback.h
```diff
@@ -0,0 +1,24 @@
...
+  Debug = 0,
+  Info = 1,
+  Warn = 2,
+  Err = 3
+} Level;
+
+/** @brief User-defined callback for logging. Returning true from this callback
+ * prevents default mavsdk`s logging to stdout. Returning false keeps it.
+ */
+typedef std::function<bool(Level level, const std::string &message,
```
julianoes, line 19: I'd prefer to use the more modern:

```
using Callback = std::function<bool(Level level, const std::string& message, const std::string& file, int line)>;`
```

julianoes (changes requested):
Very cool, I always wanted to do that but never took the time to do so. Thanks a lot for the contribution!

While we're at it. Should we also enable an option to log to a logfile instead of stdout?

src/core/log_callback.h
levevg: My bad, I didn't know enum class can have underlying type. Definitely a good idea, especially since this enum becomes public. I'll change it

src/core/log_callback.h
levevg: Given the capitalisation of Mavsdk in you message, I assume you mean static method? Because now logging isn't bound to Mavsdk instances, and probably some logging happens in places where no Mavsdk instances exist.
If my assumption is correct, then it's probably a question of style/convention - should I place callback setting into Mavsdk class as static variable/method, or place it into some namespace. I decided that logging belongs to the whole framework and not to the Mavsdk class or any other class, so placed it into separate namespace mavsdk::log. But I'm new here and I'll be glad to hear from you how it should be done correctly.

Regarding extern var vs. subscribe_log -  does subscribe imply that several callbacks are possible at once? Or subscribe_log should just overwrite old callback, being essentially a setter for single callback variable?

src/core/log_callback.h
levevg: Thanks, will fix. Also apparently some styling checks failed. I'll wait a bit for discussion here to settle and then fix everything at once

src/core/log_callback.h
levevg: Honestly, I'd better left implementation of saving messages to a logfile to the users. I believe In many cases it will suffice to just redirect stdout to a file in run time. When I proposed this change I had in mind slightly different case - when the rest of project already has some logging system and user wants to integrate MAVSDK logs into it. E.g. in the project I'm working on we use spdlog framework, and it has several sinks at once - SQLite db, stdout, WebSocket broadcast.

julianoes: > and probably some logging happens in places where no Mavsdk instances exist.

Interesting point. Are you sure? I can't think of any. Generally `Mavsdk` needs to be instantiated for the whole time.

> Or subscribe_log should just overwrite old callback, being essentially a setter for single callback variable?

Yes, basically a setter. That's how they are all implemented, for now, which is not optimal.

One problem we have with the `subscribe_log` API is that MAVSDK prints its version right on loading, so we would have to add the log callback to the constructor. Would that work?

> Honestly, I'd better left implementation of saving messages to a logfile to the users. I believe In many cases it will suffice to just redirect stdout to a file in run time.

That's fine, at least for now that's really good anyway.

>

JonasVautherin: > Interesting point. Are you sure? I can't think of any. Generally Mavsdk needs to be instantiated for the whole time.

Actually I'm thinking that one day, a `System` may be a standalone thing that does not rely on `Mavsdk` anymore. That would be such a case, right?

levevg: > Interesting point. Are you sure? I can't think of any. Generally Mavsdk needs to be instantiated for the whole time.
For example, any static helper function. Or any lambda that doesn't need to capture Mavsdk otherwise. I'm not that familiar with the codebase to say how often are such cases. After few minutes of searching I found several e.g. LogError() is used in functions 
```
static rpc::follow_me::FollowMeResult::Result translateToRpcResult(const mavsdk::FollowMe::Result& result)
```
```
bool CurlWrapper::download_file_to_path
```
Also LogDebug is used in integration tests.

> One problem we have with the subscribe_log API is that MAVSDK prints its version right on loading, so we would have to add the log callback to the constructor. Would that work?
That's what I'm talking about - because logging logically belongs to the framework and to not the Mavsdk instances (which can exist in any quantity including zero), if we bind logging to Mavsdk instances we may run into any kind of weird complications.

Setter function for the callback is good idea anyway, for the same reasons we usually make setters for class members - e.g. one day we may want to do some additional work on callback change, or validate something. I'll add it

julianoes: Hm, you're right! I did think of the integration tests. An given what @JonasVautherin said your approach then makes sense.

julianoes (approved):
I think this looks good now, thank you!

julianoes:
The only problem now is the Windows build :thinking::

```
D:\a\MAVSDK\MAVSDK\src\core\log_callback.h(11,20): error C3431: 'Level': a scoped enumeration cannot be redeclared as an unscoped enumeration 
```

JonasVautherin (approved):
LGTM

julianoes:
I'll merge this but I'd very much appreciate docs on how to use it in https://github.com/mavlink/MAVSDK-docs/blob/main/en/cpp/guide/dev_logging.md.

levevg:
@julianoes 
> I'll merge this but I'd very much appreciate docs on how to use it in https://github.com/mavlink/MAVSDK-docs/blob/main/en/cpp/guide/dev_logging.md.

Is it a good place to document it? That page describes internal logging API for plugin developers, but this change is public API for users. Maybe create a separate page? Or maybe just rename "Plugin/Test Logging" to "Logging" so it's easier to find?

julianoes:
Oh, you're right @levevg. Renaming or adding one would both be good. I can't decide :smile:.

PR #1443, opened 2021-05-24
Add acceptance_radius back to MissionItem

bazfp:
Appears this added in a previous PR but disappeared during a refactor. It's a useful option to have as sometimes the defaults are causing some items to never be achieved in our SITL testing (We've now swapped to MissionRaw)

Happy to update the factory funcs + proto to include this, unless there's a reason it was removed :)

### DIFF
--- a/src/plugins/mission/mission_impl.cpp
+++ b/src/plugins/mission/mission_impl.cpp
@@ -547,6 +547,7 @@ std::pair<Mission::Result, Mission::MissionPlan> MissionImpl::convert_to_result_
                 new_mission_item.relative_altitude_m = int_item.z;
 
                 new_mission_item.is_fly_through = !(int_item.param1 > 0);
+                new_mission_item.acceptance_radius_m = int_item.param2;
 
                 have_set_position = true;
 
--- a/src/plugins/mission/mission_impl.cpp
+++ b/src/plugins/mission/mission_impl.cpp
@@ -266,18 +266,7 @@ float MissionImpl::hold_time(const MissionItem& item)
 
 float MissionImpl::acceptance_radius(const MissionItem& item)
 {
-    float acceptance_radius_m;
-    if (std::isfinite(item.acceptance_radius_m)) {
-        acceptance_radius_m = item.acceptance_radius_m;
-    } else if (item.is_fly_through) {
-        // _acceptance_radius_m is 0, determine the radius using fly_through
-        acceptance_radius_m = 3.0f;
-    } else {
-        // _acceptance_radius_m is 0, determine the radius using fly_through
-        acceptance_radius_m = 1.0f;
-    }
-
-    return acceptance_radius_m;
+    return item.acceptance_radius_m;
 }
 
 std::vector<MAVLinkMissionTransfer::ItemInt>
--- a/src/plugins/mission/mission_impl.cpp
+++ b/src/plugins/mission/mission_impl.cpp
@@ -266,7 +266,18 @@ float MissionImpl::hold_time(const MissionItem& item)
 
 float MissionImpl::acceptance_radius(const MissionItem& item)
 {
-    return item.acceptance_radius_m;
+    float acceptance_radius_m;
+    if (std::isfinite(item.acceptance_radius_m)) {
+        acceptance_radius_m = item.acceptance_radius_m;
+    } else if (item.is_fly_through) {
+        // _acceptance_radius_m is 0, determine the radius using fly_through
+        acceptance_radius_m = 3.0f;
+    } else {
+        // _acceptance_radius_m is 0, determine the radius using fly_through
+        acceptance_radius_m = 1.0f;
+    }
+
+    return acceptance_radius_m;
 }
 
 std::vector<MAVLinkMissionTransfer::ItemInt>
--- a/src/plugins/mission/mission_impl.cpp
+++ b/src/plugins/mission/mission_impl.cpp
@@ -269,8 +269,7 @@ float MissionImpl::acceptance_radius(const MissionItem& item)
     float acceptance_radius_m;
     if (std::isfinite(item.acceptance_radius_m)) {
         acceptance_radius_m = item.acceptance_radius_m;
-    }
-    else if (item.is_fly_through) {
+    } else if (item.is_fly_through) {
         // _acceptance_radius_m is 0, determine the radius using fly_through
         acceptance_radius_m = 3.0f;
     } else {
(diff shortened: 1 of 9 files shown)


JonasVautherin (changes requested):
I am fine with adding it, but you should do it by adding it to the [`mission.proto` definition](https://github.com/mavlink/MAVSDK-Proto/blob/main/protos/mission/mission.proto#L164) and generate the code.

E.g. `mission.h` is not a file you change manually, right?

julianoes (changes requested):
Oh, did this get lost? I wonder how? Thanks for the contribution @bazfp. And as @JonasVautherin wrote, please change it in the proto repo:
https://github.com/mavlink/MAVSDK-Proto/blob/main/protos/mission/mission.proto#L175

bazfp:
Thanks guys

https://github.com/mavlink/MAVSDK-Proto/pull/229

src/plugins/mission/mission_impl.cpp
```diff
@@ -267,7 +267,9 @@ float MissionImpl::hold_time(const MissionItem& item)
 float MissionImpl::acceptance_radius(const MissionItem& item)
 {
     float acceptance_radius_m;
-    if (item.is_fly_through) {
+    if (std::isfinite(item.acceptance_radius_m)) {
+        acceptance_radius_m = item.acceptance_radius_m;
+    } else if (item.is_fly_through) {
         // _acceptance_radius_m is 0, determine the radius using fly_through
```
JonasVautherin, line 273: Not related to this PR, but I don't get this comment. @julianoes would you have some insights about it?

JonasVautherin (approved):
LGTM, thanks a lot!

JonasVautherin:
@bazfp: some integration tests are still failing, actually. Please pull my latest changes, I updated your branch and the proto submodule :+1:

bazfp:
> @bazfp: some integration tests are still failing, actually. Please pull my latest changes, I updated your branch and the proto submodule 👍

Seems to be failing due to the `NAN` `acceptance_radius` logic. It sets the radius to 1 or 3 depending on flythrough if we don't explicitly set the `acceptance_radius`, which means the downloaded mission back doesn't match the one we sent if it was `NAN`.

I could try and work backwards and set the `acceptance_radius` to `NAN` if it's 1 or 3 depending on flythrough, but then you could break it if you set it explicitly to 1!

src/plugins/mission/mission_impl.cpp
julianoes: Me neither :thinking:

julianoes (approved):

JonasVautherin:
> It sets the radius to 1 or 3 depending on flythrough if we don't explicitly set the acceptance_radius, which means the downloaded mission back doesn't match the one we sent if it was NAN.

@julianoes are 1 and 3 some "standard" defaults? I guess MAVSDK does that because sending NAN to PX4 results in an invalid waypoints?

@bazfp would you mind trying to remove the logic that sets it to 1 or 3 and leave it to NAN instead?

I would be tempted to say that if there is a default, it should be set on the PX4 side, not in MAVSDK... :thinking:

bazfp:
I've removed the offending section that was setting the acceptance radius.

I could set the acceptance radius default to be 1.0m in the proto definition. It wouldn't account for flythrough waypoints anymore but that logic could go anywhere else.

JonasVautherin:
> I could set the acceptance radius default to be 1.0m in the proto definition.

I understand the idea, but I still believe that if the user does not set it, then it should be the responsibility of the autopilot to have a sane default (which may differ from a frame to the other)

julianoes:
For some reason `SitlTest.MissionChangeSpeed` is now broken and running indefinitely. It seems like that's actually related to this PR.

julianoes:
@bazfp looks like it's still timing out. I'm gonna have a look tomorrow to see what's going on.

julianoes:
Ok, I can reproduce it with PX4 v1.10. It looks like it's something that has been fixed on master. It never accepts the first mission item as "reached" although it is basically hovering there.

julianoes:
@bazfp I've added the defaults back in. It turns out they are required for PX4 v1.9, v1.10, v1.11 because PX4 gets stuck indefinitely with NAN. In PX4 master/v1.12 the acceptance radius is completely ignored which is wrong in my opinion and I'm going to fix.

JonasVautherin:
Oh, my bad then :sweat_smile:. Do you agree that PX4 should have a sane default if NAN is sent?

julianoes:
@JonasVautherin yes, I agree. It should but it will take some time until we can rely on that. Or we need to have a version check to check for it.

JonasVautherin:
No need for a version check, but it's good to see that it is fixed upstream. Will come some day :blush:

julianoes:
Aha, now CI fails because the check back for acceptance radius is not correct. Because we don't know if 1 or 3 meters are set on purpose or because of NAN :facepalm:. This is harder than expected.

bazfp:
> Could set the acceptance radius default to be 1.0m in the proto definition. It wouldn't account for flythrough waypoints anymore but that logic could go anywhere else.

One solution?

julianoes:
@JonasVautherin the suggestion by @bazfp would work, right?

JonasVautherin:
Yes, we could define to 1 in the proto, and then should we return an error code if the user sets it (explicitly) to NAN, since we know that it crashes on the PX4 side?

julianoes:
@JonasVautherin hm, I would accept NAN but change it to NAN unless we know we are connected to a PX4 new enough?

JonasVautherin:
But in that case, should it be a default in the proto definition, or should we just hardcode that behavior? Because it seems like anyway we will hardcode parts of it...

julianoes:
Hard-coding like we do right now is fine but the problem is that the test fails which checks if the downloaded items are equal to the ones uploaded.

JonasVautherin:
But the proto file won't fix that, right?

julianoes:
We would change the test not to use NAN, I suppose, but we could also do that already.

JonasVautherin:
I would vote for that: the default in the proto file is "NAN", because it means "let PX4 choose", and that's what we want. And for versions earlier than 1.12, we have a fix that says "if it is NAN, set to 1". And in the test we don't use NAN, I guess we don't have a choice. So:

* Keep the proto as it is
* In mission_impl.cpp, change NAN for 1 (bonus: check the PX4 version and do it only if < 1.12)
* In the tests, don't use NAN

How does that sound?

julianoes:
Yes, that's great. I'll make the change.

JonasVautherin (approved):
Great! Thanks :rocket:

PR #1453, opened 2021-06-02
Identify docs generator script as Python3

hamishwillee:
This allows the docs to be built on recent Ubuntu builds, which have Python3 rather than Python. However I am not sure it is "right" fix.


JonasVautherin (approved):
Fine with me! Thanks!

julianoes:
Oh yes oh yes, thanks @hamishwillee.

PR #1455, opened 2021-06-03
Filter out command acks not targeted to us

JonasVautherin:
Because the target sysid:compid is saved in the payload of a message, it has to be filtered out after the message has been decoded. In this case, receiving an ACK targeted to another component results in a warning ("Command ack not matching our current command"), which is confusing.

This PR transparently filters out such messages.


julianoes (approved):
Ok, makes sense.

levevg:
This apparently breaks communication with Ardupilot, which sends ACKs with sysid=0 compi=0. I believe 0 0 should be considered something like a broadcast and not filtered out.
Also, the condition seems to be wrong - difference in either sysid OR compi should be enough to conclude that ack is targeted for somebody else. Now both sysid and compi must differ simultaneously.
I'll commit a fix soon together with few other changes for better Ardupilot compatibility

PR #1456, opened 2021-06-03
Expose `set_timeout_s` to language bindings

### LINKED ISSUE #1441: MAVSDK timeout tuning
It seems to me that the timeout used in MAVSDK is a bit low for bad links (I'm thinking about a slow connection to a cloud simulator), resulting in a lot of timeouts (which makes it difficult to e.g. upload a reasonably-sized mission).

@julianoes: is there a global timeout that can be tuned, and what would it imply if we increased it? Like do we want to have an option to set the timeout (or choose the link speed), or would it be fine to just increase the timeout for everybody?

From my tests uploading a big mission to a cloud simulator, each waypoint took an average 0.4s to upload... I have heard of issues when uploading from a device emulator (iOS/Android), so I wonder if that maybe went above 0.5s, triggering the timeouts?

JonasVautherin:
Resolves #1441 

* [x] Depends on https://github.com/mavlink/MAVSDK-Proto/pull/230


reedev:
Hi,
I guess you merged this one (and closed the related issue) since it is supposed to work?
I tried it out and I get the same timeout. I must do something incorrectly since I went back to an earlier version of which I remembered that the mission example was working, and no luck with that version either.
I will open a new question how to build/work with MAVSDK in qtcreator.

PR #1461, opened 2021-06-07
Ignore ghost storage_information message

JonasVautherin:
Some MAVLink systems happen to send the STORAGE_INFORMATION message to indicate that the camera has a slot for a storage even if there is no way to know anything about that storage (e.g. whether or not there is an sdcard in the slot).

In this PR I consider that a total capacity of 0 means that this is such a message, and I don't expect MAVSDK users to leverage it, which is why it is ignored.

I have already seen two MAVSDK-based apps that were failing because of that message, I think it is time we ignored it.


julianoes (approved):
Noice.

PR #1462, opened 2021-06-07
Remove assumption on sysids per udp packet

### LINKED ISSUE #1454: Should we assume that a UDP packet contains messages from only one system?
I was just reading the code, and wondered about [this](https://github.com/mavlink/MAVSDK/blob/main/src/core/udp_connection.cpp#L234-L235):

```c++
            if (!saved_remote && sysid != 0) {
                saved_remote = true;
```

This is an optimization so that it does not "save" the remote for each message of the UDP packet, right?

I'm just thinking that this assumes that a UDP packet only ever contains messages from one sysid. Not sure if that's relevant or not, but I can imagine situations where this will give weird results :thinking:.

JonasVautherin:
Resolves #1454

### DIFF
--- a/src/core/udp_connection.cpp
+++ b/src/core/udp_connection.cpp
@@ -225,14 +225,11 @@ void UdpConnection::receive()
 
         _mavlink_receiver->set_new_datagram(buffer, static_cast<int>(recv_len));
 
-        bool saved_remote = false;
-
         // Parse all mavlink messages in one datagram. Once exhausted, we'll exit while.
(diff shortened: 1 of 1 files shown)


julianoes (approved):
Oh, now I understand it. So this happens if we coalesce multiple MAVLink messages from different remotes into one UDP packet?

JonasVautherin:
Yes! Not sure how often that happens (my opinion is that it should not), but nothing prevents it anywhere...

PR #1464, opened 2021-06-09
core: fix invalid sysid/compid on trimmed messages

julianoes:
It turns out that the fields target_system and target_component can be trimmed if they are zero. In that case we need to assume zero instead of e.g. reading into the CRC of the MAVLink message.

Found together with @JonasVautherin.


JonasVautherin (approved):
Nice!

PR #1468, opened 2021-06-18
telemetry: health improvements

julianoes:
This includes several health improvements:

- The level calibration health indicator was actually hard-coded to true, and it is not actually a requirement for health or being able to arm, so it probably makes sense to just get rid of it.
- Add flag to show if system can be armed.


JonasVautherin (approved):
Nice!

JonasVautherin (approved):

PR #1469, opened 2021-06-18
Add file upload callback registration for mavlink FTP server

MatejFranceskin:
This PR adds a method to register file uploaded to Mavlink FTP server callback.
When file is successfully uploaded the callback is called with the file path as parameter.

### DIFF
--- a/src/plugins/ftp/include/plugins/ftp/ftp.h
+++ b/src/plugins/ftp/include/plugins/ftp/ftp.h
@@ -140,6 +140,17 @@ public:
     void upload_async(std::string local_file_path, std::string remote_dir, UploadCallback callback);
 
     /**
+     * @brief Callback type for new file uploaded by remote.
+     */
+
+    using FileUploadedCallback = std::function<void(std::string)>;
+
+    /**
+     * @brief Register a callback to be triggered when remote uploads a file to FTP server
+     */
+    void register_file_uploaded_callback(FileUploadedCallback callback);
+
+    /**
      * @brief Callback type for list_directory_async.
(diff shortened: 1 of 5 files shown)


JonasVautherin:
Thanks Matej!

How is that different from `upload_async(std::string local_file_path, std::string remote_dir, UploadCallback callback);`? Doesn't the callback serve the same purpose?

julianoes:
@JonasVautherin this is for the server to be notified when something has been uploaded, so it's different, I think.

src/plugins/ftp/include/plugins/ftp/ftp.h
```diff
@@ -139,6 +139,17 @@ class Ftp : public PluginBase {
...
+     * @brief Callback type for new file uploaded by remote.
+     */
+
+    using FileUploadedCallback = std::function<void(std::string)>;
+
+    /**
+     * @brief Register a callback to be triggered when remote uploads a file to FTP server
+     */
+    void register_file_uploaded_callback(FileUploadedCallback callback);
+
```
julianoes, line 152: This file is auto-generated from https://github.com/mavlink/MAVSDK-Proto/blob/main/protos/ftp/ftp.proto, you can't change it like that.

julianoes (changes requested):

MatejFranceskin:
After more testing I found some additional problems. QGC mavlink FTP client ends the upload with reset session instead of terminate and I had to find a different way to achive what I needed.
I am closing this since it is not needed anymore and it might cause problems.

PR #1471, opened 2021-06-29
Few compatibility improvements for Ardupilot

levevg:
- ACKs from Ardupilot weren't received after https://github.com/mavlink/MAVSDK/pull/1455
- Ardupilot doesn't have CAL_* params, removed requesting them from Telemetry
- some actions (e.g. Action::takeoff) caused Ardupilot to crash because of all command_long params set to NaN by default
- Action::takeoff was denied because no positive altitude was sent
- Mission uploading wasn't working because Ardupilot sends MISSION_REQUEST instead of MISSION_REQUEST_INT

Unfortunately I didn't find a good way to determine if a system is Ardupilot to apply this in real time, so I just used a macro ARDUPILOT set at compilation time. It is probably ok because it's highly unlikely that somebody will use different autopilots at the same time

### DIFF
--- a/src/core/mavlink_commands.h
+++ b/src/core/mavlink_commands.h
@@ -65,19 +65,25 @@ public:
         }
     };
 
+#ifdef ARDUPILOT
+#define COMMAND_LONG_PARAM_UNSET 0
+#else
+#define COMMAND_LONG_PARAM_UNSET NAN
+#endif
+
     struct CommandLong {
         uint8_t target_system_id{0};
         uint8_t target_component_id{0};
         uint16_t command{0};
         uint8_t confirmation = 0;
         struct Params {
-            float param1 = NAN;
-            float param2 = NAN;
-            float param3 = NAN;
-            float param4 = NAN;
-            float param5 = NAN;
-            float param6 = NAN;
-            float param7 = NAN;
+            float param1 = COMMAND_LONG_PARAM_UNSET;
+            float param2 = COMMAND_LONG_PARAM_UNSET;
+            float param3 = COMMAND_LONG_PARAM_UNSET;
+            float param4 = COMMAND_LONG_PARAM_UNSET;
+            float param5 = COMMAND_LONG_PARAM_UNSET;
+            float param6 = COMMAND_LONG_PARAM_UNSET;
+            float param7 = COMMAND_LONG_PARAM_UNSET;
         } params{};
 
         // TODO: rename to set_all
--- a/src/plugins/action/action_impl.cpp
+++ b/src/plugins/action/action_impl.cpp
@@ -350,7 +350,7 @@ void ActionImpl::takeoff_async(const Action::ResultCallback& callback) const
     command.target_component_id = _parent->get_autopilot_id();
 
 #ifdef ARDUPILOT
-    command.params.param7 = 1.0; // Ardupilot doesn't take off without altitude set
+    command.params.param7 = get_takeoff_altitude().second;
 #endif
 
     _parent->send_command_async(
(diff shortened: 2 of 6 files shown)


src/plugins/action/action_impl.cpp
```diff
@@ -349,6 +349,10 @@ void ActionImpl::takeoff_async(const Action::ResultCallback& callback) const
     command.command = MAV_CMD_NAV_TAKEOFF;
     command.target_component_id = _parent->get_autopilot_id();
 
+#ifdef ARDUPILOT
+    command.params.param7 = 1.0; // Ardupilot doesn't take off without altitude set
```
JonasVautherin, line 353: @julianoes: does PX4 use this param? If yes we could move to it instead of the px4-specific parameter, right?

JonasVautherin (approved):
Very nice. I think it would be an improvement to do the runtime check (in the heartbeat, there is a MAV_AUTOPILOT enum that says if it is Ardupilot or PX4), but this does it for now.

The Ardupilot bits are clearly separated, which does not seem to add maintenance work.

I vote for merging this, waiting for @julianoes' opinion!

src/plugins/action/action_impl.cpp
julianoes: It does and I would like to switch to it but guess what: PX4 wants AMSL! So this is something we need to fix but for now this workaround make sense.

Although, hang on. @levevg we probably should save the takeoff altitude here:
https://github.com/mavlink/MAVSDK/blob/3ef7dc3040cdb14297623e44a6f6ac42479ee326/src/plugins/action/action_impl.cpp#L583-L586
and then use it.

src/core/mavlink_commands.h
```diff
@@ -65,19 +65,25 @@ class MavlinkCommandSender {
         }
     };
 
+#ifdef ARDUPILOT
+#define COMMAND_LONG_PARAM_UNSET 0
+#else
+#define COMMAND_LONG_PARAM_UNSET NAN
+#endif
```
julianoes, line 72: I'm not a big fan of this but for now it's ok. I'll revisit it later.

julianoes (commented):
Thanks @levevg!

bazfp:
Nice changes. I have been looking at this as well.

I can see the need for Ardupilot/PX4 support to be runtime, not compile time. `#IFDEF` are fine now for highlighting the changes but I suggest we pull them out into Ardupilot only impl files. We also need to create a runtime detector to use only the correct impl.

One way to fix the `NAN` blanking is to zero command types when using them for Ardupilot classes. Something like this works: (New constructor with blanking value would be better?)

```MavlinkCommandSender::CommandLong::set_as_reserved(command.params, 0.0f);```

It looks like telemetry health will still be bad as there is no implementation for that yet.

src/core/mavlink_commands.h
levevg: > we probably should save the takeoff altitude here
  @julianoes Do you mean to save it in a private variable if it is Ardupilot? Because now it's setting MIS_TAKEOFF_ALT param which is PX4-specific

JonasVautherin:
> I can see the need for Ardupilot/PX4 support to be runtime, not compile time. #IFDEF are fine now for highlighting the changes but I suggest we pull them out into Ardupilot only impl files.

Agreed for the runtime using MAV_AUTOPILOT from the heartbeat (maybe saving that in the `System` class, with `system.is_px4()` and `system.is_ardupilot()` functions to be used instead of the `#ifdef`?

However, I don't have a very strong opinion about Ardupilot-only impl files. The way it is here is very clear to me, with small changes. If for some plugin that was to become very different, then a different impl file would make sense of course.

@levevg: Would you mind using the altitude from `set_takeoff_altitude`, as suggested by @julianoes? If PX4, `set_takeoff_altitude` does what it does now, and in any case you could add that it saves it as a class member, to be used by the `takeoff_async()` function (instead of your hardcoded value to 1.0). With that change I think we could merge the PR.

JonasVautherin (approved):

julianoes (approved):
Nice, thank you!

PR #1473, opened 2021-07-02
Parameter Server

bazfp:
Added support for MAVSDK to act as a parameter server (autopilot) using `param_server` plugin.

* Listens to `PARAM_REQUEST_VALUE` `PARAM_EXT_REQUEST_VALUE` `PARAM_REQUEST_LIST` `PARAM_EXT_REQUEST_VALUE`
* Responds with appropriate `PARAM_VALUE` ext or otherwise
* Plugin requires a `system` before setting values - Not a huge fan but should be ok.

I've tested this by getting two `mavsdk` objects to communicate to one another with `AUTOPILOT` configuration, they don't register each other correctly as they are missing `MAV_CMD_REQUEST_AUTOPILOT_CAPABILITIES` and `MAV_CMD_SET_MESSAGE_INTERVAL` responses at current. Expect some more changes from me regarding this :)

Proto here: https://github.com/mavlink/MAVSDK-Proto/pull/238


src/core/mavlink_parameters.cpp
```diff
@@ -21,13 +21,34 @@ MAVLinkParameters::MAVLinkParameters(SystemImpl& parent) : _parent(parent)
         MAVLINK_MSG_ID_PARAM_EXT_ACK,
         std::bind(&MAVLinkParameters::process_param_ext_ack, this, std::placeholders::_1),
         this);
+
+    // Parameter Server Callbacks
+    _parent.register_mavlink_message_handler(
+        MAVLINK_MSG_ID_PARAM_REQUEST_READ,
+        std::bind(&MAVLinkParameters::process_param_request_read, this, std::placeholders::_1),
```
julianoes, line 28: The idea is to use lambdas instead of `std_bind` moving forward. However, we can also change that in a PR later.

src/core/mavlink_parameters.cpp
```diff
@@ -102,6 +123,21 @@ void MAVLinkParameters::get_param_async(
...
+}
+
+std::pair<MAVLinkParameters::Result, MAVLinkParameters::ParamValue>
+MAVLinkParameters::get_server_param(const std::string& name, ParamValue value_type)
+{
+    if (_param_server_store.find(name) != _param_server_store.end()) {
+        auto value = _param_server_store.at(name);
+        return {MAVLinkParameters::Result::Success, value};
+    }
+    return {MAVLinkParameters::Result::WrongType, {}};
```
julianoes, line 138: Is `WrongType` the right error here, or should we add something like `NotFound`?

src/core/mavlink_parameters.h
```diff
@@ -546,6 +552,12 @@ class MAVLinkParameters {
     std::mutex _all_param_mutex{};
 
     // dl_time_t _last_request_time = {};
+
+    std::map<std::string, ParamValue> _param_server_store;
+    void process_param_request_read(const mavlink_message_t& message);
+    void process_param_ext_request_read(const mavlink_message_t& message);
+    void process_param_request_list(const mavlink_message_t& message);
+    void process_param_set(const mavlink_message_t& message);
```
julianoes, line 560: My main question here is whether it makes sense to add this functionality to the existing `MAVLinkParameters` class, or a separate class. To me it's not entirely clear what is client and what is server related in this class. If it's mostly separate except from the type it might make sense to have a separate class and only re-use some of the types and helper functions.

src/core/mavlink_parameters.h
```diff
@@ -546,6 +552,12 @@ class MAVLinkParameters {
     std::mutex _all_param_mutex{};
 
     // dl_time_t _last_request_time = {};
+
+    std::map<std::string, ParamValue> _param_server_store;
```
julianoes, line 556: A `std::vector<std::pair<std::string, ParamValue>>` might be faster for small numbers of params.

src/core/mavlink_parameters.h
bazfp: I agree this should be separated for clarity. With provide/retrieve naming it is perhaps a little more clear. For now it will require re-implementation of the param worker/jobs so we can do this later.

src/core/mavlink_parameters.h
bazfp: Noted. I have left it as a map at current just because of it's easy lookup api compared to a pair.

julianoes (dismissed):
Nice, thanks for the edits.

julianoes (approved):

julianoes:
Thanks a lot @bazfp!

PR #1476, opened 2021-07-08
Run commands in parallel

JonasVautherin:
Currently, the commands are stored in a queue and sent serially. Only when the first command of the queue is done (i.e. it was acked or it timed out) can the next command be sent. This is fine when we do not send too many commands and when they do not time out (or are not slow).

With the Auterion cloud simulator, we hit the problem that the commands are executed slower than the command queue grows, which makes the whole command microservice implode. This PR seems to solve this problem by running the commands in parallel.

More specifically, here are the changes:

* We keep the sending queue, but it is only used by `do_work()`. Once a command has been sent, it is saved in a map.
* The map of sent commands uses the command id as the key. This means that only one command of each id can be run in parallel, otherwise the new one will return a `CommandDenied` error. That means that if one sends `takeoff()` twice, and the second time reaches `do_work()` before the first one finished, then the second one will error. That's on purpose, because with MAVLink it is currently impossible to run twice the same command, receive two ACKs, and know which ACK goes with which command (the ACK message only contains the command id, but if the same command is sent twice, it's useless).
* `receive_command_ack()` and `receive_timeout()` now act on the map of `_sent_commands`.
* The `_timeout_cookie` is now saved in the `Work` object, because each command needs a cookie.
* `receive_timeout()` now takes the command id as a parameter, as a way to know which command did time out.

### DIFF
--- a/src/core/mavlink_commands.cpp
+++ b/src/core/mavlink_commands.cpp
@@ -292,7 +292,7 @@ void MavlinkCommandSender::do_work()
 
     {
         std::lock_guard<std::mutex> lock(_sent_commands_mutex);
-        auto was_inserted =
+        const auto was_inserted =
             _sent_commands.insert(std::make_pair(work->mavlink_command, work)).second;
 
         // The command is inserted in the list only if another command with the same id does
--- a/src/core/mavlink_commands.cpp
+++ b/src/core/mavlink_commands.cpp
@@ -5,15 +5,6 @@
 
 namespace mavsdk {
 
-// TODO: Currently the mavlink command handling is made in a way to only
-//       process one command at any time. Therefore, the work state is global
-//       for this whole class.
-//       The limitation is made because:
-//       - We are not sure what exactly will happen if the commands are sent in parallel
-//         and what kind of edge cases we might run into.
-//       - The queue used does not support going through and checking each and every
-//         item yet.
-
 MavlinkCommandSender::MavlinkCommandSender(SystemImpl& system_impl) : _parent(system_impl)
 {
     _parent.register_mavlink_message_handler(
@@ -143,17 +134,17 @@ void MavlinkCommandSender::receive_command_ack(mavlink_message_t message)
     std::pair<Result, float> temp_result{Result::UnknownError, NAN};
 
     {
-        LockedQueue<Work>::Guard work_queue_guard(_work_queue);
-        auto work = work_queue_guard.get_front();
+        std::lock_guard<std::mutex> lock(_sent_commands_mutex);
 
-        if (!work) {
+        if (_sent_commands.find(command_ack.command) == _sent_commands.end()) {
+            LogWarn() << "Received ack for unexisting command: "
+                      << static_cast<int>(command_ack.command) << "! Ignoring...";
             return;
         }
 
-        if (work->mavlink_command != command_ack.command) {
-            // If the command does not match with our current command, ignore it.
-            LogWarn() << "Command ack " << int(command_ack.command)
-                      << " not matching our current command: " << work->mavlink_command;
+        auto work = _sent_commands.at(command_ack.command);
+
+        if (!work) {
             return;
         }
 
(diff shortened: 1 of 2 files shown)


src/core/mavlink_commands.cpp
```diff
@@ -278,39 +275,64 @@ void MavlinkCommandSender::receive_timeout()
...
+            return;
         }
     }
 
-    if (temp_callback != nullptr) {
-        call_callback(temp_callback, temp_result.first, temp_result.second);
-    }
+    work->already_sent = true;
+    _parent.register_timeout_handler(
+        std::bind(&MavlinkCommandSender::receive_timeout, this, work->mavlink_command),
```
julianoes, line 331: I usually try to go to lambda instead of `std::bind`.

src/core/mavlink_commands.h
```diff
@@ -122,19 +123,20 @@ class MavlinkCommandSender {
...
     };
 
     void receive_command_ack(mavlink_message_t message);
-    void receive_timeout();
+    void receive_timeout(const uint16_t command);
 
     void call_callback(const CommandResultCallback& callback, Result result, float progress);
 
     SystemImpl& _parent;
     LockedQueue<Work> _work_queue{};
```
julianoes, line 137: Can't you now remove that one?

julianoes (commented):
Nice, that's awesome!

julianoes:
> FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory

@JonasVautherin you hit the jackpot :money_mouth_face:.

src/core/mavlink_commands.h
```diff
@@ -122,19 +123,20 @@ class MavlinkCommandSender {
...
-
-    void* _timeout_cookie = nullptr;
+    std::unordered_map<uint16_t, std::shared_ptr<Work>> _sent_commands{};
```
julianoes, line 138: Actually, for low numbers a `std::vector<std::pair<uint16_t, Work>>` would likely be faster.

src/core/mavlink_commands.cpp
```diff
@@ -278,39 +275,64 @@ void MavlinkCommandSender::receive_timeout()
...
+
+    LogDebug() << "sending it the first time (" << work->mavlink_command << ")";
+    work->time_started = _parent.get_time().steady_time();
 
     {
-        LockedQueue<Work>::Guard work_queue_guard(_work_queue);
-        auto work = work_queue_guard.get_front();
+        std::lock_guard<std::mutex> lock(_sent_commands_mutex);
+        auto was_inserted =
+            _sent_commands.insert(std::make_pair(work->mavlink_command, work)).second;
```
julianoes, line 296: I'd probably use `const` as much as possible:

```suggestion
        const auto was_inserted =
            _sent_commands.insert(std::make_pair(work->mavlink_command, work)).second;
```

As well as below.

src/core/mavlink_commands.cpp
JonasVautherin: Agreed, but here I actually just moved around the previous code (which is admittedly not reflected well in the git diff), I did not want to refactor too much in this PR (if that's okay for you).

src/core/mavlink_commands.h
JonasVautherin: It is the one used by `do_work()` to run the commands. When a command is sent, it is now saved in the `unordered_map` which is used by the receiving side. Something like this:

* `_work_queue`: sending queue
* `_sent_commands`: ACK queue

While a command is "active" (i.e. in `_sent_commands`), another command with the same cmd_id cannot be sent by `_work_queue`.

src/core/mavlink_commands.cpp
julianoes: Yap, that's fine.

src/core/mavlink_commands.h
julianoes: Ah ok.

src/core/mavlink_commands.h
julianoes: I agree that it's more expressive. So, let's leave it as you have it and profile and optimize later if required. We can for instance write an abstraction for ordered and unordered map and implement it with a vector inside.

PR #1477, opened 2021-07-08
Cancel call_every when subscribing a second time

JonasVautherin:
It's not fixing a bug I have, but I came across this while reading the code. I believe that if one subscribes twice in a row without unsubscribing, then the first subscription is never cancelled (which is therefore a leak).

The only way to cancel a subscription is by calling `subscribe_` again with `nullptr`, but I think it should support being replaced by another one.

@julianoes maybe to get your opinion on that, as it would probably affect all the subscriptions, and at that point I would maybe look into supporting multiple subscriptions? Not sure.

### DIFF
--- a/src/core/call_every_handler.cpp
+++ b/src/core/call_every_handler.cpp
@@ -56,6 +56,7 @@ void CallEveryHandler::remove(const void* cookie)
     auto it = _entries.find(const_cast<void*>(cookie));
     if (it != _entries.end()) {
         _entries.erase(const_cast<void*>(cookie));
+        cookie = nullptr;
         _iterator_invalidated = true;
     }
 }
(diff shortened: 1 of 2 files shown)


src/core/call_every_handler.cpp
```diff
@@ -56,6 +56,7 @@ void CallEveryHandler::remove(const void* cookie)
     auto it = _entries.find(const_cast<void*>(cookie));
     if (it != _entries.end()) {
         _entries.erase(const_cast<void*>(cookie));
+        cookie = nullptr;
```
julianoes, line 59: But `cookie` is `const`?

julianoes (commented):
I think this was indeed a leak because the call every creates a `shared_ptr`, and uses that as the cookie/ptr.

src/core/call_every_handler.cpp
julianoes: Nevermind, I guess the pointer is `const`, the cookie can be changed, apparently.

julianoes (approved):

PR #1479, opened 2021-07-12
core: remove deprecated UID methods

julianoes:
This removes the old UUID stuff which has been marked deprecated for a while. The new way to access the uid is using the Info plugin.

### DIFF
--- a/src/integration_tests/action_transition_multicopter_fixedwing.cpp
+++ b/src/integration_tests/action_transition_multicopter_fixedwing.cpp
@@ -7,18 +7,7 @@
 
 using namespace mavsdk;
 
-// static void connect(Mavsdk);
-static void takeoff(std::shared_ptr<Action> action, std::shared_ptr<Telemetry> telemetry);
-void wait_until_hovering(std::shared_ptr<Telemetry> telemetry);
-static void takeoff_and_transition_to_fixedwing();
-static void land_and_disarm(std::shared_ptr<Action> action, std::shared_ptr<Telemetry> telemetry);
-
 TEST_F(SitlTest, ActionTransitionSync_standard_vtol)
-{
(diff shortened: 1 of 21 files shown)


TSC21 (dismissed):

src/integration_tests/action_transition_multicopter_fixedwing.cpp
```diff
@@ -44,11 +33,35 @@ void takeoff_and_transition_to_fixedwing()
...
+    ASSERT_EQ(action_ret, Action::Result::Success);
 
     LogInfo() << "Waiting until hovering";
-    wait_until_hovering(telemetry);
+    auto prom = std::promise<void>{};
+    auto fut = prom.get_future();
+    // Wait until hovering.
+    telemetry->subscribe_flight_mode([&telemetry, &prom](Telemetry::FlightMode mode) {
+        if (mode == Telemetry::FlightMode::Hold) {
+            telemetry->subscribe_flight_mode(nullllptr);
```
JonasVautherin, line 59: `nullllptr`? :sweat_smile:

JonasVautherin (changes requested):

src/integration_tests/action_transition_multicopter_fixedwing.cpp
julianoes: Haha, that must have slipped in after testing.

julianoes: Ah yes, I remember typing `all` for `git add --all` and it was in the wrong window but I couldn't see where it went.

julianoes: Also, beware of `nullllptr`, they are extra null, at least double!

JonasVautherin (approved):

PR #1480, opened 2021-07-13
Add support for multiple batteries

### LINKED ISSUE #1365: Support multiple batteries
As discussed [in this thread](https://github.com/mavlink/MAVSDK-Swift/issues/189#issuecomment-802221631), I think that `telemetry.battery()` should expose a list of batteries (instead of only one). In the usual case with one battery, that would be a list of length one. Also, MAVSDK should use BATTERY_STATUS if it is advertised, but fallback to SYS_STATUS if it is not.

jffortuna:
Requires https://github.com/mavlink/MAVSDK-Proto/pull/239
Adds support for multiple batteries, by handling BATTERY_STATUS message.
If no BATTERY_STATUS message is received, battery.id will default to 0, and SYS_STATUS values will be used.
Closes #1365

### DIFF
--- a/src/plugins/telemetry/telemetry_impl.cpp
+++ b/src/plugins/telemetry/telemetry_impl.cpp
@@ -1082,7 +1082,7 @@ void TelemetryImpl::process_battery_status(const mavlink_message_t& message)
     new_battery.remaining_percent = bat_status.battery_remaining * 1e-2f;
 
     set_battery(new_battery);
-    
+
     {
         std::lock_guard<std::mutex> lock(_subscription_mutex);
         if (_battery_subscription) {
--- a/src/plugins/telemetry/telemetry_impl.cpp
+++ b/src/plugins/telemetry/telemetry_impl.cpp
@@ -1027,21 +1027,21 @@ void TelemetryImpl::process_sys_status(const mavlink_message_t& message)
     mavlink_msg_sys_status_decode(&message, &sys_status);
 
     if (!_has_bat_status) {
-    Telemetry::Battery new_battery;
-    new_battery.voltage_v = sys_status.voltage_battery * 1e-3f;
-    // FIXME: it is strange calling it percent when the range goes from 0 to 1.
-    new_battery.remaining_percent = sys_status.battery_remaining * 1e-2f;
+        Telemetry::Battery new_battery;
+        new_battery.voltage_v = sys_status.voltage_battery * 1e-3f;
+        // FIXME: it is strange calling it percent when the range goes from 0 to 1.
+        new_battery.remaining_percent = sys_status.battery_remaining * 1e-2f;
 
-    set_battery(new_battery);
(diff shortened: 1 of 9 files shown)


src/plugins/telemetry/telemetry_impl.h
```diff
@@ -350,5 +351,9 @@ class TelemetryImpl : public PluginImplBase {
     void* _rc_channels_timeout_cookie{nullptr};
     void* _gps_raw_timeout_cookie{nullptr};
     void* _unix_epoch_timeout_cookie{nullptr};
+
+    // Battery info can be extracted form SYS_STATUS or from BATTERY_STATUS.
```
JonasVautherin, line 355:
```suggestion
    // Battery info can be extracted from SYS_STATUS or from BATTERY_STATUS.
```

JonasVautherin (changes requested):
Nitpick on a comment :innocent:

JonasVautherin:
I see that all the tests are failing, and the proto submodule is not updated in your PR. Did you generate the code from the proto submodule? If yes, did you forget to add it to the PR?

Also you need to run clang-format for the style (`./tools/fix_style.sh .`)

src/plugins/telemetry/telemetry_impl.cpp
```diff
@@ -1061,6 +1068,31 @@ void TelemetryImpl::process_sys_status(const mavlink_message_t& message)
...
+{
+    mavlink_battery_status_t bat_status;
+    mavlink_msg_battery_status_decode(&message, &bat_status);
+
+    _has_bat_status = true;
+
+    Telemetry::Battery new_battery;
+    new_battery.id = bat_status.id;
+    new_battery.voltage_v = bat_status.voltages[0] * 1e-3f;
+    // FIXME: it is strange calling it percent when the range goes from 0 to 1.
```
julianoes, line 1081: Good point :(

jffortuna:
> I see that all the tests are failing, and the proto submodule is not updated in your PR. Did you generate the code from the proto submodule? If yes, did you forget to add it to the PR?
> 
> Also you need to run clang-format for the style (`./tools/fix_style.sh .`)

Yes, I did submit this PR at the same time as the other on proto (merged now).
Wasn't sure about the procedure, if I should update that reference, or a maintainer would.

Will do.

src/plugins/telemetry/telemetry_impl.cpp
jffortuna: I actually copied this block from the status message, but indeed, it is a good point :)

JonasVautherin (approved):

src/plugins/telemetry/telemetry_impl.cpp
julianoes: Ah nevermind then.

julianoes (approved):

PR #1489, opened 2021-07-24
core/plugins: detect autopilot at runtime

julianoes:
This is much nicer for two reasons:
- No re-compilation required.
- All code paths are built in CI.


julianoes:
CI looks unrelated, merging.

PR #1497, opened 2021-08-02
Add `SET_MESSAGE_INTERVAL` support to telemetry_server publish functions

bazfp:
Use a mavlink message cache to fulfil the `SET_MESSAGE_INTERVAL` requirements in telemetry server plugin.

This removes the `UNSUPPORTED` message from `system_impl` for `SET_MESSAGE_INTERVAL`


src/plugins/telemetry_server/telemetry_server_impl.h
```diff
@@ -62,14 +68,21 @@ class TelemetryServerImpl : public PluginImplBase {
...
     TelemetryServer::Result publish_extended_sys_state(
-        TelemetryServer::VtolState vtol_state, TelemetryServer::LandedState landed_state) const;
+        TelemetryServer::VtolState vtol_state, TelemetryServer::LandedState landed_state);
 
 private:
     std::chrono::time_point<std::chrono::steady_clock> _start_time;
 
+    std::vector<RequestMsgInterval> _interval_requests;
+    std::mutex _interval_mutex;
+    std::mutex _msg_cache_mutex;
```
julianoes, line 81: Is `_msg_cache_mutex` really required? It's only used once where we also have `_interval_mutex`.

src/plugins/telemetry_server/telemetry_server_impl.h
```diff
@@ -62,14 +68,21 @@ class TelemetryServerImpl : public PluginImplBase {
...
+    std::map<uint64_t, mavlink_message_t> _msg_cache;
```
julianoes, line 82: Can it be a `std::unordered_map`?

src/plugins/telemetry_server/telemetry_server_impl.h
bazfp: Yes I was having the same thought. I suppose I just want to make sure each data store is locking on access :)

PR #1501, opened 2021-08-05
Fix two warnings recently introduced

julianoes:
Found by AppleClang:

```
/Users/julianoes/src/MAVSDK/src/core/system_impl.cpp:1384:66: warning: self-comparison always evaluates to true [-Wtautological-compare]
    return _parent.get_base_mode() & (MAV_MODE_FLAG_SAFETY_ARMED == MAV_MODE_FLAG_SAFETY_ARMED);
```

And GCC:

```
/home/julianoes/src/MAVSDK/src/plugins/mission_raw_server/mission_raw_server_impl.cpp:315:26: warning: comparison of integer expressions of different signedness: ‘int’ and ‘std::vector<mavsdk::MAVLinkMissionTransfer::ItemInt>::size_type’ {aka ‘long unsigned int’} [-Wsign-compare]
  315 |     if (_current_seq + 1 == _current_mission.size()) {
      |         ~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/src/core/system_impl.cpp
+++ b/src/core/system_impl.cpp
@@ -1381,7 +1381,7 @@ void SystemImpl::set_server_armed(bool armed)
 
 bool SystemImpl::is_server_armed() const
 {
-    return _parent.get_base_mode() & (MAV_MODE_FLAG_SAFETY_ARMED == MAV_MODE_FLAG_SAFETY_ARMED);
+    return (_parent.get_base_mode() & MAV_MODE_FLAG_SAFETY_ARMED) == MAV_MODE_FLAG_SAFETY_ARMED;
 }
 
 void SystemImpl::set_custom_mode(uint32_t custom_mode)
--- a/src/core/mavsdk_impl.cpp
+++ b/src/core/mavsdk_impl.cpp
@@ -641,7 +641,7 @@ void MavsdkImpl::set_base_mode(uint8_t base_mode)
     _base_mode = base_mode;
 }
 
-bool MavsdkImpl::get_base_mode() const
+uint8_t MavsdkImpl::get_base_mode() const
 {
     return _base_mode;
 }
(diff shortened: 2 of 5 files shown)


src/core/system_impl.cpp
```diff
@@ -1381,7 +1381,7 @@ void SystemImpl::set_server_armed(bool armed)
 
 bool SystemImpl::is_server_armed() const
 {
-    return _parent.get_base_mode() & (MAV_MODE_FLAG_SAFETY_ARMED == MAV_MODE_FLAG_SAFETY_ARMED);
```
bazfp, line 1384: The braces are wrong.

`(_parent.get_base_mode() & MAV_MODE_FLAG_SAFETY_ARMED) == MAV_MODE_FLAG_SAFETY_ARMED)`

bazfp (changes requested):

src/core/system_impl.cpp
julianoes: But look at my change, it's now anding and then `> 0`, so it should be true if the bit is there.

bazfp: Yes but it is a looser bitmask check. https://stackoverflow.com/a/14480058

julianoes: Ah, that's a thing, ok. Let me change it and force push.

bazfp:
```
D:\a\MAVSDK\MAVSDK\src\core\system_impl.cpp(1384,39): error C2220: the following warning is treated as an error [D:\a\MAVSDK\MAVSDK\build\release\src\core\mavsdk.vcxproj]
D:\a\MAVSDK\MAVSDK\src\core\system_impl.cpp(1384,39): warning C4806: '&': unsafe operation: no value of type 'bool' promoted to type 'MAV_MODE_FLAG' can equal the given constant 
```

I think this needs a static_cast somewhere to fix the windows build. Oh bitmasks :(

bazfp (approved):

PR #1514, opened 2021-08-23
Add heading to global_position telemetry plugins

bazfp:
https://github.com/mavlink/MAVSDK-Proto/pull/246

We are not parsing or providing the heading value in the `global_position_int` message used in telemetry plugins. This add that functionality.


JonasVautherin (changes requested):
I think you are missing some mocks (see CI output)

PR #1519, opened 2021-08-30
Add `param_set` handling

bazfp:
Missed the ability to set parameters clientside when I added the `param_server` plugin. 

Handles `param_set` requests and returns the value to confirm if successful.

Fills that gap :)

### DIFF
--- a/src/core/mavlink_parameters.cpp
+++ b/src/core/mavlink_parameters.cpp
@@ -13,6 +13,11 @@ MAVLinkParameters::MAVLinkParameters(SystemImpl& parent) : _parent(parent)
         this);
 
     _parent.register_mavlink_message_handler(
+        MAVLINK_MSG_ID_PARAM_SET,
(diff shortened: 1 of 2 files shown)


src/core/mavlink_parameters.cpp
```diff
@@ -709,6 +714,34 @@ std::ostream& operator<<(std::ostream& strm, const MAVLinkParameters::ParamValue
...
 }
 
+void MAVLinkParameters::process_param_set(const mavlink_message_t& message)
+{
+    mavlink_param_set_t set_request{};
+    mavlink_msg_param_set_decode(&message, &set_request);
+
+    std::string safe_param_id = extract_safe_param_id(set_request.param_id);
+    if (safe_param_id != "") {
+        LogDebug() << "Set Param Request: " << safe_param_id;
```
julianoes, line 724: Same question. Should these be warnings or errors?

julianoes (approved):

PR #1521, opened 2021-09-01
added vtol state implementation (follow up)

mattes-bru:
vtol state implementation in the telemetry plugin


julianoes:
The proto change was already done but this we might be able to use, thank you!

mattes-bru:
Yes, I saw that @rligocki was first on the proto but for the implementation I don't see a pull request. Any chance to merge his proto and my implementation?

src/plugins/telemetry/include/plugins/telemetry/telemetry.h
```diff
@@ -148,6 +148,24 @@ class Telemetry : public PluginBase {
      */
     friend std::ostream& operator<<(std::ostream& str, Telemetry::LandedState const& landed_state);
 
+    /**
+     * @brief VTOL State enumeration
+     */
+    enum class VtolState {
```
mattes-bru, line 154: I used the same naming scheme like in the other enums although I would personally prefer using more intuitive names than Mc and Fw

rligocki:
Pull request should be already done on Proto side. I think It was today when @julianoes merged it.

julianoes:
> Yes, I saw that @rligocki was first on the proto but for the implementation I don't see a pull request. Any chance to merge his proto and my implementation?

Let's try.

julianoes:
I've force-pushed this with the latest changes from https://github.com/mavlink/MAVSDK-Proto/pull/251.

julianoes (approved):

src/plugins/telemetry/include/plugins/telemetry/telemetry.h
bazfp: Yes the enums are just based on MAVLink definitions https://mavlink.io/en/messages/common.html#MAV_VTOL_STATE

bazfp:
Looks good, should work with `telemetry_server` (I should make some tests :cry: )

PR #1524, opened 2021-09-02
mission: add support for DO_GIMBAL_MANAGER_CONFIGURE in mission download

JonasVautherin:
It should just be ignored when downloading because that's an implementation detail as far as the Mission plugin is concerned. It will be added back when uploading, if I understand the code correctly :+1:.


julianoes (approved):
I think you're right.

PR #1528, opened 2021-09-03
Fix mission download

JonasVautherin:
Whoops :sweat_smile:


julianoes:
Oops. Do we need to create a patch release after this?

PR #1529, opened 2021-09-05
Update dependencies

JonasVautherin:


JonasVautherin:
Hmm I need to update absl and its patch, I think

julianoes:
Thx!

PR #1530, opened 2021-09-05
Update proto submodule

JonasVautherin:
Goes with https://github.com/mavlink/MAVSDK-Proto/pull/254

As can be seen, it does not modify anything regarding MAVSDK's API. It just renames some implementation details (which is needed for MAVSDK-Java).


JonasVautherin:
My guess is that the CI is failing because I generated the code with https://github.com/mavlink/MAVSDK/pull/1529 locally. So I'll wait until it is merged :innocent:

PR #1533, opened 2021-09-07
Improve connection and corresponding logging

JonasVautherin:
* Discover the remote system before starting the grpc server
* Do not call notify_on_discover when the sysid is 0 (which means that MAVSDK is connecting to a remote, and therefore a system has not actually been discovered)
* Improve connection logging (stuff like "discovered 0 components" is a bit confusing IMO)

### DIFF
--- a/src/core/mavsdk_impl.cpp
+++ b/src/core/mavsdk_impl.cpp
@@ -216,7 +216,11 @@ void MavsdkImpl::receive_message(mavlink_message_t& message, Connection* connect
 
     std::lock_guard<std::recursive_mutex> lock(_systems_mutex);
 
+    // The only situation where we create a system with sysid 0 is when we initialize the connection
+    // to the remote.
     if (_systems.size() == 1 && _systems[0].first == 0) {
+        LogDebug() << "New: System ID: " << static_cast<int>(message.sysid)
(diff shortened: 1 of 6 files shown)


julianoes (approved):
Awesome!

levevg:
@JonasVautherin 
Hi! After this change I can't receive any notification when remote is actually connected.
Maybe you could add notify_on_discover() here: https://github.com/mavlink/MAVSDK/blob/ea2386c109cd8b6284416d0d24317023fd03fec7/src/core/mavsdk_impl.cpp#L225

JonasVautherin:
You mean that your remote has a sysid set to 0? :thinking:

levevg:
> You mean that your remote has a sysid set to 0? 🤔

Nope, it is 1. But the problem is that when I connect to it via UDP, a placeholder is created in _systems vector, with zero id. When the remote is connected, zero is rewritten by actual sysid (in the place I linked above). Notification isn't fired at the moment of placeholder creation nor at the moment of setting the actual sysid, so it's not fired at all

JonasVautherin:
Oh right I see! Yes that makes sense. Can you open a PR with your suggestion? :pray:

PR #1535, opened 2021-09-08
Fix stop in mavsdk_server

JonasVautherin:
Now that mavsdk_server.run() blocks until a system is discovered, mavsdk_server.stop() must cancel it.

### DIFF
--- a/src/mavsdk_server/src/connection_initiator.h
+++ b/src/mavsdk_server/src/connection_initiator.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <future>
+#include <mutex>
 #include <string>
 
 #include "connection_result.h"
@@ -16,8 +17,6 @@ public:
 
     bool start(Mavsdk& mavsdk, const std::string& connection_url)
     {
-        init_mutex();
-
         LogInfo() << "Waiting to discover system on " << connection_url << "...";
         _discovery_future = wrapped_subscribe_on_new_system(mavsdk);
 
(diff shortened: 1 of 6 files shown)


src/mavsdk_server/src/connection_initiator.h
```diff
@@ -45,29 +51,29 @@ template<typename Mavsdk> class ConnectionInitiator {
...
 
         return future;
     }
 
-    std::once_flag _discovery_flag{};
-    std::shared_ptr<std::promise<void>> _discovery_promise{};
-    std::future<void> _discovery_future{};
+    std::mutex _mutex;
+    std::atomic<bool> _is_discovery_finished = false;
+    std::shared_ptr<std::promise<bool>> _discovery_promise{};
```
julianoes, line 75: Does this actually need to be a shared pointer? And why?

JonasVautherin: Actually it's passed to a callback, so if the `ConnectionInitiator` object goes out of scope and the callback is called, we need a shared_ptr to avoid a crash. Or don't we?

julianoes: Ok, I guess so.

julianoes (approved):
I think this looks alright.

JonasVautherin:
Tested to work with MAVSDK-Swift and MAVSDK-Python :tada:

PR #1540, opened 2021-09-09
New feature extension mission item

rligocki:
Created PR one more time with on a new branch. 
Impelementation works (It is possible to add vehicle action like takeoff, land or transition for each mission waypoint)
PR contains integration tests. 

Transition test is missing check if vehicle is able to do transition. If transtion test is run with quadcopter, quad will fly after transition endlessly with transition heading.


README.md
```diff
@@ -22,6 +22,7 @@ The MAVSDK C++ part consists of:
 - The [mavsdk_server](https://github.com/mavlink/MAVSDK/tree/main/src/mavsdk_server) implementing the gRPC server for the language clients.
 
 ## Repos
+UPDATE
```
julianoes, line 25: ?

src/integration_tests/mission_takeoff_land.cpp
```diff
@@ -0,0 +1,173 @@
...
+#include <cmath>
+#include "integration_test_helper.h"
+#include "path_checker.h"
+#include "mavsdk.h"
+#include "plugins/telemetry/telemetry.h"
+#include "plugins/action/action.h"
+#include "plugins/mission/mission.h"
+
+using namespace mavsdk;
+using namespace std::placeholders; // for `_1`
```
julianoes, line 15: Unused.

src/integration_tests/mission_takeoff_land.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+void do_mission_with_takeoff_and_land(float mission_altitude_m);
+
+TEST_F(SitlTest, MissionTakeoffAndLandHigh)
```
julianoes, line 19: Why `High`? What about just:
```suggestion
TEST_F(SitlTest, MissionTakeoffAndLand)
```

src/integration_tests/mission_transition.cpp
```diff
@@ -0,0 +1,178 @@
...
+using namespace std::placeholders; // for `_1`
+
+void do_mission_takeoff_transitions_land(float mission_altitude_m);
+
+TEST_F(SitlTest, MissionTakeoffTransitionAndLandHigh)
+{
+    do_mission_takeoff_transitions_land(40);
+}
+
+void do_mission_takeoff_transitions_land(float mission_altitude_m)
```
julianoes, line 24: Same thing as above. Why `High`? And you can just put the function into the `TEST_F`. That way the `ASSERT` can actually stop the test, once they are a function deep they can't and are like `EXPECT`.

src/plugins/mission/mission_impl.cpp
```diff
@@ -467,6 +468,73 @@ MissionImpl::convert_to_int_items(const std::vector<MissionItem>& mission_items)
             int_items.push_back(next_item);
         }
 
+        if (item.vehicle_action != VehicleAction::None) {
```
julianoes, line 471: So is this added after the waypoint or before? I'm not quite sure on that yet.

julianoes (changes requested):
I think we're getting there soon.

src/plugins/mission/mission_impl.cpp
julianoes: I read through this again and I don't think this is correct. So if you say that the first item of a mission would be a takeoff, then the current code would add a `MAV_CMD_NAV_WAYPOINT` at the beginning of this function, and then later down here a `MAV_CMD_NAV_TAKEOFF`, so that would be the wrong way round.

I think the correct way would be to add either a TAKEOFF, LAND, or WAYPOINT as the first item and then add the TRANSITION ones later on, like here. And then for the inverse function this would have to be taken care of in the same way.

src/integration_tests/mission_transition.cpp
rligocki: You mean to remove do_mission_takeoff_transitions_land and put test into TEST_F?

src/plugins/mission/mission_impl.cpp
rligocki: Current implementation adds vehicle action after any other action (waypoint, gimbal, camera). I would prefer different implementation. I agree, that `MAV_CMD_NAV_TAKEOFF` action should be placed before ` MAV_CMD_NAV_WAYPOINT`. In case of LAND and TRANSITION commands, I would prefer to add them after all other actions. It is more intuitive to place waypoint and then require vehicle to land or do transition.

src/integration_tests/mission_takeoff_land.cpp
```diff
@@ -0,0 +1,168 @@
...
+#include "integration_test_helper.h"
+#include "path_checker.h"
+#include "mavsdk.h"
+#include "plugins/telemetry/telemetry.h"
+#include "plugins/action/action.h"
+#include "plugins/mission/mission.h"
+
+using namespace mavsdk;
+
+void do_mission_with_takeoff_and_land(float mission_altitude_m);
```
julianoes, line 16: Leftover:

```suggestion
```

src/integration_tests/mission_transition.cpp
```diff
@@ -0,0 +1,171 @@
...
+#include "integration_test_helper.h"
+#include "path_checker.h"
+#include "mavsdk.h"
+#include "plugins/telemetry/telemetry.h"
+#include "plugins/action/action.h"
+#include "plugins/mission/mission.h"
+
+using namespace mavsdk;
+
+void do_mission_takeoff_transitions_land(float mission_altitude_m);
```
julianoes, line 16:
```suggestion
```

src/plugins/mission/mission_impl.cpp
```diff
@@ -289,6 +290,38 @@ MissionImpl::convert_to_int_items(const std::vector<MissionItem>& mission_items)
...
+                NAN,
+                x,
+                y,
+                z,
+                MAV_MISSION_TYPE_MISSION};
+
+            _mission_data.mavlink_mission_item_to_mission_item_indices.push_back(item_i);
+            int_items.push_back(next_item);
+        }
+
```
julianoes, line 324: If you do a TAKEOFF, then you don't need a WAYPOINT afterwards, I think. I'm not 100% sure how this would behave.

src/integration_tests/mission_transition.cpp
```diff
@@ -0,0 +1,171 @@
...
+            ASSERT_EQ(result, Mission::Result::Success);
+            prom->set_value();
+            LogInfo() << "Mission uploaded.";
+        });
+
+        auto status = future_result.wait_for(std::chrono::seconds(2));
+        ASSERT_EQ(status, std::future_status::ready);
+        future_result.get();
+    }
+
```
julianoes, line 127: Same here. Download and compare.

src/integration_tests/mission_takeoff_land.cpp
```diff
@@ -0,0 +1,168 @@
...
+            ASSERT_EQ(result, Mission::Result::Success);
+            prom->set_value();
+            LogInfo() << "Mission uploaded.";
+        });
+
+        auto status = future_result.wait_for(std::chrono::seconds(2));
+        ASSERT_EQ(status, std::future_status::ready);
+        future_result.get();
+    }
+
```
julianoes, line 124: Can you also download the mission again here and check if the items match again, so that we know the inverse works?

julianoes (commented):
Can you

src/integration_tests/mission_takeoff_land.cpp
rligocki: Don't understand what exactly do you mean. When whole mission is generated and uploaded into vehicle it is also possible to download this mission?

julianoes: Basically I mean this: https://github.com/mavlink/MAVSDK/blob/ff986d803be36de0f2bed850c3adb4ade98a95a1/src/integration_tests/mission.cpp#L196-L220

rligocki:
Looks like, there is problem with transition, when version 1.11 reaches higher altitude. Setting higher margin value should be solution for all version of firmware. Next problem is with comparison of ground states. 1.12 is ok, but looks like lower version indicates ground state different way.

julianoes:
> Looks like, there is problem with transition, when version 1.11 reaches higher altitude. Setting higher margin value should be solution for all version of firmware. Next problem is with comparison of ground states. 1.12 is ok, but looks like lower version indicates ground state different way.

Not sure what you mean but you tell me what you want to do. :)

rligocki:
I don't know how to solve difference in indication of ground states. If it works for 1.12, then how to modify ground state indication to be compatible with lower version?

src/integration_tests/mission_takeoff_land.cpp
julianoes: I have added it in 00fe5786f2e3e798e2ba183bdf6c4e7c9fddbcd0

rligocki:
@julianoes What is current problem in case of this PR? I would like to finally finish this PR.

JonasVautherin:
I must say I have lost track of this PR, I don't really remember.

But it seems that the first problem is that it does not pass the CI (conflicts right now). Then I'm not sure if you addressed all of Julian's comments (above in his review). Other than that I don't see a reason for not merging it once it is ready :+1:

PR #1541, opened 2021-09-09
Robustify gimbal v2 detection

JonasVautherin:
Instead of trying to detect the gimbal protocol once with a 1s timeout, this will try 10 times with a 2s timeout.


JonasVautherin:
Actually that won't work, I don't think we can wait 20 seconds before letting the user upload a mission with v1.

PR #1542, opened 2021-09-10
Fix CI permissions when generating xcframework

JonasVautherin:
No idea why, but the `.xcframework` generated in the CI  has permission issues (missing execution permissions) while the one I build locally is fine.

This is an attempt at fixing it.


julianoes (approved):

JonasVautherin:
Doesn't work.

PR #1546, opened 2021-09-13
core: refactor MAVLink command queue

julianoes:
This changes the command queue, so that:
- Only one queue is required instead of a queue and a sent map
- SET_MESSAGE_INTERVAL (511) and REQUEST_MESSAGE (512) commands are only denied if also param1 (msgid) match, otherwise they are kept in the queue until later. This should help with initial requests of messages.


JonasVautherin (dismissed):
Makes sense

JonasVautherin (approved):

PR #1548, opened 2021-09-14
Fix possible setting options list when camera uses exclusions

JonasVautherin:
Currently, parameters are refreshed when they are changed. If the camera definition defines an "update" dependency (not sure when that's done), then a parameter that is modified triggers a refresh in those update dependencies.

However, what is currently missing is a way to refresh parameters that were excluded. Let's have a look using the python [`examples/camera_params.py`](https://github.com/mavlink/MAVSDK-Python/blob/main/examples/camera_params.py) example.

Without this PR, say the camera is in VIDEO mode, with exposure set to "Aperture Priority". The output will be:

```
=== Current settings ===

* CAM_MODE: PHOTO
* CAM_FOCUSMODE: Focus Mode
    -> Infinity
* CAM_EXPMODE: Exposure Mode
    -> Aperture Priority
* CAM_WBMODE: White Balance
    -> Auto
* CAM_EV: Exposure Compensation
    -> 0
* CAM_APERTURE: Aperture
    -> f/3.5
* CAM_ISO: ISO
    -> Auto
* CAM_IMGDEST: Image store
    -> Camera
* CAM_IMGSIZE: Image Size
    -> Large

Usage:
  p    print current (changeable) camera settings
  m    change camera mode
  s    change a setting
```

Note how there is no shutter option, because it is excluded by the aperture option. Now say you move to "Manual Exposure", the shutter option will not appear. This PR aims at fixing that, such that switching between "Manual", "Program Auto", "Shutter Priority" and "Aperture Priority" update the list accordingly.

This issue was first discovered in the MAVSDK-Swift sample app. I will test it there before setting this PR as "ready".

### DIFF
--- a/src/plugins/camera/camera_definition.cpp
+++ b/src/plugins/camera/camera_definition.cpp
@@ -113,22 +113,22 @@ bool CameraDefinition::parse_xml()
 
         const char* type_str = e_parameter->Attribute("type");
         if (!type_str) {
-            LogErr() << "type attribute missing";
+            LogErr() << "type attribute missing for " << param_name;
             return false;
         }
 
         if (strcmp(type_str, "string") == 0) {
-            LogDebug() << "Ignoring string params.";
+            LogDebug() << "Ignoring string params: " << param_name;
             continue;
         }
 
         if (strcmp(type_str, "custom") == 0) {
-            LogDebug() << "Ignoring custom params.";
+            LogDebug() << "Ignoring custom params: " << param_name;
             continue;
         }
 
         if (!new_parameter->type.set_empty_type_from_xml(type_str)) {
-            LogErr() << "unknown type attribute";
+            LogErr() << "Unknown type attribute: " << type_str;
             return false;
         }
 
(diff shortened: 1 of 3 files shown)


julianoes (approved):

src/plugins/camera/camera_definition.cpp
```diff
@@ -486,9 +486,6 @@ bool CameraDefinition::get_possible_settings(
 
     for (const auto& parameter : _parameter_map) {
         for (const auto& option : parameter.second->options) {
-            if (_current_settings[parameter.first].needs_updating) {
-                continue;
-            }
```
JonasVautherin, line 491: I removed those lines because an exclusion does not depend on `needs_updating`. I could verify in the test camera_definition that the unit test below was wrong: it was exposing `CAM_VIDFMT` when it should actually have been excluded.

src/plugins/camera/camera_impl.cpp
```diff
@@ -1633,14 +1633,15 @@ void CameraImpl::refresh_params()
         // any other possible settings to change. However, we still would
         // like to notify the current settings with this one change.
         notify_current_settings();
+        notify_possible_setting_options();
```
JonasVautherin, line 1636: This one line is the actual fix I needed :sweat_smile:

julianoes: Haha, ok!

julianoes (approved):

JonasVautherin:
Working with the MAVSDK-Swift sample app!

JonasVautherin:
Fedora CI issue is unrelated.

PR #1551, opened 2021-09-16
plugins: action: change to Hold mode before sending DO_REPOSITION

TSC21:
Go-To's are only processed in PX4 if in _Hold_ mode. So one needs to enforce changing to this mode first before sending the `DO_REPOSITION` Mavlink command.

### DIFF
--- a/src/plugins/action/action_impl.cpp
+++ b/src/plugins/action/action_impl.cpp
@@ -389,18 +389,34 @@ void ActionImpl::goto_location_async(
     const float yaw_deg,
     const Action::ResultCallback& callback)
 {
-    MavlinkCommandSender::CommandInt command{};
+    // Change to Hold mode first
+    _parent->set_flight_mode_async(
+        SystemImpl::FlightMode::Hold,
+        [this, yaw_deg, latitude_deg, longitude_deg, altitude_amsl_m, callback](
+            MavlinkCommandSender::Result mode_change_result, float) {
+            Action::Result action_result = action_result_from_command_result(mode_change_result);
 
-    command.command = MAV_CMD_DO_REPOSITION;
-    command.target_component_id = _parent->get_autopilot_id();
-    command.params.param4 = to_rad_from_deg(yaw_deg);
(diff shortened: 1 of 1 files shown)


src/plugins/action/action_impl.cpp
```diff
@@ -389,18 +389,30 @@ void ActionImpl::goto_location_async(
...
-        command, [this, callback](MavlinkCommandSender::Result result, float) {
-            command_result_callback(result, callback);
+    // Change to Hold mode first
+    _parent->set_flight_mode_async(
+        SystemImpl::FlightMode::Hold,
+        [this, yaw_deg, latitude_deg, longitude_deg, altitude_amsl_m, callback](
+            MavlinkCommandSender::Result mode_change_result, float) {
+            Action::Result action_result = action_result_from_command_result(mode_change_result);
+
+            // If successfully changed to Hold, then send the DO_REPOSITION command
```
JonasVautherin, line 399: What happens otherwise? Shouldn't there be an `else` here? :sweat_smile:

JonasVautherin (changes requested):
One question I have is... should `goto_location_async` change the flight mode, or should it return an error if it is not in the right flight mode and let the user do it? Because we already have an action to go to hold mode: https://github.com/mavlink/MAVSDK-Proto/blob/main/protos/action/action.proto#L100

src/plugins/action/action_impl.cpp
TSC21: good point. you have a suggestion?

JonasVautherin: Well yeah, you should call `callback` with the error result

TSC21: done

TSC21:
> or should it return an error if it is not in the right flight mode and let the user do it?

That seems counter-intuitive. QGC/AMC do force a mode change before sending the GoTo. So I believe this should happen here as well.

TSC21:
@JonasVautherin sorry I accidentally deleted the branch. Follow-up here: https://github.com/mavlink/MAVSDK/pull/1553.

PR #1552, opened 2021-09-16
Add health support for ardupilot

DaRavenox:
For the Ardupilot branch the following is done to set health:

1) Checks calibration of the magnetometer, gyro and accelerometer by checking the corresponding offset values. If any offset is zero then the calibration is not complete.

2) Requests the home position. The currently home position is sent only on startup for Ardupilot so we need to ask for it here in case we have missed it.

### DIFF
--- a/src/plugins/telemetry/telemetry_impl.cpp
+++ b/src/plugins/telemetry/telemetry_impl.cpp
@@ -1057,7 +1057,6 @@ void TelemetryImpl::process_sys_status(const mavlink_message_t& message)
             _parent->call_user_callback([callback, arg]() { callback(arg); });
         }
     }
-
     const bool armable = sys_status.onboard_control_sensors_health & MAV_SYS_STATUS_PREARM_CHECK;
 
     set_health_armable(armable);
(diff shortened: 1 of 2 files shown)


src/plugins/telemetry/telemetry_impl.cpp
```diff
@@ -2255,52 +2473,134 @@ void TelemetryImpl::check_calibration()
                     std::placeholders::_2),
                 this);
         }
-    } else {
-        _parent->remove_call_every(_calibration_cookie);
```
bazfp, line 2259: Is this intentional?

src/plugins/telemetry/telemetry_impl.h
```diff
@@ -381,5 +393,24 @@ class TelemetryImpl : public PluginImplBase {
...
     bool _has_received_mag_calibration{false};
+
+    // Ardupilot calibration status values
+
+    struct ap_calibration_offset {
+        float x{0};
+        float y{0};
+        float z{0};
+
+        [[nodiscard]] bool calibrated() const { return ((x != 0) && (y != 0) && (z != 0)); }
```
bazfp, line 404: Fun, never seen `nodiscard` before

src/plugins/telemetry/telemetry_impl.h
```diff
@@ -74,6 +74,7 @@ class TelemetryImpl : public PluginImplBase {
     void set_rate_unix_epoch_time_async(double rate_hz, Telemetry::ResultCallback callback);
 
     void get_gps_global_origin_async(const Telemetry::GetGpsGlobalOriginCallback callback);
+    void request_home_position_async();
```
bazfp, line 77: This should be private

src/plugins/telemetry/telemetry_impl.cpp
DaRavenox: Yes, as far as I understand this is called once we no longer want to keep calling this check_calibration(). Before we wanted to not check calibration when our system was detected as Ardupilot and thus we remove it the first time check_calibration gets called. Now we do want to keep checking calibration until we have gotten healthy values, hence this is removed.

src/plugins/telemetry/telemetry_impl.h
DaRavenox: Might not be useful, but since all my usages of this are nondiscarding I put it here.

src/plugins/telemetry/telemetry_impl.h
DaRavenox: Done

bazfp (approved):

src/plugins/telemetry/telemetry_impl.cpp
```diff
@@ -2138,6 +2263,14 @@ void TelemetryImpl::subscribe_heading(Telemetry::HeadingCallback& callback)
...
 }
 
+void TelemetryImpl::request_home_position_async()
+{
+    MavlinkCommandSender::CommandLong command_request_message{*_parent};
+    command_request_message.command = MAV_CMD_REQUEST_MESSAGE;
+    command_request_message.params.param1 = MAVLINK_MSG_ID_HOME_POSITION;
+    _parent->send_command_async(command_request_message, nullptr);
+}
+
```
JonasVautherin, line 2273: Where is the response to that request used? I think I'm missing it :thinking:

DaRavenox: We setup a message handler for the `HOME_POSITION` message here https://github.com/mavlink/MAVSDK/blob/634f171bb5b492e2d5bd6453f54e7e88c603899e/src/plugins/telemetry/telemetry_impl.cpp#L42-L45  which is used to handle the response.

JonasVautherin (approved):
Got it. So it's the same as the PX4 checks, only with the Ardupilot params. Makes sense to me!

Thanks a lot!

PR #1553, opened 2021-09-16
plugins: action: change to Hold mode before sending DO_REPOSITION

TSC21:
Go-To's are only processed in PX4 if in Hold mode. So one needs to enforce changing to this mode first before sending the DO_REPOSITION Mavlink command.

### DIFF
--- a/src/plugins/action/action_impl.cpp
+++ b/src/plugins/action/action_impl.cpp
@@ -389,19 +389,39 @@ void ActionImpl::goto_location_async(
     const float yaw_deg,
     const Action::ResultCallback& callback)
 {
-    MavlinkCommandSender::CommandInt command{};
+    auto send_do_reposition =
+        [this, callback, yaw_deg, latitude_deg, longitude_deg, altitude_amsl_m]() {
+            MavlinkCommandSender::CommandInt command{};
 
-    command.command = MAV_CMD_DO_REPOSITION;
-    command.target_component_id = _parent->get_autopilot_id();
-    command.params.param4 = to_rad_from_deg(yaw_deg);
-    command.params.x = int32_t(std::round(latitude_deg * 1e7));
-    command.params.y = int32_t(std::round(longitude_deg * 1e7));
-    command.params.z = altitude_amsl_m;
+            command.command = MAV_CMD_DO_REPOSITION;
+            command.target_component_id = _parent->get_autopilot_id();
+            command.params.param4 = to_rad_from_deg(yaw_deg);
+            command.params.x = int32_t(std::round(latitude_deg * 1e7));
+            command.params.y = int32_t(std::round(longitude_deg * 1e7));
+            command.params.z = altitude_amsl_m;
 
-    _parent->send_command_async(
-        command, [this, callback](MavlinkCommandSender::Result result, float) {
-            command_result_callback(result, callback);
-        });
+            _parent->send_command_async(
+                command, [this, callback](MavlinkCommandSender::Result result, float) {
+                    command_result_callback(result, callback);
+                });
+        };
+
+    // Change to Hold mode first
+    if (_parent->get_flight_mode() != SystemImpl::FlightMode::Hold) {
+        _parent->set_flight_mode_async(
+            SystemImpl::FlightMode::Hold,
+            [this, callback, send_do_reposition](MavlinkCommandSender::Result result, float) {
+                Action::Result action_result = action_result_from_command_result(result);
+                if (action_result != Action::Result::Success) {
+                    command_result_callback(result, callback);
+                    return;
+                }
+                send_do_reposition();
+            });
+        return;
+    }
+
+    send_do_reposition();
 }
 
 void ActionImpl::do_orbit_async(


JonasVautherin:
Same question as on the closed branch:

One question I have is... should goto_location_async change the flight mode, or should it return an error if it is not in the right flight mode and let the user do it? Because we already have an action to go to hold mode: https://github.com/mavlink/MAVSDK-Proto/blob/main/protos/action/action.proto#L100

@julianoes do you have an opinion?

TSC21:
> One question I have is... should goto_location_async change the flight mode, or should it return an error if it is not in the right flight mode and let the user do it? Because we already have an action to go to hold mode: https://github.com/mavlink/MAVSDK-Proto/blob/main/protos/action/action.proto#L100

I answered there: https://github.com/mavlink/MAVSDK/pull/1551#issuecomment-920963749

src/plugins/action/action_impl.cpp
```diff
@@ -389,18 +389,34 @@ void ActionImpl::goto_location_async(
...
+                command.params.z = altitude_amsl_m;
+
+                _parent->send_command_async(
+                    command,
+                    [this, callback](MavlinkCommandSender::Result do_reposition_result, float) {
+                        command_result_callback(do_reposition_result, callback);
+                    });
+            } else {
+                if (callback) {
+                    callback(action_result);
```
JonasVautherin, line 417: Same as above:

```
_parent->send_command_async(...
```

JonasVautherin (changes requested):

src/plugins/action/action_impl.cpp
TSC21: I don't get what you mean

JonasVautherin:
```suggestion
                    _parent->send_command_async(
                        command,
                        [this, callback](MavlinkCommandSender::Result mode_change_result, float) {
                            command_result_callback(mode_change_result_result, callback);
                        });
```

I mean something like this, where you put the callback on the queue instead of calling it directly.

But I haven't tested it :innocent:

TSC21: Ah got it

TSC21: actually I am still trying to see where that is the correct approach. I just followed for example what happens with `arm_async` and it looks similar to what I did. I ended up mimic the approach there, as you can see on the commit now. And I guess that's the correct approach considering it's also being applied to the arm command.

src/plugins/action/action_impl.cpp
```diff
@@ -389,19 +389,40 @@ void ActionImpl::goto_location_async(
...
+    // Change to Hold mode first
+    if (_parent->get_flight_mode() != SystemImpl::FlightMode::Hold) {
+        _parent->set_flight_mode_async(
+            SystemImpl::FlightMode::Hold,
+            [callback, send_do_reposition](MavlinkCommandSender::Result result, float) {
+                Action::Result action_result = action_result_from_command_result(result);
+                if (action_result != Action::Result::Success) {
+                    if (callback) {
+                        callback(action_result);
+                    }
```
JonasVautherin, line 418: Actually, I think it can be done just like in the other places in action_impl.cpp:

```suggestion
                    command_result_callback(callback, action_result);
```

JonasVautherin (changes requested):

src/plugins/action/action_impl.cpp
TSC21: So why is it different for the `arm` command? Is it because the API was just not updated?

JonasVautherin: I would say so, but let's ask @julianoes to have a second opinion :)

julianoes:
> @julianoes do you have an opinion?

My opinion is that it should be "easy" and the drone should do what is expected. So in this case, I would vote to automatically switch flight mode before sending the command.

src/plugins/action/action_impl.cpp
julianoes: I think this is also wrong for the `arm` command. The point is that `call_user_callback` needs to be used which happens in `command_result_callback`.

TSC21: Done

src/plugins/action/action_impl.cpp
```diff
@@ -389,19 +389,38 @@ void ActionImpl::goto_location_async(
...
+        };
+
+    // Change to Hold mode first
+    if (_parent->get_flight_mode() != SystemImpl::FlightMode::Hold) {
+        _parent->set_flight_mode_async(
+            SystemImpl::FlightMode::Hold,
+            [this, callback, send_do_reposition](MavlinkCommandSender::Result result, float) {
+                Action::Result action_result = action_result_from_command_result(result);
+                if (action_result != Action::Result::Success) {
+                    command_result_callback(result, callback);
```
julianoes, line 416:
```suggestion
                    command_result_callback(result, callback);
                    return;
```

julianoes (approved):

PR #1555, opened 2021-09-18
CLion/clang-tidy pass

julianoes:
More to come...

### DIFF
--- a/src/core/call_every_handler.cpp
+++ b/src/core/call_every_handler.cpp
@@ -1,17 +1,17 @@
 #include "call_every_handler.h"
 
+#include <utility>
+
 namespace mavsdk {
 
 CallEveryHandler::CallEveryHandler(Time& time) : _time(time) {}
 
-CallEveryHandler::~CallEveryHandler() {}
-
 void CallEveryHandler::add(std::function<void()> callback, double interval_s, void** cookie)
 {
     auto new_entry = std::make_shared<Entry>();
-    new_entry->callback = callback;
+    new_entry->callback = std::move(callback);
     auto before = _time.steady_time();
-    // Make sure it gets run straightaway. The epsilon seemed not enough so
+    // Make sure it gets run straightaway. The epsilon seemed not enough, so
     // we use the arbitrary value of 1 ms.
     _time.shift_steady_time_by(before, -interval_s - 0.001);
     new_entry->last_time = before;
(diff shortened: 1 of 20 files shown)


JonasVautherin (approved):

src/core/mavlink_mission_transfer.h
```diff
@@ -25,10 +25,10 @@ class Sender {
     Sender() = default;
     virtual ~Sender() = default;
     virtual bool send_message(mavlink_message_t& message) = 0;
-    virtual uint8_t get_own_system_id() const = 0;
-    virtual uint8_t get_own_component_id() const = 0;
-    virtual uint8_t get_system_id() const = 0;
-    virtual Autopilot autopilot() const = 0;
+    [[nodiscard]] virtual uint8_t get_own_system_id() const = 0;
```
JonasVautherin, line 28: Just out of curiosity: shouldn't the compiler automatically warn when the return value of a const function is ignored? Is it just good practice to use both `const` and `[[nodiscard]]` then?

src/core/http_loader_test.cpp
```diff
@@ -12,9 +12,9 @@ using namespace mavsdk;
 
 class HttpLoaderTest : public testing::Test {
 protected:
-    const std::string _file_url_1 = "http://subdomain.domain.com/some_folder/some_file";
-    const std::string _file_url_2 = "http://subdomain.domain.com/some_folder/another_file.txt";
-    const std::string _file_url_3 = "http://subdomain.domain.com/some_folder/yet_another_file.mp4";
+    const std::string _file_url_1 = "https://subdomain.domain.com/some_folder/some_file";
```
JonasVautherin, line 15: Does it actually make a difference? If yes, shouldn't we both test for `http` and `https`?

src/core/mavsdk.h
```diff
@@ -131,17 +131,6 @@ class Mavsdk {
...
-     * @brief Adds a TCP connection with a specific port number on localhost.
-     *
-     * @param remote_port The TCP port to connect to (defaults to 5760).
-     * @param forwarding_option message forwarding option (when multiple interfaces are used).
-     * @return The result of adding the connection.
-     */
-    ConnectionResult add_tcp_connection(
-        int remote_port = DEFAULT_TCP_REMOTE_PORT,
-        ForwardingOption forwarding_option = ForwardingOption::ForwardingOff);
-
```
JonasVautherin, line 144: aha, so that one was actually not implemented? :sweat_smile:

JonasVautherin (approved):

src/core/http_loader_test.cpp
julianoes: Hehe good point. That change is probably a bit over the top.

src/core/mavsdk.h
julianoes: Looks like it.

PR #1559, opened 2021-09-21
Fix list_photos

JonasVautherin:
* Don't get stuck in busy mode: when returning an error, set `is_fetching_photos` back to `false`
* Don't notify for old photos: if `last_advertised_image_index == -1` (i.e. the camera plugin is not fully initialized), don't notify for `CAMERA_IMAGE_CAPTURED` message because it may come from QGC listing old photos.
* Clear image storage when formatting, otherwise the image metadata stay for the deleted images and we end up requesting them for nothing

### DIFF
--- a/src/plugins/camera/camera_impl.cpp
+++ b/src/plugins/camera/camera_impl.cpp
@@ -865,7 +865,8 @@ void CameraImpl::process_camera_image_captured(const mavlink_message_t& message)
         _captured_request_cv.notify_all();
 
         std::lock_guard<std::mutex> lock(_capture_info.mutex);
-        if (_capture_info.last_advertised_image_index < capture_info.index &&
+        if (_capture_info.last_advertised_image_index != -1 &&
+            _capture_info.last_advertised_image_index < capture_info.index &&
             _capture_info.callback) {
             _capture_info.last_advertised_image_index = capture_info.index;
             const auto temp_callback = _capture_info.callback;
@@ -1772,7 +1773,7 @@ void CameraImpl::list_photos_async(
     }
 
     {
-        std::lock_guard<std::mutex> lock(_status.mutex);
+        std::lock_guard<std::mutex> status_lock(_status.mutex);
 
         if (_status.is_fetching_photos) {
             _parent->call_user_callback([callback]() {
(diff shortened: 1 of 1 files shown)


src/plugins/camera/camera_impl.cpp
```diff
@@ -1824,6 +1826,8 @@ void CameraImpl::list_photos_async(
                 while (cv_status == std::cv_status::timeout) {
                     request_try_number++;
                     if (request_try_number >= request_try_limit) {
+                        std::lock_guard<std::mutex> status_lock(_status.mutex);
```
julianoes, line 1833: What about having one lock only? Did you check that `_status.mutex` is now only accessed inside `_capture_request_mutex` or alone? Otherwise this could deadlock.

JonasVautherin: You mean locking `_status.mutex` for the whole `list_photos` requesting? Not sure I want to do that, it could involve thousands of round-trips :confused:.

Now `_captured_request_mutex` is used exactly once, when this thread starts. To prevent this loop from running twice in parallel. And `_status.mutex` is otherwise only used alone.

Not sure how to do otherwise... do you have a suggestion?

julianoes: as long as we don't acquire these mutexs in reverse order anywhere it's fine.

JonasVautherin: Currently we don't. I know it's not ideal, but well...

src/plugins/camera/camera_impl.cpp
```diff
@@ -1832,12 +1840,14 @@ void CameraImpl::list_photos_async(
 
                     _parent->send_command_async(
                         make_command_request_camera_image_captured(i), nullptr);
-                    cv_status =
-                        _captured_request_cv.wait_for(lock, std::chrono::duration<double>(1));
+                    cv_status = _captured_request_cv.wait_for(
+                        capture_request_lock, std::chrono::duration<double>(1));
```
julianoes, line 1844: What's this duration? Why not `std::chrono::seconds(1)`?

julianoes (approved):

PR #1560, opened 2021-09-21
Fix some scope issues in mavlink_commands.cpp

JonasVautherin:
Seems to me that the lack of `break` here ends up crashing MAVSDK when there is a timeout. At least I had that issue, and this fixed it.

At the same time I fixed some unused `temp_callback`, I think.

### DIFF
--- a/src/core/mavlink_commands.cpp
+++ b/src/core/mavlink_commands.cpp
@@ -101,7 +101,7 @@ void MavlinkCommandSender::queue_command_async(
             LogWarn() << "Dropping command " << static_cast<int>(identification.command)
                       << " that is already being sent";
             auto temp_callback = callback;
-            call_callback(callback, Result::CommandDenied, NAN);
+            call_callback(temp_callback, Result::CommandDenied, NAN);
             return;
         }
     }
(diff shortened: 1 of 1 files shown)


JonasVautherin:
This said, I also see a lot of stuff like this, where it times out for an "unexisting command". Not sure I understand what that means:

```
[01:43:07|Warn ] sending again after 0.503432 s, retries to do: 3  (2000). (mavlink_commands.cpp:283)
[01:43:07|Warn ] Timeout for unexisting command: 2000! Ignoring... (mavlink_commands.cpp:316)
[01:43:07|Warn ] sending again after 1.00773 s, retries to do: 2  (2000). (mavlink_commands.cpp:283)
[01:43:07|Warn ] Timeout for unexisting command: 2000! Ignoring... (mavlink_commands.cpp:316)
[01:43:08|Warn ] sending again after 1.51182 s, retries to do: 1  (2000). (mavlink_commands.cpp:283)
[01:43:08|Warn ] Timeout for unexisting command: 2000! Ignoring... (mavlink_commands.cpp:316)
[01:43:08|Error] Retrying failed (2000) (mavlink_commands.cpp:303)
```

src/core/mavlink_commands.cpp
```diff
@@ -305,6 +305,7 @@ void MavlinkCommandSender::receive_timeout(const CommandIdentification& identifi
             temp_callback = work->callback;
             temp_result = {Result::ConnectionError, NAN};
             _work_queue.erase(it);
+            break;
```
julianoes, line 308: I thought I had fixed that one yesterday. Strange.

julianoes (approved):

PR #1562, opened 2021-09-21
Ardupilot mission handling and vehicle type checking

DaRavenox:
Adds support to mission_raw to upload, start missions and clear missions when the autopilot is ArduPilot.

Further adds support to grab the vehicle type from the HEARTBEAT message, and to use this to determine what the correct mode command is for a ArduPilot vehicle, since this depends on the type.

### DIFF
--- a/src/core/system_impl.h
+++ b/src/core/system_impl.h
@@ -88,7 +88,6 @@ public:
 
     Autopilot autopilot() const override { return _autopilot; };
 
-
     FlightMode to_flight_mode_from_custom_mode(uint32_t custom_mode);
     static FlightMode to_flight_mode_from_px4_mode(uint32_t custom_mode);
     static FlightMode to_flight_mode_from_ardupilot_rover_mode(uint32_t custom_mode);
--- a/src/core/system_impl.h
+++ b/src/core/system_impl.h
@@ -50,6 +50,52 @@ public:
         Stabilized,
     };
 
+    // Enumeration representing the available modes for the Arudpilot rover autopilot.
+    enum class APRoverMode{
+        Manual = 0,
+        Acro = 1,
+        Steering = 3,
+        Hold = 4,
+        Loiter = 5,
+        Follow = 6,
+        Simple = 7,
+        Auto = 10,
+        RTL = 11,
+        Smart_RTL = 12,
+        Guided = 15,
+        Initializing = 16
+    };
+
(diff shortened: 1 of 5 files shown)


src/core/system_impl.h
```diff
@@ -50,6 +50,52 @@ class SystemImpl : public Sender {
...
+        Guided_No_GPS = 20,
+        Smart_RTL = 21,
+        Flow_Hold = 22,
+        Follow = 23,
+        Zigzag = 24,
+        System_ID = 25,
+        Auto_Rotate = 26,
+        Auto_RTL = 27,
+        Turtle = 28
+    };
```
julianoes, line 97: What about moving these enums to a separate header file like https://github.com/mavlink/MAVSDK/blob/main/src/core/px4_custom_mode.h?

src/plugins/mission_raw/mission_raw_impl.cpp
```diff
@@ -377,15 +399,23 @@ void MissionRawImpl::clear_mission_async(const MissionRaw::ResultCallback& callb
...
     reset_mission_progress();
 
-    _parent->mission_transfer().clear_items_async(
-        MAV_MISSION_TYPE_MISSION, [this, callback](MAVLinkMissionTransfer::Result result) {
-            auto converted_result = convert_result(result);
-            _parent->call_user_callback([callback, converted_result]() {
-                if (callback) {
-                    callback(converted_result);
-                }
+    // For ArduPilot to clear a mission we need to upload an empty mission.
```
julianoes, line 402: Ha, really? :joy: That's funny.

src/core/system_impl.h
```diff
@@ -50,6 +50,52 @@ class SystemImpl : public Sender {
         Stabilized,
     };
 
+    // Enumeration representing the available modes for the Arudpilot rover autopilot.
+    enum class APRoverMode {
```
julianoes, line 54: Usually we try the convention to lowercase acronyms like so: `ApRoverMode`, however, I assume you copied them from somewhere, so then maybe it's better leave em as is.

DaRavenox:
Heya, sorry about the number of changes. Had a chat with Seb and we decided on hiding the ardupilot mode behind flight_mode, so now I translate back and forth from it.

JonasVautherin (approved):

julianoes (approved):
Looks good to me. 

Any objections @JonasVautherin?

PR #1563, opened 2021-09-22
Add prepare() function to initialize camera plugin

JonasVautherin:
Depends on https://github.com/mavlink/MAVSDK-Proto/pull/255

The new `prepare()` function tries to fetch the camera definition and returns `Success` or `Error`. In case of `Error`, calling `prepare()` again will try again.

### DIFF
--- a/src/plugins/camera/camera_impl.cpp
+++ b/src/plugins/camera/camera_impl.cpp
@@ -120,6 +120,42 @@ void CameraImpl::deinit()
     _camera_found = false;
 }
 
+Camera::Result CameraImpl::prepare()
+{
+    auto prom = std::make_shared<std::promise<Camera::Result>>();
+    auto ret = prom->get_future();
+
+    prepare_async([&prom](Camera::Result result) { prom->set_value(result); });
+
+    return ret.get();
+}
+
+void CameraImpl::prepare_async(const Camera::ResultCallback& callback)
+{
+    auto temp_callback = callback;
+
+    std::lock_guard<std::mutex> lock(_information.mutex);
+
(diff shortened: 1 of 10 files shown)


src/plugins/camera/mocks/camera_mock.h
```diff
@@ -7,6 +7,7 @@ namespace testing {
 
 class MockCamera {
 public:
+    MOCK_CONST_METHOD0(prepare, Camera::Result()){};
```
julianoes, line 10: But there is currently no result?

src/plugins/camera/camera_impl.cpp
```diff
@@ -120,6 +120,14 @@ void CameraImpl::deinit()
     _camera_found = false;
 }
 
+void CameraImpl::prepare()
+{
+    // This is a trick to get mavsdk_server to initialize the camera plugin. Because mavsdk_server
+    // only initializes plugins when they are first used (LazyPlugin), and this may end up timing
+    // out on the first call of a camera function. By calling this before, one can ensure that the
+    // camera plugin is initialized early.
```
julianoes, line 128: Oh funny, you don't actually do anything in here. I thought you would kick off something.

JonasVautherin: Yeah I thought it was clever, but you're right. I changed it to actually return `Success` when the camera_definition has been downloaded.

JonasVautherin: And it will retry downloading the camera_definition every time you call `prepare()`.

julianoes (dismissed):

PR #1570, opened 2021-10-07
telemetry: fix target compid of two commands

julianoes:
If we send them to any component, then we might get an answer from someone else rather then the autopilot and get confused.

Found by @TSC21.


TSC21:
@julianoes should be `target_component_id`

JonasVautherin:
> It's what is defined on the MAVSDK API.

Oooh that's in the MAVSDK API! I was really confused by looking at the mavlink specs :laughing:. Thanks Nuno!

PR #1577, opened 2021-10-20
mavsdk_server: add API to set sysid/compid

julianoes:
What about something along these lines?

Instead of https://github.com/mavlink/MAVSDK-Proto/pull/256.


julianoes:
@JonasVautherin any idea what this is about:
https://github.com/mavlink/MAVSDK/runs/3953533027?check_suite_focus=true#step:7:442

PR #1581, opened 2021-10-25
Add custom action plugin

TSC21:
A custom action is an action that can be executed in a waypoint which is not standardized or handled by MAVLink and its microservices. This action is triggered by the autopilot with a command that gets sent to the onboard/mission computer, and the mission computer then handles the execution of that same action through the plugin.

On the reception of that command, the mission computer process running a MAVSDK instance with the _custom_action_ plugin reads the custom action ID it's supposed to execute, fetches the definition of the custom action sequence with that specific ID from a JSON file, and then loads the definition to the action to the client side, where it then executes a script, a command, or a sequence of staged scripts/commands. During this process, it sends also progress status updates to the autopilot so this is aware that the action is being processed - in case of a timeout, the autopilot is supposed to continue the mission.

For now. `MAV_CMD_WAYPOINT_USER_1` is being used as the command sent from the autopilot (PX4-enabled) with the information to start the action. param1 is filled with the Action ID while the param2 is filled with the Action Timeout. A follow-up to this implementation will look to renable and use https://github.com/mavlink/mavlink/pull/1516 with the appropriate modifications.

A high-level action can be composed by a sequence of stages which might result on a sequence of MAVLINK commands that can be sent to both the autopilot and/or to a specific payload, or the execution of scripts. An action can also just trigger a single global scripts, instead of being composed by stages. The transition between stages can be controlled by the `StateTransitionCondition` enum, which can go from a timeout, to a return value from a script itself. Similar conditions apply to consider an action as complete, valued by the `ActionCompleteCondition` enum.

An example JSON structure defining some actions can be found under `src/integration_tests/test_data/` dir, on the `custom_action.json` file.

The following sequence diagram explains the functionality of the plugin:
custom_action_plugin_sequence

Requires to be merged:

- [ ] Proto: https://github.com/mavlink/MAVSDK-Proto/pull/257
- [ ] PX4-Autopilot support: https://github.com/PX4/PX4-Autopilot/pull/18506


julianoes:
@TSC21 are you working on getting this into PX4 and MAVSDK? Otherwise, I'll probably have to close this soon. I'm in the process of cleaning up the PR and issues for the MAVSDK repos.

TSC21:
> @TSC21 are you working on getting this into PX4 and MAVSDK? Otherwise, I'll probably have to close this soon. I'm in the process of cleaning up the PR and issues for the MAVSDK repos.

Not at this point. I will need some time (and help) bringing this to the MAVLink spec.

julianoes:
@TSC21 ok, I'll close this for now and you can re-open it once it's accepted in mavlink.

Honestly, if we were to simplify it a bit, I'd be in favor of bringing it to the spec. I think there is value in it, however, in my opinion it should be simpler with just a 1 to 1 mapping between a mavlink custom command and one script with one return value. Right now there is logic for a state machine, parameters, multiple commands, etc. which I find quite complex and hard to grasp or explain.

PR #1596, opened 2021-11-04
Fix CI permissions when generating xcframework

julianoes:
Second try.


julianoes:
Doesn't work, it's a known issue: https://github.com/actions/upload-artifact/issues/38.

julianoes:
A workaround would be to upload the .zip in a .tar. Would that work @byuarus?

JonasVautherin:
> Doesn't work, it's a known issue: [actions/upload-artifact#38](https://github.com/actions/upload-artifact/issues/38).

aha! Good catch

julianoes:
Actually, given we use `actions/upload-release-asset@v1` for the upload as part of a release, this fix might still work, just not for the PR artifacts. So I'm going to try to merge this and make a release to test it.

PR #1597, opened 2021-11-04
camera: fix callback not called after formatting

julianoes:
This should fix the case where the camera_captured_image callback is not called after the storage has been formatted.

### DIFF
--- a/src/plugins/camera/camera_impl.cpp
+++ b/src/plugins/camera/camera_impl.cpp
@@ -1822,10 +1822,16 @@ void CameraImpl::format_storage_async(Camera::ResultCallback callback)
 
             receive_command_result(result, [this, callback](Camera::Result camera_result) {
                 if (camera_result == Camera::Result::Success) {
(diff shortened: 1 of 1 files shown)


julianoes:
@JonasVautherin not sure what CI is trying to say:
```
[ 99%] Building CXX object src/mavsdk_server/src/CMakeFiles/mavsdk_server.dir/grpc_server.cpp.o
Copying OS X content src/mavsdk_server/src/mavsdk_server.framework/Headers/mavsdk_server_api.h
Copying OS X content src/mavsdk_server/src/mavsdk_server.framework/Modules/module.modulemap
[100%] Linking CXX shared library mavsdk_server.framework/mavsdk_server
ld: malformed 32-bit x.y.z version number: 71778495.0.0
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

JonasVautherin (approved):
I hope the CI issue is one of those things that happens once in a thousand years :laughing:

PR #1599, opened 2021-11-04
camera: request dropped IMAGE_CAPTURED messages

julianoes:
This should fix holes when displaying all images that have been captured in a mission.

### DIFF
--- a/src/plugins/camera/camera_impl.cpp
+++ b/src/plugins/camera/camera_impl.cpp
@@ -69,10 +69,14 @@ void CameraImpl::init()
 
     _parent->add_call_every(
         [this]() { check_connection_status(); }, 0.5, &_check_connection_status_call_every_cookie);
+
+    _parent->add_call_every(
+        [this]() { request_missing_capture_info(); }, 0.5, &_request_missing_capture_info_cookie);
 }
 
 void CameraImpl::deinit()
 {
+    _parent->remove_call_every(_request_missing_capture_info_cookie);
     _parent->remove_call_every(_check_connection_status_call_every_cookie);
     _parent->remove_call_every(_status.call_every_cookie);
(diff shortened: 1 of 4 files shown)


src/plugins/camera/camera_impl.cpp
```diff
@@ -925,24 +929,59 @@ void CameraImpl::process_camera_image_captured(const mavlink_message_t& message)
...
+                LogDebug() << "Call user later for " << capture_info.index;
+                if (_capture_info.callback) {
+                    const auto temp_callback = _capture_info.callback;
+                    _parent->call_user_callback(
+                        [temp_callback, capture_info]() { temp_callback(capture_info); });
+                }
+                _capture_info.missing_image_indices.erase(it);
+            }
+
+            // Save captured indices that have been
```
JonasVautherin, line 962: that have been... missed?
```suggestion
            // Save captured indices that have been missed
```

JonasVautherin (approved):

src/plugins/camera/camera_impl.cpp
```diff
@@ -932,15 +936,62 @@ void CameraImpl::process_camera_image_captured(const mavlink_message_t& message)
...
+            }
+        }
+    }
+}
+
+void CameraImpl::request_missing_capture_info()
+{
+    std::lock_guard<std::mutex> lock(_capture_info.mutex);
+
+    for (auto it = _capture_info.missing_image_retries.begin();
```
JonasVautherin, line 977: I think you want a `while` here :sweat_smile:

julianoes: There is this exact pattern a couple of times in this codebase with `for` but a `while` would also work, you're right, and might make it clearer.

PR #1613, opened 2021-11-09
mission: fix jsoncpp cmake

julianoes:
This fixes the build on Arch Linux without SUPERBUILD.


src/plugins/mission/CMakeLists.txt
```diff
@@ -1,7 +1,3 @@
-
-hunter_add_package(jsoncpp)
```
bazfp, line 2: This are unneeded in mission?

julianoes: Json import was moved to mission_raw, so not needed here anymore.

JonasVautherin (approved):
Nice!

JonasVautherin (approved):

PR #1619, opened 2021-11-16
Fix failure enums (do not cast enum values directly)

JonasVautherin:
@potaito

### DIFF
--- a/src/plugins/failure/failure_impl.cpp
+++ b/src/plugins/failure/failure_impl.cpp
@@ -76,12 +76,74 @@ Failure::Result FailureImpl::inject(
     MavlinkCommandSender::CommandLong command{};
 
     command.command = MAV_CMD_INJECT_FAILURE;
-    command.params.maybe_param1 = static_cast<float>(failure_unit);
-    command.params.maybe_param2 = static_cast<float>(failure_type);
+    command.params.maybe_param1 = failure_unit_to_mavlink_enum(failure_unit);
+    command.params.maybe_param2 = failure_type_to_mavlink_enum(failure_type);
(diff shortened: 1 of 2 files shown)


potaito:
This did the trick. Motor failure injection is now working. Here's a draft of the example script that I used for testing

```python3
#!/usr/bin/env python3

import asyncio
from os import system
from mavsdk import System
from mavsdk.failure import FailureType, FailureUnit

async def run():
    # Init the drone
    drone = System()
    await drone.connect(system_address="udp://:14540")

    print("Waiting for drone to connect...")
    async for state in drone.core.connection_state():
        if state.is_connected:
            print(f"Drone discovered with UUID: {state.uuid}")
            break

    print("Enabling SYS_FAILURE_EN")
    await drone.param.set_param_int('SYS_FAILURE_EN', 1)
    
    print("Waiting for drone to have a global position estimate...")
    async for health in drone.telemetry.health():
        if health.is_global_position_ok and health.is_home_position_ok:
            print("Global position estimate ok")
            break

    print("-- Arming")
    await drone.action.arm()

    print("-- Taking off")
    await drone.action.takeoff()

    await asyncio.sleep(5)

    print("-- Injecting Motor failure")
    await drone.failure.inject(FailureUnit.SYSTEM_MOTOR, FailureType.OFF, instance=0)

if __name__ == "__main__":
    loop = asyncio.get_event_loop()
    loop.run_until_complete(run())
``` 

Prior to this PR, the value for the motor failure unit that MAVSDK-Python sent to the drone was 10. With this PR the value of the failure unit is now 101 as it should be.

PR #1621, opened 2021-11-17
Fix battery voltage

slaghuis:
Battery voltage is reported as 3.x volt on a 4S battery. This is typically one cell only.  Sum the voltage of all cells to obtain the total battery voltage.

### DIFF
--- a/src/plugins/telemetry/telemetry_impl.cpp
+++ b/src/plugins/telemetry/telemetry_impl.cpp
@@ -1076,7 +1076,13 @@ void TelemetryImpl::process_battery_status(const mavlink_message_t& message)
 
     Telemetry::Battery new_battery;
     new_battery.id = bat_status.id;
-    new_battery.voltage_v = bat_status.voltages[0] * 1e-3f;
+    new_battery.voltage_v = 0.0f;
+    for (int i = 0; i < 10; i++)
+    {
+        if (bat_status.voltages[i] == UINT16_MAX)
+            break;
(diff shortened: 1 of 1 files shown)


src/plugins/telemetry/telemetry_impl.cpp
```diff
@@ -1076,7 +1076,13 @@ void TelemetryImpl::process_battery_status(const mavlink_message_t& message)
 
     Telemetry::Battery new_battery;
     new_battery.id = bat_status.id;
-    new_battery.voltage_v = bat_status.voltages[0] * 1e-3f;
+    new_battery.voltage_v = 0.0f;
+    for (int i = 0; i < 10; i++)
```
JonasVautherin, line 1080: Nitpick: since the battery id is a `uint8_t`, it means that MAVLink supports up to 255 batteries, right?

Wouldn't it make sense to go up to 255 instead of 10 here? I know it's highly unlikely that it ever goes higher than 10, but with 255 we just follow the MAVLink specs. And because we `break` after the first `UINT16_MAX`, it doesn't add any override.

What do you think?

JonasVautherin (changes requested):
Thanks a lot!

src/plugins/telemetry/telemetry_impl.cpp
slaghuis: That is a valid suggestion Jonas.  It is always good to stay within the specs.  It will also make the code more future proof.

JonasVautherin (approved):

PR #1623, opened 2021-11-17
Use modern numeric limits

JonasVautherin:


julianoes (approved):
So verbose but I like it.

PR #1627, opened 2021-11-19
Soft remove gimbal v1 and always use v2

### LINKED ISSUE #1625: Absence of mutex locking in Gimbal plugin causing SEGV_ACCERR
While working on a new plugin mechanism for MAVSDK-Java I am getting the following error on Android: `Fatal signal 11 (SIGSEGV), code 2 (SEGV_ACCERR)`

This error comes only when the Gimbal plugin is enabled, and it signifies that two or more threads are mutating a variable without any locking. And I noticed that mutex locking is missing here:

https://github.com/mavlink/MAVSDK/blob/631f6d366a5683b773560d8abe9752b54afe98fa/src/plugins/gimbal/gimbal_impl.cpp#L68

I am not too well-versed in C++ and I am not sure if this is by design, but please look into this issue.

JonasVautherin:
* Add target_component_id to `RequestMessage`
* Default to gimbal v2 in mission plugin
* Default to gimbal v2 in gimbal plugin, requiring a newly-added `prepare()` function to detect the remote gimbal manager

Currently, both the mission plugin and the gimbal plugin detect the protocol version separately, which can end up in one going for v2 and the other going for v1. Also MAVLink cannot properly detect the protocol version and has to guess, falling back to v1 in case of timeout.

The big question here is whether or not we can drop gimbal protocol v1 in MAVSDK. This PR does not completely remove it, with the idea that we could try for a while and see if it impacts many users before completely removing the entire v1 code.

Resolves #1625 (supposedly).

### DIFF
--- a/src/plugins/gimbal/gimbal_impl.cpp
+++ b/src/plugins/gimbal/gimbal_impl.cpp
@@ -40,29 +40,37 @@ Gimbal::Result GimbalImpl::prepare()
     return ret.get();
 }
 
-void GimbalImpl::prepare_async(const Gimbal::ResultCallback& callback) _parent->request_message()
-    .request(
-        MAVLINK_MSG_ID_GIMBAL_MANAGER_INFORMATION,
-        [this, callback](MavlinkCommandSender::Result result, const mavlink_message_t& message) {
-            if (result == MavlinkCommandSender::Result::Success) {
-                mavlink_gimbal_manager_information_t gimbal_manager_information;
-                mavlink_msg_gimbal_manager_information_decode(
-                    &message, &gimbal_manager_information);
+void GimbalImpl::prepare_async(const Gimbal::ResultCallback& callback)
+{
+    if (_gimbal_protocol != nullptr) {
+        _parent->call_user_callback([callback]() { callback(Gimbal::Result::Success); });
+    } else {
+        _parent->request_message().request(
+            MAVLINK_MSG_ID_GIMBAL_MANAGER_INFORMATION,
+            [this,
+             callback](MavlinkCommandSender::Result result, const mavlink_message_t& message) {
+                if (result == MavlinkCommandSender::Result::Success) {
+                    mavlink_gimbal_manager_information_t gimbal_manager_information;
+                    mavlink_msg_gimbal_manager_information_decode(
+                        &message, &gimbal_manager_information);
 
-                LogDebug() << "Gimbal manager information for gimbal device "
-                           << static_cast<int>(gimbal_manager_information.gimbal_device_id)
-                           << " was discovered";
+                    LogDebug() << "Gimbal manager information for gimbal device "
+                               << static_cast<int>(gimbal_manager_information.gimbal_device_id)
+                               << " was discovered";
 
-                _gimbal_protocol.reset(new GimbalProtocolV2(
-                    *_parent, gimbal_manager_information, message.sysid, message.compid));
+                    _gimbal_protocol.reset(new GimbalProtocolV2(
+                        *_parent, gimbal_manager_information, message.sysid, message.compid));
 
-                _parent->call_user_callback([callback]() { callback(Gimbal::Result::Success); });
-            } else {
-                _parent->call_user_callback(
-                    [callback, result]() { callback(gimbal_result_from_command_result(result)); });
-            }
-        },
-        0); // any component
+                    _parent->call_user_callback(
+                        [callback]() { callback(Gimbal::Result::Success); });
+                } else {
+                    _parent->call_user_callback([callback, result]() {
+                        callback(gimbal_result_from_command_result(result));
+                    });
+                }
+            },
+            0); // any component
+    }
 } // namespace mavsdk
 
 void GimbalImpl::disable()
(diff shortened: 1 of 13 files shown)


JonasVautherin:
Interestingly, when using `gimbal.prepare()` with AuterionOS, I get the following:

```
[12:17:10|Warn ] command denied (512). (mavlink_commands.cpp:179)
[12:17:10|Warn ] Received ack for not-existing command: 512! Ignoring... (mavlink_commands.cpp:246)
```

Though it prepares correctly and the gimbal control works after that...

JonasVautherin:
Added some debugging out of curiosity:

```
[12:35:20|Debug] Gimbal manager information for gimbal device 0 was discovered at 1:1 (gimbal_impl.cpp:57)
[12:35:20|Warn ] command (#512) denied by 1:100 (mavlink_commands.cpp:179)
[12:35:20|Warn ] Received ack from 1/1 for not-existing command: 512! Ignoring... (mavlink_commands.cpp:239)
```

So component 1:100 replies "DENIED", I guess to say that it does not support that message (why not "UNSUPPORTED" then?), which makes MAVSDK "consume" the ACK. Right after that, 1:1 sends an ACK that supposedly corresponds to the gimbal manager information it sent above.

I don't think there is a proper fix for that, it feels like it's all expected in the mavlink world. Though what it implies is that the `RequestMessage` is nice, but it can't be solely relied upon (say the DENIED message had been received before the gimbal manager information message, then my `prepare()` function above would have been screwed, right?).

divyanshu1234:
I tested it just not using MAVSDK-Java. I'm not getting a crash a before but now there's this message:

```
[09:35:38|Error] Gimbal plugin not prepared! Call `prepare()` first! (gimbal_impl.cpp:244)
```

src/plugins/gimbal/gimbal_impl.cpp
```diff
@@ -24,162 +24,229 @@ GimbalImpl::~GimbalImpl()
...
+void GimbalImpl::prepare_async(const Gimbal::ResultCallback& callback)
 {
-    // We did not receive a GIMBAL_MANAGER_INFORMATION in time, so we have to
-    // assume Version2 is not available.
-    LogDebug() << "Falling back to Gimbal Version 1";
-    _gimbal_protocol.reset(new GimbalProtocolV1(*_parent));
-    _protocol_cookie = nullptr;
-}
+    if (_gimbal_protocol != nullptr) {
+        _parent->call_user_callback([callback]() { callback(Gimbal::Result::Success); });
```
julianoes, line 46:
```suggestion
        if (callback) {
            _parent->call_user_callback([callback]() { callback(Gimbal::Result::Success); });
        }
```

src/plugins/gimbal/gimbal_impl.cpp
```diff
@@ -24,162 +24,229 @@ GimbalImpl::~GimbalImpl()
...
+
+                    LogDebug() << "Gimbal manager information for gimbal device "
+                               << static_cast<int>(gimbal_manager_information.gimbal_device_id)
+                               << " was discovered";
+
+                    _gimbal_protocol.reset(new GimbalProtocolV2(
+                        *_parent, gimbal_manager_information, message.sysid, message.compid));
+
+                    _parent->call_user_callback(
+                        [callback]() { callback(Gimbal::Result::Success); });
```
julianoes, line 65: Also check for `callback != nullptr` here.

src/plugins/gimbal/gimbal_impl.cpp
```diff
@@ -24,162 +24,229 @@ GimbalImpl::~GimbalImpl()
...
+                } else {
+                    _parent->call_user_callback([callback, result]() {
+                        callback(gimbal_result_from_command_result(result));
```
julianoes, line 68: And here.

src/plugins/gimbal/gimbal_impl.cpp
```diff
@@ -24,162 +24,229 @@ GimbalImpl::~GimbalImpl()
...
 {
-    wait_for_protocol_async([=]() {
-        _gimbal_protocol->set_pitch_rate_and_yaw_rate_async(
-            pitch_rate_deg_s, yaw_rate_deg_s, callback);
-    });
+    if (_gimbal_protocol == nullptr) {
+        _parent->call_user_callback([callback]() {
+            LogErr() << "Gimbal plugin not prepared! Call `prepare()` first!";
+            callback(Gimbal::Result::Error);
+        });
```
julianoes, line 123: And here. But place the `LogErr()` outside of the callback.

And the same below.

src/plugins/mission/mission_impl.h
```diff
@@ -167,9 +163,8 @@ class MissionImpl : public PluginImplBase {
 
     static constexpr double RETRY_TIMEOUT_S = 0.250;
 
-    void* _gimbal_protocol_cookie{nullptr};
     enum class GimbalProtocol { Unknown, V1, V2 };
-    std::atomic<GimbalProtocol> _gimbal_protocol{GimbalProtocol::Unknown};
+    std::atomic<GimbalProtocol> _gimbal_protocol{GimbalProtocol::V2};
```
julianoes, line 167: Yer, I think that's reasonable like that.

julianoes (commented):
I think I'm ok with this PR. Just a few more checks if the callback is not nullptr, and than this is good.

julianoes:
@JonasVautherin I forgot what we decided to do about this.

JonasVautherin:
I think it won't work consistently because other components can deny the request if they want to, so the `prepare()` function may fail. That's where I stopped at least.

julianoes:
Ok, so what do you suggest to do? :man_shrugging:

PR #1628, opened 2021-11-19
New vehicle action 

rligocki:


JonasVautherin (commented):
What are those cmake.yml for? :thinking:

julianoes:
@rligocki should we try another effort to get this in?

rligocki:
Let's do it. But I will need little bit of guidance.

julianoes:
@rligocki I've squashed all your commits into one, and removed the files you had added by accident. We now need to add the actual implementation back in. I suggest you fetch this state and add a commit on top with the implementation.

The way to get to this state (without ugly merges) would be:

The steps below assume that `origin` is your fork:

```
git stash # to get rid of anything dirty in your tree, in case you need it later
git fetch origin
git checkout main
git reset --hard origin/main
```

This should wipe your changes and reset to the state of this PR. Note that it does not have any of the actual implementation here. So if you have that one somewhere locally, make sure to stash it or put it into a branch to use it later.

rligocki:
Now it should contain implementation and tests. This should work without any problem for 1.12 and 1.11 version of PX4. Only update of proto needs to be done.

julianoes:
@rligocki this is still conflicting, see what GitHub says: :arrow_down:.

rligocki:
CI did not passed. Is it possible that it is caused by merge requests, that has been done?

rligocki:
Based on latest CI from main branch, it looks like PX4 1.12 tests failed because of latest commits in main branch.

src/integration_tests/mission_takeoff_land.cpp
```diff
@@ -0,0 +1,180 @@
...
+    new_item.acceptance_radius_m = 1;
+    new_item.vehicle_action = Mission::MissionItem::VehicleAction::Takeoff;
+    mission_plan.mission_items.push_back(new_item);
+
+    new_item.latitude_deg = 47.398300;
+    new_item.vehicle_action = Mission::MissionItem::VehicleAction::Land;
+    mission_plan.mission_items.push_back(new_item);
+
+    new_item.latitude_deg = 47.398500;
+    new_item.vehicle_action = Mission::MissionItem::VehicleAction::TransitionToFw;
```
julianoes, line 102: I don't understand this action here. So we landed and now we're transitioning to fixedwing on ground? I don't think that's something PX4 supports. Usually, you need to takeoff and land in multicopter mode and then transition in air.

Also, if this test is made for a VTOL it needs to be named accordingly, like this:
https://github.com/mavlink/MAVSDK/blob/9b4b79cbdfca361fab9a528ba78e81376503cbd0/src/integration_tests/action_transition_multicopter_fixedwing.cpp#L10

src/mavsdk/plugins/mission/mission_impl.cpp
```diff
@@ -554,6 +587,70 @@ MissionImpl::convert_to_int_items(const std::vector<MissionItem>& mission_items)
...
+                    break;
+            }
+
+            const int32_t x = int32_t(std::round(item.latitude_deg * 1e7));
+            const int32_t y = int32_t(std::round(item.longitude_deg * 1e7));
+            float z = item.relative_altitude_m;
+            MAV_FRAME frame = MAV_FRAME_GLOBAL_RELATIVE_ALT_INT;
+
+            if (command == MAV_CMD_NAV_LAND) {
+                z = 0;
```
julianoes, line 628: I wonder what happens if the landing spot is below the takeoff/home altitude. I think PX4 just ignores this and lands anyway.

src/mavsdk/plugins/mission/mission_impl.cpp
```diff
@@ -630,8 +728,14 @@ std::pair<Mission::Result, Mission::MissionPlan> MissionImpl::convert_to_result_
                 new_mission_item.relative_altitude_m = int_item.z;
                 new_mission_item.yaw_deg = int_item.param4;
 
-                new_mission_item.is_fly_through = !(int_item.param1 > 0);
-                new_mission_item.acceptance_radius_m = int_item.param2;
+                if (int_item.command == MAV_CMD_NAV_TAKEOFF) {
+                    new_mission_item.acceptance_radius_m = 1;
+                    new_mission_item.is_fly_through = 0;
```
julianoes, line 733:
```suggestion
                    new_mission_item.is_fly_through = false;
```

julianoes (changes requested):

src/mavsdk/plugins/mission/mission_impl.cpp
julianoes: Actually, this altitude should be an indication of the landing altitude, so presumably we should use the altitude field here.

src/integration_tests/mission_takeoff_land.cpp
rligocki: This is mistake. There should be Takeoff action. Will fix it.

src/mavsdk/plugins/mission/mission_impl.cpp
rligocki: Will fix it.

src/mavsdk/plugins/mission/mission_impl.cpp
rligocki: So you suggest to assign min value of float to z variable, so that vehicle will descent until it will reach ground?

julianoes: Hm, so the altitude set here is the altitude of the waypoint before landing, I see. In that case we really don't know the landing altitude and 0 is probably the best guess. I'd say we leave it as is then.

rligocki: Ok

rligocki:
Now it should be ok.

julianoes:
@rligocki CI seems to fail the new test, any idea?

```
[  FAILED  ] SitlTest.MissionTakeoffAndLand
```
https://github.com/mavlink/MAVSDK/runs/5014636341?check_suite_focus=true#step:4:47103

julianoes:
Looks good. I'll merge it next time I go through all PRs.

julianoes:
Rebased and force pushed.

JonasVautherin (dismissed):

julianoes:
Thanks @rligocki and sorry it took a while.

PR #1631, opened 2021-11-23
core: only request autopilot version if autopilot

dlech:
This adds a check to only request the autopilot version on connection if the connection is actually an autopilot. This message fails, e.g. when connecting to a ground control station, which causes extra noise in the logs and extra mavlink messages that are just one more thing that has to be filtered out when debugging other problems.


JonasVautherin (approved):
Makes sense to me. Thanks!

julianoes (approved):

julianoes:
Thanks @dlech!

PR #1632, opened 2021-11-24
mission: remove TRIGGER_CONTROL

julianoes:
It turns out that you can enable distance-based triggering just using the command SET_CAM_TRIGG_DIST and you don't have to explicitly start and stop the triggering with TRIGGER_CONTROL.

This makes the logic simpler, removes some overhead of mission items having to be sent, and moves it closer to the implementation of QGroundControl.

FYI @RomanBapst

### DIFF
--- a/src/plugins/mission/mission_impl.cpp
+++ b/src/plugins/mission/mission_impl.cpp
@@ -407,35 +407,6 @@ MissionImpl::convert_to_int_items(const std::vector<MissionItem>& mission_items)
         }
 
         if (item.camera_action != CameraAction::None) {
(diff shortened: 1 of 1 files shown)


JonasVautherin:
So the fact that the parameter is set to a non-zero value also makes it take pictures (and stop when it is set back to 0)? And is that the behavior that MAVLink wants?

JonasVautherin (approved):

RomanBapst (dismissed):

julianoes:
@JonasVautherin correct!

JonasVautherin (approved):

PR #1634, opened 2021-11-25
mission: fail if mission item is unsupported

julianoes:
These two breaks were missing. If an item is unsupported we have to exit the parsing and return the result instead of continuing.


JonasVautherin (approved):
Good catch!

PR #1639, opened 2021-11-28
Merging plugin libraries into one library, directory restructure

julianoes:
This is another take at merging the plugin libraries into the main mavsdk library.

This would have a couple of benefits:

- Easier Python direct wrappers.
- Easier linking to libmavsdk, not needing to link to every plugin required.
- Making API/ABI versioning clearer in the future.

Replaces https://github.com/mavlink/MAVSDK/pull/1332.


JonasVautherin (approved):

julianoes:
@JonasVautherin I wonder if we should change the include paths.

Now it is:
```
#include <mavsdk/mavsdk.h>
#include <mavsdk/plugins/shell/shell.h>
```

But I would consider:
```
#include <mavsdk/mavsdk.h>
#include <mavsdk/action/action.h>
```

Or even:
```
#include <mavsdk/mavsdk.h>
#include <mavsdk/action.h>
```

And while at it I would want to swap to `.hpp`.
```
#include <mavsdk/mavsdk.hpp>
#include <mavsdk/action.hpp>
```

JonasVautherin:
This one seems fine to me:

```
#include <mavsdk/mavsdk.h>
#include <mavsdk/action.h>
```

I really don't have any opinion about `.h` or `.hpp`, I just don't mind :laughing:

julianoes:
We could also move the header files but create shim header files to forward to the new ones.

JonasVautherin:
> We could also move the header files but create shim header files to forward to the new ones.

Your call, I don't mind if we just move the headers. Anyway this will break the users' build system, so they'll have t make some changes, and at this point updating the headers does not feel like a big deal to me.

JonasVautherin (approved):

PR #1640, opened 2021-11-30
telemetry: use SYS_STATUS to determine calibration

julianoes:
PX4 by now sends the calibration state out using SYS_STATUS. This is to be preferred as it updates after a calibration unlike the params which are only fetched on startup.

Hopefully in the future, we can deprecate support for older PX4 versions and get rid of these params completely.


JonasVautherin (approved):
Great! :rocket:

PR #1646, opened 2021-12-07
Ignore photos with empty file_url

JonasVautherin:
It's a MAVLink feature to sometimes send photo information with `file_url=""`, but I don't know a use-case for it yet. So for now MAVSDK will just ignore it.


julianoes (commented):
What if a camera informs about the fact that it captured an image but there is not URL for it, for instance because it's on an SD card and can't be downloaded?

JonasVautherin:
You're right. But then I don't think I can make a difference, can I? :thinking:

julianoes:
I'm afraid not. I would leave it as is.

PR #1647, opened 2021-12-07
added mavsdk_connection

persuader72:
This pull request adds Mavsdk::add_mavsdk_connection function. That function permit to share the connection of another mavsdk instance. This function is useful if you want to impersonate different devices on the same connection.

For example if you want to make a single application that act as a camera, a gimbal and a client companion computer sharing that same autopilot serial connection.

### DIFF
--- a/src/core/connection.cpp
+++ b/src/core/connection.cpp
@@ -2,6 +2,7 @@
 
 #include <memory>
 #include <utility>
+#include <vector>
 #include "mavsdk_impl.h"
 #include "mavlink_channels.h"
 
@@ -10,10 +11,10 @@ namespace mavsdk {
 std::atomic<unsigned> Connection::_forwarding_connections_count = 0;
 
 Connection::Connection(receiver_callback_t receiver_callback, ForwardingOption forwarding_option) :
-    _receiver_callback(std::move(receiver_callback)),
     _mavlink_receiver(),
     _forwarding_option(forwarding_option)
 {
+    _receiver_callback[0] = std::move(receiver_callback);
     // Insert system ID 0 in all connections for broadcast.
     _system_ids.insert(0);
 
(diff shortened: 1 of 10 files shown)


julianoes:
Interesting! Thanks for the contribution.

Could you alternatively use message forwarding?
https://github.com/mavlink/MAVSDK/blob/21ac1ec4f4baabb278d774f6ec2d14a407041a64/src/core/mavsdk.h#L90

persuader72:
Yes, I honestly not thought about this solution

The following diagram shows how the problem can be solved with two kinds of solutions.
github drawio

So I think that a very similar solution can be reached with the Forwarding option
The only difference is that with the connection sharing solution you don't have to make two additional mock network connections inside the same software.

At this point, I don't know if the new solution can be useful anyway or not.

julianoes:
My aim is to keep it as simple as possible, and only have one way to do a function whenever possible, so that not multiple need to be maintained and tested.

Would you mind trying if you can make your use case work with forwarding?

Actually, there is one more way that you could try if you wanted which is to tie it all together using mavlink-router.

If you can make it work with either of my suggestions, then I'd prefer to go without your suggested changes.

julianoes:
@persuader72 were you able to come up with a solution for your problem given my suggestions, or do you think this PR is still worthwhile?

persuader72:
@julianoes sorry for the delay, yes the solution is working fine. 

The only problem is that if I add a second connection when the first Is already active (for example in the on_new_system event when the autopilot is discovered)  the mavsdk keeps logging these messages until the remote end of the second connection becomes active.

[09:11:43|Error] Message forwarding failed (mavsdk_impl.cpp:159)
[09:11:43|Error] No known remotes (udp_connection.cpp:134)

There are very fast. I seem that this message is printed for every message the autopilot has sent. It looks like that if there is a second connection with the forwarding option but the remote end is not connected an error message is logged for every message incoming in the first connection.

When the remote end of the second connection becomes online those messages disappear.
 
As a suggestion I think Would be better to to lower the priority of this message from Error to Debug because the case of a connection with a remote end disconnected is not an error but a connection state.

Regards
Stefano

julianoes:
Oh that's a good observation. I think best would be to ignore these messages in the case you're describing. Would you mind making a PR?

PR #1648, opened 2021-12-08
Publish mission current N on mission completion

bazfp:
Previously I was erroneously resetting mission current to zero on completion, this resulted in strange double publishing of mission current.

Correct path is to publish mission current N at the end of a mission and not reset. Mission upload will reset current to 0.

### DIFF
--- a/src/plugins/mission_raw_server/mission_raw_server_impl.cpp
+++ b/src/plugins/mission_raw_server/mission_raw_server_impl.cpp
@@ -311,8 +311,9 @@ MissionRawServer::MissionItem MissionRawServerImpl::current_item_changed() const
 
 void MissionRawServerImpl::set_current_item_complete()
 {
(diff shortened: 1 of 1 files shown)


src/plugins/mission_raw_server/mission_raw_server_impl.cpp
```diff
@@ -311,24 +311,28 @@ MissionRawServer::MissionItem MissionRawServerImpl::current_item_changed() const
 
 void MissionRawServerImpl::set_current_item_complete()
 {
+    if (_current_seq + 1 > _current_mission.size())
+        return;
```
julianoes, line 315:
```suggestion
    if (_current_seq + 1 > _current_mission.size()) {
        return;
    }
```

julianoes (commented):
Other than a nitpick, that's ok.

JonasVautherin (approved):

PR #1650, opened 2021-12-09
mavlink_message_handler: use std::map instead of std::vector

dayjaby:
If you have a lot of plugins activated, you naturally have a lot of mavlink message handlers (potentially even multiple handlers for the same ID). With this PR, we do not iterate through all message handlers to find the one that can handle the current messsage ID. Instead we use a std::map to have a mapping message_id -> handlers in an efficient way.

### DIFF
--- a/src/core/mavlink_message_handler.cpp
+++ b/src/core/mavlink_message_handler.cpp
@@ -8,28 +8,21 @@ void MAVLinkMessageHandler::register_one(
 {
     std::lock_guard<std::mutex> lock(_mutex);
 
-    Entry entry = {msg_id, callback, cookie};
-    _table.push_back(entry);
+    if (_map.count(msg_id) == 0) {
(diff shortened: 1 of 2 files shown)


julianoes:
Thanks for the pull request @dayjaby. Now when it comes to optimization it should be based on profiling. Did you gather any evidence that this would be faster?

I'm asking because from what I have learnt is that a vector tends to be the fastest option for low numbers, e.g. below 50 entries. That's because a vector is laid out consecutively in memory and can be nicely cached.

Therefore, I would want to see proof that using a map here is actually more efficient and/or faster.

And actually it would be really nice if we had some profiling available for some "use case". I'm just not so sure how to go about it because it's mostly defined by IO. Maybe we would have to make a test of delay between message in until the callback is called?

julianoes:
I'm closing this for now. Once we have a good benchmark of some "real world" usage and can measure CPU and/or latency I'd reconsider this.

PR #1654, opened 2021-12-23
ftp_client: fix component ID comment

dayjaby:
```Server component id is for example 1 for autopilot or 240 for companion computer.```

This sentence was misleading, as Mavsdk::Configuration::UsageType::CompanionComputer in ftp_server results in a component ID of 195 (MAV_COMP_ID_PATHPLANNER).
So I ended up trying component ID 191 (MAV_COMP_ID_ONBOARD_COMPUTER) and component ID 240 (as the comment suggests), which both did not work.

Does it make sense to invent new names for component IDs though? Why not stick to the same names as in mavlink and have a nice 1-to-1 relationship (OnboardComputer = MAV_COMP_ID_ONBOARD_COMPUTER, etc.)

### DIFF
--- a/examples/ftp_client/ftp_client.cpp
+++ b/examples/ftp_client/ftp_client.cpp
@@ -16,35 +16,35 @@ using std::chrono::seconds;
 
 void usage(const std::string& bin_name)
 {
-    std::cerr << "Usage : " << bin_name
-              << " <connection_url> <server component id> <command> <parameters>\n"
-              << '\n'
-              << "Connection URL format should be :\n"
-              << " For TCP : tcp://[server_host][:server_port]\n"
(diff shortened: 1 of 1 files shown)


julianoes:
> Does it make sense to invent new names for component IDs though? Why not stick to the same names as in mavlink and have a nice 1-to-1 relationship (OnboardComputer = MAV_COMP_ID_ONBOARD_COMPUTER, etc.)

Yes I think we should try to clear that up too.

JonasVautherin (approved):

julianoes:
Please fix style: `./tools/fix_style.sh .`

julianoes (approved):

JonasVautherin (approved):

PR #1655, opened 2021-12-23
examples: add camera_manager

dayjaby:
Tested with

```cmake -Bbuild -H. && cmake --build build -j4 && ./build/camera_manager .```

and QGroundControl version v4.2.0, which shows the camera control interface in QGC:
Screenshot from 2021-12-23 14-02-43

### CODE REVIEW

**Review by JonasVautherin — COMMENTED** [review]
Nice! Quick question: how does the client know the ip:port of the mavlinkftp server? The camera_information message says `mavlinkftp:///infos/camera_info.xml`. Is that specified in MAVLink somehow? Like I could imagine `mavlinkftp://192.168.1.12:24547/infos/camera_info.xml`...

**Comment by dayjaby** [comment]
@JonasVautherin It's just about sys ids and comp ids, not about ip addresses/ports. QGroundControl receives a HEARTBEAT from this mavlink camera manager component (with MAV_COMP_ID_CAMERA = 100). Upon receiving a heartbeat from any camera component, QGC will send a request to this component to retrieve the CAMERA_INFORATMION, which contains the `cam_definition_uri`. From exactly this camera component (same mavlink sysid/compid), QGC will start to retrieve the camera definition file via MAVLink FTP.

There is some inconsistency regarding "mavlinkftp" vs "mftp". In some commits I've seen mavlinkftp, but the mavlink documentation specifies "mftp" (https://mavlink.io/en/services/camera.html) and you can find that as well in QGroundControl: https://github.com/mavlink/qgroundcontrol/commit/4b50ffd48b1588d9e7f3a35779a5cb207384f896

Maybe it's working for me just because it still had some old data being cached, so would be good if someone else tests it (:

So basically this PR allows MAVSDK to implement the camera side of the mavlink camera protocol. There is lots of room for improvements: e.g. storing the parameters in a local file, publish all camera parameters on boot as PARAM_EXT, implement some streaming pipelines. The list is long :D

**Comment by dayjaby** [comment]
My first approach was invalid, as I first tested it against a camera_definition_uri with http, which made QGroundControl preserve that in its cache. So the mavlinkftp did not work at all, as QGroundControl never used mavlink ftp for CAMERA_INFORMATION in the first place (a PR that fixes this issue in QGC is mentioned above).

The correct URL at the moment is something like `mftp://[;comp=100]infos/camera_info.xml`.

**Comment by dayjaby** [comment]
Probably going to turn this into an own plugin, as this would allow to react to parameter changes quite easily:

```
void CameraManagerPluginImpl::init()
{
    _parent->provide_server_param_int("CAM_EV", 5);

    _parent->provide_server_param_int("CAM_ISO", 0);
    _parent->subscribe_param<uint32_t>(
        "CAM_ISO", [](uint32_t iso) { std::cout << "NEW ISO: " << iso << std::endl; }, nullptr);
}
```

Just not sure how to handle the different typing yet. E.g. one could end up using float for CAM_ISO or uint32 or int32 or whatever.

File: examples/camera_manager/camera_manager.cpp [hunk 1]
```diff
@@ -0,0 +1,124 @@
...
+        usage(argv[0]);
+        return 1;
+    }
+
+    Mavsdk mavsdk;
+    Mavsdk::Configuration configuration(Mavsdk::Configuration::UsageType::Camera);
+    mavsdk.set_configuration(configuration);
+    ConnectionResult connection_result = mavsdk.add_any_connection("udp://:24547");
+    if (connection_result != ConnectionResult::Success) {
+        std::cerr << "Error setting up Mavlink FTP server.\n";
```
julianoes, line 41:
```suggestion
        std::cerr << "Error setting up Mavlink camera manager.\n";
```

examples/camera_manager/camera_manager.cpp
```diff
@@ -0,0 +1,124 @@
...
+    auto mavlink_passthrough = MavlinkPassthrough{system};
+    mavlink_passthrough.subscribe_message_async(
+        MAVLINK_MSG_ID_COMMAND_LONG, [&](const mavlink_message_t& msg) {
+            mavlink_command_long_t cmd;
+            mavlink_msg_command_long_decode(&msg, &cmd);
+            switch (cmd.command) {
+                case MAV_CMD_REQUEST_CAMERA_INFORMATION:
+                    mavlink_camera_information_t camera_information;
+                    camera_information.time_boot_ms = 0;
+                    strncpy((char*)camera_information.vendor_name, "Foo Industries", 32);
```
julianoes, line 83:
```suggestion
                    strncpy((char*)camera_information.vendor_name, "Foo Industries", sizeof(camera_information.vendor_name));
```

examples/camera_manager/camera_manager.cpp
```diff
@@ -0,0 +1,124 @@
...
+                    strncpy((char*)camera_information.model_name, "T100", 32);
```
julianoes, line 84:
```suggestion
                    strncpy((char*)camera_information.model_name, "T100", sizeof(camera_information.model_name));
```

examples/camera_manager/camera_manager.cpp
```diff
@@ -0,0 +1,124 @@
...
+                    camera_information.resolution_h = 5456;
+                    camera_information.resolution_v = 3632;
+                    camera_information.flags =
+                        CAMERA_CAP_FLAGS_CAPTURE_VIDEO | CAMERA_CAP_FLAGS_CAPTURE_IMAGE |
+                        CAMERA_CAP_FLAGS_HAS_MODES |
+                        CAMERA_CAP_FLAGS_CAN_CAPTURE_IMAGE_IN_VIDEO_MODE |
+                        CAMERA_CAP_FLAGS_CAN_CAPTURE_VIDEO_IN_IMAGE_MODE |
+                        CAMERA_CAP_FLAGS_HAS_BASIC_ZOOM | CAMERA_CAP_FLAGS_HAS_BASIC_FOCUS |
+                        CAMERA_CAP_FLAGS_HAS_VIDEO_STREAM;
+                    camera_information.cam_definition_version = 0;
```
julianoes, line 98: I think this starts at 1.

examples/camera_manager/camera_manager.cpp
```diff
@@ -0,0 +1,124 @@
...
+                    strncpy(
+                        camera_information.cam_definition_uri,
+                        "mftp://[;compid=100]infos/camera_info.xml",
```
julianoes, line 101: What is this weird `[;compid=100]`? Is it in the specs?

examples/camera_manager/infos/camera_info.xml
```diff
@@ -0,0 +1,664 @@
...
+                            <roption name="1/30" value="0.033333" />
+                            <roption name="1/60" value="0.016666" />
+                            <roption name="1/125" value="0.008" />
+                            <roption name="1/250" value="0.004" />
+                            <roption name="1/500" value="0.002" />
+                            <roption name="1/1000" value="0.001" />
+                            <roption name="1/2000" value="0.0005" />
+                            <roption name="1/4000" value="0.00025" />
+                            <roption name="1/8000" value="0.000125" />
+                        </parameterrange>
```
julianoes, line 388: Do you need all these video formats just for an example? I would suggest to simplify a bit.

src/mavsdk/plugins/ftp/ftp_impl.cpp
```diff
@@ -36,6 +36,7 @@ FtpImpl::~FtpImpl()
 
 void FtpImpl::init()
 {
+    LogDebug() << "register mavlink message handler";
```
julianoes, line 39:
```suggestion
```

src/mavsdk/plugins/ftp/ftp_impl.cpp
```diff
@@ -774,11 +776,8 @@ void FtpImpl::process_mavlink_ftp_message(const mavlink_message_t& msg)
     if (payload->size > max_data_length) {
         error_code = ServerResult::ERR_INVALID_DATA_SIZE;
     } else {
-        /*
-            LogDebug() << "ftp - opc: " << (int)payload->opcode << " size: "
-                << (int)payload->size << " offset: " << (int)payload->offset << " seq: " <<
-        payload->seq_number;
-        */
+        LogDebug() << "ftp - opc: " << (int)payload->opcode << " size: " << (int)payload->size
```
julianoes, line 779: debugging...

src/mavsdk/plugins/ftp/ftp_impl.cpp
```diff
@@ -1306,12 +1309,99 @@ FtpImpl::ServerResult FtpImpl::_work_calc_file_CRC32(PayloadHeader* payload)
...
+            }
+        }
+    }
+
+    if (error_code != ServerResult::SUCCESS) {
+        payload->opcode = RSP_NAK;
+        payload->size = 1;
+        uint8_t* pData = &payload->data[0];
+        *pData =
+            error_code; // Straight reference to data[0] is causing bogus gcc array subscript error
```
julianoes, line 1386: What?

src/mavsdk/plugins/ftp/ftp_impl.cpp
```diff
@@ -1306,12 +1309,99 @@ FtpImpl::ServerResult FtpImpl::_work_calc_file_CRC32(PayloadHeader* payload)
...
+    ftp_msg.target_system = _session_info.stream_target_system_id;
+    ftp_msg.target_network = 0;
+    ftp_msg.target_component = _get_target_component_id();
+    mavlink_message_t msg;
+    mavlink_msg_file_transfer_protocol_encode(
+        _parent->get_own_system_id(), _parent->get_own_component_id(), &msg, &ftp_msg);
+    _parent->send_message(msg);
+
+    if (more_data)
+        send();
```
julianoes, line 1404:
```suggestion
    if (more_data) {
        send();
    }
```

src/mavsdk/plugins/ftp/ftp_impl.h
```diff
@@ -216,6 +217,7 @@ class FtpImpl : public PluginImplBase {
     void _reset_timer();
     void _stop_timer();
     void _list_directory(uint32_t offset);
+    unsigned get_size();
```
julianoes, line 220:
```suggestion
    unsigned get_size() const;
```

julianoes (changes requested):

examples/camera_manager/camera_manager.cpp
dayjaby: This makes sense. Thanks!

examples/camera_manager/camera_manager.cpp
dayjaby: Yes, https://mavlink.io/en/services/ftp.html#mavlink-ftp-url-scheme

examples/camera_manager/infos/camera_info.xml
dayjaby: It's a one-to-one copy from https://mavlink.io/en/services/camera_def.html#full_example

src/mavsdk/plugins/ftp/ftp_impl.cpp
dayjaby: The whole mavsdk ftp plugin seems to be based on the PX4 implementation. So I based the ftp burst operation also on the PX4 implementation (this bogus comment comes from PX4: https://github.com/PX4/PX4-Autopilot/blob/f8c2ee73db3654a34976291e5ac94c437e27d4a7/src/modules/mavlink/mavlink_ftp.cpp#L1126)

The implementation here is not complete yet; I transmit the entire file within one burst. Appearently, in PX4, they burst out up to 35000 bytes at once, but it's not specified yet: https://mavlink.io/en/services/ftp.html#burst-read-file

src/mavsdk/plugins/ftp/ftp_impl.h
dayjaby: 👍

src/mavsdk/plugins/ftp/ftp_impl.cpp
dayjaby: 👍

src/mavsdk/plugins/ftp/ftp_impl.cpp
julianoes: Aaaah, good point. We should do that...

examples/camera_manager/camera_manager.cpp
julianoes: I guess you're right, ok.

examples/camera_manager/camera_manager.cpp
julianoes: interesting, ok.

examples/camera_manager/infos/camera_info.xml
julianoes: Ok, fair enough.

dayjaby:
@julianoes applied your suggestions.

However, I still don't feel comfortable having this just as an example. I think it's doing a lot of similar stuff as your component_information plugin; and eventually we will need something similar for VIDEO_STREAMING_INFORMATION as well.

I think of something along the lines of:

1. We implement a camera_manager plugin
- Users can use that plugin to provide callback functions for the different camera parameters, to forward parameter changes to their non-mavlink cameras
2. We implement a video_stream_manager plugin
- Users provide callback functions to establish a video stream, for example via GStreamer

Any common functionality among the camera_manager, video_stream_manager, component_information and ftp plugin, we should put into src/mavsdk/core. D'accord?

If you have a better word than "manager", let me know ;)

dlech:
> If you have a better word than "manager", let me know ;)

All of the existing plugins that implement such features are called `xyz_server`.

Jaeyoung-Lim:
This is awesome!

julianoes:
I assume we will merge this once the #1733 is in. Or should we get it in before that?

dlech:
I don't think we want to merge it as-is, but rather reuse the ideas on top of the new changes. Don't we want a MAVSDK-Proto API for this too?

dayjaby:
@dlech we only need to add subscribe/provide parameter callbacks to the .proto.

For the ftp camera information file transmission:
I think @julianoes already moved the FTP stuff out of a plugin into something that we can include in the camera server implementation already. Just the FTP burst needs some update.

julianoes:
There has been some progress in https://github.com/mavlink/MAVSDK/pull/2088 in the same direction. I think once that other PR is merged, we can take a diff and see what is missing from this PR and pull it in.

julianoes:
Closing this, it's very outdated at this stage.

PR #1660, opened 2022-01-03
Add component information client and server

julianoes:
Depends on https://github.com/mavlink/MAVSDK-Proto/pull/270.

Probably best reviewed commit by commit.


JonasVautherin (dismissed):
Nice!

JonasVautherin (approved):

JonasVautherin (approved):

src/mavsdk/core/mavlink_request_message_handler.cpp
```diff
@@ -0,0 +1,126 @@
...
+    _system_impl(system_impl)
+{
+    _system_impl.register_mavlink_command_handler(
+        MAV_CMD_REQUEST_MESSAGE,
+        [this](const MavlinkCommandReceiver::CommandLong& command) {
+            return handle_command_long(command);
+        },
+        this);
+
+    _system_impl.register_mavlink_command_handler(
```
dayjaby, line 16: Is the duplicate register_mavlink_command_handler on purpose?

src/mavsdk/core/mavlink_request_message_handler.h
```diff
@@ -0,0 +1,46 @@
...
+class SystemImpl;
+
+class MavlinkRequestMessageHandler {
+public:
+    MavlinkRequestMessageHandler() = delete;
+    explicit MavlinkRequestMessageHandler(SystemImpl& system_impl);
+    ~MavlinkRequestMessageHandler();
+
+    using Params = std::array<float, 5>;
+    using Callback = std::function<std::optional<MAV_RESULT>(Params)>;
```
dayjaby, line 22: Consider `const Params&` instead of `Params` here.

dayjaby (commented):
Looking good so far!

src/mavsdk/core/mavlink_request_message_handler.cpp
julianoes: Ooops, good catch! I think one should be for command_int!

src/mavsdk/core/mavlink_request_message_handler.h
julianoes: Yes, would make sense. Thx.

JonasVautherin (approved):

src/mavsdk/core/mavlink_request_message_handler.h
julianoes: Changed in 71ff6ae5, finally.

julianoes:
I think this is somewhat contained and should not be interfering with the rest too much, so I'm merging it as in its current prototype state.

PR #1663, opened 2022-01-04
mavlink_parameters: handle param_ext_set

dayjaby:
This PR implements the following part of the mavlink extended parameter protocol:

image

Apart from that, this PR also allows plugins to subscribe to the extended parameter types via an overloaded callback function (based on https://en.cppreference.com/w/cpp/utility/variant/visit):

```
_parent->subscribe_param(
    "CAM_ISO",
    overloaded {
            [](auto iso) { std::cerr << "invalid parameter type for CAM_ISO" << std::endl; },
            [](uint16_t iso) { std::cout << "uint16_t: NEW ISO: " << iso << std::endl; },
            [](uint32_t iso) { std::cout << "uint32_t: NEW ISO: " << iso << std::endl; }
    },
    nullptr);
```

TODO:

- get rid of the code duplication in set_from_mavlink_param_ext_set / set_from_mavlink_param_ext_value
- ~~move `subscribe_param<T>` implementation to .cpp file and instantiate only the allowed types~~


src/mavsdk/core/mavlink_parameters.h
```diff
@@ -24,310 +24,23 @@ class MAVLinkParameters {
...
-                    return false;
-            }
-            return true;
-        }
+        bool set_from_mavlink_param_value(const mavlink_param_value_t& mavlink_value);
+        bool set_from_mavlink_param_set(const mavlink_param_set_t& mavlink_set);
+        bool set_from_mavlink_param_ext_set(const mavlink_param_ext_set_t& mavlink_ext_set);
+        bool set_from_mavlink_param_ext_value(const mavlink_param_ext_value_t& mavlink_ext_value);
+        bool set_from_xml(const std::string& type_str, const std::string& value_str);
+        bool set_empty_type_from_xml(const std::string& type_str);
```
julianoes, line 32: I think that's a nice cleanup, thanks.

src/mavsdk/core/system_impl.h
```diff
@@ -25,6 +25,12 @@
 #include <mutex>
 #include <future>
 
+template<class... Ts> struct overloaded : Ts... {
+    using Ts::operator()...;
+};
+
+template<class... Ts> overloaded(Ts...) -> overloaded<Ts...>;
+
```
julianoes, line 33: So this helper can only be used library internally. Was that the aim? Or should it be usable outside of the library in the public API?

And what was your use case for it?

julianoes (commented):
Thanks for the contribution! See comments inline.

src/mavsdk/core/mavlink_parameters.cpp
```diff
@@ -368,7 +401,7 @@ void MAVLinkParameters::do_work()
                     &work->mavlink_message,
                     param_id,
                     buf.get(),
-                    work->param_value.get_mav_param_type(),
+                    work->param_value.get_mav_param_ext_type(),
```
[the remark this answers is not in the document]
dayjaby: Yes, without that modification my camera_manager example complained that the uint32 parameter sent by QGroundControl has the wrong type. This fixes this issue.

src/mavsdk/core/system_impl.h
dayjaby: Still learning about std::variant, I found std::visit which had an example of using overloaded functions to handle the different active types of an std::variant: https://en.cppreference.com/w/cpp/utility/variant/visit

It's supposed to be used in plugin implementations, but could also be used outside if the user wants to. I'll send you an example very soon.

However, this approach has a fundamental flaw, that you could write a handler for uint32_t which would also accept types like uint8_t or uint16_t instead of dismissing it as wrong parameter type. Need to do some more testing about this.

dayjaby: An example looks like this. If the plugin doesn't know the type necessary to run the callback function, because it's supplied by the user, this approach is pretty helpful:

Plugin code:

```
    template <class T>
    void add_param(const std::string& name, T default_value, const typename std::function<void(T t)>& callback) {
        _parent->provide_server_param_int(name, default_value);
        _parent->subscribe_param(
            name,
            overloaded {  
                [&name](auto value) { std::cerr << "invalid parameter type for " << name << std::endl; },
                callback
            },
            nullptr);
    }
```

User code:

```
camera_manager.add_param<uint32_t>("CAM_ISO", 0, [](uint32_t iso) { std::cout << "NEW ISO: " << iso << std::endl; });
```

Note that `_parent->provide_server_param_int` would also have to be replaced with a `_parent->provide_server_param`.

src/mavsdk/core/mavlink_parameters.cpp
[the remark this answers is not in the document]
julianoes: Ah great, thanks.

src/mavsdk/core/system_impl.h
julianoes: I actually hate the fact that we have all types smaller than 32 bits. We could have saved quite a bit of trouble by not suggesting them in the spec. That was just stupid, I think.

Anyway, I see. In that case you need to add the `overloaded` helper template in one of the public headers, maybe in the param plugin.h.

src/mavsdk/core/mavlink_parameters.cpp
```diff
@@ -392,6 +425,30 @@ void MAVLinkParameters::do_work()
             // As we're a server in this case we don't need any response
             work_queue_guard.pop_front();
         } break;
+
+        case WorkItem::Type::Ack: {
+            if (work->extended) {
+                auto buf = std::make_unique<char[]>(128);
```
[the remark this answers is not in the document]
dayjaby: Makes perfect sense. Done!

src/mavsdk/core/system_impl.h
dayjaby: > Anyway, I see. In that case you need to add the overloaded helper template in one of the public headers, maybe in the param plugin.h.

Done!

dayjaby:
Guess it's good to merge now!

Not sure about https://github.com/mavlink/MAVSDK/pull/1663/commits/fa84fa7ff801f0870c6449c8fb7b4d39be5b4464, which shows some of the strengths of std::visit.

Probably one issue that remains is how float values are compared to each other: We need some kind of FLOAT_EPSILON there.

src/mavsdk/core/mavlink_parameters.cpp
```diff
@@ -1261,31 +1261,7 @@ void MAVLinkParameters::ParamValue::get_128_bytes(char* bytes) const
...
-    } else if (std::get_if<float>(&_value)) {
-        return std::to_string(std::get<float>(_value));
-    } else if (std::get_if<double>(&_value)) {
-        return std::to_string(std::get<double>(_value));
-    } else {
-        LogErr() << "Unknown data type for param.";
-        assert(false);
-        return {"(unknown)"};
-    }
+    return std::visit([](auto value) { return std::to_string(value); }, _value);
```
julianoes, line 1264: Nice!

julianoes (dismissed):

src/mavsdk/core/include/mavsdk/plugin_base.h
```diff
@@ -2,6 +2,12 @@
 
 namespace mavsdk {
 
+template<class... Ts> struct overloaded : Ts... {
```
julianoes, line 5: I'd add a comment what this is for, and maybe a small example how to use it.

dayjaby:
Fixed style.

> template<class... Ts> struct overloaded : Ts... {
>> I'd add a comment what this is for, and maybe a small example how to use it.

I will do it as part of the camera_manager example/plugin.

julianoes:
I would still like add a note, otherwise this is unused/confusing code that the next reader will remove.

dayjaby:
Okay, thanks for your time and guiding me through this!

Guess the sitl error is not caused by this PR? The commit before passed through all tests and checks.

julianoes (approved):

julianoes:
Thanks @dayjaby!

PR #1664, opened 2022-01-05
calibration: handle ack for gimbal calibration

julianoes:
Gimbal calibration is implemented in the "right" way with a long running command. This allows us to close the stream when we receive the final ACCEPTED ack.

FYI @potaito.


potaito:
Interesting, thanks!

Issue #1568, opened 2021-09-29
ArduPilot support limitations


bazfp:
This is used as a tracking issue for ArduPilot incompatibilities.

Some, but not all MAVSDK plugins have ArduPilot support. This ticket will list **known** compatibilities and incompatibilities. If not listed, the state is unknown - It may work, it may not.

Please comment if you have tested functions and can confirm/deny working.

If you would like an ArduPilot feature included in MAVLink, please contribute.

### Known API **functional** with ArduPilot :

- Telemetry
  - `set_rate_position_async`
  - `set_rate_gps_info_async`
  - `health_all_ok`
  - `subscribe_position`
  - `subscribe_heading`
  - `subscribe_raw_gps`
  - `armed`
- Mission Raw
  - **NOTE:** First mission item must be home position for ArduPilot
  - `upload_mission_async`
  - `clear_mission_async`
  - `start_mission_async`
- Action
  - `arm`
- MAVLink Passthrough
  - `subscribe_message_async`
  - `send_command_int`
- Param
  - `set_param_float`

### Known **Broken** API

- Mission
  - Needs to account for first mission item being home in Ardupilot
- Probably much more :)

ykhedar:
Further broken APIs discovered through running Integration Tests with APM-SITL:
- Action

```c++
- takeoff             // Results in Error:     command unsupported (21). (mavlink_commands.cpp:187)

- land                // Results in Error:  command unsupported (21). (mavlink_commands.cpp:187)

- set_takeoff_altitude // Results in Error:    Error: Retrying failed get param busy timeout: 
                                  // MIS_TAKEOFF_ALT (mavlink_parameters.cpp:793)
				 //		 sending again, retries to do: 3  (MIS_TAKEOFF_ALT). (mavlink_parameters.cpp:780) 

- goto_location  // Results in Error:  command denied (192). (mavlink_commands.cpp:180)
```

There are probably more but need to fix these basic functions before moving to more advanced ones.

LorenzMeier:
These are things that you would need to address on the Ardupilot side. The design goal of MAVSDK is to implement the MAVLink standard (and in the process address things that are not properly standardized, e.g. we're working on a scheme for common flight modes to be able to rip out stack-specific flight modes in MAVSDK).

[...]

Issue #728, opened 2019-04-23
Adding support for ArduPilot


JonasVautherin:
## Context

A while ago, @peterbarker worked on adding support for APM in the SDK ([here](https://github.com/peterbarker/DroneCore/tree/ardupilot)). More recently, @jinchengde (and others) started looking into that, again. The goal of this issue is to discuss the high-level requirements and architecture.

## Benefits for APM

The project is architectured in such a way that everything relies on a C++ MAVLink implementation. Making this implementation compatible with APM means that APM will also get the language bindings ([Python](https://github.com/dronecode/dronecodesdk-python), [Swift](https://github.com/dronecode/dronecodesdk-swift), [Java](https://github.com/dronecore/dronecore-java) and future ones like [Javascript](https://github.com/mrpollo/DronecodeSDK-JavaScript)).

The interface common to all our bindings is defined [here](https://github.com/Dronecode/DronecodeSDK-Proto/tree/master/protos) and is used to auto-generate language-specific APIs (e.g. using RxSwift in Swift, RxJava in Java, and asyncio in Python).

## SDK architecture (overview)

The SDK can been seen as composed from 4 parts:

1. The [interface definition](https://github.com/Dronecode/DronecodeSDK-Proto/tree/master/protos) (protobuf files), common to all the languages (C++, Swift, Java, Python). Each proto file is considered a "plugin", which is a group of (MAVLink) features.
2. The C++ [MAVLink implementation](https://github.com/Dronecode/DronecodeSDK), which actually maintains a connection to the drone and does the MAVLink logic.
3. The [backend](https://github.com/Dronecode/DronecodeSDK/tree/develop/backend), which exposes the C++ implementation of (2) as a messaging server through the interface of (1).
4. The language bindings ([Python](https://github.com/dronecode/dronecodesdk-python), [Swift](https://github.com/dronecode/dronecodesdk-swift), [Java](https://github.com/dronecore/dronecore-java)), which implement (1) to communicate with the backend (3).

At the lowest level, each plugin (defined by one proto file) is a C++ library. The backend uses those libraries, as can been seen [here](https://github.com/Dronecode/DronecodeSDK/blob/develop/backend/src/CMakeLists.txt#L38-L46).

JonasVautherin:
I can see different ways to achieve that (and I'm looking for more thoughts). I believe that for each feature, there can be 3 situations:

1. The feature has (can have) the same interface between APM and PX4, but not the same MAVLink implementation
[...]

PR #1669, opened 2022-01-15
Adding Ardupilot SITL to MAVSDK CI 

ykhedar:
Following discussion on Issue #728 for adding Ardupilot support in MAVSDK (related issues #1562 and #1568), this pull request adds Ardupilot SITL (Copter, Rover) along with necessary changes needed to run it successfully. Currently onnly 3/34 Integrations tests pass. Some notes about changes made:

1. Added Dockerfiles each for Copter and Rover Firmware versions 4.1.2
2. Edited the script for building and pushing the new docker files.
3. Edited `run-sitl-tests.sh`, `start_px4_sitl.sh` and `stop_px4_sitl.sh` to include ardupilot related sitl codes. May need to change the file names in the future. e.g. `start_px4_sitl.sh` -> `start_sitl.sh`.
4. Added code block for running ardupilot sitl into the github actions file.

Requesting Review and Suggestions,
ykhedar

P.S. Congrats on the v1.0 release of MAVSDK :)


tools/run-sitl-tests.sh
```diff
@@ -7,9 +7,29 @@ if [ "$#" -ne 1 ]; then
     exit 1
 fi
 
-PX4_FIRMWARE_DIR=$1
+echo "PX4 Version is: " ${PX4_VERSION}
```
JonasVautherin, line 10: Should that line exist? Feels like it's already done below (line 13), right?

JonasVautherin (commented):
Nice! It requires the APM containers to be pushed on the mavsdk docker registry, right? I see that the CI is failing to pull them at the moment.

tools/run-sitl-tests.sh
ykhedar: my bad :) .Removed it in the latest commit.

ykhedar:
> Nice! It requires the APM containers to be pushed on the mavsdk docker registry, right? I see that the CI is failing to pull them at the moment.

exactly. I did all the tests on my account with all images pushed to my docker hub repo. After everything was working, i quashed all the commits into one so as to keep the pull request clean.

tools/start_px4_sitl.sh
```diff
@@ -25,61 +25,119 @@ if [[ $AUTOSTART_SITL != 1 ]]; then
     exit 0
 fi
 
-if [[ -n "$PX4_FIRMWARE_DIR" ]]; then
-    px4_firmware_dir=$PX4_FIRMWARE_DIR
+# PX4 specific SITL Run Code
```
julianoes, line 28: I agree we should change the name of these scripts.

julianoes (commented):
Looks quite clean, thank you @ykhedar. Do the tests actually succeed?

julianoes (commented):
Looks quite clean, thank you @ykhedar. Do the tests actually succeed?

julianoes:
I have built and pushed the docker images.

ykhedar:
> Looks quite clean, thank you @ykhedar. Do the tests actually succeed?


No. The tests are going on forever on the CI so I stopped the workflow. When I checked the logs, this was due to bad telemetry health condition inside a while loop in one of the tests for VTOL. Here is the link to the code:

[action_transition_multicopter_fixedwing](https://github.com/mavlink/MAVSDK/blob/cfd9feaa2786bb3850813ed3ea0ab27041ee7742/src/integration_tests/action_transition_multicopter_fixedwing.cpp#L37)

Should we add maybe a break condition in this test instead of letting the while loop run forever? Maybe do the same for all such while loops in all the tests? Just make the test fail if the while condition is not met after 60 seconds or 120 seconds?

JonasVautherin:
> Maybe do the same for all such while loops in all the tests?

Generally I think it would make sense if the tests did not hang forever :+1:.

This said, for the health, is it failing because of a difference with Ardupilot? Can we fix `health_all_ok()` easily?

julianoes:
We should limit the tests to 45mins, like the PX4 SITL ones.

And if you want to go through all of them, feel free, but I think it's quite a bit of work.

Do the tests run through for you locally?

ykhedar:
> We should limit the tests to 45mins, like the PX4 SITL ones.
> 
> And if you want to go through all of them, feel free, but I think it's quite a bit of work.
> 
> Do the tests run through for you locally?

How does PX4 SITL implement the limit of 45 minutes? Cant seem to find the setting for this. 

On my local machine i saw 3 tests pass (i swear :D) which now seems to be incorrect. My bad :( . 

Should I add a simple timer using std::chrono in the while loop to wait for 30 seconds or so before the loop exits with a failed test? Or do you suggest something better?

JonasVautherin:
> How does PX4 SITL implement the limit of 45 minutes? Cant seem to find the setting for this.

I think @julianoes was referring to [this](https://github.com/mavlink/MAVSDK/blob/main/.github/workflows/main.yml#L634).

julianoes:
@ykhedar it would be best to use this function for checks with timeout:
https://github.com/mavlink/MAVSDK/blob/6ad7dbd424a5be32c0d4f3e61590fa6c1de3cb83/src/integration_tests/integration_test_helper.h#L69-L84

julianoes (dismissed):
Nice!

ykhedar:
> Nice!

Just made some changes to the commits (squashed them into one) and your review was made stale by git. Also fetched upstream and merged the new commits into ykhedar/MAVSDK.

Apparently 2 out of 34 tests are passing.

JonasVautherin (dismissed):

ykhedar:
Thanks @JonasVautherin for your review. Do we need to get all the Ardupilot related Integration tests to succeed before merging this pull request?

JonasVautherin:
We already use `--gtest_filter` in the [script running the test](https://github.com/mavlink/MAVSDK/blob/main/tools/run-sitl-tests.sh#L15). What about adding `PX4` in the tests that only pass with PX4, and filter them out in the Ardupilot runs? Enabling a test for Ardupilot would then just require us to remove "PX4" from the test name, so they could be incrementally activated. What do you think?

ykhedar:
Sounds like a good idea to me. I would do the necessary changes. Something like this:

`TEST_F(SitlTest, ActionGoto)` -> `TEST_F(SitlTest, PX4ActionGoto)` ?

ykhedar:
Okay, I did the changes but it seems now some checks related to macOS-framework and ios are failing for some reason and i cannot think of why. Any ideas? :)

JonasVautherin:
Grmbl I don't get why the ios build sometimes does that :disappointed:

JonasVautherin (approved):

julianoes (approved):
Thanks a lot for your work @ykhedar!

PR #1671, opened 2022-01-17
docker: add arm build for library only

julianoes:
This adds dockcross builds for armv6, armv7, arm64 to generate .deb
files to be used on RPi and similar.


JonasVautherin (approved):
Nice!

PR #1672, opened 2022-01-17
cmake: fix installed include path

julianoes:
This only mattered if you installed the library without the mavsdk_server since the mavsdk_server had the correct include path.


JonasVautherin (dismissed):
Good catch

bazfp (dismissed):

PR #1686, opened 2022-01-21
core: fix linking on Windows

### LINKED ISSUE #1679: MAVSDK Windows Installation
Hi, 
When installing MAVSDK on Windows, are we given the option to either 1) locally install mavsdk-windows-x64-release.zip or 2) clone the repository and cmake build and install? The zip folder does not contain a CMakeLists.txt so do we have to make one ourselves? Because of this uncertainty I tried the second approach by using:
git clone https://github.com/mavlink/MAVSDK.git
git submodule update --init --recursive
cmake -DCMAKE_BUILD_TYPE=Debug -Bbuild/default -S. 
cmake --build build/default -j8 
And was given this error message on the last command:
Capture1
Note the message is a lot longer but the errors all involve unresolved external symbols.

julianoes:
This might fix #1679.


JonasVautherin (approved):
Ok :sweat_smile:

PR #1690, opened 2022-01-25
Fixed XCode warning

byuarus:
Fixed XCode warning: "declaration of "Struct MavsdkServer" will not be visible outside of this function"

### DIFF
--- a/src/mavsdk_server/src/mavsdk_server_api.h
+++ b/src/mavsdk_server/src/mavsdk_server_api.h
@@ -16,7 +16,8 @@ extern "C" {
 #define DLLExport __attribute__((visibility("default")))
 #endif
 
-    struct MavsdkServer;
+struct MavsdkServer;
+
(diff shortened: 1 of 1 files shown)


src/mavsdk_server/src/mavsdk_server_api.h
```diff
@@ -16,6 +16,7 @@ extern "C" {
 #define DLLExport __attribute__((visibility("default")))
 #endif
 
+    struct MavsdkServer;
```
JonasVautherin, line 19:
```suggestion
struct MavsdkServer;

```

Just a nitpick, for the style check :laughing:

JonasVautherin (changes requested):

src/mavsdk_server/src/mavsdk_server_api.h
byuarus: 👍

PR #1693, opened 2022-02-10
mavlink-ftp: Do not continue on receiving an old ack

ThomasDebrunner:
This fixes a bug that could cause sender and receiver to get out of sync.
When the ack from receiver is delayed, the sender re-transmits the package. When the ack then arrives anyways, the sender continues with the next package. Since we already re-sent the previous package, another ack for the previous package is received. This makes the sender again move to the next package. From here on, the receiver and sender are out of sync.

This is generally not an issue, but when now an ack gets lost, the sender no longer realizes, since it is already ahead of time and receives enough acks to finish.

This PR fixes the issue by checking the sequence number of the ack. If the ack is for a package with a sequence number lower than our current, the ack is ignored, since we already must have received a previous ack for this sequence.
--> Only consider the first ack for a sequence number.

### DIFF
--- a/src/mavsdk/plugins/ftp/ftp_impl.cpp
+++ b/src/mavsdk/plugins/ftp/ftp_impl.cpp
@@ -18,6 +18,13 @@
 
 namespace mavsdk {
 
+bool seq_lt(uint16_t a, uint16_t b)
+{
+    // From https://en.wikipedia.org/wiki/Serial_number_arithmetic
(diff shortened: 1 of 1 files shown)


TSC21 (dismissed):

julianoes (dismissed):

ThomasDebrunner:
done

julianoes (approved):

julianoes:
The Windows CI failure looks unrelated.

Issue #1683, opened 2022-01-20
Handling of MAVLINK params


ykhedar:
I just stumbled upon the function which handles the MAV_PARAM types and found that the function in :

https://github.com/mavlink/MAVSDK/blob/a6836be251e216f9967183a7551cc6ccb2a60382/src/mavsdk/core/mavlink_parameters.cpp#L923

handles it in a different way then the function for handling MAV_PARAM_EXT types. Is there a reason for this? Even the setting of `_value` in this function seems to be odd. Since the `temp.int32_value` is never assigned a value, using it to assign it to `_value` seems to give it garbage values. Or am I missing something?

julianoes:
Can you describe the problem that you're trying to solve? I'm not sure I understand what you mean.

The `int32_value` is a union with the `float_value`, so it's accessing the same bits but with case to a different type.

julianoes:
@bazfp no, why? So you're saying we should use `std::variant` instead? I mean we should anyway but I don't understand why it's UB.

ykhedar:
> Can you describe the problem that you're trying to solve? I'm not sure I understand what you mean.
> 
> The `int32_value` is a union with the `float_value`, so it's accessing the same bits but with case to a different type.

I was exploring why getting all parameters from ardupilot sitl using MAVSDK was failing for some type of parameters. Since PX4 has only int32 and float param types (apm has in addition int8, int16 as well), the int8 and int16 param type was not being handled in the function tagged in my OP. 

Doing more investigations, I will post in more detail once the tests are done completely.

julianoes:
> apm has in addition int8, int16 as well), the int8 and int16 param type was not being handled in the function tagged in my OP.

Aha, I understand! I was under the impression ArduPilot only has float and nothing else.

ykhedar:
Hi again, 

In the past few days I did some more research on the parameter handling related to PX4, Ardupilot and the current state in MAVSDK.

MAVLINK standard supports: uint8, int8, uint16, int16, uint32, int32, uint64, int64, real32, real64
PX4 supports/uses:                int32 and real32(float)
as found from [PX4 Param Reference](https://docs.px4.io/v1.9.0/en/advanced_config/parameter_reference.html)

Ardupilot supports/uses:         int8, int16, int32, real32(float)
as found from [APM Copter Param Reference](https://ardupilot.org/dev/docs/mavlink-get-set-params.html#setting-a-parameter)

MAVSDK supports int32 and float(real32)

[...]

PR #1703, opened 2022-02-21
param: changes for making param set/get work with ArduPilot

ykhedar:
This pull request contains the minimal necessary changes needed in the MAVSDK to make param set/get work with Ardupilot. Following the discussion from https://github.com/mavlink/MAVSDK/issues/1683, I decided to try the suggestion B from @julianoes 

> B) Always download all params on startup to then have the type cached when the function is used. This would take longer but make the API easier. What do you think?

I had to, however create a different function to deal with the param set for APM

```bash
float MAVLinkParameters::ParamValue::get_4_float_bytes() const
float MAVLinkParameters::ParamValue::get_4_float_bytes_apm() const
```
As mentioned before, Ardupilot likes to have direct cast to float instead of the re-interpret cast used in MAVSDK for PX4 param set. More details can be found in https://mavlink.io/en/services/parameter.html#ardupilot

Awaiting valued feedback on this.

### DIFF
--- a/src/mavsdk/core/mavlink_parameters.cpp
+++ b/src/mavsdk/core/mavlink_parameters.cpp
@@ -937,13 +937,28 @@ bool MAVLinkParameters::ParamValue::set_from_mavlink_param_value(
     const mavlink_param_value_t& mavlink_value)
 {
     switch (mavlink_value.param_type) {
+        case MAV_PARAM_TYPE_UINT8: {
+            uint8_t temp = mavlink_value.param_value;
+            _value = temp;
+        } break;
+
         case MAV_PARAM_TYPE_INT8: {
             int8_t temp = mavlink_value.param_value;
             _value = temp;
         } break;
 
+        case MAV_PARAM_TYPE_UINT16: {
+            uint16_t temp = mavlink_value.param_value;
+            _value = temp;
+        } break;
+
         case MAV_PARAM_TYPE_INT16: {
             int16_t temp = mavlink_value.param_value;
+            _value = temp;
+        } break;
+
+        case MAV_PARAM_TYPE_UINT32: {
+            uint32_t temp = mavlink_value.param_value;
             _value = temp;
         } break;
 
@@ -959,7 +974,8 @@ bool MAVLinkParameters::ParamValue::set_from_mavlink_param_value(
 
         default:
             // This would be worrying
-            LogErr() << "Error: unknown mavlink param type: ";
+            LogErr() << "Error: unknown mavlink param type: "
+                     << std::to_string(mavlink_value.param_type);
             return false;
     }
     return true;
@@ -1279,12 +1295,18 @@ bool MAVLinkParameters::ParamValue::set_as_same_type(const std::string& value_st
 {
     if (std::get_if<float>(&_value)) {
         return std::get<float>(_value);
+    } else if (std::get_if<uint32_t>(&_value)) {
+        return static_cast<float>(std::get<uint32_t>(_value));
     } else if (std::get_if<int32_t>(&_value)) {
-        return static_cast<int32_t>(std::get<float>(_value));
+        return static_cast<float>(std::get<int32_t>(_value));
+    } else if (std::get_if<uint16_t>(&_value)) {
+        return static_cast<float>(std::get<uint16_t>(_value));
     } else if (std::get_if<int16_t>(&_value)) {
-        return static_cast<int16_t>(std::get<float>(_value));
+        return static_cast<float>(std::get<int16_t>(_value));
+    } else if (std::get_if<uint8_t>(&_value)) {
+        return static_cast<float>(std::get<uint8_t>(_value));
     } else if (std::get_if<int8_t>(&_value)) {
-        return static_cast<int8_t>(std::get<float>(_value));
+        return static_cast<float>(std::get<int8_t>(_value));
     } else {
         LogErr() << "Unknown type";
         assert(false);
--- a/src/mavsdk/core/mavlink_parameters.cpp
+++ b/src/mavsdk/core/mavlink_parameters.cpp
@@ -311,7 +311,19 @@ void MAVLinkParameters::do_work()
                     work->param_value.get_mav_param_ext_type());
             } else {
                 // Param set is intended for Autopilot only.
-                mavlink_msg_param_set_pack(
+                if (_parent.autopilot() == SystemImpl::Autopilot::ArduPilot) {
+                    mavlink_msg_param_set_pack(
+                    _parent.get_own_system_id(),
+                    _parent.get_own_component_id(),
+                    &work->mavlink_message,
+                    _parent.get_system_id(),
+                    _parent.get_autopilot_id(),
+                    param_id,
+                    work->param_value.get_4_float_bytes_apm(),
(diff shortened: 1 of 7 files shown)


src/mavsdk/core/mavlink_parameters.cpp
```diff
@@ -311,7 +311,19 @@ void MAVLinkParameters::do_work()
                     work->param_value.get_mav_param_ext_type());
             } else {
                 // Param set is intended for Autopilot only.
-                mavlink_msg_param_set_pack(
+                if (_parent.autopilot() == SystemImpl::Autopilot::ArduPilot) {
```
JonasVautherin, line 314: Maybe a nitpick, but what about having the `if` just set a temporary variable like this?

```c++
float my_param;
if (_parent.autopilot() == SystemImpl::Autopilot::ArduPilot) {
    my_param = work->param_value.get_4_float_bytes_apm();
} else {
    my_param = work->param_value.get_4_float_bytes();
}
```

This makes it clear that the difference between APM and PX4 is only for this param `my_param` (well, it needs a better name), and all the other parameters given to `mavlink_msg_param_set_pack` are the same.

What do you think?

src/mavsdk/core/system_impl.cpp
```diff
@@ -767,15 +791,29 @@ std::pair<MAVLinkParameters::Result, int> SystemImpl::get_param_int(const std::s
     auto res = prom.get_future();
 
     MAVLinkParameters::ParamValue value_type;
-    value_type.set<int32_t>(0);
-
+    //value_type.set<int32_t>(0);
```
JonasVautherin, line 794:
```suggestion

```

JonasVautherin (changes requested):

src/mavsdk/core/system_impl.cpp
```diff
@@ -643,14 +646,35 @@ MAVLinkParameters::Result SystemImpl::set_param_float(const std::string& name, f
...
+    // for (auto const& x : _param_map)
+    // {
+    //     std::cout << x.first  // ParamValue ID
+    //             << ':' 
+    //             << x.second // ParamValue Value 
+    //             << ':' 
+    //             << x.second.get_mav_param_type() // ParamValue Type
+    //             << std::endl;
+    // }
+    return _param_map;
```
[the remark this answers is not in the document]
ykhedar: Sadly yes, the problem is we need to keep a cache of the parameter and I could only think of this way right now. So as soon as the System is created and it inits the plugins e.g. param plugin, the param plugin in its init method sends a request to get_all_params() which then populates the _param_map in System class and then also returns it to the param plugin.
let me know if you have a better idea ;).

src/mavsdk/core/system_impl.cpp
ykhedar: Done.

src/mavsdk/core/system_impl.cpp
```diff
@@ -643,14 +644,25 @@ MAVLinkParameters::Result SystemImpl::set_param_float(const std::string& name, f
 MAVLinkParameters::Result SystemImpl::set_param_int(const std::string& name, int32_t value)
 {
     MAVLinkParameters::ParamValue param_value;
-    param_value.set<int32_t>(value);
+    //param_value.set<int32_t>(value);
```
JonasVautherin, line 647:
```suggestion

```

JonasVautherin:
The CI is not happy, and you should fix the style with something like `./tools/fix_style.sh .` :+1: :blush:

julianoes:
@ykhedar I've rebased your PR, fixed style issues, and force pushed it to your branch.

I also added a test to check `get_all_params`:
```
cmake --build build/default/ -j6 && build/default/src/integration_tests/integration_tests_runner --gtest_filter="SitlTest.GetAllParams"
```
(For ArduPilot, don't forget to change `add_udp_connection()` to `add_udp_connection(14550)`.

The test fails for me for me for ArduPilot,
```
[05:29:00|Info ] Int params: 0 (param.cpp:149)
[05:29:00|Info ] Float params: 0 (param.cpp:155)
[05:29:00|Info ] Combined params: 0 (param.cpp:161)
```

and PX4:
```
[05:30:13|Error] Error: unknown mavlink param type:  (mavlink_parameters.cpp:965)
[05:30:13|Error] Error: unknown mavlink param type:  (mavlink_parameters.cpp:965)
[05:30:14|Info ] Int params: 0 (param.cpp:149)
[05:30:14|Info ] Float params: 0 (param.cpp:155)
[05:30:14|Info ] Combined params: 0 (param.cpp:161)
```

so something is not quite right yet.
It works for PX4 before that.

One thing that I think of changing is to only get the param that you want to set. That way we don't have to get all params which is not necessarily something that everyone wants to do, always.

julianoes:
FYI @ykhedar I'm currently reworking this.

ykhedar:
> @ykhedar I've rebased your PR, fixed style issues, and force pushed it to your branch.
> 
> I also added a test to check `get_all_params`:
> 
> ```
> cmake --build build/default/ -j6 && build/default/src/integration_tests/integration_tests_runner --gtest_filter="SitlTest.GetAllParams"
> ```
> 
> (For ArduPilot, don't forget to change `add_udp_connection()` to `add_udp_connection(14550)`.
> 
> The test fails for me for me for ArduPilot,
> 
> ```
> [05:29:00|Info ] Int params: 0 (param.cpp:149)
> [05:29:00|Info ] Float params: 0 (param.cpp:155)
> [05:29:00|Info ] Combined params: 0 (param.cpp:161)
> ```
> 
> and PX4:
> 
> ```
> [05:30:13|Error] Error: unknown mavlink param type:  (mavlink_parameters.cpp:965)
> [05:30:13|Error] Error: unknown mavlink param type:  (mavlink_parameters.cpp:965)
> [05:30:14|Info ] Int params: 0 (param.cpp:149)
> [05:30:14|Info ] Float params: 0 (param.cpp:155)
> [05:30:14|Info ] Combined params: 0 (param.cpp:161)
> ```
> 
> so something is not quite right yet. It works for PX4 before that.
> 
> One thing that I think of changing is to only get the param that you want to set. That way we don't have to get all params which is not necessarily something that everyone wants to do, always.

Hi @julianoes Thanks for taking a look at the PR. I am not able to work on it for at least a week more due to the other project so please go ahead and rework if you have the time. 

Last time I checked the tests not passing for PX4 set param was most probably because the params were not being cached at the right time. I was looking into it but then dragged into the other project.

I believe caching all params on first connection is not a bad idea. I have seen this behaviour in Mavproxy and really like that, as soon as it is connected to the vehicle it saves all params to a a file. Since MAVSDK is usually connected over a wire to the Autopilot, should not be a huge problem speed or lost packet wise. What do you think?

ykhedar:
@julianoes There was a small bug(wrong function name) which was causing issues for CI. I fixed it in my last commit and CI seems to be happy now. Are you still working on something for this pull request or we can rebase and merge it into the main branch?

julianoes:
@ykhedar I'm having a look now, sorry for the delay.

julianoes:
@ykhedar thanks for fixing that, I was wondering what was going on!

julianoes (approved):
Reviewed again, let's get this in and start using it.

PR #1705, opened 2022-02-27
Support more than one camera

persuader72:
This PR is about the ability of the mavsdk::Camera plugin to handle more than one camera on the same mavlink channel.

The main change is in **MAVLinkMessageHandler** in which has been added an optional filter to select only themessages that came from a specific component_id. The MAVLinkMessageHandler::register_one function has been overloaded to accept the component id parameter.

In the **System** module  a DiscoverIdCallback callback has been added to be able to receive the **component_id** value  inside the callback when a component is discovered .

The **Camera** plugin has been changed to accept the **camera_id** in the constructor and bind only to the message coming to the selected camera component.  

It is now possible to use the  **component_id** received in the **component_discovered** callback to create an instance of the Camera plugin binded to a specific camera.

The full PR is backward compatible with the current library behavior.


julianoes (commented):
Thanks for the contribution! This is an interesting approach.

Now one issue with it is that the plugin API (the main header file `camera.h` is auto-generated, so we can't easily add the component ID as a constructor parameter as you did. If you check the file it says: Auto-generated, don't change.

So we need to find another way to do that, or add support for a notion of component IDs in general, which then all plugins can make use of.

I need to think about this some more. In the meantime, would an approach with a setter like `set_camera_id` work instead of using the constructor?

@JonasVautherin do you have an idea what we could do?

persuader72:
Yes, I miss the fact that the Camera.h is auto-generated.

The reason because i put the camera_id in the constructor is that the plugin must known the "component_id" before it call the register_mavlink_message_handler functions to register the messages callback, This happen in the Init function that is invoked by plugin's constructor so the information must be know in the constructor.

Another possibility is: add a new method to the MAVLinkMessageHandler  class that will change all  "component_id" (Entry::cmp_id) fields of the entries records that have the same cookie.

In this way the Camera plugin can init itself in the usual way without a specific "component_id" and change all callbacks previously registered by the plugin itself when the function set_camera_id id called.

persuader72:
The code has been changed in order to remove the camera_id argument in the Camera plugin constructor.

Now is possible to use the already existent select_camera function  to change the component_id of the Camera.
I made a [PR](https://github.com/mavlink/MAVSDK-Proto/pull/275) to update the Proto of the public interface of the plugin.

The Camera::select_camera can be used anytime after the Camera instance has ben constructed.

JonasVautherin:
> @JonasVautherin do you have an idea what we could do?

Unfortunately nothing simple :sweat_smile:

julianoes:
Note, style/docs here fails because of missing docstrings in https://github.com/mavlink/MAVSDK-Proto/pull/275.

src/mavsdk/core/mavlink_message_handler.cpp
```diff
@@ -8,7 +8,16 @@ void MAVLinkMessageHandler::register_one(
 {
     std::lock_guard<std::mutex> lock(_mutex);
 
-    Entry entry = {msg_id, callback, cookie};
+    Entry entry = {msg_id, -1, callback, cookie};
+    _table.push_back(entry);
+}
+
+void MAVLinkMessageHandler::register_one(
+    uint16_t msg_id, uint8_t cmp_id, const Callback &callback, const void *cookie)
```
julianoes, line 16:
```suggestion
    uint16_t msg_id, uint8_t component_id, const Callback &callback, const void *cookie)
```

src/mavsdk/core/mavlink_message_handler.cpp
```diff
@@ -65,4 +74,15 @@ void MAVLinkMessageHandler::process_message(const mavlink_message_t& message)
 #endif
 }
 
+void MAVLinkMessageHandler::update_component_id(uint16_t msg_id, uint8_t cmp_id, const void* cookie)
```
julianoes, line 77:
```suggestion
void MAVLinkMessageHandler::update_component_id(uint16_t msg_id, uint8_t component_id, const void* cookie)
```

src/mavsdk/core/mavlink_message_handler.cpp
```diff
@@ -8,7 +8,16 @@ void MAVLinkMessageHandler::register_one(
 {
     std::lock_guard<std::mutex> lock(_mutex);
 
-    Entry entry = {msg_id, callback, cookie};
+    Entry entry = {msg_id, -1, callback, cookie};
```
julianoes, line 11: I think I would prefer to use a `std::optional<uint8_t>` instead of a `int16_t` with `-1` as the magic value.

src/mavsdk/plugins/camera/camera_impl.cpp
```diff
@@ -219,6 +219,31 @@ void CameraImpl::manual_disable()
     _camera_found = false;
 }
 
+void CameraImpl::update_component()
```
julianoes, line 222: Where is this called from?

julianoes (commented):
I made a couple of comments. In general this should be ok and work.

src/mavsdk/plugins/camera/camera_impl.cpp
persuader72: It's called on the `CameraImpl::select_camera(const size_t id)` (line 260 of this file)  this method in on the public interface and can be called anytime after the constructor.

src/mavsdk/core/mavlink_message_handler.cpp
persuader72: Yes it looks better

persuader72:
I think I completed the requested changes, I left these changes in separate commits,  it's better to compact it  in only two commits?

julianoes:
I rebased this on top of `main`, ran the generation and style fix and force pushed it. Once CI looks ok, we can merge it.

Thanks a lot @persuader72, I think this looks really clean like this actually, I like it.

julianoes (approved):

PR #1709, opened 2022-03-02
FTP: Set size in message on read

ThomasDebrunner:
Since https://github.com/PX4/PX4-Autopilot/pull/18704, sending a 0 length on download does not make PX4 send any data. 
The download gets stuck then.

This just always requests the maximum number of bytes, except for the last packet.


julianoes (approved):
Thanks! And I fixed style so it should be good to go.

PR #1718, opened 2022-03-10
Add set function to system status

bastianhjaeger:


TSC21:
@bastianhjaeger can you run `tools/run-docker.sh tools/fix_style.sh .`?

julianoes (dismissed):
I think that makes sense like this, thanks @bastianhjaeger.

@JonasVautherin any objections?

JonasVautherin:
@julianoes: I was thinking about exposing it in the core plugin, for the language wrappers. But actually that could be another PR, maybe :thinking:

JonasVautherin (approved):

PR #1726, opened 2022-03-22
action: add API to set current speed

julianoes:
This is an API that allows to temporarily set the speed of the drone, e.g. for mission mode or in goto/reposition mode.

FYI @reedev


reedev:
sounds good, I will test it when it's merged

JonasVautherin (approved):

PR #1731, opened 2022-04-07
core: fix uninitialized command

### LINKED ISSUE #1729: std::optional for CommandLong always returns true. Causes fault on Ardupilot
[Line 429](https://github.com/mavlink/MAVSDK/blob/4dcee2b14945a80276f1b30abd99088ed7009d9f/src/mavsdk/core/mavlink_command_sender.cpp#L440) of mavlink_commands.cpp,  is a check to see if particular parameter on the command long has been set using the contextural conversion of std::optional. If it has not been set, the the value is initialised to certain values depending on whether the auto pilot is ardupilot or not. 

It does appear that this check always returns true, regardless if the param has been set. This maybe due to the initialisation of the values in CommandLong to NAN. This will result in ardupilot autopilots throwing a SIGFPE and closing the connection. This may be the cause of some of the problems in #728 and elsewhere particularly when SITL is used. 

I'm not sure of the correct solution but perhaps not initialising those parameters is probably an option?

julianoes:
Fixes #1729.

### DIFF
--- a/src/mavsdk/core/system_impl.cpp
+++ b/src/mavsdk/core/system_impl.cpp
@@ -512,7 +512,7 @@ bool SystemImpl::send_message(mavlink_message_t& message)
 void SystemImpl::send_autopilot_version_request()
 {
     // We don't care about an answer, we mostly care about receiving AUTOPILOT_VERSION.
-    MavlinkCommandSender::CommandLong command;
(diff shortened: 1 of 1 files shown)


JonasVautherin (approved):
Nice catch

julianoes:
@JonasVautherin credits to @PeterQFR.

Issue #1728, opened 2022-04-03
Camera::set_setting_async send request only to the first camera


persuader72:
Working with a system with two cameras (#1705) i noticed that the camera settings requests using (`Camera::set_setting_async`) are  always sent to the first camera, regardless the real camera target.

The problem resides in the  `MAVLinkParameters` class that in the` MAVLinkParameters::do_work` function use predefined constants depending on the `work->defined` value has target components.

I want to try to add a more generic way to set the component target in  the MAVLinkParameters class.  

My question is: It's better to (1) add a component id argument (with a default value) to already existent functions like `SystemImpl::set_param_async`, or is better to (2) add a new whole set of functions that will accept also a component id?

julianoes:
Ah yes, that sounds like a correct assessment.

Mmm, overloading with default arguments or add more functions? :thinking: I would probably use these functions:
https://github.com/mavlink/MAVSDK/blob/4dcee2b14945a80276f1b30abd99088ed7009d9f/src/mavsdk/core/system_impl.h#L169-L172
And then add the overload with default. I wouldn't use the functions which already have the default for `extended`.

PR #1735, opened 2022-04-08
Fix multicamera options

persuader72:
As seen in #1728 the MAVLinkParameters is using some magic numbers to define the designated target component of the message,

This prevents the Camera plugin to read the parameters of the any camera other than the first.

This PR add some overloaded functions for the `set_param` and `get_params` actions that accept an additional target `component_id` parameter. 

The Camera plugin as been updated to make use of this new functions.


julianoes:
@persuader72 I changed the APIs a little bit, I hope your ok with my suggestion?

persuader72:
Perfect!  Thanks

PR #1741, opened 2022-04-18
Mavlink: support for MAVLink dialects

julianoes:
This swaps from using the [repo of generated MAVLink headers](https://github.com/mavlink/c_library_v2/) to the [repo containing the MAVLink xml specs](https://github.com/mavlink/mavlink/).

This allows to select the MAVLink dialect at configure time like this:

```
cmake -Bbuild -H. -DMAVLINK_DIALECT=common
```

Update:
- the headers are generated at configure time (int the same way as SUPERBUILD but not depending on the SUPERBUILD variable)
- If Python (and python-future) are not available during configure/build, the mavlink headers can be supplied using `-DMAVLINK_HEADERS=path/to/headers`.


JonasVautherin:
_Warning: annoying opinion :see_no_evil:_

If I understand correctly, this says: "instead of taking the headers from the mavlink submodule, generate them from it", with the intent: "one can easily make a fork, point the mavlink submodule to somewhere else, and use mavlink-passthrough with custom mavlink messages".

I would find it cleaner to treat mavlink as a dependency like the others [here](https://github.com/mavlink/MAVSDK/tree/main/third_party). Instead of dealing with the submodule, it would then just be a URL in the corresponding CMakeLists.

This, way, we would move the fairly convoluted logic of adding a custom cmake command to the dependency script (which is a helper, but not a mandatory part of the build process, as opposed to this solution), and it would also make the mavlink dependency more consistent with the other (instead of linking from the cmake binary folder, it would just take mavlink from `third_party`).

But I understand this is completely different from this approach, so that's just my opinion. If this works, let's go for it.

julianoes:
I was thinking about that and decided against it, for now. My reasoning was that:
- It is a bit more difficult to do because the paths are a bit trickier. Not that I have to install the headers publicly for mavlink_passthrough. I want to get rid of that at some point but not quite yet.
- It's actually a bit funny to treat mavlink as a "third party" here, when really MAVSDK is sorta built around it, and is literally in the same GitHub organziation :smile:. Given we also have the proto files as a submodule I think we are ok with submodules, otherwise we would have to find a similar way for that submodule as well.

So maybe the takeaway is that it shouldn't be called third_party?

julianoes:
FYI I changed it so:
- the headers are generated at configure time (int the same way as SUPERBUILD but not depending on the SUPERBUILD variable)
- If Python (and python-future) are not available during configure/build, the mavlink headers can be supplied using `-DMAVLINK_HEADERS=path/to/headers`.

PR #1747, opened 2022-04-20
workflows: workaround for git trust CI issue

julianoes:
Let's see if this fixes it...

Related to https://github.com/actions/checkout/issues/760.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -173,6 +173,10 @@ jobs:
       - uses: actions/checkout@v2
         with:
           submodules: recursive
+      - name: git permission workaround
+        run: |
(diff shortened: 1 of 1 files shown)


JonasVautherin:
Did we have that issue? Haven't seen it. But if it's needed...

I would have hoped that github could fix it by setting those variables automatically (since github does not seem to be vulnerable), but maybe it's just a matter of time :innocent:.

julianoes:
I rebuilt the docker images in #1741 which might have also played a role. And then there are some things executed as a user and some as root. In theory the action/checkout thing deals with it but it doesn't seem to work for containers with git inside newer or older? :thinking:

PR #1752, opened 2022-04-22
Fix vehicle_type detection for systems

bazfp:
It seems like `vehicle_type` can jump around a lot if multiple systems are reporting vehicle_types differently. This allows only autopilots the ability to set vehicle types for system.


julianoes (approved):
That looks right.

PR #1760, opened 2022-05-03
Fix to allow non-flight controller heartbeats to be processed

pweids:
The change from https://github.com/mavlink/MAVSDK/pull/1752 inadvertantly threw a warning for all heartbeats that came from an "INVALID" autopilot. However, this is acceptable according to the [Heartbeat protocol.](https://mavlink.io/en/messages/common.html#HEARTBEAT) This small commit fixes that issue.

### DIFF
--- a/src/mavsdk/core/system_impl.cpp
+++ b/src/mavsdk/core/system_impl.cpp
@@ -216,15 +216,15 @@ void SystemImpl::process_heartbeat(const mavlink_message_t& message)
     // Only set the vehicle type if the heartbeat is from an autopilot component
     // This check only works if the MAV_TYPE::MAV_TYPE_ENUM_END is actually the
     // last enumerator.
-    if (heartbeat.autopilot != MAV_AUTOPILOT_INVALID &&
-        MAV_TYPE::MAV_TYPE_ENUM_END > heartbeat.type) {
+    if (MAV_TYPE::MAV_TYPE_ENUM_END < heartbeat.type) {
+        LogErr() << "type received in HEARTBEAT was not recognized";
+    } else {
         auto new_vehicle_type = static_cast<MAV_TYPE>(heartbeat.type);
-        if (_vehicle_type != new_vehicle_type)
+        if (heartbeat.autopilot != MAV_AUTOPILOT_INVALID && _vehicle_type != new_vehicle_type) {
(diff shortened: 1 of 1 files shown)


JonasVautherin (approved):

bazfp:
Looks good to me!

bazfp (changes requested):
Actually, this reverts my changes and allows vehicle type to be changed by non autopilot heartbeats again, which makes no sense. Can you brace the vehicle change so it only happens if the autopilot sends it

pweids:
You’re right, it would. I misinterpreted the intent of your change. Thought
you just wanted to give a warning. I’ll update. And thanks for the review!

pweids:
@bazfp please confirm the update fixes it

pweids:
I'm sure I'm missing something, but the logic that prints the error message is the same (heartbeat is from an autopilot and the vehicle type has changed). Furthermore, it's odd that the types are not printed out by the warnings. And it works just fine with you in the other version? How can I replicate your test?

See anything I'm missing?

julianoes:
> Is the intent of MAVSDK to only support autopilot types? Why would a system not want to receive non autopilot components heartbeats? For example, how would you connect to component 192, which is QGC's component ID with this implementation?

No, that's certainly not the intent. A `System` should represent all components, and they can all send heartbeats.

julianoes:
I will revisit this in the next days, thanks for the PR.

julianoes:
I've added two commits to prevent it from warning initially, and to fix the actual warning so it makes more sense.

@bazfp @JonasVautherin could you check if this works for you like that?

julianoes (approved):

JonasVautherin (approved):
Seems to be working fine for me, both on sim and on the bench :+1:

julianoes:
Ok, I'm going to merge it and revisit later if this doesn't work for you @bazfp.

bazfp:
This looks like it works for me. The intent of the original change was to stop rogue MAVLink components incorrectly setting the entire system vehicle type for MAVSDK. It was my mistake that it blocked them :)

pweids:
Thanks, everyone!

bazfp (commented):
Just ran into some issues while testing this:

julianoes:
@bazfp ok what about it? :thinking:

PR #1764, opened 2022-05-05
Return Timeout code when timing out, instead of ConnectionError

JonasVautherin:


julianoes (approved):
I actually stumbled on this as well just a few days ago. Thank you!

PR #1767, opened 2022-05-06
core: Add callback list

julianoes:
This allows to replace callback subscriptions with lists, so that multiple callbacks can be registered.

FYI @devbharat: I think this might be the minimal implementation of what we need for multiple subscriptions.

### DIFF
--- a/src/mavsdk/core/callback_list.h
+++ b/src/mavsdk/core/callback_list.h
@@ -1,0 +1,94 @@
+#pragma once
+
+#include <cstdint>
+#include <functional>
+#include <memory>
+#include <mutex>
+#include <utility>
+#include <vector>
+
+namespace mavsdk {
+
+template<typename... Args> class CallbackList;
+
+template<typename... Args> class Handle {
+public:
+    explicit Handle(uint64_t id) : _id(id) {}
+
+private:
+    uint64_t _id;
+
+    friend CallbackList<Args...>;
+};
+
+template<typename... Args> class CallbackList {
+public:
+    Handle<Args...> subscribe(const std::function<void(Args...)>&& callback)
+    {
+        check_removals();
+
+        std::lock_guard<std::mutex> lock(_mutex);
+        auto handle = Handle<Args...>(_last_id++);
+        _list.emplace_back(handle, callback);
+        return handle;
+    }
+
+    void unsubscribe(Handle<Args...> handle)
+    {
+        // We want to allow unsubscribing while in a callback. This would
+        // normally lead to a deadlock. Therefore, we just try to grab the
+        // lock here, and if it's already taken, we schedule the removal
+        // for later.
+        std::unique_lock<std::mutex> lock(_mutex, std::try_to_lock);
+        if (lock.owns_lock()) {
+            _list.erase(
+                std::remove_if(
+                    _list.begin(),
+                    _list.end(),
+                    [&](auto& pair) { return pair.first._id == handle._id; }),
+                _list.end());
+        } else {
+            std::lock_guard<std::mutex> remove_later_lock(_remove_later_mutex);
+            _remove_later.push_back(handle._id);
+        }
+    }
+
+    void operator()(Args... args)
+    {
+        check_removals();
+
+        std::lock_guard<std::mutex> lock(_mutex);
+        for (const auto& pair : _list) {
+            pair.second(args...);
+        }
+    }
+
+private:
+    void check_removals()
+    {
+        std::lock_guard<std::mutex> remove_later_lock(_remove_later_mutex);
+
+        // We could probably just grab the lock here but it's safer not to
+        // acquire both locks to avoid deadlocks.
+        if (_mutex.try_lock()) {
+            for (auto& id : _remove_later) {
+                _list.erase(
+                    std::remove_if(
+                        _list.begin(),
+                        _list.end(),
+                        [&](auto& pair) { return pair.first._id == id; }),
+                    _list.end());
+            }
+            _mutex.unlock();
+        }
+    }
+
+    mutable std::mutex _mutex{};
+    uint64_t _last_id{0};
+    std::vector<std::pair<Handle<Args...>, std::function<void(Args...)>>> _list{};
+
+    mutable std::mutex _remove_later_mutex{};
+    std::vector<uint64_t> _remove_later{};
+};
+
+} // namespace mavsdk
--- a/src/mavsdk/core/CMakeLists.txt
+++ b/src/mavsdk/core/CMakeLists.txt
@@ -120,6 +120,7 @@ install(FILES
 )
 
 list(APPEND UNIT_TEST_SOURCES
+    ${PROJECT_SOURCE_DIR}/mavsdk/core/callback_list_test.cpp
     ${PROJECT_SOURCE_DIR}/mavsdk/core/mavsdk_time_test.cpp
     ${PROJECT_SOURCE_DIR}/mavsdk/core/mavsdk_math_test.cpp
     ${PROJECT_SOURCE_DIR}/mavsdk/core/mavlink_channels_test.cpp
--- a/src/mavsdk/core/callback_list_test.cpp
+++ b/src/mavsdk/core/callback_list_test.cpp
@@ -1,0 +1,71 @@
+#include "callback_list.h"
+#include "log.h"
+#include <gtest/gtest.h>
+
+using namespace mavsdk;
+
+TEST(CallbackList, SubscribeCallUnsubscribe)
+{
+    unsigned first_called = 0;
+    unsigned second_called = 0;
+
+    CallbackList<int, double> cl;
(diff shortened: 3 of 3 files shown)


src/mavsdk/core/callback_list.h
```diff
@@ -0,0 +1,94 @@
...
+
+private:
+    uint64_t _id;
+
+    friend CallbackList<Args...>;
+};
+
+template<typename... Args> class CallbackList {
+public:
+    Handle<Args...> subscribe(const std::function<void(Args...)>&& callback)
```
devbharat, line 26: Is it necessary to already fix the return type?

src/mavsdk/core/callback_list.h
```diff
@@ -0,0 +1,94 @@
...
+#include <utility>
+#include <vector>
+
+namespace mavsdk {
+
+template<typename... Args> class CallbackList;
+
+template<typename... Args> class Handle {
+public:
+    explicit Handle(uint64_t id) : _id(id) {}
```
devbharat, line 16: There should perhaps be a way to couple the handle to a callbacklist, so they can't be shuffled(use the wrong handle with the wrong cblist). Something like a random key that the cblist generates and gives to all its handles that is also verified when using the handle to remove it. That would prevent people from trying to remove callback with the wrong handle at runtime. You could also use the function signature and that'd add some more protection at compile time.

devbharat: Anther thing that's perhaps needed is a way to know if the handle is 'alive', of if it has been used to free up the callback already.

src/mavsdk/core/callback_list.h
devbharat: Also, would you be able to copy the callbacklist and handle?

devbharat:
Just thinking about it here, would it not be possible to simply wrap eventpp under-the-hood in this class's implementation, and only expose out wrapped handles (keep a map of eventpp handles to your handles internally)? Then anyone inside mavsdk has the full eventpp library to work with, and from outside you only have whatever you expose with the wrapper.

src/mavsdk/core/callback_list.h
julianoes: > Something like a random key that the cblist generates and gives to all its handles that is also verified when using the handle to remove it.

I thought about that but then decided it was probably overkill because the handle is already type checked, as the handle depends on the callback `Args`. The only way to mix it up is if you have e.g. two instances of mavsdk and subscribe to two methods with the same signature. You could - at runtime - avoid it using a random key, indeed.

> Anther thing that's perhaps needed is a way to know if the handle is 'alive',

Hm, you could yes, What would you need it for? I would argue that if you just want to unsubscribe you can just unsubscribe multiple times and it will ignore it. Alternatively, we could make it so that you have to `std::move` the handle into the unsubscribe method, so that the compiler might warn you, if you use it again afterwards, so:

```
void unsubscribe_foo(Handle&& handle);
```

julianoes:
> Then anyone inside mavsdk has the full eventpp library to work with, and from outside you only have whatever you expose with the wrapper.

Yes, I thought about that. The question is what the benefits would be, or whether we need them. If yes, then it's worthwhile.

> I prototype the [3rd most popular answer](https://stackoverflow.com/a/41292751) (imo the only 'real' suggestion) and it works well, the user of the library (unit test in this case) just needs to add a file with the template instantiation for the types of callbacks they need. Its a bit special and probably not everyone likes it, but I tested it and it works.

Thanks a lot for that! So it looks like we would have to list all instantiated templates in some cpp file. Since the API is auto-generated, this should be easy, I think.

I think the question is now whether the added complexity is worthwhile.

Note: I'm gonna fetch your commit and push it to this branch.

src/mavsdk/core/callback_list.h
julianoes: The return type is to have better type safety, but we could just use generic handles instead. Depends what we prefer.

src/mavsdk/core/callback_list.h
devbharat: > Hm, you could yes, What would you need it for?

I think the obvious way to use it would be to let it carry some state, then you could assign some meaning to if(handle.empty()) do_something(); later in the code. But I can also see why you might explicitly not want to allow such meanings to sit on top of the handle and say use another variable to do it. Its a nice to have.

JonasVautherin: Is there another use-case than doing something like below?

```c++
if (handle.empty()) {
    unsubscribe(handle);
}
```

Because then, I think it's nicer to just say that calling `unsubscribe(handle)` with an already unsubscribed handle is simply a noop.

I can't imagine a scenario where I would do:

```
if (handle.empty()) {
  // do some complex logic here that relies on the handle having been unsubscribed already
}
```

:thinking:

JonasVautherin:
> Yes, I thought about that. The question is what the benefits would be, or whether we need them. If yes, then it's worthwhile.

Yeah, I would like to question the benefits of that. Adding eventpp as a dependency adds maintenance cost for everyone using MAVSDK. If in return it is used heavily by MAVSDK contributors in a way that makes them more efficient than without eventpp, then of course it's worth considering. But if that's mostly because eventpp is cool and makes a few lines look better internally, then I am a bit concerned about adding it as a dependency.

So maybe I could ask this: what complex features are planned that will benefit from eventpp?

JonasVautherin (approved):

devbharat:
> what complex features are planned that will benefit from eventpp?

@JonasVautherin perhaps I can elaborate on what I am working on and why I got started on this thing. I am refactoring the mavlink part of a sort of massive monolithic MAVLink handling class on companion compute side that does RPC with a bunch of backend servers that carry out various tasks. The idea is to refactor the mavlink side into a bunch of different plugins on MAVSDK by functionality, and have them do RPC with the backend servers. That sounded great to begin, but we quickly ran into some limitations. The interaction between the different plugins is almost non-existent on MAVSDK (I understand until now this perhaps has been intentional). I'll come back to this.
But in fact, interactions even with the external world via the gRPC interface was severely limited, due to having only one client being served, even though many can connect. The reason was, every time some client attached a callback, it overwrote the callback that the previous client had attached. Take for example, 3 clients connecting to the core service, while all 3 stay connected, only the last one gets at the actual updates. This prompted to begin using callbacklists instead of callbacks, which was actually pretty easy to switch to as the api of interacting with callbackslists from eventpp is almost exactly the same as that with a normal callback.
Coming back to the plugins, the only way they can interact right now, is by the system class or the mavsdk class (and the server plugin impl base class), but I think that is quite limiting as well. Plugin A can consume some mavlink msgs have some complicated logic that leads to some state transitions (say different stages of preflight checks) and at each of those transitions, other plugins B and C (say camera or storage or PPK or RemoteID needs to do something) might want to take some actions on, or transition their own states etc. That was neither possible via the external gRPC interface nor internally within the plugins inside MAVSDK.  Putting all that into the system/mavsdk/pluginbase class felt a bit backwards. So I decided to provide an internal API for plugins to register with the base class and be 'discoverable' from one another, and attack callbacks to each other similar in appearance to how external consumers would attach callbacks via gRPC. That now enabled several external consumers and internal plugins to all be able to latch on to the API that any plugin provides.
I can imagine this goes perhaps 'a bit sideways' to how MAVSDK has been growing until now, but this was the least convoluted way I could find to enable the kind of functionality we had in the spaghetti class, and still have somewhat clear interfaces.

JonasVautherin:
@devbharat I see, you need to have a way for plugins to talk to other plugins :thinking:.

I haven't looked into it, so I'm trying to learn from your experience: say plugin Foo needs plugin Bar internally, doesn't it "just" work to pass a reference to Bar when instantiating Foo? Then Foo can call Bar methods, right? I'm sure there is an issue with that but I don't see it :sweat_smile:.

> The reason was, every time some client attached a callback, it overwrote the callback that the previous client had attached.

Yes, that's a problem, I'm very happy it's getting fixed now! In Java and Swift it is not a problem because we can share the callback with Rx, but e.g. [the oldest mavsdk-python open issue](https://github.com/mavlink/MAVSDK-Python/issues/60) is about this :grin:.

Though I like @julianoes version in this PR, that does not need eventpp :innocent:. But then if eventpp can help improve the internal event loop and all, why not!

julianoes:
> doesn't it "just" work to pass a reference to Bar when instantiating Foo?

It would, but the plugins are auto-generated, so you can't easily change the constructors to pass in other plugins.

julianoes:
I'm going to merge this as is for now, and then I'll continue with PRs on top of it, maybe even replacing this implementation with eventpp since @devbharat demonstrated that we can hide it away.

JonasVautherin:
> It would, but the plugins are auto-generated, so you can't easily change the constructors to pass in other plugins.

Oh right... I wonder if we could use some kind of annotation in the proto files to define such dependencies :thinking:. And then how that would look in the language bindings...

PR #1770, opened 2022-05-11
Update Follow me code to match upstream changes

junwoo091400:
## Context
Recently merged PR : https://github.com/PX4/PX4-Autopilot/pull/19260, introduced whole new set of parameters to the follow-me (called FollowTarget in upstream) code. Therefore the aim is along with https://github.com/mavlink/MAVSDK-Proto/pull/287, update the Follow Me code in MAVSDK

## What changed
1. Renaming of Parameters "NAV_*" into "FLW_TGT_*"
2. Addition of two parameters : FLW_TGT_ALT_M and FLW_TGT_MAX_VEL


junwoo091400:
@JonasVautherin Can we have a generic function that handles these basic Parameter get behavior? It's a lot of copy & paste in both get & set in the implementation code :face_with_thermometer: 

https://github.com/mavlink/MAVSDK/pull/1770/files#diff-0e2a47a501e78daff4d3fed04b889fb571366cacae01774b72da340d893fc9edR45-R53

src/mavsdk/plugins/follow_me/follow_me_impl.cpp
```diff
@@ -219,7 +258,7 @@ bool FollowMeImpl::is_config_ok(const FollowMe::Config& config) const
     if (config.min_height_m < CONFIG_MIN_HEIGHT_M) {
         LogErr() << debug_str << "Err: Min height must be atleast 8.0 meters";
     } else if (config.follow_distance_m < CONFIG_MIN_FOLLOW_DIST_M) {
-        LogErr() << debug_str << "Err: Min Follow distance must be atleast 1.0 meter";
+        LogErr() << debug_str << "Err: Min Follow distance must be atleast 2.0 meter";
```
JonasVautherin, line 261: Nitpick:

```suggestion
        LogErr() << debug_str << "Err: Min Follow distance must be at least 2.0 meter";
```

JonasVautherin: BTW can we print `CONFIG_MIN_FOLLOW_DIST_M` here? Something like

```
LogErr() << debug_str << "Err: Min Follow distance must be at least " << CONFIG_MIN_FOLLOW_DIST_M << " meter";
```

JonasVautherin:
> Can we have a generic function that handles these basic Parameter get behavior?

I don't see a way :sweat_smile:

junwoo091400:
Updated the implementation code & proto commit!

src/mavsdk/plugins/follow_me/follow_me_impl.cpp
junwoo091400: Changes applied!

JonasVautherin (dismissed):

src/mavsdk/plugins/follow_me/follow_me.cpp
```diff
@@ -57,42 +57,44 @@ FollowMe::Result FollowMe::stop() const
 }
 
 std::ostream&
-operator<<(std::ostream& str, FollowMe::Config::FollowDirection const& follow_direction)
+operator<<(std::ostream& str, FollowMe::Config::FollowAltitudeMode const& follow_altitude_mode)
 {
-    switch (follow_direction) {
-        case FollowMe::Config::FollowDirection::None:
```
potaito, line 63: @JonasVautherin Is there a way to keep these for now, but mark them as deprecated?
It would be fairly straight forward to map the directions to an angle as a float.

junwoo091400: There was a discussion about this and that was the first plan, but since MAVSDK v2 is being released, it was decided that breaking the compatability would be ok 🤔

https://github.com/mavlink/MAVSDK-Proto/pull/287#issuecomment-1132930474

JonasVautherin:
>  however, It seems like there's a merge conflict within MAVSDK due to the difference in the generated files:

@junwoo091400 yeah it's not your fault, you just got unlucky because I updated MAVSDK's dependencies (including protobuf) and it changed the generated code. I rebased `main` and generated the code again, I believe this should be fine.

On your end you'll have to make a clean build (because you need to rebuild the dependencies) :+1:.

I hope this helps :blush:

JonasVautherin:
> and I totally didn't expect the generated files to be different

Are your generated files different? :thinking:

junwoo091400:
> @junwoo091400 So you'd say it's ready to be merged?

Yes it's ready to be merged 👍 

p.s. SITL tests failing, but that's expected since the container used in v1.11 and v1.12 doesn't have the current upstream PR's implementations in Follow Me (new Parameters): https://github.com/mavlink/MAVSDK/blob/main/.github/workflows/main.yml#L683

This FollowMe test failure will keep show up in other PRs in the MAVSDK, how should we deal with this?

potaito:
> This FollowMe test failure will keep show up in other PRs in the MAVSDK, how should we deal with this?

What do you mean @junwoo091400? It's not merged yet, so it should not be causing issues in any other MAVSDK PRs :thinking:

julianoes:
We need to merge https://github.com/mavlink/MAVSDK-Proto/pull/287 and then update the submodule here, and then we can do the final check and merge here.

JonasVautherin:
> It needs to go into PX4 first, usually. Otherwise we end up with stuff that's not supported.

Agree. Let's wait until https://github.com/PX4/PX4-Autopilot/pull/18026 is merged then :+1:

junwoo091400:
> What do you mean @junwoo091400? It's not merged yet, so it should not be causing issues in any other MAVSDK PRs thinking

Oh what I meant was that once this is merged, the follow me tests will fail since the MAVSDK always tests against the v1.11 and v1.12 PX4 SITL

julianoes:
> Oh what I meant was that once this is merged, the follow me tests will fail since the MAVSDK always tests against the v1.11 and v1.12 PX4 SITL

Right. So we need to ignore these tests for v1.11 and v1.12.

For now we can disable the integration test but we should come up with some sort of "test matrix", so a table which says which test works for which PX4 and ArduPilot version.

junwoo091400:
Upstream PR has been merged. Good for merge to MAVSDK as well 👍

junwoo091400:
Regarding the 'proto' having conflicts, should I:
1. Rebase the proto PR
2. Reference that submodule commit?

junwoo091400:
`proto` merge conflict solved by updating the proto submodule commit in the first commit of this PR during rebase

junwoo091400:
Quick question, what is the point of having an integration test when PX4 has it's own MAVSDK SITL tests as well? Since I noticed that they are doing similar tests but are maintained in separate repositories 🤷

junwoo091400:
> For now, what you could try is to check the version in the integration test using `info->get_version()` or something like that and then exit or do the test.

With the latest commit, this has been applied, and should skip the SITL tests for v1.11 and v1.12 for FollowMe (and hence CI should pass).

I went for this change as I wasn't sure rightaway how to actually make the matrix criteria work:

I thought of modifying the executable `integration_tests_runner` for each PX4 version SITL tests, but this seemed a bit **inefficient** considering for every SITL test (for v1.11, v1.12, v1.13), the `integration_tests_runner` would need to be re-built. So I wanted to ask if this is ok or not before making changes 👍 Any thoughts on this?

potaito:
> Quick question, what is the point of having an integration test when PX4 has it's own MAVSDK SITL tests as well? Since I noticed that they are doing similar tests but are maintained in separate repositories shrug

What do you mean exactly? I extened the existing PX4 SITL tests for follow-me

junwoo091400:
> > Quick question, what is the point of having an integration test when PX4 has it's own MAVSDK SITL tests as well? Since I noticed that they are doing similar tests but are maintained in separate repositories shrug
> 
> What do you mean exactly? I extened the existing PX4 SITL tests for follow-me

I mean that the integration tests in mavsdk (Used in Mavsdk CI) is totally separate from mavsdk tests in PX4 (Used in PX4 CI)

potaito:
Ah sorry, I thought we were in the PX4 repo :sweat_smile: 
Yes, this is a legitimate question. I would assume that MAVSDK and PX4 test different things, so the tests will never be fully identical.

julianoes:
Thanks @junwoo091400 for the contribution!

PR #1772, opened 2022-05-17
Refactor MAVLinkParameters into client and server classes

julianoes:
This splits the MAVLinkParameters class into MavlinkParameterClient and MavlinkParameterServer.

This brings:
- Support to provide parameters in components.
- Get/set parameters for a specific component.
- Both with normal as well as extendended parameters.

Big thanks to @Consti10 for starting this work!


julianoes:
The current state is that it compiles again and the tests seem to pass but now it needs simplifying and cleanup.

src/mavsdk/plugins/calibration/calibration_impl.cpp
```diff
@@ -407,18 +407,18 @@ void CalibrationImpl::receive_statustext(const MavlinkStatustextHandler::Statust
...
-                    _system_impl->param_changed("CAL_MAG0_ID");
+                    //_system_impl->param_changed("CAL_MAG0_ID");
                     break;
                 case State::LevelHorizonCalibration:
-                    _system_impl->param_changed("SENS_BOARD_X_OFF");
-                    _system_impl->param_changed("SENS_BOARD_Y_OFF");
-                    _system_impl->param_changed("SENS_BOARD_Z_OFF");
+                    //_system_impl->param_changed("SENS_BOARD_X_OFF");
+                    //_system_impl->param_changed("SENS_BOARD_Y_OFF");
+                    //_system_impl->param_changed("SENS_BOARD_Z_OFF");
```
JonasVautherin, line 421: hmm? :sweat_smile:

JonasVautherin (commented):
Wow that's a big one, with quite some improvements! It depends on https://github.com/mavlink/MAVSDK-Proto/pull/296, right?

Just a couple questions:
1. Is that right that it is the responsibility of the drone to now have conflicting param names between components? It seems to me that we save them all together, so if a camera has a param called `MODE` and the autopilot also has it, then they will conflict. Correct?
2. What happens when `select_component` is called? Does that impact the params already in the queue, or only the new ones? Say I select the autopilot, set a param (which puts it on the worker queue), then select the camera to set another param. Am I guaranteed that the first param will be sent to the autopilot?

julianoes:
@JonasVautherin thanks for the review!

> 1. Is that right that it is the responsibility of the drone to now have conflicting param names between components? It seems to me that we save them all together, so if a camera has a param called MODE and the autopilot also has it, then they will conflict. Correct?

That's a good point I haven't thought about. Sounds like I ought to change it.

> 2. What happens when select_component is called? Does that impact the params already in the queue, or only the new ones? Say I select the autopilot, set a param (which puts it on the worker queue), then select the camera to set another param. Am I guaranteed that the first param will be sent to the autopilot?

Good point. I need to check again what I did but better would definitely be if it only affects the new ones. I'll see if I can change it.

src/mavsdk/plugins/calibration/calibration_impl.cpp
julianoes: Hmm indeed :thinking:

src/mavsdk/core/mavlink_parameter_cache.cpp
```diff
@@ -0,0 +1,153 @@
...
+std::vector<MavlinkParameterCache::Param>
+MavlinkParameterCache::all_parameters(bool including_extended) const
+{
+    if (including_extended) {
+        return _all_params;
+    } else {
+        // TODO: we might want to cache this
+        std::vector<MavlinkParameterCache::Param> vec{};
+        for (const auto& entry : _all_params) {
+            if (entry.value.needs_extended() && !including_extended) {
```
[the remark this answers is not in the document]
julianoes: Right, thanks.

JonasVautherin (dismissed):

PR #1773, opened 2022-05-17
Add version number fallback

bazfp:
It's impossible to build from source.tar.gz or .zips at current as it is dependent on the repo being a git repository. This adds a fallback to `0.0.0` in this case as version number is not known and allows dependent modules to build

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,7 +14,12 @@ execute_process(
     WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
     OUTPUT_VARIABLE VERSION_STR
     OUTPUT_STRIP_TRAILING_WHITESPACE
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -14,7 +14,12 @@ execute_process(
     WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
     OUTPUT_VARIABLE VERSION_STR
     OUTPUT_STRIP_TRAILING_WHITESPACE
+    RESULT_VARIABLE GITVER
 )
+if(NOT GITVER EQUAL "0")
```
JonasVautherin, line 19: Nitpick: to be consistent with the rest of the file

```suggestion

if (NOT GITVER EQUAL "0")
```

JonasVautherin (approved):

PR #1774, opened 2022-05-17
Fix system discovery

JonasVautherin:
Without this fix, MAVSDK has issues detecting more than one system (it seems like one `notify_on_discover` is lost, because one of the systems was `always_connected` and is renamed when a new one is detected).

This fix just clears the "fake system" when a new one is discovered, so that the new one triggers the discovery callback.

### DIFF
--- a/src/mavsdk/core/mavsdk_impl.cpp
+++ b/src/mavsdk/core/mavsdk_impl.cpp
@@ -300,13 +300,10 @@ void MavsdkImpl::receive_message(mavlink_message_t& message, Connection* connect
 
     std::lock_guard<std::recursive_mutex> lock(_systems_mutex);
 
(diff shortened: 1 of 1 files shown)


julianoes (approved):
This should be ok, however, in the future we should not even require any of this anymore as you can send out heartbeats to a remote port just from a MAVSDK server component without having to have a system as such.

JonasVautherin:
Right, but what about client components that want to initiate the discovery? Are you saying that the client components should always listen and the server components should always broadcast?

julianoes:
For what it's worth: I'm all for separating discovery and connection, best with backwards compatibility.

Yer something like that might work indeed. There are however a couple more dependencies from systems to mavsdk that would have to moved as well (e.g. timeout or call every handlers).

PR #1776, opened 2022-05-18
Generate grpc_server.h/grpc_server.cpp, support selectively building plugins

julianoes:
[cmake/tools: generate grpc_server, list plugins](https://github.com/mavlink/MAVSDK/commit/ebf1400bd3f238f06ba3193a0e73dfd6d7cde837)

This is refactoring around the way the plugins are built into the mavsdk
library as well as the mavsdk_server executable.

The changes contain the following:
- Auto-generation of grpc_server.h and grpc_server.cpp from jinja2 templates. This means that the mavsdk_server is always in sync and new plugins don't have to be added to it manually anymore.
- As a side effects this removes the ugly rewriting of CMakeLists.txt based on the auto-generated plugins and instead uses a separate plugins.txt file that is used throughout.
- The cmake variable ENABLED_PLUGINS can be used to manually select the plugins to be built for users requiring to do so.

To use this, the enabled plugins can be passed at configure time:
```
cmake -DENABLED_PLUGINS:list="action;telemetry;info;" ...
```

FYI @devbharat


JonasVautherin (approved):
Nice! Just not passing the CI (yet), but otherwise it's pretty cool :sunglasses:

src/mavsdk_server/src/grpc_server.cpp
```diff
@@ -19,33 +21,126 @@ int GrpcServer::run()
...
+
+#ifdef CALIBRATION_ENABLED
     builder.RegisterService(&_calibration_service);
+#endif
+
+#ifdef CAMERA_ENABLED
     builder.RegisterService(&_camera_service);
+#endif
+
+#ifdef CAMERA_SERVER_ENABLED
```
devbharat, line 41: It looks a bit not elegant, maybe the 3 lines can be wrapped into some macro Call like REGISTER_IF_ENABLE(…) but it’s auto generated so also does not matter so much.

julianoes: Right, it's not pretty but it's just auto-generated, and at least it's not out of sync like it often was in the past because someone would add a plugin and they and us maintainers would both forget to update it.

The plugin header includes are not there at all, so not sure how it would be there.

devbharat: Yeah I agree.  The only way i could imagine moving this into the plugin would be through lazy plugin somehow, am not sure if that’s possible, is also rather cosmetic and not so important.

JonasVautherin (approved):

devbharat (approved):

src/mavsdk_server/src/grpc_server.cpp
julianoes: I don't quite understand how you imagine it. Feel free to make a PR or code example to show it. This can also happen later, of course.

devbharat: Something like this https://diffy.org/diff/43d4582733066

julianoes: @devbharat I think what you suggest makes sense, thanks!

@JonasVautherin can you check that change in diffy as well? You know this part of the code.

JonasVautherin: I'll check ASAP :+1:

devbharat: I was also not sure of that, but the unit test thing seems to work. Worst case we can have like a two step constructor that creates the service and registers it and then returns the object.

JonasVautherin (dismissed):

JonasVautherin (approved):

PR #1777, opened 2022-05-19
Fix FTP test segfault

julianoes:
@JonasVautherin I think the fix you suggested in #1774 does not work in general.

Removing a system is not really supported, so in the case of the FTP test, the ftp plugin is already created from the fake plugin and removing it has bad effects, in this case a segfault.

This PR reverts your commit and adds a commit which instead triggers a discovery notification when a fake system changes to a real system.

However, in the future, given the server components - I think - we can remove the fake system completely. However, this requires us to split the FTP plugin into client and server. That's something I wanted to do but haven't gotten to yet.


JonasVautherin (approved):
Oh right. I thought about that, and then settled on the other way, thinking that nobody would use the "fake system". But a server component does, indeed :man_facepalming:.

Thanks a lot!

PR #1779, opened 2022-05-19
Generate from protos

JonasVautherin:
For some reason, when generating the code from `main`, I get this diff.

@julianoes am I missing something, or should that be merged? I am a bit surprised that the CI did not catch it :thinking:


JonasVautherin:
Oh, @julianoes did you change `protoc-gen-mavsdk` and did not release it on PyPi?

julianoes:
Sorry, but the dependency with that is a little tricky, isn't it.

PR #1787, opened 2022-05-22
info: updated the message to obtain flight and autopilot info as per …

ykhedar:
…the mavlink spec.
The following messages are deprecated in the MAVLINK protocol

[MAV_CMD_REQUEST_FLIGHT_INFORMATION](https://mavlink.io/en/messages/common.html#MAV_CMD_REQUEST_FLIGHT_INFORMATION)
[MAV_CMD_REQUEST_AUTOPILOT_CAPABILITIES](https://mavlink.io/en/messages/common.html#MAV_CMD_REQUEST_AUTOPILOT_CAPABILITIES)

The suggested way to get the same information is using the following command message:
[MAV_CMD_REQUEST_MESSAGE](https://mavlink.io/en/messages/common.html#MAV_CMD_REQUEST_MESSAGE)

with appropriate message ID for requested message which would be for flight information the following message: 
[MAVLINK_MSG_ID_FLIGHT_INFORMATION](https://mavlink.io/en/messages/common.html#MAV_CMD_REQUEST_MESSAGE) 

_Note APM does not implement this yet. See Issue      https://github.com/ArduPilot/ardupilot/issues/15217_

and for autopilot version: 
[MAVLINK_MSG_ID_AUTOPILOT_VERSION 148](https://mavlink.io/en/messages/common.html#AUTOPILOT_VERSION)

This PR resolves this incompatibility with the MAVLINK protocol.

Happy to hear your thoughts :)

### DIFF
--- a/src/mavsdk/core/system_impl.cpp
+++ b/src/mavsdk/core/system_impl.cpp
@@ -469,7 +469,7 @@ void SystemImpl::send_autopilot_version_request()
     MavlinkCommandSender::CommandLong command{};
 
     command.command = MAV_CMD_REQUEST_MESSAGE; // MAV_CMD_REQUEST_AUTOPILOT_CAPABILITIES;
-    command.params.maybe_param1 = MAVLINK_MSG_ID_AUTOPILOT_VERSION;
+    command.params.maybe_param1 = {static_cast<float>(MAVLINK_MSG_ID_AUTOPILOT_VERSION)};
     command.target_component_id = get_autopilot_id();
 
     send_command_async(command, nullptr);
@@ -481,7 +481,7 @@ void SystemImpl::send_flight_information_request()
     MavlinkCommandSender::CommandLong command{};
 
     command.command = MAV_CMD_REQUEST_MESSAGE; // MAV_CMD_REQUEST_FLIGHT_INFORMATION;
-    command.params.maybe_param1 = MAVLINK_MSG_ID_FLIGHT_INFORMATION;
+    command.params.maybe_param1 = {static_cast<float>(MAVLINK_MSG_ID_FLIGHT_INFORMATION)};
     command.target_component_id = get_autopilot_id();
 
     send_command_async(command, nullptr);
--- a/src/mavsdk/core/system_impl.cpp
+++ b/src/mavsdk/core/system_impl.cpp
@@ -468,7 +468,7 @@ void SystemImpl::send_autopilot_version_request()
     // We don't care about an answer, we mostly care about receiving AUTOPILOT_VERSION.
     MavlinkCommandSender::CommandLong command{};
 
-    command.command = MAV_CMD_REQUEST_MESSAGE; //MAV_CMD_REQUEST_AUTOPILOT_CAPABILITIES;
+    command.command = MAV_CMD_REQUEST_MESSAGE; // MAV_CMD_REQUEST_AUTOPILOT_CAPABILITIES;
     command.params.maybe_param1 = MAVLINK_MSG_ID_AUTOPILOT_VERSION;
     command.target_component_id = get_autopilot_id();
 
--- a/src/mavsdk/core/system_impl.cpp
+++ b/src/mavsdk/core/system_impl.cpp
@@ -487,6 +487,7 @@ void SystemImpl::send_autopilot_version_request()
     if (fut.get() == MavlinkCommandSender::Result::Unsupported) {
         _old_message_520_supported = false;
         LogWarn() << "Trying alternative command (512).";
+        send_autopilot_version_request();
     }
 }
 
@@ -514,6 +515,7 @@ void SystemImpl::send_flight_information_request()
     if (fut.get() == MavlinkCommandSender::Result::Unsupported) {
         _old_message_528_supported = false;
         LogWarn() << "Trying alternative command (512)..";
+        send_flight_information_request();
     }
 }
 
--- a/src/mavsdk/core/system_impl.cpp
+++ b/src/mavsdk/core/system_impl.cpp
@@ -465,26 +465,56 @@ bool SystemImpl::send_message(mavlink_message_t& message)
 
 void SystemImpl::send_autopilot_version_request()
 {
-    // We don't care about an answer, we mostly care about receiving AUTOPILOT_VERSION.
-    MavlinkCommandSender::CommandLong command{};
+    auto prom = std::promise<MavlinkCommandSender::Result>();
+    auto fut = prom.get_future();
 
-    command.command = MAV_CMD_REQUEST_MESSAGE; // MAV_CMD_REQUEST_AUTOPILOT_CAPABILITIES;
-    command.params.maybe_param1 = {static_cast<float>(MAVLINK_MSG_ID_AUTOPILOT_VERSION)};
+    MavlinkCommandSender::CommandLong command{};
     command.target_component_id = get_autopilot_id();
 
-    send_command_async(command, nullptr);
+    if (_old_message_520_supported) {
+        // Note: This MAVLINK message is deprecated and would be removed from MAVSDK in a future
+        // release.
+        command.command = MAV_CMD_REQUEST_AUTOPILOT_CAPABILITIES;
+        command.params.maybe_param1 = 1.0f;
+    } else {
+        command.command = MAV_CMD_REQUEST_MESSAGE;
+        command.params.maybe_param1 = {static_cast<float>(MAVLINK_MSG_ID_AUTOPILOT_VERSION)};
+    }
+
+    send_command_async(command, [&prom](MavlinkCommandSender::Result result, float _val) {
+        prom.set_value(result);
+    });
+    if (fut.get() == MavlinkCommandSender::Result::Unsupported) {
+        _old_message_520_supported = false;
+        LogWarn() << "Trying alternative command (512).";
+    }
 }
 
 void SystemImpl::send_flight_information_request()
 {
-    // We don't care about an answer, we mostly care about receiving FLIGHT_INFORMATION.
-    MavlinkCommandSender::CommandLong command{};
+    auto prom = std::promise<MavlinkCommandSender::Result>();
+    auto fut = prom.get_future();
 
-    command.command = MAV_CMD_REQUEST_MESSAGE; // MAV_CMD_REQUEST_FLIGHT_INFORMATION;
-    command.params.maybe_param1 = {static_cast<float>(MAVLINK_MSG_ID_FLIGHT_INFORMATION)};
+    MavlinkCommandSender::CommandLong command{};
     command.target_component_id = get_autopilot_id();
 
-    send_command_async(command, nullptr);
+    if (_old_message_528_supported) {
+        // Note: This MAVLINK message is deprecated and would be removed from MAVSDK in a future
+        // release.
+        command.command = MAV_CMD_REQUEST_FLIGHT_INFORMATION;
+        command.params.maybe_param1 = 1.0f;
+    } else {
+        command.command = MAV_CMD_REQUEST_MESSAGE;
+        command.params.maybe_param1 = {static_cast<float>(MAVLINK_MSG_ID_FLIGHT_INFORMATION)};
+    }
+
+    send_command_async(command, [&prom](MavlinkCommandSender::Result result, float _val) {
+        prom.set_value(result);
+    });
+    if (fut.get() == MavlinkCommandSender::Result::Unsupported) {
+        _old_message_528_supported = false;
+        LogWarn() << "Trying alternative command (512)..";
+    }
 }
 
 void SystemImpl::set_connected()
@@ -532,7 +562,7 @@ void SystemImpl::set_connected()
     }
     if (enable_needed) {
         if (has_autopilot()) {
-            send_autopilot_version_request();
+            // send_autopilot_version_request();
         }
 
         std::lock_guard<std::mutex> lock(_plugin_impls_mutex);
--- a/src/mavsdk/core/system_impl.cpp
+++ b/src/mavsdk/core/system_impl.cpp
@@ -468,6 +468,19 @@ void SystemImpl::send_autopilot_version_request()
     auto prom = std::promise<MavlinkCommandSender::Result>();
     auto fut = prom.get_future();
 
+    send_autopilot_version_request_async(
+        [&prom](MavlinkCommandSender::Result result, float _val) { prom.set_value(result); });
+
+    if (fut.get() == MavlinkCommandSender::Result::Unsupported) {
+        _old_message_520_supported = false;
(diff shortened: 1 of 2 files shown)


julianoes (commented):
Thanks for that. It's good to update MAVSDK and make sure we don't use deprecated messages, however, this - unfortunately - would break compatibility with PX4 v1.10 where REQUEST_MESSAGE was not yet implemented.

Therefore, what would be optimal is to use REQUEST_MESSAGE by default but fall back to the older message if we don't get a response.

Easiest would probably be if we add some sort of support for Commands with fallbacks (so then the command sender would internally switch between the new and fallback commands.

JonasVautherin:
My concern with the fallback is this: what if the drone supports the new `REQUEST_MESSAGE` message, but the request is lost? MAVSDK will time out, not because it is unsupported, but because of an actual timeout. If MAVSDK now falls back to the old message and the drone does not support it, it will timeout again, this time because it is not supported. How do we handle that?

ykhedar:
If a particular mavlink command is unsupported, does PX4 send an error about command not being supported? At least with APM i get this error thrown in the MAVSDK logs like this:

`[09:59:06|Warn ] command unsupported (528). (mavlink_command_sender.cpp:198)`

Maybe this could be used as deciding criteria on using fallback?

JonasVautherin:
The "UNSUPPORTED" error code does exist, but as far as I can tell, it is not mandatory (at least I have seen many places where it is not used). So if you receive "UNSUPPORTED", you can conclude that it is not supported, but if you don't receive anything, you cannot make the difference between a timeout and an unsupported message (at least in PX4).

That essentially makes "UNSUPPORTED" useless for us, unfortunately.

julianoes:
> My concern with the fallback is this: what if the drone supports the new REQUEST_MESSAGE message, but the request is lost? MAVSDK will time out, not because it is unsupported, but because of an actual timeout. If MAVSDK now falls back to the old message and the drone does not support it, it will timeout again, this time because it is not supported. How do we handle that?

We try harder :grin:, so wit 5 retries, we basically try both commands 3 times, hopefully one goes through.

julianoes:
> That essentially makes "UNSUPPORTED" useless for us, unfortunately.

I disagree. We can test against PX4 v1.10 for instance and check, and if we get UNSUPPORTED, then we know, go to fallback. If we time out, then we try the fallback anyway, and then try the normal one again. UNSUPPORTED is nice, because it means we don't have to wait for the timeout, so we are much quicker.

JonasVautherin:
> I disagree. We can test against PX4 v1.10 for instance and check, and if we get UNSUPPORTED, then we know, go to fallback.

But is there a PX4 version that responds UNSUPPORTED to the newer REQUEST_MESSAGE instead of implementing it? And is it really worth removing the older message from the PX4 codebase, just for the sake of it?

I don't know, my feeling is that PX4 should support both (I don't see what we gain by removing the old one), and then MAVSDK has essentially two possibilities:

1. Support both with a more or less advanced fallback mechanism.
2. Just use the old one until it does not have to support an older version of PX4, at which point it can switch to the new one.

ykhedar:
I would prefer the 1st option since that would allow us to be not waiting for the EOL of PX4 1.10 for which we probably don't have a timeline yet.

What about keeping the older message as the default one and only try the newer message if we receive `UNSUPPORTED `response from the autopilot. 

`PX4 1.10, 1.11 and 1.12` still support the older message so the fallback would not be triggered. 

`APM 4.1.2` does not support the older message so will receive `UNSUPPORTED `response which will trigger the fallback and the newer message would be used for further requests.

This would allow us to slowly transition to the newer message.  What do you think?

JonasVautherin:
> APM 4.1.2 does not support the older message

Oh, I had not realized that. Then your suggestion makes complete sense to me:

> What about keeping the older message as the default one and only try the newer message if we receive UNSUPPORTED response from the autopilot?

julianoes:
> What about keeping the older message as the default one and only try the newer message if we receive UNSUPPORTED response from the autopilot?

What about the other way round? Try the newer message and fallback to the old one for PX4 v1.10? PX4 v1.10 is basically EOL but that's just my opinion. There is no support window or dates as far as I know, it's up to us MAVSDK maintainers what we choose to support. All I know is that there are still plenty forks around based on v1.10.

JonasVautherin:
> Try the newer message and fallback to the old one for PX4 v1.10?

My understanding is that PX4 v1.10 does not send UNSUPPORTED for the REQUEST_MESSAGE, does it? If it doesn't, then we have to rely on timeouts, which requires weird heuristics like "let's try this one 2 times, then the other one 2 times, then go back to the first one, then wait 10 seconds (who knows?), and during that time monitor the other mavlink messages to guess if it's more likely to be a timeout or an unsupported message". Of course most of the time it will work, so we will end up having heuristics that we believe work "fairly well" but without really having a clue how often it can result in problems, with convoluted code that will stay around for who knows how long, all that just to use a preferred message with no real benefit (because PX4 1.12 supports both, rights?).

Using the old one first seems less ambiguous: PX4 v1.10+ supports it, and when support is dropped in newer versions of PX4 we can make PX4 return "UNSUPPORTED", just like Ardupilot. That's still not perfect w.r.t. the mavlink specs (that don't enforce the use of UNSUPPORTED), but at least that seems reliable for the versions of PX4 and Ardupilot we want to support.

ykhedar:
> Using the old one first seems less ambiguous: PX4 v1.10+ supports it, and when support is dropped in newer versions of PX4 we can make PX4 return "UNSUPPORTED", just like Ardupilot. That's still not perfect w.r.t. the mavlink specs (that don't enforce the use of UNSUPPORTED), but at least that seems reliable for the versions of PX4 and Ardupilot we want to support.

I agree. Even though its not the perfect way to adopt the new message, its much easier to implement and maintain. 

However it is also not easy to make sure that any future versions of PX4 would implement UNSUPPORTED for the old message so we might end up coming back to this issue again. Could it be an idea to decide on a release number (say v1.5.0) for MAVSDK which just supports the new message and not the old message? Such a raodmap would also better prepare such transitions of deprecated standard messages. PX4 v1.10.2 was released almost 26 months before and i can imagine only a very small number of users could be still using it at the moment.

julianoes:
> PX4 v1.10.2 was released almost 26 months before and i can imagine only a very small number of users could be still using it at the moment.

For single users, I agree. However, many companies just for off PX4 and then don't necessarily merge back, or not often.

julianoes:
> with convoluted code that will stay around for who knows how long, all that just to use a preferred message with no real benefit 

The thing is that it's not just about this one. We have the same problem with camera commands.

To me, if we don't use the "right"/"standard" thing first, then we don't push in the right direction. For instance, if someone looks at the code, or develops by looking at what is sent, they will by default implement the old way, so then you can never get out of the rut.

julianoes:
> My understanding is that PX4 v1.10 does not send UNSUPPORTED for the REQUEST_MESSAGE, does it?

Let's check that.

JonasVautherin:
WARNING: strong opinion follows :laughing: 

> For instance, if someone looks at the code, or develops by looking at what is sent, they will by default implement the old way, so then you can never get out of the rut.

The typical way I expect deprecation to work is that for some update of the library, a function is marked as "deprecated" (meaning "it will disappear and break your build some day in the future"). During the transition period, the library *still supports the deprecated way*, and obviously supports the new way. After it has been decided that the transition period has been long enough (for some projects it takes many years), then the deprecated function is just removed. Code depending on the deprecated function just breaks.

In this particular case, that ship has sailed, because the message has been deprecated for many years already and Ardupilot has chosen to drop it. But luckily Ardupilot answers UNSUPPORTED (the fact that UNSUPPORTED is not mandatory is a fundamental flaw in MAVLink anyway, IMO), so we seem to have a chance to still support Ardupilot and PX4 by using the old message first, and falling back to the new message when UNSUPPORTED is sent. Some day, PX4 _can_ decide to drop support for the old message (ideally by sending UNSUPPORTED, too), and at this point we can just drop the old message in MAVSDK. At this point, we break:

1. The old systems that never got updated and still rely on the old message. But that's fine because we decided to drop them, and they had years to update. They can choose to update or just stay with an older MAVSDK, that's not our problem.
2. The new systems that were implemented using messages that _have been marked as deprecated for years_. Those are screwed and need to update or stay with an older MAVSDK, and again that's not our problem.

But trying to "show by example" to developers who don't read the specs by implementing a solution that sometimes fails (though maybe rarely, who knows) _by design_ is not a good idea IMO: I'd rather have those who "do it wrong" be screwed and have to put resources into updating their system than punishing those who "do it right" by having a flaky, convoluted detection system that may fail in some cases.

Those messages have been deprecated for as long as I have known PX4 (that's 5 years now), and I don't think I know a single drone from that era that is still being produced and sold and at the same time did not have a chance to update. So I would be fine dropping the old messages. Still I'm fine supporting both if we fine a way, but I will always vote against heuristics that try to guess whether it was an actual timeout or if that's just an unsupported message :innocent:.

ykhedar:
Allright, i did a rough implementation of the "try old message first and fallback to new message" idea.

src/mavsdk/core/system_impl.cpp
```diff
@@ -532,7 +562,7 @@ void SystemImpl::set_connected()
     }
     if (enable_needed) {
         if (has_autopilot()) {
-            send_autopilot_version_request();
+            // send_autopilot_version_request();
```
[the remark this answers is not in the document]
JonasVautherin: Hmm that's weird :thinking:. Though I think it's needed, right?

src/mavsdk/core/system_impl.cpp
```diff
@@ -465,26 +465,56 @@ bool SystemImpl::send_message(mavlink_message_t& message)
...
+        command.command = MAV_CMD_REQUEST_MESSAGE;
+        command.params.maybe_param1 = {static_cast<float>(MAVLINK_MSG_ID_AUTOPILOT_VERSION)};
+    }
+
+    send_command_async(command, [&prom](MavlinkCommandSender::Result result, float _val) {
+        prom.set_value(result);
+    });
+    if (fut.get() == MavlinkCommandSender::Result::Unsupported) {
+        _old_message_520_supported = false;
+        LogWarn() << "Trying alternative command (512).";
```
JonasVautherin, line 489: It's not really "trying" the alternative... maybe it should?

```suggestion
        LogWarn() << "Trying alternative command (512).";
        send_autopilot_version_request();
```

src/mavsdk/core/system_impl.cpp
```diff
@@ -465,26 +465,56 @@ bool SystemImpl::send_message(mavlink_message_t& message)
...
+        command.command = MAV_CMD_REQUEST_MESSAGE;
+        command.params.maybe_param1 = {static_cast<float>(MAVLINK_MSG_ID_FLIGHT_INFORMATION)};
+    }
+
+    send_command_async(command, [&prom](MavlinkCommandSender::Result result, float _val) {
+        prom.set_value(result);
+    });
+    if (fut.get() == MavlinkCommandSender::Result::Unsupported) {
+        _old_message_528_supported = false;
+        LogWarn() << "Trying alternative command (512)..";
```
JonasVautherin, line 516: Same here

```suggestion
        LogWarn() << "Trying alternative command (512)..";
        send_flight_information_request();
```

JonasVautherin (changes requested):

src/mavsdk/core/system_impl.cpp
ykhedar: Will update.

src/mavsdk/core/system_impl.cpp
ykhedar: You are right. Will update.

src/mavsdk/core/system_impl.cpp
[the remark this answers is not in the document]
ykhedar: I am not sure. But it could be that before adding the result callback to the `send_autopilot_version_request() ` it was basically acting like an async call. So it was exiting quickly. But now since we wait for the result its blocking for some time? 

On the other side i cannot figure it out if sending the autopilot version request is needed for anything which comes after. Maybe you can comment on that?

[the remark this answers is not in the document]
ykhedar: The following is the output with the autopilot version command DISABLED.
```
[09:39:51|Debug] Callback debugging is on. (mavsdk_impl.cpp:25)
Connection Success. 
Waiting to discover system...
[09:39:51|Info ] New system on: 127.0.0.1:54634 (with system ID: 1) (udp_connection.cpp:192)
[09:39:51|Debug] New system ID: 1 Comp ID: 1 (mavsdk_impl.cpp:601)
[09:39:51|Debug] Component Autopilot (1) added. (system_impl.cpp:376)
[09:39:51|Debug] Discovered 1 component(s) (system_impl.cpp:530)
Discovered autopilot
Got System. 
[09:39:51|Warn ] command unsupported (528). (mavlink_command_sender.cpp:202)
[09:39:51|Warn ] Trying alternative command (512).. (system_impl.cpp:517)
```
The following is the output with the autopilot version command ENABLED.

```
[09:55:14|Debug] Callback debugging is on. (mavsdk_impl.cpp:25)
Connection Success. 
Waiting to discover system...
[09:55:14|Info ] New system on: 127.0.0.1:54634 (with system ID: 1) (udp_connection.cpp:192)
[09:55:14|Debug] New system ID: 1 Comp ID: 1 (mavsdk_impl.cpp:601)
[09:55:14|Debug] Component Autopilot (1) added. (system_impl.cpp:376)
[09:55:14|Debug] Discovered 1 component(s) (system_impl.cpp:530)
[09:55:14|Warn ] sending again after 0.501881 s, retries to do: 3  (520). (mavlink_command_sender.cpp:291)
[09:55:15|Warn ] Callback called from mavsdk_impl.cpp:617 took more than 1 second to run. (mavsdk_impl.cpp:707)
Aborted (core dumped)
```

Any clues here?

src/mavsdk/core/system_impl.cpp
ykhedar: updated.

src/mavsdk/core/system_impl.cpp
ykhedar: updated

src/mavsdk/core/system_impl.cpp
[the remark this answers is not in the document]
ykhedar: created `send_autopilot_version_request_async()` version so that it will not block in the `void SystemImpl::set_connected()` function. I am still not sure if we care about the success of this message in this function or it is more left to the plugins to take care of that. At least in the Info plugin the message is sent again on init.

JonasVautherin (approved):

julianoes (approved):
Sorry for the very delayed review. This looks perfect!

PR #1792, opened 2022-05-26
PR for BATTERY_STATUS_V2

hamishwillee:
This is  PR for a new battery status message `BATTERY_STATUS_V2` that is being added in (under discussion) RFC: https://github.com/mavlink/rfcs/pull/19

This message has been added in https://github.com/mavlink/mavlink/pull/1846 into development.xml

Note, first version just allows easy testing. 
- It pulls in the mavlink git_tag in a commit that can later be removed.
- adds handler for new message. Does not update proto with new fields.

### DIFF
--- a/src/mavsdk/plugins/telemetry/telemetry_impl.cpp
+++ b/src/mavsdk/plugins/telemetry/telemetry_impl.cpp
@@ -86,6 +86,11 @@ void TelemetryImpl::init()
         this);
 
     _parent->register_mavlink_message_handler(
+        MAVLINK_MSG_ID_BATTERY_STATUS_V2,
+        [this](const mavlink_message_t& message) { process_battery_status_v2(message); },
+        this);
+
+    _parent->register_mavlink_message_handler(
         MAVLINK_MSG_ID_HEARTBEAT,
         [this](const mavlink_message_t& message) { process_heartbeat(message); },
         this);
@@ -1163,6 +1168,42 @@ void TelemetryImpl::process_battery_status(const mavlink_message_t& message)
     }
     // FIXME: it is strange calling it percent when the range goes from 0 to 1.
     new_battery.remaining_percent = bat_status.battery_remaining * 1e-2f;
+
+    set_battery(new_battery);
+
+    {
+        std::lock_guard<std::mutex> lock(_subscription_mutex);
+        if (_battery_subscription) {
+            auto callback = _battery_subscription;
+            auto arg = battery();
+            _parent->call_user_callback([callback, arg]() { callback(arg); });
+        }
+    }
+}
+
+void TelemetryImpl::process_battery_status_v2(const mavlink_message_t& message)
+{
+    mavlink_battery_status_v2_t bat_status;
+    mavlink_msg_battery_status_v2_decode(&message, &bat_status);
+
+    _has_bat_status = true;
+
+    Telemetry::Battery new_battery;
+    new_battery.id = bat_status.id;
+    new_battery.voltage_v = bat_status.voltage; 
+    //Is it correct to set this to NaN for UINT32_MAX? HOw do you specify = NaN?
+
+    // FIXME: it is strange calling it percent when the range goes from 0 to 1.
+    // Can we fix this? What about if value not provided?
+    new_battery.remaining_percent = bat_status.percent_remaining; // to test * 1e-2f;
+
+    //bat_status also has:
+    //uint32_t current; /*< [mA] Battery current (through all cells/loads). UINT32_MAX: field not provided.*/
+    // uint32_t current_consumed; /*< [mAh] Consumed charge (since vehicle powered on). UINT32_MAX: field not provided. Note: For power modules the expectation is that batteries are fully charged before turning on the vehicle.*/
+    // uint32_t current_remaining; /*< [mAh] Remaining charge (until empty). UINT32_MAX: field not provided. Note: Power monitors should not set this value.*/
+    // uint32_t status_flags; /*<  Fault, health, and readiness status indications.*/
+    // int16_t temperature; /*< [cdegC] Temperature of the whole battery pack (not internal electronics). INT16_MAX field not provided.*/
+
 
     set_battery(new_battery);
 
--- a/src/mavsdk/plugins/telemetry/telemetry_impl.h
+++ b/src/mavsdk/plugins/telemetry/telemetry_impl.h
@@ -201,6 +201,7 @@ private:
     void process_fixedwing_metrics(const mavlink_message_t& message);
     void process_sys_status(const mavlink_message_t& message);
     void process_battery_status(const mavlink_message_t& message);
+    void process_battery_status_v2(const mavlink_message_t& message);
     void process_heartbeat(const mavlink_message_t& message);
     void process_rc_channels(const mavlink_message_t& message);
     void process_unix_epoch_time(const mavlink_message_t& message);
--- a/third_party/mavlink/CMakeLists.txt
+++ b/third_party/mavlink/CMakeLists.txt
@@ -18,7 +18,9 @@ find_package(Python3 COMPONENTS Interpreter REQUIRED)
 ExternalProject_add(
     mavlink
     GIT_REPOSITORY https://github.com/mavlink/mavlink
-    GIT_TAG 3b52eac09c2e37325e4bc49cd2667ea37bf1d7d2
+    GIT_TAG a5556816af29e348875cbcc26a1d36b42dc5e921
+    # This GIT_TAG will need to update as it is for a temporary branch containing the new battery
+    # https://github.com/mavlink/mavlink/commit/a5556816af29e348875cbcc26a1d36b42dc5e921
     PREFIX mavlink
     CONFIGURE_COMMAND Python3::Interpreter
         -m pymavlink.tools.mavgen


src/mavsdk/plugins/telemetry/telemetry_impl.cpp
```diff
@@ -1176,6 +1181,42 @@ void TelemetryImpl::process_battery_status(const mavlink_message_t& message)
...
+{
+    mavlink_battery_status_v2_t bat_status;
+    mavlink_msg_battery_status_v2_decode(&message, &bat_status);
+
+    _has_bat_status = true;
+
+    Telemetry::Battery new_battery;
+    new_battery.id = bat_status.id;
+    new_battery.voltage_v = bat_status.voltage; 
+    //Is it correct to set this to NaN for UINT32_MAX? HOw do you specify = NaN?
```
hamishwillee, line 1194: @julianoes The old version seems to ignore possibility that these values might be arriving as UINT32_MAX or whatever. 
The [doc says](https://mavsdk.mavlink.io/main/en/cpp/api_reference/structmavsdk_1_1_telemetry_1_1_battery.html#structmavsdk_1_1_telemetry_1_1_battery_1aefd1d5569ff2a69bf507b422cbe75742)
> `float mavsdk::Telemetry::Battery::voltage_v {float(NAN)}`

Does that mean we need to check for a UINT32_MAX and set to NaN? Hints?

src/mavsdk/plugins/telemetry/telemetry_impl.cpp
```diff
@@ -1176,6 +1181,42 @@ void TelemetryImpl::process_battery_status(const mavlink_message_t& message)
...
+
+    // FIXME: it is strange calling it percent when the range goes from 0 to 1.
+    // Can we fix this? What about if value not provided?
```
hamishwillee, line 1197: See this fixme? Would be good to break this and send a percentage at some point. How are you managing those kinds of changes?

src/mavsdk/plugins/telemetry/telemetry_impl.cpp
```diff
@@ -1176,6 +1181,42 @@ void TelemetryImpl::process_battery_status(const mavlink_message_t& message)
...
+    new_battery.remaining_percent = bat_status.percent_remaining; // to test * 1e-2f;
+
+    //bat_status also has:
```
hamishwillee, line 1200: Other fields are shown below. I think it worth supplying everything. Thoughts? 

My understanding is I'd update the proto and rebuild the file for that.

src/mavsdk/plugins/telemetry/telemetry_impl.cpp
julianoes: Yes, good point. We should check for it and set NAN if it's `std::limits<uint32_t>::max`, the C++ way or UINT32_MAX :joy:.

src/mavsdk/plugins/telemetry/telemetry_impl.cpp
julianoes: Yes, there are a few that are wrong like that. Thanks for pointing it out. Version 2 is coming soon, so now is the time to break/fix it.

Here is where it needs fixing first:
https://github.com/mavlink/MAVSDK-Proto/blob/e8526d1fd8505fa3bc47eb1ea5edb31e58cb3be0/protos/telemetry/telemetry.proto#L571

src/mavsdk/plugins/telemetry/telemetry_impl.cpp
julianoes: The current seems like a good indicator, and status flags, and temperature is also important.
I would want to ignore the mAh fields as they are somewhat battery specific/internal, at least for smart batteries.

src/mavsdk/plugins/telemetry/telemetry_impl.cpp
```diff
@@ -1176,6 +1181,42 @@ void TelemetryImpl::process_battery_status(const mavlink_message_t& message)
...
+    //uint32_t current; /*< [mA] Battery current (through all cells/loads). UINT32_MAX: field not provided.*/
+    // uint32_t current_consumed; /*< [mAh] Consumed charge (since vehicle powered on). UINT32_MAX: field not provided. Note: For power modules the expectation is that batteries are fully charged before turning on the vehicle.*/
+    // uint32_t current_remaining; /*< [mAh] Remaining charge (until empty). UINT32_MAX: field not provided. Note: Power monitors should not set this value.*/
+    // uint32_t status_flags; /*<  Fault, health, and readiness status indications.*/
+    // int16_t temperature; /*< [cdegC] Temperature of the whole battery pack (not internal electronics). INT16_MAX field not provided.*/
+
+
+    set_battery(new_battery);
```
JonasVautherin, line 1208: Small concern here: is it the drone's responsibility to not send "conflicting" battery messages (I think there are 3 ways: battery_v1, battery_v2 (this one) and something going through the heartbeat (or a similar message)?). Or should MAVSDK somehow prioritize ("if I receive battery_v1, ignore the one in the heartbeat, if I receive battery_v2, ignore the heartbeat and battery_v1")?

src/mavsdk/plugins/telemetry/telemetry_impl.cpp
hamishwillee: @julianoes Thanks, but I think I need a bit more info.
This is what it looks like now:
```
    float remaining_percent = 2 [(mavsdk.options.default_value)="NaN"]; // Estimated battery remaining (range: 0.0 to 1.0)
```
1. As I understand it, to allow us to break this, first I change the name. So 
   ```
   float percent_remaining = 2 [(mavsdk.options.default_value)="NaN"]; // Estimated battery percentage remaining (range: 0 to 100).
   ```
   - What is the value after `=`? In this case the value is 2. My guess would be the default initialisation, but then we afterwards state `mavsdk.options.default_value`=NaN. That would imply to me that the default is NaN?
   - This is sent as a uint8. It's an integer with an invalid value of UINT8_MAX. Can we do something like (?):
     ```
     uint8 percent_remaining = 255 [(mavsdk.options.default_value)="UINT8_MAX"]; // Estimated battery percentage remaining (range: 0 to 100).
     ```

src/mavsdk/plugins/telemetry/telemetry_impl.cpp
hamishwillee: Really depends on what you expect your consumers to be able to do with the info. Right now, we're saying that the only thing of interest to a user is the % battery remaining. [technically we're also giving them the voltage, but this is of no use without context provided by SMART_BATTERY_INFO].

With the update I think you're saying that we might also give them current. But what does that actually tell them - without knowing capacity and consumption, you can't do anything with it. So we really need to supply "all or nothing". 

The same thing is true of temperature. If we just want to know that we're over temp we might as well just pass through the status bits, which tell you that the system thinks the battery has a high/low temperature. I guess passing through the actual temperature is useful if we want someone to be able to do something to control it - i.e. know it is not just faulting, but is really about to catch fire.

So I think I need more direction about how you see the users and exactly what fields you would be happy with.

One option might be just a little more - status bits and temperature.

If we're not providing all the extra info then we probably also don't need https://mavlink.io/en/messages/common.html#SMART_BATTERY_INFO, except perhaps for "function".

src/mavsdk/plugins/telemetry/telemetry_impl.cpp
hamishwillee: It isn't really a concern.

You should be able to map either or the two battery status messages to the same values. If a battery or the flight stack is sending out different values for the same battery in the same field purpose that would have to be a bug. There are some modes/faults (that we are not exposing yet) that might appear in one message and not the other, but not "clashes".

The SYS_STATUS (not heartbeat) is not indexed and that can be a problem. In a single-battery system all the messages should have battery values for the same name fields. But in a multi battery system the value of SYS_STATUS is "undefined". For MAVSDK we should never use that message for battery information. I haven't been able to deprecate it but we don't need it.

IMO we should take either v1 or v2 and use both if both come in. Why? Well, as I said there should be no clash. Further, I think most systems will only be outputting both messages for a small time, if at all. So there isn't much point putting in code to just use one or the other.

src/mavsdk/plugins/telemetry/telemetry_impl.cpp
```diff
@@ -1176,6 +1181,42 @@ void TelemetryImpl::process_battery_status(const mavlink_message_t& message)
...
+void TelemetryImpl::process_battery_status_v2(const mavlink_message_t& message)
+{
+    mavlink_battery_status_v2_t bat_status;
+    mavlink_msg_battery_status_v2_decode(&message, &bat_status);
+
+    _has_bat_status = true;
+
+    Telemetry::Battery new_battery;
+    new_battery.id = bat_status.id;
+    new_battery.voltage_v = bat_status.voltage;
```
hamishwillee, line 1193: I think this does it right?

```suggestion
    new_battery.voltage_v = (std::numeric_limits<uint32_t>::max() == bat_status.voltage) ? NAN : bat_status.voltage * 1e-3f;
```

src/mavsdk/plugins/telemetry/telemetry_impl.cpp
hamishwillee: Cool. voltage fixed above. But leaving battery remaining until I understand if we can/should make that an int

src/mavsdk/plugins/telemetry/telemetry_impl.cpp
JonasVautherin: Not exactly :slightly_smiling_face:. I believe that the only thing to change there is the docs (`((range: 0.0 to 1.0)` to `(range: 0 to 100)`). Let me explain:

* `remaining_percent` follows the convention that the "unit" goes at the end, like `velocity_m_s` and `time_s`. I don't have a strong opinion here so if you insist, we can rename it :+1:.
* The number after `=` is for protobuf. In protobuf, each field has an "id", which is used in the binary representation. Something like the `msgid` in MAVLink, I guess. We never want to change it.
* `[(mavsdk.options.default_value)="NaN"]` This is not used by protobuf, just by our MAVSDK code generators. The thing is that "UINT8_MAX" is a C definition, and this value could be used by the templates in all languages. We could have the templates translate that to the right value for each language, but... why would we set a float default value to uint8_max? Feels like `NaN` makes more sense for an undefined float :thinking:.

src/mavsdk/plugins/telemetry/telemetry_impl.cpp
JonasVautherin: Ok :+1:, thanks! :blush:

src/mavsdk/plugins/telemetry/telemetry_impl.cpp
julianoes: Yes, that's good.

src/mavsdk/plugins/telemetry/telemetry_impl.cpp
julianoes: I would be ok with just adding the failure/status flags. I wouldn't necessarily add everything until someone needs/wants it.

reedev: Hi,
Sounds like a nice improvement, I was thinking about an extended parameter set too.

What I would like to have, as an end user, is how much capacity of my battery I have used.

Also, current is another one. Maybe not as end user parameter but as power consumption, since it's a hint how healthy a setup is.

cheers

hamishwillee: @julianoes OK, I'll add the flags. Request here, but 

@reedev That said, I stick by my all or nothing :-)

1. Current isn't actually a great measure (anecdotally) of health. It's a good measure of dynamic power consumption, so we would send that for systems that want to do their own estimates of system load/consumption.
2. "How much capacity of my battery I have used" is _problematic_. We can and do supply `current_consumed` but that is the amount consumed _since the battery was turned on_. If you started with a full battery it will be the same thing, but not otherwise.
   - a power module will make the assumption the battery was full, which is how it calculates the remaining charge and 
3. `current_remaining` is the capacity remaining in mAH. It should only sent by a smart battery, which is capable of working out the charge state of the battery reliably. A power module should not send this - a guess based on  `current_consumed` could be widely off.
  - @julianoes I'm wondering if we'd be better off setting a flag to indicate that current_remaining is an estimate based on current consumed - rather than this kind of assumtion?

src/mavsdk/plugins/telemetry/telemetry_impl.cpp
julianoes: I would say `_percent` should go from 0..100. If there is no unit, it should go from 1.0. I know this might not be inline with what I have implemented earlier (case and point here) but I'd say that's the most intuitive.

julianoes: I agree with @JonasVautherin that we can't have magic numbers in the API except NAN.

hamishwillee: @julianoes @JonasVautherin So the reason I wanted to change the name was that Julian said that we change the name on API break - I presume the assumption being that it fails a compile and users check what's changed.

So I am happy to change `(range: 0.0 to 1.0) ` to `(range: 0 to 100)` but don't we need to rename it something? I could call it `charge_percent` if you prefer?

JonasVautherin: > that we change the name on API break

We _can_ change the name on API break if we want to. It's just an opportunity, it's not mandatory. I think there was a misunderstanding there, right? :grin:

It's just that for semantic versioning, every time we break the API, we have to increase the major version. So we don't want to be on MAVSDK v503 just because we rename stuff :grin:. However, we will soon move to MAVSDK v2 (we had a real need to break the API), so that's an opportunity to rename small stuff (and break more of the API on this occasion :laughing:).

Does that make sense?

src/mavsdk/plugins/telemetry/telemetry_impl.cpp
hamishwillee: OK,

> Current consumed, and total current is something that needs to be kept internal to a smart battery, in my opinion, because otherwise it will - more often than not - be wrong.

A smart battery can report consumption (since full) and remaining battery capacity reliably -they will not be wrong, even after many power cycles and for aged batteries.
Power monitors only measure consumption since power on, and estimate power remaining based on assumption of a full battery at power on. 

We are going to allow a consumer to determine which case they have using a new flag. If the flag is set (by a smart battery) the values are assumed to be accurate and consumed is "relative to a full battery". If not set (default, or by a power module) only the current consumed can be trusted.

So there is the option to supply this information only if it is known to be accurate, or to supply it and state that it might not be accurate. 

Just FYI!

julianoes: When you have a smart battery you don't need consumption since full and capacity, that's why a smart battery gives you percentage.

reedev: I agree with only exposing the current. I am not sure how QGC handles the 'current consumed' but that's not something for an FC to keep track with. Current is a raw value and works well. Until now I was peeking the mavlink message itself to calculate the power consumption and I find it very valuable.

I am looking forward to a proper smart battery. I've once purchased a BatMon but never got into it seriously. Are their changes in the PX4?

hamishwillee: The battery message updates are "in discussion". There are no changes to PX4 yet, but this is part of what might end up happening.

julianoes:
@hamishwillee do you still want to get BATTERY_STATUS_V2 in? I'm confused here.

hamishwillee:
Yes I do, but there is no point until a corresponding PR goes into PX4 or ArduPilot. Closing.

PR #1793, opened 2022-05-29
More command result options

julianoes:
Requires https://github.com/mavlink/MAVSDK-Proto/pull/289.


src/mavsdk/core/mavlink_command_sender.cpp
```diff
@@ -79,13 +79,11 @@ void MavlinkCommandSender::queue_command_async(
     CommandIdentification identification = identification_from_command(command);
 
     for (const auto& work : _work_queue) {
-        if (work->identification == identification) {
+        if (work->identification == identification && callback == nullptr) {
```
JonasVautherin, line 82: Is `callback == nullptr` compatible with the error message "command is already being sent"? :thinking:

JonasVautherin (dismissed):

src/mavsdk/core/mavlink_command_sender.cpp
julianoes: Hm, I don't understand exactly what you mean. Does my commit message make sense to you?

JonasVautherin: Yeah sorry, I had missed it :see_no_evil:

JonasVautherin (approved):

PR #1794, opened 2022-05-29
telemetry: sys_status for position health flags

julianoes:
With this commit we start to use the sys_status flags to determine whether local and global position are ok. If the flags don't appear to be set, we fall back to counting GPS satellites like before.

### DIFF
--- a/src/mavsdk/plugins/telemetry/telemetry_impl.h
+++ b/src/mavsdk/plugins/telemetry/telemetry_impl.h
@@ -236,6 +236,9 @@ private:
     void request_home_position_async();
     void check_calibration();
 
+    static bool sys_status_present_enabled_health(
+        const mavlink_sys_status_t& sys_status, MAV_SYS_STATUS_SENSOR flag);
+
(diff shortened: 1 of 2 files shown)


src/mavsdk/plugins/telemetry/telemetry_impl.h
```diff
@@ -416,5 +419,12 @@ class TelemetryImpl : public PluginImplBase {
         OffsetStatus gyro_offset;
 
     } _ap_calibration{};
+
+    enum class SysStatusUsed {
```
JonasVautherin, line 423: Nitpick: this could be a `SysStatus { Unknown, Used, Unused }` instead of something that sounds like an augmented boolean :grin:

JonasVautherin (approved):

src/mavsdk/plugins/telemetry/telemetry_impl.h
julianoes: Oh yes, good point.

PR #1797, opened 2022-06-01
Link mavsdk to SSL (fix for Ubuntu 22.04)

potaito:
This fixes compiling issues on Ubuntu 22.04.

This was the error I got with current main branch:
```
[ 61%] Linking CXX executable integration_tests_runner
/usr/bin/ld: ../mavsdk/libmavsdk.so.1.4.0: undefined reference to `EVP_PKEY_get_id'
/usr/bin/ld: ../mavsdk/libmavsdk.so.1.4.0: undefined reference to `SSL_CTX_load_verify_file'
/usr/bin/ld: ../mavsdk/libmavsdk.so.1.4.0: undefined reference to `SSL_CTX_load_verify_dir'
/usr/bin/ld: ../mavsdk/libmavsdk.so.1.4.0: undefined reference to `SSL_get1_peer_certificate'
/usr/bin/ld: ../mavsdk/libmavsdk.so.1.4.0: undefined reference to `EVP_PKEY_get_bn_param'
gmake[2]: *** [src/system_tests/CMakeFiles/system_tests_runner.dir/build.make:229: src/system_tests/system_tests_runner] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:1030: src/system_tests/CMakeFiles/system_tests_runner.dir/all] Error 2
gmake[1]: *** Waiting for unfinished jobs....
/usr/bin/ld: mavsdk/libmavsdk.so.1.4.0: undefined reference to `EVP_PKEY_get_id'
/usr/bin/ld: mavsdk/libmavsdk.so.1.4.0: undefined reference to `SSL_CTX_load_verify_file'
/usr/bin/ld: mavsdk/libmavsdk.so.1.4.0: undefined reference to `SSL_CTX_load_verify_dir'
/usr/bin/ld: mavsdk/libmavsdk.so.1.4.0: undefined reference to `SSL_get1_peer_certificate'
/usr/bin/ld: mavsdk/libmavsdk.so.1.4.0: undefined reference to `EVP_PKEY_get_bn_param'
gmake[2]: *** [src/CMakeFiles/unit_tests_runner.dir/build.make:409: src/unit_tests_runner] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:815: src/CMakeFiles/unit_tests_runner.dir/all] Error 2
/usr/bin/ld: ../mavsdk/libmavsdk.so.1.4.0: undefined reference to `EVP_PKEY_get_id'
/usr/bin/ld: ../mavsdk/libmavsdk.so.1.4.0: undefined reference to `SSL_CTX_load_verify_file'
/usr/bin/ld: ../mavsdk/libmavsdk.so.1.4.0: undefined reference to `SSL_CTX_load_verify_dir'
/usr/bin/ld: ../mavsdk/libmavsdk.so.1.4.0: undefined reference to `SSL_get1_peer_certificate'
/usr/bin/ld: ../mavsdk/libmavsdk.so.1.4.0: undefined reference to `EVP_PKEY_get_bn_param'
gmake[2]: *** [src/integration_tests/CMakeFiles/integration_tests_runner.dir/build.make:821: src/integration_tests/integration_tests_runner] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:1001: src/integration_tests/CMakeFiles/integration_tests_runner.dir/all] Error 2
gmake: *** [Makefile:136: all] Error 2
```

Also tested with Ubuntu 20.04 in docker and it's still compiling :)


src/mavsdk/CMakeLists.txt
```diff
@@ -15,6 +15,7 @@ target_link_libraries(mavsdk
     PRIVATE
     JsonCpp::JsonCpp
     tinyxml2::tinyxml2
+    ssl
```
JonasVautherin, line 18: Does the target work (see [here](https://cmake.org/cmake/help/latest/module/FindOpenSSL.html#imported-targets))?

```suggestion
    OpenSSL::SSL
```

JonasVautherin (commented):
I'm wondering who uses it in mavsdk :thinking:. I would have thought that e.g. curl would link it anyway...

JonasVautherin (changes requested):
Yeah it's lacking `find_package(OpenSSL)`. 

Still I would like to understand why Ubuntu 22.04 is not happy because of OpenSSL :thinking:. Marking "request for changes" to make sure I don't merge before that :laughing:.

julianoes:
"Dependencies are easy" has no one said ever.

potaito:
> Yeah it's lacking `find_package(OpenSSL)`.

I did add `hunter_add_package(openssl)` and `find_package(openssl), but this is the first time I saw this hunter thing :sweat_smile: 
But seems that the way you added it, it's working fine now :pray: 

> Still I would like to understand why Ubuntu 22.04 is not happy because of OpenSSL thinking. Marking "request for changes" to make sure I don't merge before that laughing.

Question is probably more why 20.04 was happy without explicit linking to OpenSSL

JonasVautherin:
> Question is probably more why 20.04 was happy without explicit linking to OpenSSL

I can only think of curl using OpenSSL (I don't think we use it anywhere else), and so it should be linked from the Curl target as a transitive dependency :thinking:

julianoes:
When I try this in docker it "just works" without changes. :thinking: 

That's with these things installed (which shouldn't really matter):

https://github.com/mavlink/MAVSDK/blob/12cf8cfb0a6e8775175c32979a6ffc0ca9a27972/docker/Dockerfile-Ubuntu-20.04#L13-L39

JonasVautherin:
> That's with these things installed (which shouldn't really matter):

Yet it works this time. Is it somehow (wrongly) taking curl from the system?

JonasVautherin:
Why does it seem to work when Julian touches it? @julianoes can you stay 24/7 in the CI and touch all the Ubuntu 22.04 jobs, pleeaaaaase? :laughing:

JonasVautherin:
Ok I'm confused. Let me try to just add 22.04 [here](https://github.com/mavlink/MAVSDK/pull/1801).

julianoes:
Maybe this is now captured in the cache of the configure step... I guess we need to disable the cache for this test.

julianoes:
Now it seems consistently not working, except that locally in docker, the same builds just fine.

potaito:
> Now it seems consistently not working, except that locally in docker, the same builds just fine.

Sorry, I'm super swamped this and next week :see_no_evil:

julianoes:
Fixed in #1801.

potaito:
Awesome, thank you very much :tada:

PR #1801, opened 2022-06-08
Add Ubuntu 22.04 to CI

JonasVautherin:
@potaito: FYI


JonasVautherin:
Here, curl finds openssl on the system (which we don't expect for the superbuild):

```
2022-06-08T22:17:01.1231890Z -- Found OpenSSL: /usr/lib/x86_64-linux-gnu/libcrypto.so (found version "3.0.2")  
```

julianoes:
> but that does not tell me why @julianoes changes in https://github.com/mavlink/MAVSDK/pull/1797 work thinking.

My changes don't work.

JonasVautherin:
@julianoes @potaito: I think that was it :blush:

julianoes (approved):
Nice catch!

potaito:
Thanks for the digging @JonasVautherin, I appreciate it!!

Issue #1805, opened 2022-06-13
Param Server - exception when using custom string parameter


Consti10:
I am trying to accomplish the following:
1) Have a parameter server (supposed to run on the OHD Ground station) for OHD related settings

2) test out the set / get -equivalents inside QOpenHD for then changing the settings.

So far I am just trying things out, but while doing so I stumpled across the following issue:

I add a custom (string) parameter to the server side:
https://github.com/OpenHD/Open.HD/blob/consti-test/OpenHD/ohd_telemetry/tests/test_mavsdk.cpp#L40

And just for testing, request all the parameters from the Ground controll application side: 
` auto result=paramOhdGround->get_all_params();`

Both the server and client side crash. On the server side, the output is:
```
test_mavsdk: /media/consti10/ade719d9-e80a-46da-8802-58ca3b8e08bc1/1_ohd-dev/MAVSDK/src/mavsdk/core/mavlink_parameters.cpp:1999: float mavsdk::MAVLinkParameters::ParamValue::get_4_float_bytes_bytewise() const: Assertion `false' failed.
[12:24:24|Error] Unknown type (mavlink_parameters.cpp:1998)
```

On the client side the output is less usefull:
```
See: https://mavsdk.mavlink.io/main/en/cpp/troubleshooting.html#user_callbacks
terminate called after throwing an instance of 'std::future_error'
  what():  std::future_error: No associated state
```

Any ideas ?

Consti10:
The issue above can be easily reprodcued by adding 2 lines of code to the autopilot_server example:

https://github.com/OpenHD/MAVSDK/commit/753f6553970fbda491fc66f4711a193dcce940db

Consti10:
Also, a question not a bug:

For commands or parameters set - aka stuff that is acknowledged - is there some type of re-transmission inside mavsdk ?

JonasVautherin:
> For commands or parameters set - aka stuff that is acknowledged - is there some type of re-transmission inside mavsdk ?

Yes those are the [different](https://mavlink.io/en/services/command.html) [mavlink](https://mavlink.io/en/services/parameter.html) [microservices](https://mavlink.io/en/services/parameter_ext.html) :+1:.

> The issue above can be easily reprodcued by adding 2 lines of code to the autopilot_server example

Nice, thanks for sharing a reproducible example! This looks like a bug indeed. Would you be willing to look into debugging it inside MAVSDK?

Consti10:
[...]

PR #1807, opened 2022-06-16
Consti test

Consti10:
Opening this pull request to show the fixes, **not ready to be merged yet.**.

Related to https://github.com/mavlink/MAVSDK/issues/1805

Fixes for mavlink_param when used as 
1) Server:
Do not forward parameters that need extended protocoll (string-values) on non-ext param requests, this made the original impl crash at run time.

The fix I went with seems to work but is not the nicest in my opinion - I basically try and hide parameters that only work with extended from a non-extended perspective.

2) Client:
Fix bug where reference goes out of scope, which resulted in the client crashing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,7 +22,7 @@ venv/
 cmake-build-*
 
 # Build folder when using the command line (as recommended from the tutorial)
-build
+build/*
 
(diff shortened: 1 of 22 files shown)


Consti10:
has been moved into new, cleaner pull request

PR #1809, opened 2022-06-20
Improve mavlink parameters after split

Consti10:
This builds up on the work already done by @julianoes in https://github.com/mavlink/MAVSDK/pull/1772

And replaces this pull request:
https://github.com/mavlink/MAVSDK/pull/1807

Goals:
Refactor mavlink parameters to reduce duplicated code, eliminate bugs(relates to https://github.com/mavlink/MAVSDK/issues/1805) and document the code without changing the public api's.

In general, the work flow for refactoring is:
1) Implement a generic way (for example a templated `retrieve_server_param` ) method
2) refactor the fixed type methods (for example `retrieve_server_param_float, retrieve_server_param_int, ...`

Not all of the refactorings are that simple, but in general the workflow is the same.


src/mavsdk/core/mavlink_parameter_receiver.cpp
```diff
@@ -106,176 +118,177 @@ std::map<std::string, ParamValue> MavlinkParameterReceiver::retrieve_all_server_
...
+            }
+            auto new_work = std::make_shared<WorkItem>(WorkItem::Type::Value,safe_param_id,false);
             new_work->param_value = _all_params.at(safe_param_id);
-            new_work->extended = false;
             _work_queue.push_back(new_work);
         } else {
             LogDebug() << "Missing Param " << safe_param_id;
         }
+    }else{
+        // TODO the server should be able to handle requests with a int index instead of string as param_id.
```
julianoes, line 257: :thinking:

src/mavsdk/core/mavlink_parameter_receiver.h
```diff
@@ -23,38 +36,50 @@ class MavlinkParameterReceiver {
...
+     * It is recommended to not change the parameter set after first communicating with any client.
+     * (https://mavlink.io/en/services/parameter_ext.html#parameters_invariant).
+     * @param name the unique id for this parameter
+     * @param param_value the value for this parameter
+     * @return Result::ParamNameTooLong if the parameter name is too long,
+     * Result::WrongType if the same parameter name is provided with a different type (aka updating the parameter would mutate the type
+     * of an already provided parameter), Result::ParamValueTooLong if the parameter type is std::string but the value is longer than
+     * the extended protocol allows and
+     * Result::Success otherwise.
+     */
```
julianoes, line 57: While this is nice I wouldn't bother too much with docstrings for internal methods.

src/mavsdk/core/mavlink_parameter_subscription.cpp
```diff
@@ -0,0 +1,80 @@
...
+{
+    std::lock_guard<std::mutex> lock(_param_changed_subscriptions_mutex);
+    if (callback != nullptr) {
+        ParamChangedSubscription subscription{name, callback, cookie};
+        // This is just to let the upper level know of what probably is a bug, but we check again when actually calling the callback We also cannot assume here that the user called provide_param before subscribe_param_changed, so the only thing that makes sense is to log a warning, but then continue anyways.
+        /*std::lock_guard<std::mutex> lock2(_all_params_mutex);
+        if (_all_params.find(name) != _all_params.end()) {
+            const auto curr_value = _all_params.at(name);
+            if (!curr_value.template is_same_type_templated<T>()) {
+                LogDebug()
```
julianoes, line 21: I would say make this a `LogWarn` and enable it.

src/mavsdk/core/mavlink_parameter_subscription.h
```diff
@@ -0,0 +1,64 @@
+//
+// Created by consti10 on 21.06.22.
+//
+
```
julianoes, line 4: Remove.

src/mavsdk/core/mavlink_parameter_subscription.h
```diff
@@ -0,0 +1,64 @@
...
+#ifndef MAVSDK_SUPERBUILD_SRC_MAVSDK_CORE_MAVLINK_PARAMETER_SUBSCRIPTION_H_
+#define MAVSDK_SUPERBUILD_SRC_MAVSDK_CORE_MAVLINK_PARAMETER_SUBSCRIPTION_H_
```
julianoes, line 6: Use `#pragma once` instead.

src/mavsdk/core/mavlink_parameter_subscription.h
```diff
@@ -0,0 +1,64 @@
...
+        variant<ParamFloatChangedCallback, ParamIntChangedCallback, ParamCustomChangedCallback>;
+    struct ParamChangedSubscription {
+        const std::string param_name;
+        const ParamChangedCallbacks callback;
+        const void* const cookie;
+        explicit ParamChangedSubscription(std::string param_name1,ParamChangedCallbacks callback1,const void* cookie1):
+            param_name(std::move(param_name1)),callback(std::move(callback1)),cookie(cookie1){};
+    };
+    std::mutex _param_changed_subscriptions_mutex{};
+    std::list<ParamChangedSubscription> _param_changed_subscriptions{};
```
julianoes, line 59: We now have `CallbackList<Type>` for this.

src/mavsdk/core/param_value.h
```diff
@@ -109,6 +118,13 @@ class ParamValue {
         double,
         std::string>
         _value{};
+
+    // Consti10: hacky, returns true if this parameter needs the extended parameters' protocol
+    // (which is the case when its value is represented by a string)
```
julianoes, line 123: I think that's fine.

julianoes (commented):
Nice cleanup and changes! I left a couple of comments.

julianoes:
Note that I changed the base branch to my PR, so only your diff is visible here.

src/mavsdk/core/mavlink_parameter_sender.cpp
```diff
@@ -111,43 +114,83 @@ void MavlinkParameterSender::set_param_int_async(
...
+        set_param_async(name,value_to_set,callback,cookie,maybe_component_id,extended);
+    }else{
+        // Here I implemented the same behaviour as the original author, but by using the same "do the checking in the callback" - trick
+        // the duplicated code can be reduced nicely.
+        // NOTE that his behaviour is not exactly to the mavlink specifications.
+        // The issue here is: The user of the parameters library called set_int - the expected behaviour here would be to just send
+        // a set - message with type==int aka int32_t. However, in this case, if the server actually stores (for example) and int8_t we will
+        // get a invalid value back from the server, and the parameter is not changed. To work around this issue, here we try and find out what
+        // the server stores, then cast the user-provided int parameter into this exact type and send that instead. Not to the specs, but I didn't want to
+        // break things here.
```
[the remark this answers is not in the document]
Consti10: yeah, it is a bit messy but in my opinion easier to understand since it makes the already complicated WorkItem 's much simpler.

src/mavsdk/core/mavlink_parameter_subscription.cpp
Consti10: This comes from before I refactored mavlink_parameter_subscription out. R.n. mavlink_parameter_subscription does not hold the parameter set, and therefore cannot do this check anymore.

src/mavsdk/core/mavlink_parameter_subscription.h
Consti10: sure

src/mavsdk/core/mavlink_parameter_subscription.h
Consti10: I'l have a look at it.

src/mavsdk/core/param_value.h
Consti10: yeah, the hacky part is not there in param_value but in mavlink_parameter_receiver that must return different numbers for param_count on extended or non-extended requests.

src/mavsdk/core/mavlink_parameter_receiver.cpp
Consti10: Actually, to show why this becomes so complicated with the current support for extended and non-extended parameters - we'd need.
```
struct Parameter{
        // unique parameter id
        const std::string param_id;
        // unique parameter index when looked at from a non-extended param perspective
        const int param_idx_non_extended;
        // unique parameter index when looked at from a extended param perspective
        const int param_idx_extended;
        ParamValue param_value;
    };
```

src/mavsdk/core/mavlink_parameter_subscription.h
Consti10: can't find it in this branch, I think we should wait with that until we can merge with master / the pr branch for that.

src/mavsdk/core/mavlink_parameter_subscription.cpp
julianoes: Oh ok. So the code can be deleted?

src/mavsdk/core/mavlink_parameter_subscription.h
julianoes: I'd say that's in theory true but not very practical.

I'm open to changing it as soon as someone requires a compiler that does not support it:
https://en.wikipedia.org/wiki/Pragma_once#Portability

src/mavsdk/core/param_value.h
julianoes: So we have to have two storage containers for it?

src/mavsdk/core/mavlink_parameter_receiver.cpp
julianoes: I think we need to lists. One normal one and one for the extended ones.

src/mavsdk/core/mavlink_parameter_receiver.h
```diff
@@ -6,13 +6,26 @@
...
 
 #include <map>
 #include <string>
+#include <list>
 
 namespace mavsdk {
 
-class MavlinkParameterReceiver {
+/**
+ * One could also call this MavlinkParameterProvider or MavlinkParameterServer but all these names don't really fit perfectly.
```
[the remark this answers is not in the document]
julianoes: Right.

src/mavsdk/core/mavlink_parameter_sender.cpp
```diff
@@ -82,27 +84,28 @@ void MavlinkParameterSender::set_param_async(
...
+
+
 void MavlinkParameterSender::set_param_int_async(
     const std::string& name,
     int32_t value,
     const SetParamCallback& callback,
     const void* cookie,
     std::optional<uint8_t> maybe_component_id,
-    bool extended)
+    bool extended,bool adhere_to_mavlink_specs)
```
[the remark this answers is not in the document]
julianoes: Yer, I don't know. So I don't want to leak types like uint8_t to the public API of the Param plugin. I find that just ugly and I don't see the point of a user having to think about that.

That being said, we can expose other types internally of MAVSDK if we really require it. Again, I would prefer not to if possible.

I wouldn't call it unsafe, I would instead just return a "safe" error if a value is out of bounds of a type that it actually fits into. Does that make sense?

So then, we only really need the check for what type the counter part is for the case where it's not PX4. For PX4, we can just assume int32_t and don't need to do the "get first dance", right?

src/mavsdk/core/mavlink_parameter_subscription.h
Consti10: I see

src/mavsdk/core/mavlink_parameter_subscription.cpp
Consti10: Well, it would have been a nice (additional) check for the user, but as of r.n it is impractical - but that might change at some point.

src/mavsdk/core/mavlink_parameter_receiver.cpp
Consti10: Well, that is also not ideal, since then on each set / get operation we'd have to check 2 lists, comlicating code even further. R.n I am experimenting on the server side, if using a different type of data layout would make more sense:
`std::vector<ParamValue> _all_params`
and 
```
std::map<std::string,uint32_t> _string_id_to_index_non_extended.
std::map<std::string,uint32_t> _string_id_to_index_extended.
```

The reasoning behind that: We need 2 types of indexing anyways, if we want the param_server to be fully compatible with everything the mavlink protocoll says - by unique index or unique string id. But I am not quite sure yet what creates the most simple code in this regard.

Consti10:
So right now my goal is to make the server / provider exactly adhere to all functionalities of the mavlink protocoll (as already said earlier, the get_by_index instead of get_by_string_id part was/ is missing. While I am not sure if we will ever get the client then that far (there it really really gets complicated) that'd still mean though that other libraries who perhaps use features (like get param by index) can work with it.

I also found out that after all the refactoring, on the providing side the WorkItem has basically become a re-implementation of the mavlink Value / Ack messages.

We can also just as well create the mavlink responses as mavlink messages, then push the message onto the work queue. This makes the code less error prone, since we create the response right at the place where we have all the data available, instead of basically creating a copy of the data that is then needed for the mavlink message, and creating the mavlink message on the work queue.

Consti10:
So after running around in circles multiple times, I think the issue with the param_count/param_index for extended vs non-extended is actually not really solvable with how the current public api is written.

In short, if you look at the server / param provider -
The public api does not protect against the following issues:
1) The parameter set might change after initial operation, and there is no way to use the server in a safe way to avoid this
=> I think the only way to solve this issue is to change the public api such that either `provide_server_param()` is removed and the parameter set is provided in the constructor, or we add a public method in the likes of `ready_for_communication()` - note that both these solutions are breaking downwards compability, since by default `ready_for_communication` must be off

2) It is impossible to avoid changing the indices of parameters when talking both extended and non-extended protocoll - I don't think it is a good idea to mix these 2 up. It would be possible to just provide a immutable flag when creating the server and/or client (use_extended) and then the client / server either speaks extended or non-extended protocol, but not both.

julianoes:
> `ready_for_communication()`

I think that would be a valid approach. It's indeed a bit of a problem with params, and in PX4 this is done similar, in fact.

Consti10:
2 more simple questions:
1) How do I create a simple unit test for the mavlink core libary (I want to test mavlink_paramater_xxx directly without going through the public api route).

2) How do I use the mavlink_passthrough plugin in the "system_tests" to emulate packet loss ? It doesn't compile.

Consti10:
> 2 more simple questions:
> 
> 1. How do I create a simple unit test for the mavlink core libary (I want to test mavlink_paramater_xxx directly without going through the public api route).
> 2. How do I use the mavlink_passthrough plugin in the "system_tests" to emulate packet loss ? It doesn't compile.

Found the unit test(s)

julianoes:
Just to make sure, is the ID numbering now like this?

Normal params:
```
WIDGET_SO  0
WIDGET_BLA 1
```

Extended params:
```
GADGET_FOO 0
GADGET_PI  1
GADGET_PA  2
GADGET_PO  3
```

I believe that's how it should be, right?

julianoes (commented):
There are quite a few changes here by now which makes it hard to review...

I think it would be good if we can get CI passing and then merge it into my PR, and then merge mine (both) into main and start to use and test it.

Don't forget to fix style:
```
tools/fix_style.sh .
# or
tools/run-docker.sh tools/fix_style.sh .
```

julianoes:
@Consti10 thanks. I will have to think about what you wrote. I hope we can find away to enable your use case.

Consti10:
> @Consti10 thanks. I will have to think about what you wrote. I hope we can find away to enable your use case.

Thanks, that's nice to hear. I think it would be great for MAVSDk to fully support a seperation by components, this is nothing special of OpenHD, but rather a "feature" of mavlink itself OpenHD needs to use (and MAVSDK - probably for simlicity - so far doesn't really differentiate there).

julianoes:
@Consti10 I finally spent some time today to continue here. Sorry for the delay. First step was to rebase my PR and yours on top of it. I had a hard time with a few conflicts, so I decided to squash your (many!) commits, I hope that's ok.

Since I don't have permission to this PR, I'm closing this one and manually merging it into https://github.com/mavlink/MAVSDK/pull/1772.

Thanks again for all your great work here! I have some different opinions on some of the pieces but I think most of your work is very useful, also some of the comments.

And lastly, I think I found a way to support your use case with multiple components that need to be addressed. :)

PR #1811, opened 2022-06-21
Update openssl to 3.0.5

JonasVautherin:
- [x] Fedora fails because it is missing perl dependencies
- [x] Manylinux seems to fail for the same reason as Fedora (do we actually still need manylinux?)
- [x] Ubuntu 20.04 fails with `undefined reference to std::__throw_bad_array_new_length()` (edit: that was due to the dependencies cache being shared with Ubuntu 22.04)
- [x] SITL fails because it does not find our OpenSSL when building curl: `Found OpenSSL: /usr/lib/x86_64-linux-gnu/libcrypto.so`
- [x] The musl builds fail because of this:
```
threads_pthread.c:(.text+0x228): undefined reference to `__atomic_is_lock_free'
/usr/xcc/armv7l-linux-musleabihf-cross/bin/../lib/gcc/armv7l-linux-musleabihf/11.2.1/../../../../armv7l-linux-musleabihf/bin/ld: threads_pthread.c:(.text+0x278): undefined reference to `__atomic_load_8'
```
- [ ] iOS fails because of this (I asked about it [here](https://github.com/openssl/openssl/issues/18754)):
```
[ 75%] Performing build step for 'openssl'
ld: -bundle and -bitcode_bundle (Xcode setting ENABLE_BITCODE=YES) cannot be used together
```


julianoes:
Out of curiosity. Why? Or what's the difference from version 1.1 to 3?

JonasVautherin:
I figured it may be worth trying, to get an idea. I can see potential issues that we may encounter:

* Ubuntu 22.04 comes with OpenSSL 3. Not sure if we can make a deb package without supporting OpenSSL 3 ourselves.
* At some point one of our dependencies depends on OpenSSL 3.
* What if somebody uses OpenSSL 3 and MAVSDK in their project, and MAVSDK builds against OpenSSL 1?

An open question is therefore whether we should support both OpenSSL 1 _and_ OpenSSL 3, or if we could just move to OpenSSL 3, I guess? :thinking:

julianoes:
Do we have openssl 3 on Ubuntu 20.04?

JonasVautherin:
No it's libssl1.1. But it seems like MAVSDK builds with both in many cases, so maybe we're lucky and it's fine to just update to 3 in our third parties and let the system take 1 when packaging?

julianoes (commented):
Having fun? :joy:

julianoes:
@JonasVautherin can we somehow go to XCode 14 where bitcode is apparently no longer a thing?

JonasVautherin:
Let me have a look :+1:

julianoes:
Merging with openssl for iOS still at 1.1.1.

PR #1812, opened 2022-06-21
Transponder: manager altitude type to be more accurate

chep:
There are 2 types of altitude: from baro source (AMSL) and from gnss source (using geoid)

We need to inform user of what type is used because sometimes there is a huge difference.

Generated files come from my pull request here: https://github.com/mavlink/MAVSDK-Proto/pull/291


JonasVautherin:
I tried generating the code again, I'm not sure why it's changing so many files. But the CI was not happy before. Let's see...

PR #1813, opened 2022-06-22
integration_tests: disable RC failsafe for PX4

julianoes:
Without setting this parameter, the drone does not actually take off and will enter the RC failsafe instead.

I need to go through all other integration tests and check for similar problems while I'm at it.


.github/workflows/main.yml
```diff
@@ -86,7 +86,7 @@ jobs:
         id: cache
         with:
           path: ./build/release/third_party/install
-          key: ${{ github.job }}-${{ hashFiles('./third_party/**') }}
+          key: ${{ github.job.name }}-${{ hashFiles('./third_party/**') }}
```
JonasVautherin, line 89: Good catch! Maybe worth putting in a separate PR and merge it right away?

JonasVautherin (dismissed):

.github/workflows/main.yml
julianoes: https://github.com/mavlink/MAVSDK/pull/1814

julianoes:
Closing. The integration tests need a pass at some point anyway.

PR #1816, opened 2022-06-23
Retry home position until set. 

ykhedar:
Sometimes, randomly the home position is not received/processed by MAVSDK which leads to `all_health_okay` to be never `true`. 
This has been often the case in integration tests, where one or two tests would randomly fail waiting for the health to be okay which never happens. 

In this PR, the retry mechanism is added for requesting the home position until it is received/set.

Glad to hear your feedback :)

### DIFF
--- a/src/mavsdk/plugins/telemetry/telemetry_impl.cpp
+++ b/src/mavsdk/plugins/telemetry/telemetry_impl.cpp
@@ -198,7 +198,7 @@ void TelemetryImpl::enable()
     _parent->add_call_every([this]() { request_home_position_again(); }, 2.0f, &_homepos_cookie);
 }
 
-void TelemetryImpl::disable() 
+void TelemetryImpl::disable()
 {
     _parent->remove_call_every(_calibration_cookie);
     _parent->remove_call_every(_homepos_cookie);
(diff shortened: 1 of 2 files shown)


src/mavsdk/plugins/telemetry/telemetry_impl.h
```diff
@@ -390,6 +393,7 @@ class TelemetryImpl : public PluginImplBase {
     bool _has_bat_status{false};
 
     void* _calibration_cookie{nullptr};
+    void* _call_every_cookie{nullptr};
```
JonasVautherin, line 396: Shouldn't this be called something like `home_position_cookie`?

JonasVautherin (commented):
Makes sense to me :+1:

JonasVautherin (dismissed):

src/mavsdk/plugins/telemetry/telemetry_impl.cpp
```diff
@@ -193,10 +193,26 @@ void TelemetryImpl::enable()
...
 }
 
 void TelemetryImpl::disable() {}
 
+void TelemetryImpl::request_home_position_again()
+{
+    {
+        std::lock_guard<std::mutex> lock(_request_home_position_mutex);
+        if (_health.is_home_position_ok) {
+            _parent->remove_call_every(_homepos_cookie);
```
julianoes, line 208: Don't forget to also call this `_parent->remove_call_every(_homepos_cookie);` in `disable()`, and maybe add the one for the calibration as well, thanks!

julianoes (approved):

src/mavsdk/plugins/telemetry/telemetry_impl.cpp
ykhedar: Thanks :) just added to the latest commit.

PR #1819, opened 2022-06-27
cmake: install mavlink headers in correct dir

JonasVautherin:
Taken from https://github.com/mavlink/MAVSDK/commit/0c8e2367952b3233669739d439a13f2ccb568470, as it feels useful to merge it ASAP.

### DIFF
--- a/src/mavsdk/core/CMakeLists.txt
+++ b/src/mavsdk/core/CMakeLists.txt
@@ -119,6 +119,11 @@ install(FILES
     DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/mavsdk"
 )
 
+install(DIRECTORY
+    ${MAVLINK_HEADERS}/mavlink
+    DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/mavsdk
+)
+
 list(APPEND UNIT_TEST_SOURCES
     ${PROJECT_SOURCE_DIR}/mavsdk/core/callback_list_test.cpp
     ${PROJECT_SOURCE_DIR}/mavsdk/core/mavsdk_time_test.cpp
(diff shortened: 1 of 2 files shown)


Katawann:
I work with a deb package build locally on my side it's all good, but if others try to work with the deb provided on github it won't work

Katawann:
@julianoes Would it be possible to do a patch release ? Or do you have a plan to integrate this correction in v1.5.0 and when could we except to be released ?

julianoes:
Thanks for the note @Katawann. Here comes a patch for v1.4: #1911.

julianoes:
I have the feeling this probably is not required for v1.4. It seems to be broken in v1.4.7. I'll revert it.

The changes in `main` will eventually be released as v2.

Katawann:
Arf too bad. I'll wait for new release to test the installation then. Thank you for your help anyway !

JonasVautherin:
But it works in `main`, doesn't it?

Katawann:
I need to test again because I need to build the package to include it in my docker. I come back to you when I found some time to test it

JonasVautherin:
Got it! Let us know if you find a patch, would be helpful :blush:

Katawann:
I just build debian package with `main` branch but it still fail with the same error:
```
/usr/include/mavsdk/plugins/mavlink_passthrough/mavlink_include.h:3:10: fatal error: mavlink/v2.0/common/mavlink.h: No such file or directory
#include "mavlink/v2.0/common/mavlink.h"
```

Now let's found a fix which does not break installation for others...

julianoes:
Ok, so takeaway: main is broken, release should (again be fine) as it was.

PR #1821, opened 2022-06-27
Need to reinstall libudev1 for ubuntu22

JonasVautherin:
Looking into what makes the CI fail in https://github.com/mavlink/MAVSDK/pull/1816


JonasVautherin:
Seems like the problem got solved in the github images. No need for this fix anymore.

PR #1824, opened 2022-06-28
Unregister to prevent fulfilling promise twice

JonasVautherin:
Not sure if that was this one [here](https://github.com/mavlink/MAVSDK/runs/7082798986?check_suite_focus=true), since I would have expected to see "Waiting to discover vehicle" in the logs. But I would expect even more log output for the next ones, so... :sweat_smile:


julianoes (approved):
Nice, seems about right.

PR #1829, opened 2022-06-28
Fix build_and_push_docker_images.sh

JonasVautherin:
That's more how it's supposed to be, right? :thinking:


julianoes (commented):
Whoops.

PR #1835, opened 2022-07-04
Set default MAVLINK_DIALECT

JonasVautherin:
This CMakeLists.txt should work standalone, and therefore it should define the `MAVLINK_DIALECT` if unset.

### DIFF
--- a/third_party/mavlink/CMakeLists.txt
+++ b/third_party/mavlink/CMakeLists.txt
@@ -7,6 +7,10 @@ include(ExternalProject)
 set(Python3_FIND_REGISTRY "NEVER")
 find_package(Python3 COMPONENTS Interpreter REQUIRED)
 
+if (NOT MAVLINK_DIALECT)
+    set(MAVLINK_DIALECT common)
+endif()
+
 ExternalProject_add(
     mavlink
     GIT_REPOSITORY https://github.com/mavlink/mavlink


julianoes:
Does this not work?

JonasVautherin:
Not if you use `third_party/mavlink` standalone (as in, `cd third_party/mavlink` and do your thing from there, without MAVSDK at all) :see_no_evil:

julianoes:
Why would you do that? :thinking:

JonasVautherin:
For debugging/development, I run only the dependency I want to build.

Also it's actually equivalent to running [generate_mavlink_headers.sh](https://github.com/mavlink/MAVSDK/blob/main/tools/generate_mavlink_headers.sh). I would personally tend to use that rather than the script, because it's faster for me (the script I need to review before I run it to know what it will change on my system).

julianoes:
Ok fair. Should we remove the other default then? I think it's still required isn't it? I just don't like having it duplicate and potentially conflicting.

JonasVautherin:
The other one is required indeed. That's the option for the MAVSDK project, which we propagate to the MAVLink subproject. Just like CMAKE_TOOLCHAIN_FILE and others, after all, I don't see anything weird with that :thinking:.

julianoes:
Ok, fair enough.

PR #1839, opened 2022-07-05
Fix warning and improve integration test

JonasVautherin:
Also I believe this may fix a potential crash where the promise is fulfilled twice.


julianoes (commented):
Thanks!

PR #1844, opened 2022-07-07
Fix cache

JonasVautherin:
hmm `github.job.name` is not a thing, I believe :sweat_smile:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -98,8 +98,13 @@ jobs:
         run: cmake --build build/release -j2
       - name: install
         run: sudo cmake --build build/release --target install
+      - name: temporary workaround for Ubuntu 22.04
+        if: ${{ matrix.ubuntu_image }} == "ubuntu-22.04"
+        run: |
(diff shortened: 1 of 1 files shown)


JonasVautherin:
Ubuntu 22 is failing because of this damn issue I don't get:

```
Run sudo apt-get install --reinstall -y libudev1 libsdl2-dev
Reading package lists...
Building dependency tree...
Reading state information...
Reinstallation of libudev1 is not possible, it cannot be downloaded.
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:
The following packages have unmet dependencies:
 libudev-dev : Depends: libudev1 (= 249.11-0ubuntu3) but 249.11-0ubuntu3.1 is to be installed
E: Unable to correct problems, you have held broken packages.
Error: Process completed with exit code 100.
```

julianoes (commented):
Odd.

PR #1847, opened 2022-07-08
Only enable follow-me test if the improved followme' parameter exists

### LINKED ISSUE #1845: FollowMe integration tests
The [integration tests](https://github.com/mavlink/MAVSDK/blob/main/src/integration_tests/follow_me.cpp#L19) say:

> /* Updated FollowMe only works from PX4 v1.13, as it went through refactoring */

But I thought it had not been merged in v1.13.0. Has it?

@junwoo091400 FYI

junwoo091400:
As discussed in the [Dev-Call](https://discuss.px4.io/t/px4-dev-call-july-06-2022/28093#q2-julianoes-is-reworked-follow-me-available-before-or-after-113-release-8), as the [improved Follow-me](https://github.com/PX4/PX4-Autopilot/pull/18026) didn't get into PX4 v1.13, the SITL tests were failing.

This PR adds capability to do `param_get` to PX4 to check if the improved Follow-me is supported, and if not, it would skip the test (this would also skip the test for flash constrained builds, and is more proper way of handling the test skipping criteria, as the version tag itself is not sufficient indicator of the features the vehicle supports)

Fixes #1845

### DIFF
--- a/src/integration_tests/follow_me.cpp
+++ b/src/integration_tests/follow_me.cpp
@@ -230,7 +230,7 @@ bool autopilot_has_improved_followme(const std::shared_ptr<Param> param)
     // Check if the newly added parameter in the improved follow-me exists
     // Improved Follow-Me PR: https://github.com/PX4/PX4-Autopilot/pull/18026
     const std::pair<Param::Result, float> get_result = param->get_param_float("FLW_TGT_MAX_VEL");
-    
+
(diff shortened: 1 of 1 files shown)


JonasVautherin (dismissed):
Thanks!

JonasVautherin (approved):

julianoes (commented):
Thanks @junwoo091400!

PR #1848, opened 2022-07-08
Improve thirdparty mavlink

JonasVautherin:
Mostly playing.

I do believe that MAVLink should be a cmake target, linked with `MAVLink::mavlink`. And that's what I implement here.

Not saying that we should go for that, but I like the idea of doing `find_package(MAVLink REQUIRED)` and then have only one use of `MAVLink::mavlink` in our CMakeLists (instead of including `MAVLINK_HEADERS` a bit everywhere). That makes MAVLink a pure dependency, it's not because it's a header library that it is different from any other library, after all.

It keeps one instance of `MAVLINK_HEADERS`: when installing MAVSDK. So that people using MAVSDK don't need to depend on the MAVLink target because it was "absorbed" into MAVSDK. `MAVLINK_HEADERS` is set when calling `find_package(MAVLink REQUIRED)`.

Again, just playing.


julianoes:
I played with this recently but gave up after a while. I think it requires some changes in mavlink itself.

This was a first step I had to do:
https://github.com/mavlink/mavlink/pull/1859

JonasVautherin:
Oh, I did not know that mavlink had a CMakeLists.txt. But as always, it seems to be doing a lot of stuff, including running Python scripts :see_no_evil:.

My approach is slightly different: it goes in two steps:
1. Generate the project with pymavlink. The result is a simple header library with a simple CMakeLists (here CMakeLists.txt.in) that create an installable target out of it.
2. Build and install this simple cmake project. It results in the mavlink headers being installed where specified, together with a MAVLinkConfig.cmake/MAVLinkTargets.cmake, so that one can `find_package(MAVLink)` and link `MAVLink::mavlink`.

The first step is what I would expect upstream mavlink to do: generate a simple cmake project and release it (e.g. with a version number). This way we would just use it like any other dependency.

> I think it requires some changes in mavlink itself.

Well, I think this works :see_no_evil:. The only thing is that then MAVLink::mavlink is a dependency, so examples need to treat it as a dependency (like curl). However I could link it as PRIVATE in our projects and install the headers with the MAVSDK headers, such that MAVSDK would "absorb" MAVLink entirely.

The cool thing about treating MAVLink as an actual dependency is that one could update MAVLink without re-building MAVSDK (unless MAVLink breaks the API), and I believe dialects could be used in mavlink_passthrough without rebuilding MAVSDK.

julianoes:
> it seems to be doing a lot of stuff, including running Python scripts :see_no_evil:.

You can disable that now: https://github.com/mavlink/mavlink/pull/1859

julianoes:
> install the headers with the MAVSDK headers, such that MAVSDK would "absorb" MAVLink entirely.

That's sorta how it is right now, right?

JonasVautherin:
> That's sorta how it is right now, right?

Yes, kind of. The approach in this PR would only have benefits if MAVLink generated a proper library (instead of just header files that people copy-paste inside their project). But given that probably all projects are used to copy-pasting the headers (well, admittedly a fancy copy-paste with a git submodule), users would probably be confused if MAVSDK required them to install MAVLink as a dependency :thinking:.

Ah well that was a good exercise.

julianoes:
I still like your changes though, I just would love it if the actual MAVLink library would build and install like this.

PR #1850, opened 2022-07-10
added optional mavsdk server built with grpc proto reflection. This a…

ykhedar:
…llows quick debugging using tools like grpcurl. The feature can be tested as follows:

1. Build the server with the following command from the src directory:

`cmake -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=OFF -DBUILD_MAVSDK_SERVER=ON -DBUILD_WITH_PROTO_REFLECTION=ON -Bbuild/default -H.
`

2. Run the SITL with this server.
3. Use [grpcurl](https://github.com/fullstorydev/grpcurl) like follows:

`./grpcurl -plaintext localhost:50051 mavsdk.rpc.telemetry.TelemetryService/SubscribePosition`

This should print the telemetry:

`{
  "position": {
    "latitude_deg": 47.397742099999995,
    "longitude_deg": 8.5455937,
    "absolute_altitude_m": 488.07004,
    "relative_altitude_m": 0.010000001
  }
}`
`
{
  "position": {
    "latitude_deg": 47.397742099999995,
    "longitude_deg": 8.5455937,
    "absolute_altitude_m": 488.07004,
    "relative_altitude_m": 0.010000001
  }
}`
`
{
  "position": {
    "latitude_deg": 47.397742099999995,
    "longitude_deg": 8.5455937,
    "absolute_altitude_m": 488.07004,
    "relative_altitude_m": 0.010000001
  }
}`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,6 +3,7 @@ cmake_minimum_required(VERSION 3.13)
 option(HUNTER_ENABLED "Enable Hunter package manager support" OFF)
 option(SUPERBUILD "Build dependencies" ON)
 option(BUILD_MAVSDK_SERVER "Build mavsdk_server" OFF)
+option(BUILD_WITH_PROTO_REFLECTION "Build MAVSDK Server with Proto Reflection" OFF)
 option(BUILD_SHARED_LIBS "Build core as shared libraries instead of static ones" ON)
 
 if(SUPERBUILD AND HUNTER_ENABLED)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,7 +3,7 @@ cmake_minimum_required(VERSION 3.13)
 option(HUNTER_ENABLED "Enable Hunter package manager support" OFF)
 option(SUPERBUILD "Build dependencies" ON)
 option(BUILD_MAVSDK_SERVER "Build mavsdk_server" OFF)
-option(BUILD_WITH_PROTO_REFLECTION "Build MAVSDK Server with Proto Reflection" OFF)
+option(BUILD_WITH_PROTO_REFLECTION "Build mavsdk_server with proto reflection" OFF)
 option(BUILD_SHARED_LIBS "Build core as shared libraries instead of static ones" ON)
 
 if(SUPERBUILD AND HUNTER_ENABLED)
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -3,6 +3,7 @@ cmake_minimum_required(VERSION 3.13)
 option(HUNTER_ENABLED "Enable Hunter package manager support" OFF)
 option(SUPERBUILD "Build dependencies" ON)
 option(BUILD_MAVSDK_SERVER "Build mavsdk_server" OFF)
+option(BUILD_WITH_PROTO_REFLECTION "Build MAVSDK Server with Proto Reflection" OFF)
```
JonasVautherin, line 6: Nitpick, for consistency with the option above :innocent:.

```suggestion
option(BUILD_WITH_PROTO_REFLECTION "Build mavsdk_server with proto reflection" OFF)
```

src/mavsdk_server/src/CMakeLists.txt
```diff
@@ -43,10 +44,15 @@ else()
     add_library(mavsdk_server ${MAVSDK_SERVER_SOURCES})
 endif()
 
+if(BUILD_WITH_PROTO_REFLECTION)
```
JonasVautherin, line 47: Just wondering: instead of having that split in 3 places, what about simply adding another `target_link_libraries` below the current one? Like this:

```cmake
if(BUILD_WITH_PROTO_REFLECTION)
    target_link_libraries(mavsdk_server
        PRIVATE
        gRPC::grpc++_reflection
    )
endif()
```

Feels more direct/clearer to me. Do you have an opinion?

JonasVautherin (changes requested):
Looks nice!

Just a nitpick and an open question for you :blush:

src/mavsdk_server/src/CMakeLists.txt
ykhedar: Agree with you there :) and updated accordingly.

src/mavsdk_server/src/grpc_server.cpp
```diff
@@ -142,6 +142,10 @@ int GrpcServer::run()
     builder.RegisterService(&_tune_service);
 #endif
 
+#ifdef BUILD_WITH_PROTO_REFLECTION
```
JonasVautherin, line 145: Wait, does this have an effect? I don't think cmake will propagate it from `-DBUILD_WITH_PROTO_REFLECTION` to C++. You have to explicitly set it from the CMakeLists, right? Or does this work? :thinking:

ykhedar: You caught me there ;). commited with the fix.

JonasVautherin (approved):
Thanks a lot!

julianoes (approved):

PR #1854, opened 2022-07-13
Just for testing

JonasVautherin:
There is this weird crash on iOS when the wifi hotspot (of Herelink, not on the iOS device) is stopped. Not when e.g. mavlink-router is stopped, but really when the wifi hotspot is stopped. I don't get what and can't manage to get a good stack trace.

I have no clue how the wifi connection would be related to a crash (mavsdk_server receives datagrams over UDP, it should not care about the wifi connection at all).

Anyway, I am wondering if maybe there is something in the gRPC server. This is completely a guess, I want to see what happens if the gRPC server only accepts localhost connections.


JonasVautherin:
That's not it, but it was worth a try :(

PR #1855, opened 2022-07-14
Local protoc-gen-mavsdk install

julianoes:
@devbharat suggested to install the protoc-gen-mavsdk package locally and therefore avoiding some of the potential headaches when the system wide package is not the correct version, or not installed.

Objections @JonasVautherin?


JonasVautherin (approved):
Seems alright to me :+1:

PR #1861, opened 2022-07-15
Make commands thread-safe

JonasVautherin:
We cannot send twice the same command in MAVLink (because there is no way to differentiate the answers), and therefore MAVSDK drops a new command if an older similar command is already running. MAVSDK returns the failing code "Denied" in that case.

However, the MAVSDK command functions are not thread-safe at the moment. So if I call `land_async()` concurrently, there is a chance that two commands will get in the queue at the same time, creating problems.

I believe this is what I observe from MAVSDK-Swift (because mavsdk_server calls the MAVSDK functions concurrently).

This tries to fix it at the MAVSDK-C++ level, by making the "request" functions thread-safe.

Still needs further testing to check that it fixes my MAVSDK-Swift issue :grin:.

### DIFF
--- a/src/mavsdk/plugins/action/action.cpp
+++ b/src/mavsdk/plugins/action/action.cpp
@@ -3,6 +3,7 @@
 // (see https://github.com/mavlink/MAVSDK-Proto/blob/master/protos/action/action.proto)
 
 #include <iomanip>
+#include <mutex>
 
 #include "action_impl.h"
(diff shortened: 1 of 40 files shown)


julianoes:
I don't think this is required as the step which adds the commands on to the queue is thread safe.

JonasVautherin:
Yes, I think your fixes #1862 and #1863 solved my issue. Thanks again!

PR #1862, opened 2022-07-15
core: second in flight command fix

julianoes:
In order to prevent too many commands from clogging the queue, we remove them if they look identical.

This changes it so that we only drop a command if it does not come with a callback. If it has a callback associated with it, it's better to call it anyway, otherwise odd races or just Denied results when you don't expect it tend to be the result.

This is the same as I fixed previously, however, this time also for CommandLong which I forgot last time.

### DIFF
--- a/src/mavsdk/core/mavlink_command_sender.cpp
+++ b/src/mavsdk/core/mavlink_command_sender.cpp
@@ -107,13 +107,11 @@ void MavlinkCommandSender::queue_command_async(
     CommandIdentification identification = identification_from_command(command);
 
     for (const auto& work : _work_queue) {
-        if (work->identification == identification) {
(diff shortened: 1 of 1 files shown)


JonasVautherin (commented):
But then isn't there the problem that when we run something like `REQUEST_MESSAGE` every second, but it takes 3 seconds to time out, then we keep growing the queue?

julianoes:
> then we keep growing the queue?

Yes, but for requests that we do every few seconds, we usually do it without callback. And if we do it with callback then we would realize that it doesn't work or takes a long time, and could fix it.

julianoes:
I'm merging this so that we at least match `CommandLong` and `CommandInt`.

PR #1864, opened 2022-07-19
Bump DEPLOYMENT_TARGET to 14.0

JonasVautherin:
There seems to be issues with iOS 15.5. Just trying to bump the deployment target to see if it helps...

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -573,7 +573,7 @@ jobs:
       - name: install pymavlink dependencies
         run: python3 -m pip install future
       - name: configure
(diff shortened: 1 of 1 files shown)


julianoes (approved):
No objections from me. @byuarus?

byuarus:
It looks that you also need to change the minimum iOS version number here:
https://github.com/mavlink/MAVSDK/blob/631f6d366a5683b773560d8abe9752b54afe98fa/src/mavsdk_server/cmake/MacOSXFrameworkInfo.plist.in#L24

From my tests I've seen that if the info.plist file shows a different min iOS version then the framework supports, Apple will reject the build.

JonasVautherin:
Actually I would not merge that unless we establish that it's needed. I thought I had opened it as a draft PR, sorry :sweat_smile:

PR #1877, opened 2022-08-02
mavsdk_server: stop and destroy server and mavsdk

julianoes:
This fixes the problem where mavsdk_server_stop does not actually stop anything and then fails with a bind error when you start it again.

I'm not sure if this is the correct approach though.

Might fix https://github.com/mavlink/MAVSDK-Java/issues/96.

### DIFF
--- a/src/mavsdk_server/src/mavsdk_server.cpp
+++ b/src/mavsdk_server/src/mavsdk_server.cpp
@@ -15,13 +15,19 @@ public:
 
     bool connect(const std::string& connection_url)
     {
-        _connection_initiator.start(_mavsdk, connection_url);
+        if (_mavsdk == nullptr) {
+            _mavsdk = std::make_unique<mavsdk::Mavsdk>();
+        }
+        _connection_initiator.start(*_mavsdk, connection_url);
         return _connection_initiator.wait();
     }
 
     int startGrpcServer(const int port)
     {
-        _server = std::make_unique<GrpcServer>(_mavsdk);
+        if (_mavsdk == nullptr) {
+            _mavsdk = std::make_unique<mavsdk::Mavsdk>();
+        }
+        _server = std::make_unique<GrpcServer>(*_mavsdk);
         _server->set_port(port);
         _grpc_port = _server->run();
         return _grpc_port;
@@ -35,19 +41,21 @@ public:
 
         if (_server != nullptr) {
             _server->stop();
+            _server.reset();
         }
+        _mavsdk.reset();
     }
 
     int getPort() { return _grpc_port; }
 
     void setMavlinkIds(uint8_t system_id, uint8_t component_id)
     {
-        _mavsdk.set_configuration(mavsdk::Mavsdk::Configuration{system_id, component_id, false});
+        _mavsdk->set_configuration(mavsdk::Mavsdk::Configuration{system_id, component_id, false});
     }
 
 private:
-    mavsdk::Mavsdk _mavsdk;
     ConnectionInitiator<mavsdk::Mavsdk> _connection_initiator;
+    std::unique_ptr<mavsdk::Mavsdk> _mavsdk;
     std::unique_ptr<GrpcServer> _server;
     int _grpc_port;
 };


src/mavsdk_server/src/mavsdk_server.cpp
```diff
@@ -15,13 +15,19 @@ class MavsdkServer::Impl {
...
         return _connection_initiator.wait();
     }
 
     int startGrpcServer(const int port)
     {
-        _server = std::make_unique<GrpcServer>(_mavsdk);
+        if (_mavsdk == nullptr) {
+            _mavsdk = std::make_unique<mavsdk::Mavsdk>();
+        }
+        _server = std::make_unique<GrpcServer>(*_mavsdk);
```
JonasVautherin, line 30: Does that work? What happens if `_mavsdk` has not discovered any System when `_server` is created? It's supposed to take the first system and talk to it, if I remember correctly. The reason why `connect()` is blocking is that `_mavsdk_` needs to have a `System` discovered before the gRPC server can work (otherwise what should the gRPC server do with requests?)

JonasVautherin (commented):
When is it the case that `stop()` does not stop anything? I would think that the `attach()` function returns when the server is stopped (so calling `stop()` is not enough, you know that it was stopped when `attach` returned).

src/mavsdk_server/src/mavsdk_server.cpp
julianoes: This is like before, isn't it? You create Mavsdk but you lazily create the plugins once you require them and have a sytem to pass to them

JonasVautherin: Well it has changed back and forth a few times over the years, but it seems like the lazy plugins now allow that again. My concern is that every time it changed, it broke the API (`mavsdk_server.run()` was blocking, then not, then blocking again, etc).

Now with the solution throwing `NoSystem`, making it non-blocking means that the user now has to try repeatedly and start working once a system is there. So I honestly like the blocking call better, I think :thinking:.

src/mavsdk_server/src/mavsdk_server.cpp
```diff
@@ -35,19 +41,21 @@ class MavsdkServer::Impl {
 
         if (_server != nullptr) {
             _server->stop();
+            _server.reset();
         }
+        _mavsdk.reset();
```
JonasVautherin, line 46: Isn't that similar to calling [`destroy`](https://github.com/mavlink/MAVSDK/blob/main/src/mavsdk_server/src/mavsdk_server_api.cpp#L53)?

I realise that MAVSDK should probably be stopped on `stop()`, but why doing it on `start()`? :thinking:

src/mavsdk_server/src/mavsdk_server.cpp
julianoes: Did I change that though? Maybe I don't understand what I changed.

src/mavsdk_server/src/mavsdk_server.cpp
julianoes: This is in `stop()`, it's just GitHub swallowing a few lines.

JonasVautherin: Oh, my bad

src/mavsdk_server/src/mavsdk_server.cpp
JonasVautherin: Ooooh, I get it now. You essentially just changed it such that `MAVSDK` is a pointer that you can reset in order to "stop" it. Well I guess that works, too. I would like it better if `stop()` called `mavsdk.stop()` or `mavsdk.remove_all_connections()`, but probably this works :+1:

JonasVautherin (approved):

src/mavsdk_server/src/mavsdk_server.cpp
julianoes: Yes, I will implement, so that you can remove connections.

julianoes:
So we don't need this anymore, right @JonasVautherin!?

Issue #1569, opened 2021-10-06
Telemetry.set_rate_attitude() seems to have no effect


fas2037:
'm trying to set the rate of the attitude messages. I tried setting it to a lower hertz rate, for example 10 Hz. But the messages still seem to get published in a very high rate (100 Hz)

I subscribe to Telemetry.subscribe_attitude_quaternion()

The function should set the rate of this MAVLINK Message.
https://github.com/mavlink/MAVSDK/blob/ee84bd93b7e0ef802963503a34783772e31eb10e/src/plugins/telemetry/telemetry_impl.cpp#L215-L219

So I took a look at this:
https://github.com/mavlink/MAVSDK/blob/ee84bd93b7e0ef802963503a34783772e31eb10e/src/plugins/telemetry/telemetry_impl.cpp#L640-L720
I assume that both of this functions are able to publish the quaternion. Is there any cause for that? When checking with QGC both MAVLINK_MSG_ID_ATTITUDE_QUATERNION and MAVLINK_MSG_ID_ATTITUDE are published with 50 Hz. I guess as MAVSDK takes both messages into account and therefore the receiving rate of 100 Hz is archived.

But when using Telemetry.set_rate_attitude() I assume that this will set the rate of the subscription to a attitude message, for example the quaternion, to the desired one. This is not the case right now.

fas2037:
I'm trying to set the rate of the attitude messages. I tried setting it to a lower hertz rate, for example 10 Hz. But the messages still seem to get published in a very high rate (100 Hz)

I subscribe to Telemetry.subscribe_attitude_quaternion()

The function should set the rate of this MAVLINK Message.
https://github.com/mavlink/MAVSDK/blob/ee84bd93b7e0ef802963503a34783772e31eb10e/src/plugins/telemetry/telemetry_impl.cpp#L215-L219

So I took a look at this:
https://github.com/mavlink/MAVSDK/blob/ee84bd93b7e0ef802963503a34783772e31eb10e/src/plugins/telemetry/telemetry_impl.cpp#L640-L720
I assume that both of this functions are able to publish the quaternion. Is there any cause for that? When checking with QGC both MAVLINK_MSG_ID_ATTITUDE_QUATERNION and MAVLINK_MSG_ID_ATTITUDE are published with 50 Hz. I guess as MAVSDK takes both messages into account and therefore the receiving rate of 100 Hz is archived.

[...]

PR #1878, opened 2022-08-02
telemetry: split attitude into Euler + Quaternion

julianoes:
Before we processed whatever attitude was being sent which meant that both MAVLink messages ATTITUDE and ATTITUDE_QUATERNION would trigger both callbacks. This was often ok but can be very confusing when trying
to set the rate of the messages.

This commit decouples this magic and provides setters for both the rates.

Depends on https://github.com/mavlink/MAVSDK-Proto/pull/293.
Closes https://github.com/mavlink/MAVSDK/issues/1569.


JonasVautherin (approved):

julianoes:
Tested, works, merging.

PR #1880, opened 2022-08-11
 mission_raw: implement MISSION_CHANGED 

julianoes:
This implements the new message MISSION_CHANGED which is useful to know
if a mission has been updated on the vehicle.


julianoes:
This will be replaced by MISSION_CHECKSUM.

PR #1887, opened 2022-08-24
tools: fix clang-format error message

julianoes:
This was hard-coded when it should not have been.


JonasVautherin (approved):
:sweat_smile:

PR #1888, opened 2022-08-24
Add the ability to send commands which does not receive ack

chep:
Hi,

I have a device which, for some undocumented reason, does not send ack after some commands.
I need to add an extra argument to the send_command functions and return immediately after sending if this argument is `true`.

I don't know if you are interested in this patch but I think it may help someone one day.

Sync functions are not tested, sorry.

### DIFF
--- a/src/mavsdk/core/mavlink_command_sender.cpp
+++ b/src/mavsdk/core/mavlink_command_sender.cpp
@@ -28,7 +28,8 @@ MavlinkCommandSender::~MavlinkCommandSender()
 }
 
 MavlinkCommandSender::Result
-MavlinkCommandSender::send_command(const MavlinkCommandSender::CommandInt& command)
+MavlinkCommandSender::send_command(
+    const MavlinkCommandSender::CommandInt& command, bool no_ack)
 {
     // We wrap the async call with a promise and future.
     auto prom = std::make_shared<std::promise<Result>>();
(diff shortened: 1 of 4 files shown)


JonasVautherin:
I see where you come from, but I think that this is a bug on the device side, and I would be in favour of not supporting that bug in MAVSDK upstream :sweat_smile:.

If the drone does not answer to the command, then it is not really a command anymore: it's just an unreliable message :thinking:.

@julianoes any opinion there?

julianoes:
I agree with @JonasVautherin. The whole point of a command is that it is reliable by having an ack.

What is exactly your use case, or your device?

Alternatively, you can at least enable this by sending the command using mavlink passthrough.

chep:
I agree with you too.
My device (gimbal) uses a custom DO_DIGICAM_CONTROL command  which do not conform to Mavlink protocol. Usually, it sends ack back but with some parameters, it executes the command (I can see it) but does not send ack. I know this is a bug of the device but I need to handle this case.

I understand that you can't merge this, I think you should not merge it. It was just in case someone need this hack too...

julianoes:
Which gimbal is that? Any chance they want to talk to me and I would help them with implementing the proper MAVLink gimbal v2 protocol?

julianoes:
As discussed: we don't want to support non-standard use of commands. The whole point of a command is that it comes with an ack. Implementations that use commands without ack should be adapted.

Closing this, sorry, but thanks for the PR!

PR #1893, opened 2022-09-05
mission_raw: add home position at 0 for ArduPilot

julianoes:
ArduPilot stores the home position in mission item 0. This means that we need to add the home position as saved in the .plan file to the mission raw list of mission items.

### DIFF
--- a/src/mavsdk/core/mavlink_mission_transfer.cpp
+++ b/src/mavsdk/core/mavlink_mission_transfer.cpp
@@ -237,11 +237,27 @@ void MavlinkMissionTransfer::UploadWorkItem::start()
         return;
     }
 
-    int count = 0;
-    for (const auto& item : _items) {
-        if (count++ != item.seq) {
-            callback_and_reset(Result::InvalidSequence);
-            return;
+    // Mission items for ArduPilot are off by one because the home position is
+    // item 0, and the sequence starts counting at 0 from 1.
+    // This is only for missions, rally points, and geofence items are normal.
+    const bool is_ardupilot_mission =
+        _sender.autopilot() == Sender::Autopilot::ArduPilot && _type == MAV_MISSION_TYPE_MISSION;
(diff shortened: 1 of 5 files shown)


src/mavsdk/core/mavlink_mission_transfer.cpp
```diff
@@ -237,11 +237,27 @@ void MavlinkMissionTransfer::UploadWorkItem::start()
...
 
-    int count = 0;
-    for (const auto& item : _items) {
-        if (count++ != item.seq) {
-            callback_and_reset(Result::InvalidSequence);
-            return;
+    // Mission items for ArduPilot are off by one because the home position is
+    // item 0, and the sequence starts counting at 0 from 1.
+    // This is only for missions, rally points, and geofence items are normal.
+    const bool one_off =
```
JonasVautherin, line 243: Nitpick, trying to keep the ardupilot-specific bits very clear:

```suggestion
    const bool is_ardupilot_mission =
```

src/mavsdk/core/mavlink_mission_transfer.cpp
```diff
@@ -237,11 +237,27 @@ void MavlinkMissionTransfer::UploadWorkItem::start()
...
+        _sender.autopilot() == Sender::Autopilot::ArduPilot && _type == MAV_MISSION_TYPE_MISSION;
+
+    if (one_off) {
```
JonasVautherin, line 246:
```suggestion
    if (is_ardupilot_mission) {
```

JonasVautherin (approved):

src/mavsdk/core/mavlink_mission_transfer.cpp
julianoes: Much better, thanks!

JonasVautherin (approved):

PR #1897, opened 2022-09-10
ArduPlane Feature added.

bahadirbuber:
Arduplane flight modes added.


JonasVautherin:
Looks promising! Failing some tests though.

Also do not forget to fix the style before pushing, with `./tools/fix_style.sh .` :innocent:

src/mavsdk/core/flight_mode.h
```diff
@@ -7,6 +7,9 @@ namespace mavsdk {
 
 enum class FlightMode {
     Unknown,
+    FBWA,
```
julianoes, line 10:
```suggestion
    Fbwa,
```

src/mavsdk/core/system_impl.cpp
```diff
@@ -844,6 +844,16 @@ SystemImpl::make_command_ardupilot_mode(FlightMode flight_mode, uint8_t componen
                     static_cast<float>(flight_mode_to_ardupilot_rover_mode(flight_mode));
             }
             break;
+        case MAV_TYPE::MAV_TYPE_FIXED_WING:
+            if (flight_mode_to_ardupilot_plane_mode(flight_mode) == ardupilot::PlaneMode::Unknown) {
+                LogErr() << "Cannot translate flight mode to ardupilot pplane mode.";
```
julianoes, line 849:
```suggestion
                LogErr() << "Cannot translate flight mode to ArduPlane mode.";
```

src/mavsdk/core/flight_mode.h
julianoes: And I'd probably add the new modes at the bottom rather than the top, so we don't change the existing indices.

julianoes (commented):
Thanks for the PR. See inline comments.

julianoes:
I've added a commit to clean this up and fix style according to the MAVSDK conventions.

JonasVautherin (approved):
Great! :rocket:

PR #1904, opened 2022-09-23
core: remove wrong/confusing warning

julianoes:
This must have snucked in during some merge.


JonasVautherin (approved):
Whoops :see_no_evil:

PR #1923, opened 2022-11-02
tools: Refactor changelog generation

Andreea-L:
This PR reworks the `generate_debian_changelog.sh`. Namely it:

* Adds args for author and email, so we stop tagging every mavsdk dev with Helge's name; when not provided, by default the author is "MAVSDK Team"
* Replaces the previous git describe used for the default tag version with either the current commit hash (+ "0v" prepended) or the tag which points to the commit hash. This is to prevent `dpkg-buildpackage` errors, since in the cases where a commit is not tagged and its hash does not start with a digit, the deb build fails:
```
dpkg-buildpackage: error: version number does not start with digit
```
This meant you basically couldn't reliably issue a build on a non-tagged mavsdk commit.

### DIFF
--- a/tools/generate_debian_changelog.sh
+++ b/tools/generate_debian_changelog.sh
@@ -1,20 +1,71 @@
 #!/usr/bin/env bash
 
-# We want to extract "1.2.3" from "v1.2.3-5-g123abc".
-tag_version=`git describe --always --tags | sed 's/v\([0-9]*\.[0-9]*\.[0-9]*\).*$/\1/'`
-
-# Default to 1 for package version
-if [ "$#" == 1 ]; then
-    package_version=$1
-else
-    package_version="1"
-fi
+set -e
 
 # Date according to RFC 5322
-date=`date -R`
+DATE=$(date -R)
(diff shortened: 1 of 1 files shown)


tools/generate_debian_changelog.sh
```diff
@@ -1,20 +1,71 @@
...
+        usage_and_exit 1
+        ;;
+    esac
+done
 
-echo "mavsdk ($tag_version-$package_version) unstable; urgency=medium"
+echo "mavsdk ($VERSION-$PRE_RELEASE) unstable; urgency=medium"
 echo ""
-echo "  * Initial release"
+echo "  * Auterion MAVSDK release"
```
JonasVautherin, line 69:
```suggestion
echo "  * MAVSDK release"
```

Andreea-L: Good catch! :sweat_smile:

JonasVautherin (approved):
Thanks! :rocket:

julianoes (approved):
Amazing, thanks for the PR!

PR #1906, opened 2022-09-23
mavlink: install pymavlink dependencies

julianoes:
This automatically installs the pip packages lxml and future as required.

PR #1926, opened 2022-11-09
Move openssl dep to MAVSDK_SERVER only

JonasVautherin:
Let's see if that works.


julianoes:
Let's merge it on top of #1906, I'm fixing the Python deps and Fedora and what not there.

PR #1927, opened 2022-11-14
info: add protocol capabilities to the Info plugin

TSC21:
We were apparently missing a way of getting the capabilities through this plugin.


TSC21:
Info plugin integration tests pass as expected. The other failures are unrelated.

julianoes:
I don't think it makes sense to expose the MAVLink internals that MAVSDK tries to take care of. See bigger discussion in https://github.com/mavlink/MAVSDK-Proto/pull/301.

PR #1929, opened 2022-11-15
Use latest MAVLink definitions

TSC21:
- Fetch latest git commit tag for MAVLink.
- Update the code with the new definitions (MISSION_CURRENT, TIMESYNC and ODOMETRY).

### DIFF
--- a/third_party/mavlink/CMakeLists.txt
+++ b/third_party/mavlink/CMakeLists.txt
@@ -14,7 +14,7 @@ endif()
 ExternalProject_add(
     mavlink
     GIT_REPOSITORY https://github.com/mavlink/mavlink
-    GIT_TAG 0b5b5dabf9c6ae22f2f93f0db9a4e19dcf6f19ac
+    GIT_TAG c65c498efd2d9673e1d7da37025686daa5a7d568
     PREFIX mavlink
     CONFIGURE_COMMAND Python3::Interpreter
         -m pymavlink.tools.mavgen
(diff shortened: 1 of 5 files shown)


third_party/mavlink/CMakeLists.txt
```diff
@@ -14,7 +14,7 @@ endif()
 ExternalProject_add(
     mavlink
     GIT_REPOSITORY https://github.com/mavlink/mavlink
-    GIT_TAG 0b5b5dabf9c6ae22f2f93f0db9a4e19dcf6f19ac
```
JonasVautherin, line 17: I think I disagree with this. The effect will be that it will randomly break the build for users, who will at least be frustrated and hopefully will open an issue, which will put pressure on the maintainers (because "fix it ASAP it's broken now!").

Also ideally that should not be needed without adding a new feature to MAVSDK unless MAVLink broke backward compatibility, right? And if the problem is that MAVLink keeps breaking backward compatibility, I would say this is a MAVLink problem :innocent:.

JonasVautherin (changes requested):
Thanks a lot @TSC21!

I would vote for keeping it pinned to the commit hash you used instead of `master`, but the rest is great!

third_party/mavlink/CMakeLists.txt
TSC21: Alright, fair enough. I will revert this back to the "latest" commit hash in MAVLink.

TSC21: @JonasVautherin done

JonasVautherin (approved):
Thanks! CI is unrelated

PR #1933, opened 2022-11-24
geofence: Add support for circular geofences

YalamareddyChaitanya:
At the moment, Geofence plugin supports only polygon fences. This pull request adds support for circular fences.

### DIFF
--- a/examples/geofence_inclusion/geofence_inclusion.cpp
+++ b/examples/geofence_inclusion/geofence_inclusion.cpp
@@ -102,15 +102,18 @@ int main(int argc, char** argv)
 
     std::vector<Geofence::Polygon> polygons;
     Geofence::Polygon new_polygon{};
-    new_polygon.fence_type = Geofence::Polygon::FenceType::Inclusion;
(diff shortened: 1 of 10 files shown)


julianoes (commented):
Amazing, thanks for the contribution!

I'm going to hold off merging here until the review is done in the proto repo.

Does this work with PX4? Or just ArduPilot?

And would you mind adding an integration test or example to use this?

JonasVautherin (dismissed):
Nice!

YalamareddyChaitanya:
@julianoes I tested the new geofence api with PX4 1.13.1. Its working well.  An integration test and example are added now.

After running geofence example, if you open Qgroundcontrol, you would see similar to following image.

Screenshot from 2022-11-25 14-48-50

julianoes:
Ok, so I've merged the Proto PR, this means you need to update this branch again so that the proto submodule points to latest main, and then we can merge it here too.

YalamareddyChaitanya:
@julianoes Updated the proto submodule.

julianoes:
Nice, thanks!

julianoes (approved):

PR #1942, opened 2022-12-13
fixed log_files proto

dayjaby:
Follow-up on https://github.com/mavlink/MAVSDK-Proto/pull/304

### DIFF
--- a/examples/logfile_download/logfile_download.cpp
+++ b/examples/logfile_download/logfile_download.cpp
@@ -97,11 +97,22 @@ int main(int argc, char** argv)
         for (auto entry : get_entries_result.second) {
             std::cerr << "Got log file with ID " << entry.id << " and date " << entry.date
                       << std::endl;
-            auto result =
-                log_files.download_log_file(entry, std::string("log-") + entry.date + ".ulg");
-            if (result.first != LogFiles::Result::Success) {
+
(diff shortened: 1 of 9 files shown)


julianoes (commented):
Hm, now this breaks the API, so we can't release it for v1.4 but it would go into v2.

Is there nothing we can do to fix this in a way not to break the API? :thinking:

JonasVautherin (approved):

dayjaby:
> Is there nothing we can do to fix this in a way not to break the API? thinking

We could change the way how the API gets generated for Python/Java/... if there is a sync and async function with the same name, only expose the async one - or expose the sync one with a different name.

JonasVautherin:
> We could change the way how the API gets generated for Python/Java/... if there is a sync and async function with the same name, only expose the async one - or expose the sync one with a different name.

That seems overkill to if that's just for v1.4...

julianoes:
@JonasVautherin  had the idea of only using that fix for Python and Java, but not change for C++ here.

I create a branch with the commit: https://github.com/mavlink/MAVSDK-Proto/tree/v1.4-log-files-fix

julianoes:
> @JonasVautherin had the idea of only using that fix for Python and Java, but not change for C++ here.

Nevermind, this only goes into v2.0, so that's fine, merging.

PR #1944, opened 2022-12-15
Fix iOS CI

JonasVautherin:
Not sure why this was here in the first place, but removing it seems to solve the iOS issue


JonasVautherin:
Well then it needs to build stuff, and it's setup with the cross-compilation toolchain.

Grmbl.

PR #1949, opened 2022-12-22
Fix CI for iOS

JonasVautherin:
Let's try again, from another angle.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -526,7 +526,7 @@ jobs:
           python3 -m pip install wheel
           python3 -m pip install future
       - name: configure
-        run: cmake $superbuild $cmake_prefix_path -DENABLE_STRICT_TRY_COMPILE=ON -DCMAKE_BUILD_TYPE=Release -DCMAKE_TOOLCHAIN_FILE=$(pwd)/tools/ios.toolchain.cmake -DPLATFORM=${{ matrix.platform }} -DDEPLOYMENT_TARGET=13.0 -DBUILD_MAVSDK_SERVER=ON -DBUILD_SHARED_LIBS=OFF -DWERROR=OFF -Bbuild/${{ matrix.name }} -H.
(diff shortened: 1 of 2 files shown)


JonasVautherin:
I haven't found a better way, but also it feels like the system `python3` should be a reasonable default. And I don't know what kind of magic CMake does when it tries to find it, but that depends on the CMake version, which may be a problem.

julianoes:
I've looked into the CMake magic at some point and had to come up with the magic.

I'm a bit astonished this doesn't break for Windows. I think there, in some cases, python3 is not a thing. Let me have one more go at this before merging.

.github/workflows/main.yml
```diff
@@ -526,7 +526,7 @@ jobs:
           python3 -m pip install wheel
           python3 -m pip install future
       - name: configure
-        run: cmake $superbuild $cmake_prefix_path -DENABLE_STRICT_TRY_COMPILE=ON -DCMAKE_BUILD_TYPE=Release -DCMAKE_TOOLCHAIN_FILE=$(pwd)/tools/ios.toolchain.cmake -DPLATFORM=${{ matrix.platform }} -DDEPLOYMENT_TARGET=13.0 -DBUILD_MAVSDK_SERVER=ON -DBUILD_SHARED_LIBS=OFF -DWERROR=OFF -Bbuild/${{ matrix.name }} -H.
+        run: cmake $superbuild $cmake_prefix_path -DENABLE_STRICT_TRY_COMPILE=ON -DCMAKE_BUILD_TYPE=Release -DCMAKE_TOOLCHAIN_FILE=$(pwd)/tools/ios.toolchain.cmake -DPLATFORM=${{ matrix.platform }} -DDEPLOYMENT_TARGET=13.0 -DBUILD_MAVSDK_SERVER=ON -DBUILD_SHARED_LIBS=OFF -DWERROR=OFF -Bbuild/${{ matrix.name }} -DPython_FIND_STRATEGY=LOCATION -DPython3_EXECUTABLE=$(command -v python3) -H.
```
JonasVautherin, line 529: I hate this line :laughing:. But I guess we don't really have a choice here if we want CMake to detect python.

julianoes: But this line is mostly just for CI, so it can be ugly. If only it worked though...

JonasVautherin: Seems like it does!

julianoes: Oh I did it! Holy moly.

julianoes: Let's see if we can without that and just the "don't use framework" option.

PR #1950, opened 2022-12-22
mission_raw: add method to import plan from string

### LINKED ISSUE #1948: Directly import a mission as character data
I'm currently developing a mission management for an autonomous drone in Python, using MAVSDK's Python bindings. A mission is sent as a string to the drone's companion computer via LTE, gets imported using the command `drone.mission.import_qgroundcontrol_mission` and is then uploaded to the flight controller.
The inconvenience here is, that the function `import_qgroundcontrol_mission` only accepts a filename as input, so I have to write the received mission string to a file, Import that file and delete it again afterwards.This is especially inconvenient since most companion computers will use SD cards as their memory, which are notoriously short lived when receiving frequent write operations. Every write operation to disk you can avoid should be avoided.

Could we have an alternative import operation that accepts the plan data directly as string? I attempted integrate this function myself, but been struggling with correctly building everything for some time now. For a developer who has the build environment readily set up and understands the workings of MAVSDK, the effort should be very small.

julianoes:
This enables to import mission plan files without having to write them to some location first.

Depends on https://github.com/mavlink/MAVSDK-Proto/pull/305.

Closes #1948.

### DIFF
--- a/src/mavsdk/plugins/mission_raw/include/plugins/mission_raw/mission_raw.h
+++ b/src/mavsdk/plugins/mission_raw/include/plugins/mission_raw/mission_raw.h
@@ -417,7 +417,7 @@ public:
     void unsubscribe_mission_changed(MissionChangedHandle handle);
 
     /**
-     * @brief Import a QGroundControl missions in JSON .plan format.
+     * @brief Import a QGroundControl missions in JSON .plan format, from a file.
      *
      * Supported:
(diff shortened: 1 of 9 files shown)


JonasVautherin (dismissed):

JonasVautherin (approved):

julianoes:
I'll just merge it and assume it works.

PR #1951, opened 2023-01-03
Use CMAKE_MAKE_PROGRAM instead of hardcoding make and nmake

JonasVautherin:
Let's see if that works.


JonasVautherin:
Not working for Windows :sweat_smile:

JonasVautherin:
For OpenSSL we use `nmake`

julianoes:
Ok, I guess not :smirk:

PR #1955, opened 2023-01-11
Add option to build without cURL

### LINKED ISSUE #830: Add option to build without curl

julianoes:
This adds the cmake configure option BUILD_WITHOUT_CURL which allows building MAVSDK without the cURL dependency.

This has been requested numerous times, so it's probably worthwhile adding as an option.

For now it means that camera definition files can't be downloaded via http.

In the future we will also add the option to download the definition files via MAVLink FTP.

Closes #830.

### DIFF
--- a/src/mavsdk/core/CMakeLists.txt
+++ b/src/mavsdk/core/CMakeLists.txt
@@ -17,14 +17,12 @@ target_sources(mavsdk
     call_every_handler.cpp
     connection.cpp
     connection_result.cpp
-    curl_wrapper.cpp
     crc32.cpp
     system.cpp
     system_impl.cpp
     flight_mode.cpp
     mavsdk.cpp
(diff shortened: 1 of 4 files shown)


JonasVautherin (approved):
Ok :+1:.

Out of curiosity, why is that required? To have fewer dependencies? Also if curl is not built, is `zlib` required?

julianoes:
> Also if curl is not built, is zlib required?

Good point. I don't know where zlib is even required?

JonasVautherin:
> Good point. I don't know where zlib is even required?

At least Protobuf, I would say. And otherwise I would say curl (for e.g. compressed HTTP responses) :thinking:.

PR #1968, opened 2023-01-28
core: always send heartbeat on serial connections

### LINKED ISSUE #1967: MAVSDK Only Connects to PX4 Autopilot AFTER QGroundControl is Launched!
# Problem Description
Hi, I am having an issue with connecting to the autopilot on an OrangeCube flight controller (running PX4 latest stable firmware version) with MAVSDK. Strangely, MAVSDK **can** find the user defined serial port of the OrangeCube, but fails to find an autopilot when subscribing to the system.

Even more strangely, if I first open QGroundControl (QGC) and allow it to connect to the flight controller, then close QGC, then re-run the MAVSDK executable (exactly as before), the autopilot is found and a connection is successfully made! 

I have also determined that connecting the OrangeCube to QGC through my laptop, then disconnecting the OrangeCube (making sure that it still has power) and then connecting the usb cable to another machine (specifically, Raspberry Pi 4B+) will actually allow the autopilot subscription to succeed when running any MAVSDK example. My deduction here is that QGC is doing something to the OrangeCube/ PX4 firmware which allows the autopilot to be found. 

I have also found a [PX4 forum post](https://discuss.px4.io/t/connection-only-works-after-opening-qgc/19085/2) from others with the same issue. They found a work around by running a pymavlink script prior to their MAVSDK code (which I tried but didn't work). I suspect that MAVSDK does not include a connection step it the codebase (which QGC does) and could be a potential development activity for the next release.

# The Step-by-Step
After physically connecting the OrangeCube microUSB port to my laptop (```uname -a``` gives ```Linux <machine> 5.15.0-58-generic #64~20.04.1-Ubuntu SMP Fri Jan 6 16:42:31 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux``` via usb, I run the following command:

```sh
[IN] cd /dev/serial/by-id && ls
[OUT] usb-CubePilot_CubeOrange_0-if00
```

Next I go to the [Battery example code](https://github.com/mavlink/MAVSDK/tree/main/examples/battery) in MAVSDK, build it and pass in the serial usb connection, with the _baud rate = 57600_ as an input as follows:

```sh
[IN] cmake -S . -B build
[IN] make -j4 -S build/
[IN] ./build/battery serial:///dev/serial/by-id/usb-CubePilot_CubeOrange_0-if00:57600

[OUT]
[12:31:21|Info ] MAVSDK version: v1.4.7 (mavsdk_impl.cpp:20)
Waiting to discover system...
No autopilot found.
```

Next, I launch QGroundControl. Wait for the flight controller to be detected and then quit it.
```sh
[IN] ./QGroundControl
[IN] ctrl+c
```
Following the same steps as before, I re-run the battery example. The autopilot connects!
```sh
[IN] ./build/battery serial:///dev/serial/by-id/usb-CubePilot_CubeOrange_0-if00:57600

[OUT]
[12:35:02|Info ] MAVSDK version: v1.4.7 (mavsdk_impl.cpp:20)
Waiting to discover system...
[12:35:02|Debug] New: System ID: 1 Comp ID: 1 (mavsdk_impl.cpp:482)
[12:35:02|Debug] Component Autopilot (1) added. (system_impl.cpp:377)
[12:35:02|Warn ] Vehicle type changed (new type: 10, old type: 0) (system_impl.cpp:225)
[12:35:02|Debug] Discovered 1 component(s) (system_impl.cpp:578)
Discovered autopilot
disarmed
disarmed
disarmed
```

# Code Flow Analysis - MAVSDK Battery C++ Example
As the failed autopilot connection is printed to the console ```no autopilot found```, we know that the code passes the ```mavsdk.add_any_connection(argv[1])``` connection step in the code below.

```cpp
    Mavsdk mavsdk;
    ConnectionResult connection_result = mavsdk.add_any_connection(argv[1]);

    if (connection_result != ConnectionResult::Success) {
        std::cerr << "Connection failed: " << connection_result << '\n';
        return 1;
    }
```
Next, the get_system function is called...
```cpp
    auto system = get_system(mavsdk);
    if (!system) {
        return 1;
    }
```
This is where the issue is! We can see that the mavsdk.subscribe_on_new_system(...) step does not discover an autopilot in the 3 seconds which is allocated before a timeout. Feel free to bump that number up to 10 seconds, this still doesn't make a difference for me.
```cpp
std::shared_ptr<System> get_system(Mavsdk& mavsdk)
{
    std::cout << "Waiting to discover system...\n";
    auto prom = std::promise<std::shared_ptr<System>>{};
    auto fut = prom.get_future();

    // We wait for new systems to be discovered, once we find one that has an
    // autopilot, we decide to use it.
    mavsdk.subscribe_on_new_system([&mavsdk, &prom]() {
        auto system = mavsdk.systems().back();

        if (system->has_autopilot()) {
            std::cout << "Discovered autopilot\n";

            // Unsubscribe again as we only want to find one system.
            mavsdk.subscribe_on_new_system(nullptr);
            prom.set_value(system);
        }
    });

    // We usually receive heartbeats at 1Hz, therefore we should find a
    // system after around 3 seconds max, surely.
    if (fut.wait_for(seconds(3)) == std::future_status::timeout) {
        std::cerr << "No autopilot found.\n";
        return {};
    }

    // Get discovered system now.
    return fut.get();
}
```
As mentioned above, after QGC connects to the OrangeCube, you are free to connect it to any other device and run the battery example - it will work as you'd expect. This implies that QGC is changing a setting on the OrangeCube to enable autopilot subscription.

# No Issue with Earlier PX4 Firmware
I had another OrangeCube flight controller to hand running an earlier version of PX4. For completeness, I tested the same processes out with it and had no issue connecting to the autopilot with MAVSDK! Unfortunately, I upgraded the firmware and, therefore, cannot remember what version it was that worked. Hope this point supports my observations - that the issue lies somewhere in the PX4 firmware settings.

# Hardware and Software Versions
1) Linux <machine> 5.15.0-58-generic ~20.04.1-Ubuntu SMP Fri Jan 6 16:42:31 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
2) MAVSDK v1.4
3) PX4 firmware latest:stable
4) QGroundControl [v4.2.4](https://github.com/mavlink/qgroundcontrol/releases/tag/v4.2.4)

julianoes:
PX4 starting with v1.13 does not send heartbeats by default, so we need to initiate the MAVLink connection by sending heartbeats. Therefore, we override the default here and enable sending heartbeats.

Fixes #1967.

### DIFF
--- a/src/mavsdk/core/mavsdk_impl.cpp
+++ b/src/mavsdk/core/mavsdk_impl.cpp
@@ -532,6 +532,15 @@ ConnectionResult MavsdkImpl::add_serial_connection(
     if (ret == ConnectionResult::Success) {
         add_connection(new_conn);
     }
+
+    auto new_configuration = get_configuration();
+
+    // PX4 starting with v1.13 does not send heartbeats by default, so we need
+    // to initiate the MAVLink connection by sending heartbeats.
+    // Therefore, we override the default here and enable sending heartbeats.
+    new_configuration.set_always_send_heartbeats(true);
+    set_configuration(new_configuration);
+
     return ret;
 }
 
@@ -539,6 +548,11 @@ void MavsdkImpl::add_connection(const std::shared_ptr<Connection>& new_connectio
 {
     std::lock_guard<std::mutex> lock(_connections_mutex);
     _connections.push_back(new_connection);
+}
+
+Mavsdk::Configuration MavsdkImpl::get_configuration() const
+{
+    return _configuration;
 }
 
 void MavsdkImpl::set_configuration(Mavsdk::Configuration new_configuration)
--- a/src/mavsdk/core/mavsdk_impl.h
+++ b/src/mavsdk/core/mavsdk_impl.h
@@ -68,6 +68,7 @@ public:
     std::vector<std::shared_ptr<System>> systems() const;
 
     void set_configuration(Mavsdk::Configuration new_configuration);
+    Mavsdk::Configuration get_configuration() const;
 
     uint8_t get_own_system_id() const;
     uint8_t get_own_component_id() const;


JonasVautherin (approved):
Nice!

PR #1982, opened 2023-02-07
core/plugins: fix includes

julianoes:
It's worth a try.


JonasVautherin (approved):
Seems like it still builds the examples :+1:

PR #1983, opened 2023-02-08
examples: add example to set SiK NET ID

julianoes:
This depends on https://github.com/mavlink/mavlink/pull/1948.

And goes together with https://github.com/PX4/PX4-Autopilot/pull/21082.


JonasVautherin (approved):

julianoes:
@JonasVautherin the problem is that this depends on development.xml, so we have to hard-code the magic number of the command like this which is not great.

JonasVautherin:
On the other hand, adding `development.xml` to the MAVSDK headers would make it more unstable, right? At least here it kind of shows that it is "unsafe"...

julianoes:
Yer, I don't want to include `development.xml`, at least not for releases.

julianoes:
Meh.

PR #1998, opened 2023-03-08
add telemetry set_rate_sys_status()

41Mo:
This pr implements the set_rate_sys_status() function and changes the message sent by the set_rate_battery() function to MSG_ID_BATTERY. 

Previously, the MSG_ID_SYS_STATUS rate was set using the set_rate_battery() function, which can be misleading, this was described in issue #1996.

### DIFF
--- a/src/mavsdk/plugins/telemetry/include/plugins/telemetry/telemetry.h
+++ b/src/mavsdk/plugins/telemetry/include/plugins/telemetry/telemetry.h
@@ -2367,8 +2367,20 @@ public:
      */
     std::pair<Result, Telemetry::GpsGlobalOrigin> get_gps_global_origin() const;
 
+    /**
+     * @brief Set rate to 'SYS_STATUS' updates.
+     * SYS_STATUS message allows to recieve vehicle HEALTH information.
+     * This function is non-blocking. See 'set_rate_sys_status' for the blocking counterpart.
+     */
     void set_rate_sys_status_async(double rate_hz, const ResultCallback callback);
 
+    /**
+     * @brief Set rate to 'SYS_STATUS' updates.
+     * SYS_STATUS message allows to recieve vehicle HEALTH information.
+     * This function is non-blocking. See 'set_rate_sys_status_async' for the non-blocking counterpart.
+     * 
+     * @return Result of request.
+     */
     Result set_rate_sys_status(double rate_hz) const;
 
     /**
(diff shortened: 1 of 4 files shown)


src/mavsdk/plugins/telemetry/include/plugins/telemetry/telemetry.h
```diff
@@ -2367,6 +2367,22 @@ class Telemetry : public PluginBase {
...
+
+    /**
+     * @brief Set rate to 'SYS_STATUS' updates.
+     * SYS_STATUS message allows to recieve vehicle HEALTH information.
+     * This function is non-blocking. See 'set_rate_sys_status_async' for the non-blocking counterpart.
+     * 
+     * @return Result of request.
+     */
+    Result set_rate_sys_status(double rate_hz) const;
+
```
julianoes, line 2385: Note that this API is generated automatically from https://github.com/mavlink/MAVSDK-Proto/blob/main/protos/telemetry/telemetry.proto so the change needs to happen there first.

Sorry, I should have shared this earlier:
https://mavsdk.mavlink.io/main/en/cpp/contributing/plugins.html#add-api-to-proto

julianoes (changes requested):

src/mavsdk/plugins/telemetry/telemetry_impl.cpp
```diff
@@ -355,7 +355,7 @@ Telemetry::Result TelemetryImpl::set_rate_gps_info(double rate_hz)
 Telemetry::Result TelemetryImpl::set_rate_battery(double rate_hz)
 {
     return telemetry_result_from_command_result(
-        _parent->set_msg_rate(MAVLINK_MSG_ID_SYS_STATUS, rate_hz));
+        _parent->set_msg_rate(MAVLINK_MSG_ID_BATTERY_STATUS, rate_hz));
```
ykhedar, line 358: This message change was recently merged into main via PR https://github.com/mavlink/MAVSDK/pull/2021 . Maybe rebase current PR and attempt to merge again?

julianoes:
I'm sorry I didn't act on this earlier. Sometimes life takes over.

Anyway, it looks like the battery rate has been fixed by https://github.com/mavlink/MAVSDK/pull/2021 and I doubt we need health information faster than 1 Hz. I believe PX4 won't send it out quicker anyway as the pre-flight checks don't run faster than 1 Hz anyway.

 If I'm wrong with that let me know.

PR #2005, opened 2023-03-21
mavsdk_server: fix linking

julianoes:
This fixes linking on Debian Bullseye for me with
cmake 3.26.0
GCC 10.2.1

The error was:
```
/usr/bin/ld: libmavsdk_server.so.1.4.0: undefined reference to `dlsym'
/usr/bin/ld: libmavsdk_server.so.1.4.0: undefined reference to `dlopen'
/usr/bin/ld: libmavsdk_server.so.1.4.0: undefined reference to `dlclose'
/usr/bin/ld: libmavsdk_server.so.1.4.0: undefined reference to `dlerror'
/usr/bin/ld: libmavsdk_server.so.1.4.0: undefined reference to `dladdr'
```

### DIFF
--- a/src/mavsdk_server/src/CMakeLists.txt
+++ b/src/mavsdk_server/src/CMakeLists.txt
@@ -93,7 +93,7 @@ if(NOT IOS AND NOT ANDROID)
         mavsdk_server_bin.cpp
     )
 
-    target_link_libraries(mavsdk_server_bin
+    target_link_libraries(mavsdk_server_bin PRIVATE
         mavsdk_server
         mavsdk
     )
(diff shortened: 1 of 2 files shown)


src/mavsdk_server/src/CMakeLists.txt
```diff
@@ -112,6 +112,9 @@ if(NOT IOS AND NOT ANDROID)
         set_target_properties(mavsdk_server_bin PROPERTIES
             OUTPUT_NAME mavsdk_server
         )
+
+        # Fix undefined reference to `dlsym' on Linux
+        set(CMAKE_EXE_LINKER_FLAGS "-Wl,--no-as-needed -ldl")
```
JonasVautherin, line 117: Hmm feels weird to have to manually set the flag like this :thinking:

julianoes: I know. I tried using `target_link_options` but I couldn't get it right.

JonasVautherin: Is it not just a `target_link_libraries` with `dl`?

julianoes: Hmm. What's `dl`? And why would we need it?

julianoes: Hmm, dynamic linking, but where do we do that?

julianoes: Ok, I think I was able to do it using your suggestion.

JonasVautherin (approved):

PR #2010, opened 2023-03-29
Expose params through mavlink_passthrough

JonasVautherin:
The use-case is to request params from a component that is not exposed in a plugin (`param` or `camera`).

Since the command microservice is exposed in mavlink_passthrough, it feels like it makes sense to expose the params one as well :sweat_smile:


julianoes (approved):
Looks fine!

PR #2026, opened 2023-04-22
Easier discovery

julianoes:
This adds an easy blocking API which allows to wait for the first system. This should cover a big percentage of use cases, especially for the integration tests and examples.

### DIFF
--- a/src/mavsdk/core/include/mavsdk/mavsdk.h
+++ b/src/mavsdk/core/include/mavsdk/mavsdk.h
@@ -2,6 +2,7 @@
 
 #include <string>
 #include <memory>
+#include <optional>
 #include <vector>
 #include <functional>
 
@@ -173,6 +174,17 @@ public:
     std::vector<std::shared_ptr<System>> systems() const;
 
     /**
+     * @brief Get the first system that has been discovered.
+     *
+     * @param timeout_s A timeout in seconds.
+     *                  A timeout of 0 will not wait and return immediately.
+     *                  A negative timeout will wait forever.
+     *
+     * @return A system or nothing if nothing was discovered within the timeout.
+     */
+    std::optional<std::shared_ptr<System>> first_system(double timeout_s) const;
+
+    /**
      * @brief Possible configurations.
      */
     class Configuration {
@@ -303,15 +315,16 @@ public:
      *
      * This gets called whenever a system is added.
      *
-     * @note Only one subscriber is possible at any time. On a second
-     * subscription, the previous one is overwritten. To unsubscribe, pass nullptr;
-     *
      * @param callback Callback to subscribe.
+     *
+     * @return A handle to unsubscribe again.
      */
     NewSystemHandle subscribe_on_new_system(const NewSystemCallback& callback);
 
     /**
      * @brief unsubscribe from subscribe_on_new_system.
+     *
+     * @param handle Handle received on subscription.
      */
     void unsubscribe_on_new_system(NewSystemHandle handle);
 
--- a/src/mavsdk/core/include/mavsdk/mavsdk.h
+++ b/src/mavsdk/core/include/mavsdk/mavsdk.h
@@ -174,7 +174,10 @@ public:
     std::vector<std::shared_ptr<System>> systems() const;
 
     /**
-     * @brief Get the first system that has been discovered.
+     * @brief Get the first autopilot that has been discovered.
+     *
+     * @note This requires a MAVLink component with component ID 1 sending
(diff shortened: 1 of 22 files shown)


JonasVautherin (commented):
Doesn't that change the behavior of all the examples? They were connecting to the first autopilot, and now they connect to the first system, right?

julianoes:
You're right. I need to fix that. Would you add a component filter as an argument to the method, or just call it `first_autopilot`?

JonasVautherin:
> Would you add a component filter as an argument to the method, or just call it first_autopilot?

Good question. Thinking about it, doesn't it actually feel weird to filter based on another component? We had the discussion for filtering for a ground station, and I think we never really want to do that. Either you are a server and should listen to everybody, or a ground station and want to listen only to the autopilot, right?

That would be going in the direction of `first_autopilot` then. And if you need to write a server, then you have the lower level interface for that.

What's your opinion on that?

reedev:
fwiw, I tried out this branch and it's working for me. I search the systems for the system with autopilot.

julianoes:
The only other case would be to listen for a camera or gimbal.

JonasVautherin:
> The only other case would be to listen for a camera or gimbal.

Right, but usually they come in the system of an autopilot, don't they? :grin:

julianoes:
Yes, you're right!

src/mavsdk/core/mavsdk_impl.cpp
```diff
@@ -119,6 +119,49 @@ std::vector<std::shared_ptr<System>> MavsdkImpl::systems() const
     return systems_result;
 }
 
+std::optional<std::shared_ptr<System>> MavsdkImpl::first_autopilot(double timeout_s)
+{
+    {
+        std::lock_guard<std::recursive_mutex> lock(_systems_mutex);
+        if (!_systems.empty()) {
+            return {_systems[0].second};
```
JonasVautherin, line 127: Shouldn't it go through all the systems and return the first autopilot? This will just return the first system, right?

src/mavsdk/core/mavsdk_impl.cpp
```diff
@@ -119,6 +119,49 @@ std::vector<std::shared_ptr<System>> MavsdkImpl::systems() const
...
+        }
+    }
+
+    if (timeout_s == 0.0) {
```
JonasVautherin, line 131: Shouldn't this be `<= 0.0`? Would make sense to me :thinking:

src/mavsdk/core/mavsdk_impl.cpp
```diff
@@ -119,6 +119,49 @@ std::vector<std::shared_ptr<System>> MavsdkImpl::systems() const
...
+    auto handle = subscribe_on_new_system([this, &prom, &flag]() {
+        const auto system = systems().at(0);
+        if (system->is_connected() && system->has_autopilot()) {
+            std::call_once(flag, [&prom, &system]() { prom.set_value(system); });
+        }
+    });
+
+    auto fut = prom.get_future();
+
+    if (timeout_s > 0.0) {
```
JonasVautherin, line 148: Because of the condition above, I think this should always be true...

JonasVautherin (commented):
I like it, it actually removes a lot more boilerplate than I thought it would!

src/mavsdk/core/mavsdk_impl.cpp
julianoes: Whoops.

src/mavsdk/core/mavsdk_impl.cpp
julianoes: A negative timeout should wait forever.

src/mavsdk/core/include/mavsdk/mavsdk.h
```diff
@@ -172,6 +173,20 @@ class Mavsdk {
...
+     *
+     * @note This requires a MAVLink component with component ID 1 sending
+     *       heartbeats.
+     *
+     * @param timeout_s A timeout in seconds.
+     *                  A timeout of 0 will not wait and return immediately.
+     *                  A negative timeout will wait forever.
+     *
+     * @return A system or nothing if nothing was discovered within the timeout.
+     */
```
julianoes, line 187: @JonasVautherin see here.

src/mavsdk/core/mavsdk_impl.cpp
JonasVautherin: Oh, I had missed that :see_no_evil:

PR #2028, opened 2023-05-04
Update mavlink and use its new CMakeLists

JonasVautherin:
Note that non-superbuild users will have to handle MAVLink manually (e.g. by installing it on the system).


julianoes:
> Note that now, non-superbuild users will have to handle MAVLink manually

If that's only for v2, that's ok. However, we'll have to update the docs.

.github/workflows/main.yml
```diff
@@ -302,7 +312,7 @@ jobs:
     container: alpine:3.16.0
     steps:
       - name: install tools
-        run: apk update && apk add build-base cmake git linux-headers perl tar python3 py3-pip
+        run: apk update && apk add build-base cmake git linux-headers perl tar python3 py3-pip py3-future
```
julianoes, line 315: future should come with pip, no need to install it.

julianoes (commented):
Nice!

third_party/mavlink/mavlink.patch
```diff
@@ -0,0 +1,21 @@
...
++# Needed to find the right Python on macOS
++if (NOT DEFINED Python_FIND_FRAMEWORK)
++    set(Python_FIND_FRAMEWORK "LAST")
++endif()
++
++# Needed to find the right Python on Windows
++if (NOT DEFINED Python_FIND_REGISTRY)
++    set(Python_FIND_REGISTRY "LAST")
++endif()
++
```
JonasVautherin, line 18: So that's a patch for now, I don't know if it should go upstream or not. I need that for the github actions to work (otherwise Windows/macOS don't detect the right Python version, and so `pip install future` has not been run for the one it picks).

The thing is that I tried manually on my Windows and I don't need that: I can just `pip install future` and then it works...

JonasVautherin:
> If that's only for v2, that's ok. However, we'll have to update the docs.

I don't think we have docs explaining the non-superbuild, do we :see_no_evil:? Can't we just say that the point of the non-superbuild is to be a standard CMake setup where people have to deal with the dependencies themselves? Because that's exactly what it is: if you set `SUPERBUILD=OFF`, it means that you are responsible for making the `find_package` calls pass. We should not have to explain what a Linux distribution is, what a package is, what installing a package on a Linux distribution means, and how CMake finds them when using `find_package` :thinking:.

On the contrary, we have the superbuild exactly for the people who don't want to learn about handling dependencies...

julianoes:
If we don't have docs, then we probably should have them. But then this PR is fine :D.

.github/workflows/main.yml
julianoes: Ah, it's this step:
https://github.com/mavlink/mavlink/blob/master/.github/workflows/test_and_deploy.yml#L67-L68

JonasVautherin: Sorry I'm confused, you're pointing to a workflow where future is installed with pip instead of the system package manager. Do you mean that installing with `apk` is a bad idea and it should be installed with `pip` instead? I really don't mind in the CI :thinking:

JonasVautherin:
@julianoes kindly pinging here, it's not entirely clear to me what I should change to get it approved :innocent:

JonasVautherin:
Hmm, feels like `pip install future` from CMake does not work for iOS for some reason, I don't really get why.

.github/workflows/main.yml
```diff
@@ -100,9 +104,9 @@ jobs:
       - name: install examples dependencies
         run: sudo apt-get install libsdl2-dev
       - name: configure examples
-        run: (cd examples && cmake -DCMAKE_BUILD_TYPE=Release -DWERROR=OFF -Bbuild -H.)
+        run: cmake -DCMAKE_PREFIX_PATH=$(pwd)/build/release/third_party/install -DCMAKE_BUILD_TYPE=Release -DWERROR=OFF -Bexamples/build -Hexamples
```
julianoes, line 107: Nice, I think that's cleaner actually.

.github/workflows/main.yml
```diff
@@ -595,9 +593,7 @@ jobs:
             choco install nasm
             echo "C:\Program Files\NASM" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
       - name: install pymavlink dependencies
-        run: |
-          python3 -m pip install --upgrade pip
-          python3 -m pip install wheel
+        run: pip install future
```
julianoes, line 596: Still required?

JonasVautherin:
I fixed the iOS build by patching mavlink and removing some of the custom commands there.

julianoes (approved):

PR #2034, opened 2023-05-15
Remove 0 in debian package name

JonasVautherin:
I don't really understand this 0. If it is to make a difference between major versions of MAVSDK, why don't we do that in the other package, `libmavsdk-dev`? It should be `libmavsdk0-dev`, right? But then:

* `libmavsdk0-dev` and `libmavsdk1-dev` would conflict on the system anyway, because the headers don't have the version number (we don't `#include mavsdk0/action.h`).
* The `MAVSDKConfig.cmake` and `mavsdk.pc` files would conflict, unless we rename them `MAVSDK0Config.cmake` and `mavsdk0.pc`

But it feels a bit weird to me.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -16,7 +16,7 @@ Depends: libmavsdk0 (= ${binary:Version}), ${misc:Depends}
 Description: MAVSDK API for Mavlink-controlled vehicles
  This provides the development headers for the mavsdk C++ library.
 
-Package: libmavsdk0
+Package: libmavsdk
 Architecture: any
 Multi-Arch: same
 Depends: ${shlibs:Depends}, ${misc:Depends}, libjsoncpp1, libcurl4, libtinyxml2-6a


julianoes:
Hm. So this has to do with the so version, and we should probably fix it before releasing v2.

The way I see it done in the debian world is that the developer package (headers + library) would be called `libmavsdk-dev` and that in turn depends on the actual library which is `libmavsdk1` (or 0?). That way you can then potentially have more than one library version installed.

That said I'm not very fond of having to create two debian packages.

JonasVautherin:
> That way you can then potentially have more than one library version installed.

How does that work with the issues I mentioned above? I.e. conflicting header package, conflicting MAVSDKConfig.cmake, conflicting mavsdk.pc?

> That said I'm not very fond of having to create two debian packages.

This makes sense to me because you don't need the dev package at runtime, you just want the `.so`.

julianoes:
> How does that work with the issues I mentioned above? I.e. conflicting header package, conflicting MAVSDKConfig.cmake, conflicting mavsdk.pc?

You can only have one header package installed at a time but you could have multiple .so versions of a library installed. 

> This makes sense to me because you don't need the dev package at runtime, you just want the .so. Though I don't really get it because what we provide in the releases is only one package, with something that [looks like a hack](https://github.com/mavlink/MAVSDK/blob/main/tools/create_packages.sh#L69)? thinking

Right, I think the proper way would be to somehow add a ppa for the deb packages but that seemed way too hard whenever I had a look at it. So if we don't have that, we might as well just have the one library without that so number.

JonasVautherin:
> You can only have one header package installed at a time but you could have multiple .so versions of a library installed.

But what's the point of that? If the `.so` is compatible, you can just update the package, no need to install two in parallel, right? A major update would break both the header compatibility and the corresponding `.so`... So you need to be able to say "I want the headers + lib of MAVSDK2", probably by having the whole project renamed to `MAVSDK2` (mavsdk2.pc, MAVSDK2Config.cmake, etc).

Or am I missing something?

julianoes:
I have no strong opinion. I'm just pointing out how it's done for packages in Debian/Ubuntu as far as I'm aware. There you can have two ABI versions of a library as executables might depend on either.

JonasVautherin:
Oh right, I think I get it. When we update `libmavsdk1` to `libmavsdk2`, then the header package will be for `libmavsdk2` and at build time everything has to be against v2.

But older packages that link v1 can still do it and don't have to be updated. That makes sense!

julianoes:
But we probably should still change the 0 to 1 for v1.4, and then rename it to 2 with v2.

Or we don't use that and remove the 0.

PR #2042, opened 2023-05-19
Add camera client sample code for camera server example

tbago:
* Add camera client sample for camera server example;
    * Remove global using namespace int camera_server.cpp;
    * Remove useless include header in camera_server.cpp;

### DIFF
--- a/examples/camera_server/camera_client.cpp
+++ b/examples/camera_server/camera_client.cpp
@@ -1,0 +1,60 @@
+#include <chrono>
+#include <future>
+#include <iostream>
+
+#include <mavsdk/mavsdk.h>
+#include <mavsdk/plugins/camera/camera.h>
+
+int main(int argc, const char* argv[])
+{
+    // we run client plugins to act as the GCS
+    // to communicate with the camera server plugins.
+    mavsdk::Mavsdk mavsdk;
+    mavsdk::Mavsdk::Configuration configuration(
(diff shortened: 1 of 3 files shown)


examples/camera_server/camera_client.cpp
```diff
@@ -0,0 +1,60 @@
...
+        std::cout << info << std::endl;
+    });
+
+    camera.take_photo_async([](mavsdk::Camera::Result result) {
+        std::cout << "take photo return : " << result << std::endl;
+    });
+
+    while (true) {
+        std::this_thread::sleep_for(std::chrono::seconds(1));
+    }
```
julianoes, line 58: We should probably just take one photo and then exit.

examples/camera_server/camera_server.cpp
```diff
@@ -23,7 +11,7 @@ int main(int argc, char** argv)
 
     // 14030 is the default camera port for PX4 SITL
     // FIXME: the camera should probably initiate the connection and not just listen
-    auto result = mavsdk.add_any_connection("udp://:14030");
+    auto result = mavsdk.add_any_connection("udp://127.0.0.1:14030");
```
julianoes, line 14: I think that's correct. Thanks! 

Can you please remove the comment above it though?

julianoes (changes requested):
Thanks for that! See 2 comments inline.

julianoes (approved):

JonasVautherin (approved):

PR #2049, opened 2023-05-23
Pos vel acc

alireza787b:
I have created acceleration feedforeward among velocity and position offboard mode. It works. I will create an example code that in python in future as well to add to the examples folder. untill then just wanted to add this here.
we need to build the mavsdk again as well to this changes take effect
the edoted branch for protos are available in this fokrder repo pull request

https://github.com/alireza787b/MAVSDK-Proto/tree/pos-vel-acc

sorry if I am not familliar withyour workflow. actually first git pull experience for me, anything else I should do?


julianoes:
Thanks. Now you need to rebase this on top of main because quite a few changes have gone in with the param refactor PR. And also you need to run the `tools/fix_style.sh .` script afterwards, otherwise you end up with a big diff.

PR #2052, opened 2023-05-24
Minor refactoring in ardupilot flight modes

JonasVautherin:
Actually I think it is worth making it a bit more consistent, even if that's not a bugfix.


julianoes (approved):
Nice!

Issue #2046, opened 2023-05-22
Unknown type: bool (param_value.cpp:317) error


tlapik123:
I'm getting the `Unknown type: bool (param_value.cpp:317)` when running the camera example.

I think, I've identified the issue. 

As stated here:
https://mavlink.io/en/services/camera_def.html#parameter-types
the bool type in the camera definition should be supported.

the code in `param_value.cpp`:
```cpp
if (strcmp(type_str.c_str(), "uint8") == 0) {
        _value = static_cast<uint8_t>(std::stoi(value_str));
    } else if (strcmp(type_str.c_str(), "int8") == 0) {
        _value = static_cast<int8_t>(std::stoi(value_str));
    } else if (strcmp(type_str.c_str(), "uint16") == 0) {
        _value = static_cast<uint16_t>(std::stoi(value_str));
    } else if (strcmp(type_str.c_str(), "int16") == 0) {
        _value = static_cast<int16_t>(std::stoi(value_str));
    } else if (strcmp(type_str.c_str(), "uint32") == 0) {
        _value = static_cast<uint32_t>(std::stoi(value_str));
    } else if (strcmp(type_str.c_str(), "int32") == 0) {
        _value = static_cast<int32_t>(std::stoi(value_str));
    } else if (strcmp(type_str.c_str(), "uint64") == 0) {
        _value = static_cast<uint64_t>(std::stoll(value_str));
    } else if (strcmp(type_str.c_str(), "int64") == 0) {
        _value = static_cast<int64_t>(std::stoll(value_str));
    } else if (strcmp(type_str.c_str(), "float") == 0) {
        _value = static_cast<float>(std::stof(value_str));
    } else if (strcmp(type_str.c_str(), "double") == 0) {
        _value = static_cast<double>(std::stod(value_str));
    } else {
        LogErr() << "Unknown type: " << type_str;
        return false;
    }
    return true;
```
It also throws the error `Unknown type attribute: bool (camera_definition.cpp:130)` Which I'm not sure if its a separate issue or fixed by modifying the above code. 

Just a sidenote - why is the `strcmp` used here when we have `std::string`? Why not use `==`?

julianoes:
Good points.

According to the MAVLink enum there is no bool (see https://mavlink.io/en/messages/common.html#MAV_PARAM_EXT_TYPE) even though the docs say that. I would probably suggest to just use `uint8_t`, and I'll try to change the docs because that's confusing.

[...]

PR #2056, opened 2023-05-26
The bool parameter now automatically creates on/off options.

tlapik123:
When a bool parameter is encountered the on/off options should be created automatically as per [documentation](https://mavlink.io/en/services/camera_def.html#parameter-definition).
This should solve the second part of issue #2046.

### DIFF
--- a/src/mavsdk/plugins/camera/camera_definition.cpp
+++ b/src/mavsdk/plugins/camera/camera_definition.cpp
@@ -199,6 +199,17 @@ bool CameraDefinition::parse_xml()
             continue;
         }
 
+        auto get_default_opt = [&](){
+            auto maybe_default = find_default(new_parameter->options, default_str);
+
+            if (!maybe_default.first) {
+                LogWarn() << "Default not found for " << param_name;
+                return std::optional<Option>{};
+            }
+
+            return std::optional{maybe_default.second};
+        };
+
         auto e_options = e_parameter->FirstChildElement("options");
         if (e_options) {
             auto maybe_options = parse_options(e_options, param_name, type_map);
@@ -207,14 +218,11 @@ bool CameraDefinition::parse_xml()
             }
             new_parameter->options = maybe_options.second;
             
-            auto maybe_default = find_default(new_parameter->options, default_str);
-            
-            if (!maybe_default.first) {
-                LogWarn() << "Default not found for " << param_name;
+            if (auto default_option = get_default_opt()) {
+                new_parameter->default_option = *default_option;
+            } else {
                 return false;
             }
-            
-            new_parameter->default_option = maybe_default.second;
         } else if (type_str == "bool") {
             // Automaticaly create bool options if the parameter type is bool as per documentation.
             Option true_option;
@@ -229,14 +237,11 @@ bool CameraDefinition::parse_xml()
                 std::make_shared<Option>(std::move(false_option))
             };
 
-            auto maybe_default = find_default(new_parameter->options, default_str);
-
-            if (!maybe_default.first) {
-                LogWarn() << "Default not found for " << param_name;
+            if (auto default_option = get_default_opt()) {
+                new_parameter->default_option = *default_option;
+            } else {
                 return false;
             }
-
-            new_parameter->default_option = maybe_default.second;
         } else {
             auto maybe_range_options = parse_range_options(e_parameter, param_name, type_map);
             if (!std::get<0>(maybe_range_options)) {
--- a/src/mavsdk/plugins/camera/camera_definition.cpp
+++ b/src/mavsdk/plugins/camera/camera_definition.cpp
@@ -116,12 +116,12 @@ bool CameraDefinition::parse_xml()
             return false;
         }
 
-        if (strcmp(type_str, "string") == 0) {
+        if (std::string(type_str) == "string") {
             LogDebug() << "Ignoring string params: " << param_name;
             continue;
         }
 
-        if (strcmp(type_str, "custom") == 0) {
+        if (std::string(type_str) == "custom") {
             LogDebug() << "Ignoring custom params: " << param_name;
             continue;
         }
@@ -135,7 +135,7 @@ bool CameraDefinition::parse_xml()
         new_parameter->is_control = true;
         const char* control_str = e_parameter->Attribute("control");
         if (control_str) {
-            if (strcmp(control_str, "0") == 0) {
+            if (std::string(control_str) == "0") {
                 new_parameter->is_control = false;
             }
         }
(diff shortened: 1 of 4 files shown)


src/mavsdk/plugins/camera/camera_definition.cpp
```diff
@@ -203,6 +206,28 @@ bool CameraDefinition::parse_xml()
...
             new_parameter->options = maybe_options.second;
+            
+            auto maybe_default = find_default(new_parameter->options, default_str);
+            
+            if (!maybe_default.first) {
+                LogWarn() << "Default not found for " << param_name;
+                return false;
+            }
+            
+            new_parameter->default_option = maybe_default.second;
```
JonasVautherin, line 217: Isn't that copy-pasted from below? Could we put that into a separate function?

JonasVautherin (commented):
Thanks for the contribution! :rocket:

src/mavsdk/plugins/camera/camera_definition.cpp
tlapik123: I actually had that done via lambda, but it didn't save much code. 

The lambda:
```cpp
auto get_default_opt = [&](){
          auto maybe_default = find_default(new_parameter->options, default_str);

          if (!maybe_default.first) {
              LogWarn() << "Default not found for " << param_name;
              return std::optional<Option>{};
          }

          return std::optional{maybe_default.second};
      };
```

Usages:
```cpp
            if (auto default_option = get_default_opt()) {
                new_parameter->default_option = default_option;
            } else {
                return false;
            }
```
The main abstraction "issue" is the `return false;` without it you wouldn't need to check the result of a function.

If you feel that that is a better way to abstract that I'll change it. :)

JonasVautherin: I think it's better! It makes an implementation detail of what is inside `get_default_opt()`, and it avoids some duplication (like the log message).

tlapik123: Okay! Thx for the feedback, I've used the lambda.

JonasVautherin (approved):
I fixed the style. Next time please run `./tools/fix_style.sh .` :blush:.

Thanks for the contribution! :rocket:

src/mavsdk/plugins/camera/camera_definition.cpp
```diff
@@ -110,18 +110,21 @@ bool CameraDefinition::parse_xml()
             return false;
         }
 
-        const char* type_str = e_parameter->Attribute("type");
-        if (!type_str) {
+        const char* type_str_res = e_parameter->Attribute("type");
+        // The check for null pointer is required before the creation of std::string.
```
julianoes, line 114:
```suggestion
```

src/mavsdk/plugins/camera/camera_definition.cpp
```diff
@@ -110,18 +110,21 @@ bool CameraDefinition::parse_xml()
...
+        if (!type_str_res) {
             LogErr() << "type attribute missing for " << param_name;
             return false;
         }
 
-        if (std::string(type_str) == "string") {
+        // Convert to std::string only once.
```
julianoes, line 120:
```suggestion
```

src/mavsdk/plugins/camera/camera_definition.cpp
```diff
@@ -196,6 +199,17 @@ bool CameraDefinition::parse_xml()
...
+        auto get_default_opt = [&]() {
+            auto maybe_default = find_default(new_parameter->options, default_str);
+
+            if (!maybe_default.first) {
+                LogWarn() << "Default not found for " << param_name;
+                return std::optional<Option>{};
+            }
+
+            return std::optional{maybe_default.second};
+        };
```
julianoes, line 211: Nice!

julianoes (commented):
Small comments, otherwise this looks good.

I assume you have tested it?

tlapik123:
Yes I have tested it with the camera/definition file I had on hand and it did not produce the errors when bool type parameter without options was present.

src/mavsdk/plugins/camera/camera_definition.cpp
tlapik123: I thought this comment would be good to have, since if someone would remove the `if` or somehow change it, the runtime exception from creating a std::string from NULL could happen.

src/mavsdk/plugins/camera/camera_definition.cpp
tlapik123: Thx! :) About the std::optional - I noticed that there are some usages of `std::pair<bool` which could be done (semantically) better with `std::optional`. I might give it a shot changing it if you think it's a good idea.

julianoes: Internally, `std::optional` is good. However, in the external (auto-generated) API it's not an option. I think there we only have `std::pair`.

src/mavsdk/plugins/camera/camera_definition.cpp
julianoes: The comment was good for me to review actually but I would expect someone to work on this code to see all the examples of checking for non-null and therefore not require it but it's fine, I'm happy to leave it.

src/mavsdk/plugins/camera/camera_definition.cpp
tlapik123: Got it, I rather won't touch it then 👍.

src/mavsdk/plugins/camera/camera_definition.cpp
tlapik123: Good point, I suppose no one would delete a null check just for fun. I'll remove it.

src/mavsdk/plugins/camera/camera_definition.cpp
tlapik123: Removed in [12a8070](https://github.com/mavlink/MAVSDK/pull/2056/commits/12a807014c365b61db423a6ff6ea94348167e060)

src/mavsdk/plugins/camera/camera_definition.cpp
tlapik123: Removed in [12a8070](https://github.com/mavlink/MAVSDK/pull/2056/commits/12a807014c365b61db423a6ff6ea94348167e060)

JonasVautherin:
@tlapik123 I rebased `main` but somehow now there is a conflict. Would you mind looking into that? :innocent:

julianoes (approved):

julianoes:
Sorry, somehow this got auto-closed because it was against my branch which I now merged.

Could you just open that PR (or a new one) again against main? Sorry about that!

PR #2061, opened 2023-05-29
Generate code from protos

JonasVautherin:
Not sure what happened, it seems like the code is out of sync in `main` :thinking:.


julianoes:
Probably my fault, sorry.

julianoes:
I think I forgot to update the submodule, I will update it now and the force push this.

PR #2062, opened 2023-05-30
CI: disable integration tests

julianoes:
Some of the tests are more often failing than passing and we stopped even checking the results. So we might as well disable the tests for now, so that it at least looks better in the README badge and PR checks.

Also, we save some of Azures CPU cycles just burns energy without any gain at this point.

I think in the future we can discuss bringing back some tests, or just have some examples executable as tests instead but heavily rely on the system-tests instead.

The example tests added back in would be to test a few cases that cover whether things are roughly still connecting and running, however, it won't be feasible to have integration tests for all sorts of combinations of PX4 and ArduPilot versions against all of the various features because we just run into combinatorial explosion.


JonasVautherin (approved):
Makes sense :+1:

PR #2060, opened 2023-05-29
Split Ftp plugin into Ftp and FtpServer

julianoes:
The FTP plugin is a weird combination of client and server in one. Now that we have Components additional to Systems we can split it into client and server.

This includes:
- Complete refactor (almost rewrite) of server and client
- Slight API changes
- Using std::filesystem as much as possible everywhere
- Adding system tests

Depended on https://github.com/mavlink/MAVSDK-Proto/pull/319.

PR #2074, opened 2023-06-09
UDP connection can disconnect when heartbeats are not received

jperez-droneup:
Creating a UDP connection to a vehicle/simulator via `setup_udp_remote` or `add_any_connection` will set the `always_connected` flag to true in all cases. This means that even though MAVSDK could be sending heartbeats (configured in the `Mavsdk::Configuration::_always_send_heartbeats`), the lack of response on those heartbeats does not lead to realizing/announcing the disconnect. With this change, if the Mavsdk will properly use heartbeats to detect connection and disconnection if it is configured to send those heartbeats. This means that the system will not be considered connected (and fire the `subscribe_on_new_system` callback) until the first heartbeat response is received, and will be considered disconnected (and fire the `subscribe_is_connected` callback) when the configured amount of time passes without a heartbeat response.

### DIFF
--- a/src/mavsdk/core/mavsdk_impl.cpp
+++ b/src/mavsdk/core/mavsdk_impl.cpp
@@ -533,7 +533,7 @@ ConnectionResult MavsdkImpl::setup_udp_remote(
         new_conn->add_remote(remote_ip, remote_port);
         add_connection(new_conn);
         std::lock_guard<std::recursive_mutex> lock(_systems_mutex);
-        make_system_with_component(0, 0, true);
+        make_system_with_component(0, 0, !_configuration.get_always_send_heartbeats());
     }
     return ret;
 }


JonasVautherin:
Hmm does that work? It seems weird to set `always_connected` to `!always_send_heartbeats`, doesn't it?

And when `always_connected` is set to false, is the system still sending heartbeats even if nobody is answering? I thought it was the point of that flag :thinking:

julianoes:
Thanks @jperez-droneup!

@JonasVautherin I wonder why we need that `_always_connected` flag at all? It looks like it was added in https://github.com/mavlink/MAVSDK/pull/774 presumably because of the problems trying to create an FTP client and server which wasn't really properly supported. It is now with System vs Component, and the FTP plugin is split in #2060.

I'm gonna try to make a PR where I just remove that flag alltogether and see where it leads.

julianoes:
This is my alternative fix: https://github.com/mavlink/MAVSDK/pull/2077

@jperez-droneup it would be great if you could test this one to check whether it works for you.

jperez-droneup:
> This is my alternative fix: #2077
> 
> @jperez-droneup it would be great if you could test this one to check whether it works for you.

Your fix does indeed fix the issue, so the connection status change callback is properly firing. However, I am still seeing the issue from case1 of https://github.com/mavlink/MAVSDK/issues/2075, just FYI.

JonasVautherin:
> However, I am still seeing the issue from case1 of https://github.com/mavlink/MAVSDK/issues/2075, just FYI.

Just to be sure I understand correctly. Case one of #2075 is not fixed by this PR #2074, right? In other words, does that mean that this (#2074) could be closed in favor of #2077, but issue #2075 is still open?

jperez-droneup:
> > However, I am still seeing the issue from case1 of #2075, just FYI.
> 
> Just to be sure I understand correctly. Case one of #2075 is not fixed by this PR (#2074), is it? In other words, does that mean that this (#2074) could be closed in favor of #2077, but issue #2075 is still open because of its case 1?

That is correct. I would rather you just remove the always_connected flag altogether than my approach to work around it. And yes #2075 is still present in case1.

julianoes:
#2075 is "user error" as I understand it, or at least I wouldn't know how to avoid it. So that's where I suggest #2078.

julianoes:
Thanks for the PR but I'm closing it because I prefer removing the `_always_connected` flag fully, which was done in #2077.

JonasVautherin:
Yeah thanks a lot @jperez-droneup! The whole discussion and your insights were very useful!

PR #2077, opened 2023-06-09
core: remove always connected flag

### LINKED ISSUE #2076: UDP to remote port does not recognize disconnect
Configuring the mavsdk to `always_send_heartbeats` will properly send heartbeats to the simulator/vehicle, however [this line](https://github.com/mavlink/MAVSDK/blob/ea0b720538cff8e51947557234cb9c4300edadc3/src/mavsdk/core/mavsdk_impl.cpp#L536) sets the system to always consider itself as connected. So even though the mavsdk is sending heartbeats, if enough are missed to normally signal a disconnect, that disconnect is never recognized and any `subscribe_is_connected` callbacks on the system are not fired. I raised https://github.com/mavlink/MAVSDK/pull/2074 to fix this issue.

julianoes:
This flag was added when we didn't have the notion of a component, and therefore it was required for the FTP server implementation. I believe we should be able to remove it now.

Fixes #2076.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -113,8 +113,6 @@ jobs:
         run: ./build/release/src/system_tests/system_tests_runner
       - name: test (mavsdk_server)
         run: ./build/release/src/mavsdk_server/test/unit_tests_mavsdk_server
-      - name: test FTP server
-        run: ./build/release/src/integration_tests/integration_tests_runner --gtest_filter="FtpTest.TestServer"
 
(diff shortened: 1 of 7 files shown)


JonasVautherin:
Oh, if it works like that, it's even nicer!

julianoes:
The FTP server test fails. That's expected. It'll be removed and replaced with a system test anyway.

JonasVautherin (approved):

PR #2079, opened 2023-06-10
core: add way to remove a connection

julianoes:
This adds the method add_any_connection_with_handle which returns a handle that can later be used to remove a connection again using remove_connection.

Untested, @jperez-droneup if you want to give this a try.


jperez-droneup:
This does indeed fix my issue with reconnecting on a new IP address, and avoids destructing the mavsdk altogether. Thanks for your fast responses! 5/5 stars

julianoes:
Thanks for testing @jperez-droneup!

JonasVautherin (approved):

PR #2081, opened 2023-06-13
core: don't create a null system if not required

### LINKED ISSUE #2080: Mavsdk::remove_connection does not remove system
I believe there are still some subscriptions on the MAVSDK side for the removed connection. When removing and re-adding a connection, I get multiple printouts of messages (see heartbeat, MAVLink: info:, and Received ack messages in attached log).

It looks like the connection is removed, but the system itself is not. So when we reconnect we get duplicate messages.

At this point, I am not subscribed to anything but `subscribe_is_connected` (which is unsubscribed when a disconnect happens) and `subscribe_on_new_system` (which is unsubscribed when the connection is made). Unsubscribing everything on my end does not fix the multiple messages. For example, the only time I see `_heartbeat_timeout_cookie` ("heartbeats timed out" message) unsubscribed is when the `SystemImpl` is destructed, but `_systems` in `MavskImpl` is only cleared on destruction and I don't see any way to remove the system after removing the connection.

I found that the system id of my first system is changed to 1 in [this line](https://github.com/mavlink/MAVSDK/blob/ed7ac0777f97a84531c1790cbbc2b0de0438225b/src/mavsdk/core/mavsdk_impl.cpp#L356). However, subsequent connections (after disconnect of initial one and removal of connection) create systems with ID zero and are not changed to 1 because the systems list is not empty. Seems like we need to properly remove the systems - and all internal subscriptions to them - in order for it to properly remove and add a new remote system and have it work correctly.

julianoes:
When we use setup_udp_remote twice, we end up creating a null system every time which didn't make sense.

Closes #2080

### DIFF
--- a/src/mavsdk/core/mavsdk_impl.cpp
+++ b/src/mavsdk/core/mavsdk_impl.cpp
@@ -535,7 +535,9 @@ std::pair<ConnectionResult, Mavsdk::ConnectionHandle> MavsdkImpl::setup_udp_remo
         new_conn->add_remote(remote_ip, remote_port);
         auto handle = add_connection(new_conn);
         std::lock_guard<std::recursive_mutex> lock(_systems_mutex);
-        make_system_with_component(0, 0);
+        if (_systems.empty()) {
+            make_system_with_component(0, 0);
+        }
         return {ret, handle};
     } else {
         return {ret, Mavsdk::ConnectionHandle{}};


jperez-droneup:
This looks good to me, on reconnecting I no longer see duplicate messages.

JonasVautherin (approved):

PR #2084, opened 2023-06-14
build: Install MAVLink headers within MAVSDK

julianoes:
This reverts a previous change which requires MAVLink to be installed standalone next to MAVSDK. Instead, the headers are, for now included as part of the MAVSDK install.

While this is not the right way to go about dependencies it avoids two problems:
1. It behaves the same as before. No MAVLink dependency is suddenly required disrupting an existing workflow.
2. It avoids a potential pitfall where a user of the library installs the MAVSDK library compiled with MAVLink from the common dialect but then later tries to use MAVLink headers containing their custom dialect. Intuitively, one would expect the message of the custom dialect to work via the mavlink_passthrough plugin, however, because the MAVlink parser has a static CRC_EXTRA table it won't parse the added custom messages unless the MAVSDK library is re-compiled with that custom dialect included.

### DIFF
--- a/src/mavsdk/core/CMakeLists.txt
+++ b/src/mavsdk/core/CMakeLists.txt
@@ -3,6 +3,12 @@ find_package(Threads REQUIRED)
 
 find_package(MAVLink REQUIRED)
 
+get_target_property(
+    MAVLINK_INCLUDE_DIR
+    MAVLink::mavlink
+    INTERFACE_INCLUDE_DIRECTORIES
+)
+
 configure_file(version.h.in version.h)
 configure_file(include/mavsdk/mavlink_include.h.in include/mavsdk/mavlink_include.h)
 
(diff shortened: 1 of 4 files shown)


src/mavsdk/core/CMakeLists.txt
```diff
@@ -102,6 +107,7 @@ target_include_directories(mavsdk
     PUBLIC
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include/mavsdk>
     $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/include/mavsdk>
+    $<BUILD_INTERFACE:${MAVLINK_INCLUDE_DIR}>
```
JonasVautherin, line 110: Is that required? It should come with MAVLink::mavlink, right?

julianoes: I just tried without and it seems to be required for other targets like integration_tests, etc.

JonasVautherin: :thinking:

julianoes: Because MAVLink is now a private dependency but the integration tests still require the headers.

PR #2087, opened 2023-06-15
Build with gcc13 without errors

tlapik123:
It looks like gcc13 libs don't include `<cstdint>` header in some of the other `<>` headers and that breaks the build or there is some other change with the libs. 
Either way adding `#include <cstdint>` to the affected files fixes the issue and allows for the build with gcc13.


julianoes (approved):
Thanks!

PR #2106, opened 2023-07-30
Flight info & requesting messages improvements

holden-zenith:
Ardupilot does not support MAVLINK_MSG_ID_FLIGHT_INFORMATION or MAV_CMD_REQUEST_FLIGHT_INFORMATION. Also there was some sort of deadlock occurring from the _system_impl->send_autopilot_version_request() in enable(), so just let system or the period caller request it slightly later.

### DIFF
--- a/src/mavsdk/plugins/info/info_impl.cpp
+++ b/src/mavsdk/plugins/info/info_impl.cpp
@@ -46,40 +46,33 @@ void InfoImpl::deinit()
 
 void InfoImpl::enable()
 {
-    // We can't rely on System to request the autopilot_version,
-    // so we do it here, anyway.
-    _system_impl->send_autopilot_version_request();
-
     // We're going to retry until we have the version.
-    _system_impl->add_call_every([this]() { request_version_again(); }, 1.0f, &_call_every_cookie);
+    _system_impl->add_call_every(
+        [this]() { request_autopilot_version(); }, 1.0f, &_autopilot_info_call_every_cookie);
 
     // We're going to periodically ask for the flight information
-    if (_system_impl->autopilot() != SystemImpl::Autopilot::ArduPilot) {
-        _system_impl->send_flight_information_request();
-
-        _system_impl->add_call_every(
-            [this]() { request_flight_information(); }, 1.0f, &_flight_info_call_every_cookie);
-    }
+    _system_impl->add_call_every(
+        [this]() { request_flight_information(); }, 1.0f, &_flight_info_call_every_cookie);
 }
 
 void InfoImpl::disable()
 {
-    _system_impl->remove_call_every(_call_every_cookie);
+    _system_impl->remove_call_every(_autopilot_info_call_every_cookie);
     _system_impl->remove_call_every(_flight_info_call_every_cookie);
 
     {
         std::lock_guard<std::mutex> lock(_mutex);
-        _information_received = false;
+        _autopilot_version_received = false;
         _flight_information_received = false;
     }
 }
 
-void InfoImpl::request_version_again()
+void InfoImpl::request_autopilot_version()
 {
     {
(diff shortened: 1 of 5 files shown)


src/mavsdk/plugins/info/info_impl.cpp
```diff
@@ -46,17 +48,16 @@ void InfoImpl::deinit()
 
 void InfoImpl::enable()
 {
-    // We can't rely on System to request the autopilot_version,
-    // so we do it here, anyway.
-    _system_impl->send_autopilot_version_request();
```
JonasVautherin, line 51: Shouldn't this one stay for PX4?

src/mavsdk/plugins/info/info_impl.cpp
```diff
@@ -46,17 +48,16 @@ void InfoImpl::deinit()
...
+        _system_impl->add_call_every(
+            [this]() { request_flight_information(); }, 1.0f, &_flight_info_call_every_cookie);
+    }
 
     // We're going to retry until we have the version.
     _system_impl->add_call_every([this]() { request_version_again(); }, 1.0f, &_call_every_cookie);
-
-    // We're going to periodically ask for the flight information
-    _system_impl->add_call_every(
-        [this]() { request_flight_information(); }, 1.0f, &_flight_info_call_every_cookie);
```
JonasVautherin, line 59: Shouldn't this one stay for PX4?

holden-zenith: This one wasn't removed, it was moved into the if statement

src/mavsdk/plugins/info/info_impl.cpp
holden-zenith: So the deadlock I mentioned didn't seem Ardupilot specific. Also, this message is called immediately before this is enabled [here](https://github.com/mavlink/MAVSDK/blob/7b30f7383934823ae564ed23788d0f2fc1dfa47c/src/mavsdk/core/system_impl.cpp#L582) anyways. This happens whenever info plugin is created too quickly after the system.

src/mavsdk/plugins/info/info_impl.cpp
JonasVautherin: Oh, right :man_facepalming:

ykhedar:
Hi, thanks for the PR. My thoughts on this:

I think even if we send the request to get the FLIGHT_INFORMATION, for ardupilot it simply sends unsupported which the client can simply ignore assuming it might be supported from ArduPilot in the future. There is an open issue on Ardupilot firmware repo: https://github.com/ArduPilot/ardupilot/issues/15217 which could potentially add the support. So it would be better to support the message on ardupilot side rather than remove it on MAVSDK if the message sending is not producing any negative side effects. 

Thanks,
khedar

holden-zenith:
In general I would agree, however at the moment the plugin isn't even usable as it stands currently in my testing. Also, only flight_uuid and time_boot_ms are collected from the information message. flight_uuid is a parameter for ardupilot and time_boot_ms is in many other messages, so if somebody needs flight_uuid then they can just grab the parameter. I'd be happy to re-add it personally if that PR actually does ever progress though.

src/mavsdk/plugins/info/info_impl.cpp
holden-zenith: Okay thank you, you just reminded exactly what the problem is. This is the only plugin you can't initialize in a callback.

holden-zenith:
I will be on vacation for the rest of the month so I committed where I last left off. I have made adjustments so that using the RequestMessage class will no longer repeatedly request messages that are denied, which means that you don't have to accommodate for specific flight stacks by avoiding sending requests for unsupported messages.

Issue #2108, opened 2023-08-09
FTP Server file descriptor leak


Zhumos:
**Version**: 1.4.16 as well as main

The FTP Server currently leaks a file descriptor every time MavlinkFtp::_work_list is called as it does not call closedir() on the descriptor used.

If there is a system limit for file descriptors, the server will eventually start failing to open any further descriptors. This was confirmed by watching "/proc/<pid>/fd" and observing errors when 1024 descriptors was reached.

This issue causes failures such as when requesting a directory listing, the client reports "Ftp::Result::ProtocolError" with the server reporting "OPC:CMD_OPEN_FILE_RO" followed by "FTP: Open failed"

Line where descriptor is allocated:
https://github.com/mavlink/MAVSDK/blob/fadfe22e7c053945107201e2638be9a7f8b8ee5c/src/mavsdk/core/mavlink_ftp.cpp#L989

julianoes:
Fixed, thanks @Zhumos!

PR #2109, opened 2023-08-09
Fix FTP file descriptor leak

Zhumos:
This PR addresses the issue as described in #2108


julianoes:
Thanks @Zhumos. I'm pretty sure this is already resolved in #2060. We might want to make a bugfix against the v1.4 branch though. Do you mind making a PR against v1.4 instead?

PR #2114, opened 2023-08-14
Fix battery voltage, included extended voltages

julianoes:
See commits.


JonasVautherin (approved):
Fine for me, but... what is the problem with the 255 batteries?

julianoes:
I don't know where you got the 255 for. The array we iterate over is only 10 entries long.

JonasVautherin:
I don't know, I just see now that it is there in the code :rofl:. Good for me then!

Issue #2110, opened 2023-08-09
ArduPilot - Problem identified that may be causing the mission module to fail (fly_mission example)


trolledbypro:
Hello, I have found a possible cause of this issue of the mission module failing in ArduPilot. 
Using Wireshark, I analyzed the network traffic between my ArduPilot SITL (running inside a WSL2 Docker container that runs the latest ArduPilot Docker image) and the MAVSDK example (running inside WSL2 as I was having issues running it in Windows 10). My MAVSDK library and my fly_mission binary were built in _debug_ mode using CMake through CMake Tools in VSCode (using VSCode WSL Remote session).

Here is my terminal output:  
![image](https://github.com/mavlink/MAVSDK/assets/57928210/31595d2e-923d-4817-aeea-85eb264c2683)

There is a debug message saying the gimbal version is v1, and that the mission is unsupported.
Examining Wireshark, the last message sent before the MISSION_ACK message containing the MAV_MISSION_UNSUPPORTED payload was MAV_CMD_DO_MOUNT_CONFIGURE (204). 

When gimbal angles are defined in a mission_item, the gimbal version is verified: https://github.com/mavlink/MAVSDK/blob/fadfe22e7c053945107201e2638be9a7f8b8ee5c/src/mavsdk/plugins/mission/mission_impl.cpp#L460
We then use the version above and send a version specific message containing the gimbal configuration. Below, we see that the unsupported message is sent if the gimbal version is assumed to be v1, as is the case in my console output:
https://github.com/mavlink/MAVSDK/blob/fadfe22e7c053945107201e2638be9a7f8b8ee5c/src/mavsdk/plugins/mission/mission_impl.cpp#L1188-L1221

It appears that MAVSDK is not finding out the supported gimbal version. Wireshark shows that no GIMBAL_MANAGER_INFORMATION (280) messages were sent from the AP and no MAV_CMD_REQUEST_MESSAGE (512) message from MAVSDK requesting the gimbal version was sent.

JonasVautherin:
Thanks @trolledbypro for looking into this! :rocket: 

I think I would be in favour of a third version: drop gimbal v1 entirely and always assume v2. There are many problems with the detection anyway.

[...]

PR #2118, opened 2023-08-18
Mission module: Updated camera action commands (param4) to be compliant with mavlink standard

trolledbypro:
Hello,
As I continue to incorporate ArduPilot compatibility with MAVSDK Mission module (https://github.com/mavlink/MAVSDK/issues/2110), I noticed some discrepancies with how the camera action commands are formatted vs. the Mavlink standard for the commands. ArduPilot was rejecting the commands due to invalid parameters:

MAV_CMD_IMAGE_START_CAPTURE:
param 4 was set to NaN, [must be set to 0 instead](https://mavlink.io/en/messages/common.html#MAV_CMD_IMAGE_START_CAPTURE)

MAV_CMD_VIDEO_START_CAPTURE
param 2 was set to NaN, [must be set to 0 instead](https://mavlink.io/en/messages/common.html#MAV_CMD_VIDEO_START_CAPTURE)

In terms of testing, the only testing I have done in ArduPilot SITL is to see if the commands are accepted. These changes make both of these commands accepted. I have not tested in PX4 nor with real hardware.

### DIFF
--- a/src/mavsdk/plugins/mission/mission_impl.cpp
+++ b/src/mavsdk/plugins/mission/mission_impl.cpp
@@ -526,18 +526,21 @@ MissionImpl::convert_to_int_items(const std::vector<MissionItem>& mission_items)
             float param1 = NAN;
             float param2 = NAN;
             float param3 = NAN;
+            float param4 = NAN;
             switch (item.camera_action) {
                 case CameraAction::TakePhoto:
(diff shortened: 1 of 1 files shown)


julianoes (approved):
Thanks!

JonasVautherin (commented):
Amazing, thanks for that!

Issue #2102, opened 2023-07-19
Module range mismatch in "ActionImpl::set_actuator_async" function


elbarto1980:
Hi,
I think I might have just found a minor bug in the ["set_actuator_async"](https://github.com/mavlink/MAVSDK/blob/fadfe22e7c053945107201e2638be9a7f8b8ee5c/src/mavsdk/plugins/action/action_impl.cpp#L516) function.
The range in the switch...case statement goes from 1 to 6 while the range of the module function (index % 6) goes from 0 to 5.
So, param6 will never be set and if (index==6), then param7 will be equal to 1 and the whole command will be ignored by PX4 [handle_message_command_both](https://github.com/PX4/PX4-Autopilot/blob/1c8ab2a0d7db2d14a6f320ebd8766b5ffaea28fa/src/modules/mavlink/mavlink_receiver.cpp#L557) function.

If I understood correctly, at the moment PX4 only pays attention to 3 AUX outputs, namely param1, param2 and param3... so this bug should not have any effect on the whole system behavior.

julianoes:
Hi, thanks for the issue. That looks indeed a bit odd.
I can see the discussion back here: https://github.com/mavlink/MAVSDK-Proto/pull/217/files#r596590938 

What would be your suggestion how to fix it?

elbarto1980:
I think that something like this (I have not tested it yet) could solve the issue. It uses a secondary zero-based numbering.
Do you think this could work?

```
void ActionImpl::set_actuator_async(const int index, const float value,
                                    const Action::ResultCallback& callback) {
  MavlinkCommandSender::CommandLong command{};

  auto zbIndex = index - 1;  // zero-based index
  if (zbIndex >= 0) {
    command.command = MAV_CMD_DO_SET_ACTUATOR;
    command.target_component_id = _system_impl->get_autopilot_id();

    switch (zbIndex % 6) {
      case 0:
        command.params.maybe_param1 = value;
        break;
      case 1:
        command.params.maybe_param2 = value;
        break;
      case 2:
        command.params.maybe_param3 = value;
        break;
      case 3:
        command.params.maybe_param4 = value;
        break;
      case 4:
        command.params.maybe_param5 = value;
        break;
      case 5:
        command.params.maybe_param6 = value;
        break;
    }
    command.params.maybe_param7 = static_cast<float>(zbIndex) / 6.0f;

    _system_impl->send_command_async(
[...]

PR #2122, opened 2023-08-27
Action plugin 'set_actuator_async' switch statement index changed to zero-based

elbarto1980:
The pull request is related to this [issue](https://github.com/mavlink/MAVSDK/issues/2102)

### DIFF
--- a/src/mavsdk/plugins/action/action_impl.cpp
+++ b/src/mavsdk/plugins/action/action_impl.cpp
@@ -524,28 +524,31 @@ void ActionImpl::set_actuator_async(
         command.params.maybe_param1 = static_cast<float>(index);
         command.params.maybe_param2 = value;
     } else {
-        command.command = MAV_CMD_DO_SET_ACTUATOR;
-        switch (index % 6) {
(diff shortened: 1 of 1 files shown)


julianoes (approved):
Thanks!

Should we return an error with something like invalid argument if someone supplies an index of 0?

elbarto1980:
Actually I thought about returning an error at first, but then I saw that neither the original method did any check on the parameters range (index is a const int). So I kept it that way.

I don't have full knowledge of how errors are handled in the async commands but, looking at the source code of other methods of the same class (e.g. `goto_location_async`), I presume that, in the else branch, I should use the method:
`command_result_callback(Action::Result::CommandDenied, callback);`
and then return 

Is it right?

julianoes:
Yes, but `CommandDenied` is only for the case where the vehicle denies it, and it looks like we don't have an `InvalidArgument` for action. Let's merge it like this and add that later.

Thanks again.

elbarto1980:
I’m happy to help. Thank you!

PR #2123, opened 2023-08-27
telemetry: Be compatible for battery_status

julianoes:
It turns out that PX4 v1.13 and likely PX4 v1.14 send out UINT16_MAX instead of 0 in the extensions. We can at least catch this and prevent a wrong battery voltage from slipping through.

Follow-up to #2114.


JonasVautherin (approved):

julianoes:
Oh, you're back :wave:

PR #2125, opened 2023-08-29
Do not set GCS mode as default

ThomasDebrunner:
Setting the GCS mode as default has hidden side-effects. Most notably that the connection is considered a valid datalink, and thus affects datalink failsafes in Autopilots. 

When used on-board, this is an unwanted side-effect. While you can configure this, the default is set to the "less-safe" option, which could leave a careless user in the situation where datalink failsafes get disabled unintentionally.

### DIFF
--- a/src/mavsdk/core/mavsdk_impl.h
+++ b/src/mavsdk/core/mavsdk_impl.h
@@ -141,7 +141,7 @@ private:
 
     Time _time{};
 
-    Mavsdk::Configuration _configuration{Mavsdk::Configuration::UsageType::GroundStation};
+    Mavsdk::Configuration _configuration{Mavsdk::Configuration::UsageType::Custom};
 
     struct UserCallback {
         UserCallback() = default;


JonasVautherin:
I am probably a bit biased because I always saw MAVSDK _primarily_ as an API for ground stations, which can also be used on the other side (be it a companion or an autopilot). To me it feels like whoever uses MAVSDK for something other than a ground station should know that in the MAVLink world, they should care about their sysid/compid. MAVSDK provides a way to set them, and it changes the type accordingly (you can say that you want to be an autopilot or a companion computer, for instance).

In any case, it feels weird to set the sysid/compid to values reserved for a ground station (which is the default), and at the same time to not set the type to "ground station" :thinking:. I agree that the specs are not completely clear (some sysid/compids seem to be attributed to some types, which probably should never have been the case), but that is what we have.

On top of that, it is a breaking change for all the current users of MAVSDK as a ground station... Maybe that could go in MAVSDK v2?

@julianoes what do you think?

julianoes:
I can see the drawbacks of having a default that causes surprises, and I would always try to err on the side of safety/caution.

However, I'm not convinced changing it to `Custom` as custom implies something is customized, and not default/standard. My plan was to overhaul the configuration and component type stuff to sync the two concepts and share the enums, etc. It seems like it would also be a good idea to make the selection required in the constructor, to avoid defaults that don't work. This would be a change we could do with v2.

julianoes:
@ThomasDebrunner what about something like this?

```
Mavsdk mavsdk{Configuration{Camera}};
CameraServer camera_server{mavsdk.default_component()}; // using MAV_COMP_ID_CAMERA1
WinchServer winch_server{mavsdk.component_by_id(MAV_COMP_ID_WINCH)}; // or whatever, if additional componens within a mavsdk server are required.
```

julianoes:
Done in #2169.

PR #2126, opened 2023-09-01
sample coro

jperez-droneup:


julianoes (commented):
Interesting! It would be nice indeed to have most (if not all) work on one thread and get rid of all the locking. MAVSDK isn't really doing any compute but just IO. And coroutines could be one way to do that.

By the way, I assume you're aware: the API files like action.h is generated, so eventually changes would have to be in the template. And some of the promise/future stuff is right now in the impl, so you could move your implementation there:
https://github.com/mavlink/MAVSDK/blob/main/src/mavsdk/plugins/action/action_impl.cpp#L55-L63

One more comment: we need to check when/if we can switch to C++20 because we have to compile for all sorts of architectures (e.g. cross compilation for Android, RPi, etc.), so we need to check what compilers are available there.

jperez-droneup:
I wasn't sure if this would be visible, how embarrassing haha. Yeah I am experimenting to see if there was an easy way to adapt at least the `_aync` functions to work with coroutines. We managed to get it working on our end with `boost::asio`, so it should be possible to port over eventually. Though of course the real goal would be to drop the threading done in MAVSDK and to use coroutines instead. This was just an attempt to shim it to allow clients to call in with coroutines.

julianoes:
I'll close this PR but let me know when you find more or have suggestions.

PR #2132, opened 2023-09-08
Fix calibration in telemetry health

JonasVautherin:
The calibration check should be stopped when the drone is actually found to be running in hitl mode, and not when the hitl param is received (because the hitl param may say that the drone is not running hitl, in which case we need to continue checking the calibration).

Without this fix, the health booleans related to the calibration tend to be "stuck" on the default `false` value (because MAVSDK stops checking and never gets to the point where it realizes that the check is actually fine).


julianoes (approved):
Right, I think that makes more sense.

PR #2133, opened 2023-09-08
[v1.4] Fix conflict in camera subscriptions

JonasVautherin:
The fact that `_status.call_every_cookie` is shared between two subscriptions (camera status and camera information) makes them conflict: when unsubscribing from one, it cancels the call_every for both.

Also, depending on the order, it will `add_call_every` twice with the same cookie, which I believe ends up leaking one.

This PR makes it simpler by requesting the status updates in the plugin directly. I don't think it is worth trying to optimize the number of requests here (and I am pretty sure it ended up doubling them in SiteScan).

### DIFF
--- a/src/mavsdk/plugins/camera/camera_impl.cpp
+++ b/src/mavsdk/plugins/camera/camera_impl.cpp
@@ -200,6 +200,8 @@ void CameraImpl::manual_enable()
     _parent->add_call_every(
         [this]() { request_camera_information(); }, 10.0, &_camera_information_call_every_cookie);
 
+    _parent->add_call_every([this]() { request_status(); }, 5.0, &_status.call_every_cookie);
+
     // for backwards compatibility with Yuneec drones
     if (_parent->has_autopilot()) {
         request_flight_information();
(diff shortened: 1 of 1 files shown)


src/mavsdk/plugins/camera/camera_impl.cpp
```diff
@@ -222,6 +225,7 @@ void CameraImpl::manual_disable()
 {
     invalidate_params();
     _parent->remove_call_every(_camera_information_call_every_cookie);
+    // TODO: should this remove_call_every for _status.call_every_cookie?
```
JonasVautherin, line 228: @julianoes: not sure about what `manual_disable()` does... should it `remove_call_every(_status.call_every_cookie)` here?

julianoes: It's used when you select another camera. Not sure, I'd probably add it for consistency.

julianoes: Up to you, otherwise this is fine.

JonasVautherin: Oh right, it should otherwise it will leak a callback when `manual_enable()` is called again :thinking:. I'll change that!

julianoes (approved):

PR #2136, opened 2023-09-09
Ardupilot fix goto location

### LINKED ISSUE #2129: When running MAVSDK against ArduCopter SITL the Action/goto_location function always responds with Command Denied
I'm running a software based on latest MAVSDK code i.e. I built library on Windows 64-bit from Main branch on GitHub. I'm using two different ArduCopter SITL, one running on Linux and one directly on a Cube Orange autopilot. So far I've tested take-off, landing, MissionRaw and Offboard mode using a gamepad. All working pretty good after some tweeks. The problem is the goto_location and goto_location_async functions which always return Command Denied no matter what I do. The command is issued when ArduCopter is in "Hold" mode in the air i.e. UAV keeping its current position. Do you have any suggestion why these MAVSDK functions might not work together with ArduCopter?

JonasVautherin:
Applying the patch from https://github.com/mavlink/MAVSDK/issues/2129 (@paulholmstedt).

Resolves #2129

### DIFF
--- a/src/mavsdk/plugins/action/action_impl.cpp
+++ b/src/mavsdk/plugins/action/action_impl.cpp
@@ -450,6 +450,10 @@ void ActionImpl::goto_location_async(
 
             command.command = MAV_CMD_DO_REPOSITION;
             command.target_component_id = _system_impl->get_autopilot_id();
+			// When ArduPilot we need to set frame to Global i.e. positive altitude over mean sea level (MSL)
+			if (_system_impl->autopilot() != SystemImpl::Autopilot::Px4) {
+				command.frame = MAV_FRAME_GLOBAL;
+			}
             command.params.maybe_param4 = static_cast<float>(to_rad_from_deg(yaw_deg));
             command.params.x = int32_t(std::round(latitude_deg * 1e7));
             command.params.y = int32_t(std::round(longitude_deg * 1e7));
@@ -460,21 +464,25 @@ void ActionImpl::goto_location_async(
                     command_result_callback(result, callback);
                 });
         };
-
-    // Change to Hold mode first
-    if (_system_impl->get_flight_mode() != FlightMode::Hold) {
-        _system_impl->set_flight_mode_async(
-            FlightMode::Hold,
-            [this, callback, send_do_reposition](MavlinkCommandSender::Result result, float) {
-                Action::Result action_result = action_result_from_command_result(result);
-                if (action_result != Action::Result::Success) {
-                    command_result_callback(result, callback);
-                    return;
-                }
-                send_do_reposition();
-            });
-        return;
-    }
+	// If PX4 change to Hold mode first
+	FlightMode gotoFlightMode = FlightMode::Hold;
+	// If ArduPilot change to Offboard mode first
+	if (_system_impl->autopilot() != SystemImpl::Autopilot::Px4) {
+		gotoFlightMode = FlightMode::Offboard;
+	}
+	if (_system_impl->get_flight_mode() != gotoFlightMode) {
+		_system_impl->set_flight_mode_async(
+			gotoFlightMode,
+			[this, callback, send_do_reposition](MavlinkCommandSender::Result result, float) {
+				Action::Result action_result = action_result_from_command_result(result);
+				if (action_result != Action::Result::Success) {
+					command_result_callback(result, callback);
+					return;
+				}
+				send_do_reposition();
+			});
+		return;
+	}  
 
     send_do_reposition();
 }
--- a/src/mavsdk/plugins/action/action_impl.cpp
+++ b/src/mavsdk/plugins/action/action_impl.cpp
@@ -450,10 +450,11 @@ void ActionImpl::goto_location_async(
 
             command.command = MAV_CMD_DO_REPOSITION;
             command.target_component_id = _system_impl->get_autopilot_id();
-			// When ArduPilot we need to set frame to Global i.e. positive altitude over mean sea level (MSL)
-			if (_system_impl->autopilot() != SystemImpl::Autopilot::Px4) {
-				command.frame = MAV_FRAME_GLOBAL;
-			}
+            // When ArduPilot we need to set frame to Global i.e. positive altitude over mean sea
+            // level (MSL)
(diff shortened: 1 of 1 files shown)


src/mavsdk/plugins/action/action_impl.cpp
```diff
@@ -450,6 +450,11 @@ void ActionImpl::goto_location_async(
 
             command.command = MAV_CMD_DO_REPOSITION;
             command.target_component_id = _system_impl->get_autopilot_id();
+            // When ArduPilot we need to set frame to Global i.e. positive altitude over mean sea
+            // level (MSL)
+            if (_system_impl->autopilot() != SystemImpl::Autopilot::Px4) {
```
JonasVautherin, line 455: Does it make a difference for PX4? I'm thinking that maybe we could just drop that `if` and set `MAV_FRAME_GLOBAL` for both PX4 and Ardupilot?

Probably a question for @julianoes :innocent:

src/mavsdk/plugins/action/action_impl.cpp
```diff
@@ -460,11 +465,15 @@ void ActionImpl::goto_location_async(
...
         };
-
-    // Change to Hold mode first
-    if (_system_impl->get_flight_mode() != FlightMode::Hold) {
+    // If PX4 change to Hold mode first
+    FlightMode gotoFlightMode = FlightMode::Hold;
+    // If ArduPilot change to Offboard mode first
+    if (_system_impl->autopilot() != SystemImpl::Autopilot::Px4) {
+        gotoFlightMode = FlightMode::Offboard;
+    }
```
JonasVautherin, line 473: Hmm I think it would read better like this:

```suggestion
    FlightMode goto_flight_mode;
    if (_system_impl->autopilot() == SystemImpl::Autopilot::Px4) {
        goto_flight_mode = FlightMode::Hold;
    } else {
        gotoFlightMode = FlightMode::Offboard;
    }
```

@paulholmstedt: What do you think? I won't ask you to resend a patch with those changes, I can do it myself if you agree with the suggestion :blush:.

JonasVautherin (commented):
Thanks a lot @paulholmstedt for the patch! Note that it is a little bit unorthodox, and for a more complex PR it would be nice if you opened it from a fork of yours, but in this case I think it's easier.

:rocket:

src/mavsdk/plugins/action/action_impl.cpp
julianoes: We are sending a COMMAND_INT, so I would argue it should be `MAV_FRAME_GLOBAL_INT`. Objections?

And from what I can see PX4 ignores the frame, so we might as well set it as it's the right thing to do.

src/mavsdk/plugins/action/action_impl.cpp
julianoes: Agreed.

julianoes: Small variable case fixup:

```suggestion
    FlightMode goto_flight_mode;
    if (_system_impl->autopilot() == SystemImpl::Autopilot::Px4) {
        goto_flight_mode = FlightMode::Hold;
    } else {
        goto_flight_mode = FlightMode::Offboard;
    }
```

julianoes (commented):
:ship:

src/mavsdk/plugins/action/action_impl.cpp
JonasVautherin: @paulholmstedt: Can you confirm that Ardupilot is fine with `MAV_FRAME_GLOBAL_INT`? :innocent:

src/mavsdk/plugins/action/action_impl.cpp
```diff
@@ -460,11 +465,15 @@ void ActionImpl::goto_location_async(
...
+    FlightMode goto_flight_mode;
+    if (_system_impl->autopilot() == SystemImpl::Autopilot::Px4) {
+        goto_flight_mode = FlightMode::Hold;
+    } else {
+        goto_flight_mode = FlightMode::Offboard;
+    }
+    if (_system_impl->get_flight_mode() != gotoFlightMode) {
         _system_impl->set_flight_mode_async(
-            FlightMode::Hold,
+            gotoFlightMode,
```
JonasVautherin, line 476:
```suggestion
            goto_flight_mode,
```

julianoes (dismissed):

paulholmstedt:
I tested against my ArduCopter SITL using `MAV_FRAME_GLOBAL_INT` (which sounds like the correct Frame based on Lon/Lat scaling) and it works the same as when `MAV_FRAME_GLOBAL` is set. So MAV_FRAME_GLOBAL_INT sounds like the right way to go. @julianoes as you pointed out PX4 doesn't care about frame. You might say it's a bug in PX4, or? I used goto_locatio_async with PX4 and MAVSDK for a long time without issues.

src/mavsdk/plugins/action/action_impl.cpp
```diff
@@ -450,6 +450,11 @@ void ActionImpl::goto_location_async(
 
             command.command = MAV_CMD_DO_REPOSITION;
             command.target_component_id = _system_impl->get_autopilot_id();
+            // When ArduPilot we need to set frame to Global i.e. positive altitude over mean sea
+            // level (MSL)
+            if (_system_impl->autopilot() != SystemImpl::Autopilot::Px4) {
+                command.frame = MAV_FRAME_GLOBAL;
+            }
```
JonasVautherin, line 457:
```suggestion
            command.frame = MAV_FRAME_GLOBAL_INT;
```

JonasVautherin:
I updated the code to use MAV_FRAME_GLOBAL_INT. @paulholmstedt any chance you could pull this branch and test it to confirm that it works?

After that we can squash and merge, as far as I am concerned :+1:.

paulholmstedt:
@JonasVautherin I pulled the `ardupilot-fix-goto-location` branch from GitHub, rebuild all and tested our software against the ArduPilot SITL and the goto_location_async works as expected with `MAV_FRAME_GLOBAL_INT` set as Frame.
Thanks for the excellent work you guys do on MAVSDK!

JonasVautherin:
Thanks a lot!

@julianoes ok to squash and merge for you?

PR #2140, opened 2023-09-13
telemetry: do not interrupt calibration checks too early

JonasVautherin:
Bringing https://github.com/mavlink/MAVSDK/pull/2132 upstream.


julianoes (approved):
Thanks!

PR #2148, opened 2023-10-06
docker/CI: remove Fedora images

julianoes:
It doesn't look like anyone is using them given the images are outdated. Also, I'm not using Fedora myself these days, so I'm not testing and maintaining it. Therefore, I'm removing it until someone adds it again and commits to maintaining it.


JonasVautherin:
Makes sense :+1:

PR #2150, opened 2023-10-10
v1.4 fix storage type translation (STORAGE_TYPE_OTHER)

JonasVautherin:
We should not use `static_cast` to translate between a MAVLink enum and a MAVSDK enum, because their values don't necessarily correspond. Here STORAGE_TYPE_OTHER has a MAVLink value of 254 and fails to translate.

```c++
_status.data.storage_type = static_cast<Camera::Status::StorageType>(storage_information.type);
```

### DIFF
--- a/src/mavsdk/plugins/camera/camera_impl.cpp
+++ b/src/mavsdk/plugins/camera/camera_impl.cpp
@@ -917,13 +917,35 @@ void CameraImpl::process_storage_information(const mavlink_message_t& message)
         _status.data.used_storage_mib = storage_information.used_capacity;
         _status.data.total_storage_mib = storage_information.total_capacity;
         _status.data.storage_id = storage_information.storage_id;
-        _status.data.storage_type =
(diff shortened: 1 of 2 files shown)


src/mavsdk/plugins/camera/camera_impl.cpp
```diff
@@ -894,37 +894,58 @@ void CameraImpl::process_storage_information(const mavlink_message_t& message)
...
-                break;
-        }
-
+        _status.data.storage_status = storage_status_from_mavlink(storage_information.status);
         _status.data.available_storage_mib = storage_information.available_capacity;
         _status.data.used_storage_mib = storage_information.used_capacity;
         _status.data.total_storage_mib = storage_information.total_capacity;
         _status.data.storage_id = storage_information.storage_id;
-        _status.data.storage_type =
-            static_cast<Camera::Status::StorageType>(storage_information.type);
```
JonasVautherin, line 921: This is the bug.

julianoes (approved):

Issue #1499, opened 2021-08-03
Sequence number wrong, per message instead of per channel


julianoes:
It looks like the sequence is defined per message (or maybe per plugin?) instead of per mavlink component/instance.

```
seq:   0, msgid:    4
seq:   0, msgid:    0
seq:   0, msgid:   76
seq:   0, msgid:   20
seq:   1, msgid:   20
seq:   2, msgid:   20
seq:   3, msgid:   20
seq:   1, msgid:    4
seq:   1, msgid:    0
seq:   2, msgid:    0
seq:   2, msgid:    4
seq:   3, msgid:    0
seq:   3, msgid:    4
seq:   4, msgid:    4
seq:   4, msgid:    0
seq:   5, msgid:    4
seq:   5, msgid:    0
seq:   6, msgid:    4
seq:   6, msgid:    0
seq:   7, msgid:    4
seq:   7, msgid:    0
seq:   8, msgid:    4
seq:   8, msgid:    0
seq:   9, msgid:    4
seq:   9, msgid:    0
seq:  10, msgid:    4
seq:  11, msgid:    4
seq:  10, msgid:    0
seq:  12, msgid:    4
seq:   0, msgid:    4
seq:   0, msgid:    0
seq:   0, msgid:   20
seq:   0, msgid:   76
seq:   1, msgid:   20
seq:   2, msgid:   20
seq:   3, msgid:   20
```

Maybe related to #1467.

PR #2155, opened 2023-10-16
Threading fixes, MAVLink sequence number cleanup

julianoes:
These fixes quite a few potential issues pointed out by running the system tests with thread sanitizer.

This is also a first step towards #1499.


JonasVautherin (commented):
So now `queue_message` takes a lambda that will generate the message right when it is about to be sent?

julianoes:
@JonasVautherin yes, and most importantly there is a lock around it.

JonasVautherin (dismissed):

julianoes:
I should probably review this when I have fresh eyes. In the meantime I'll base the #2060 PR on top of it. I'm hoping it'll help with random segfaults in CI.

PR #2157, opened 2023-10-17
Preserve symlinks in zip file

tthbalazs:
Hi!

When trying to build an app for macOS, Xcode complains about not being able to follow a symlink. After some investigation it turns out that the `Current` folder (`/mavsdk_server.xcframework/macos-x86_64/mavsdk_server.framework/Versions/Current`) is in fact not a symlink, but it should be.

Updating the `zip` command to preserve symlinks should fix the issue.


julianoes (approved):
Thanks for the fix! I'll also let @JonasVautherin review it.

JonasVautherin (approved):
Thanks a lot!

PR #2159, opened 2023-10-19
mavsdk_server: default to the first autopilot 

JonasVautherin:
Currently, `mavsdk_server` defaults to being a ground station (i.e. it does not expose the `Configuration`). When started, `mavsdk_server` connects to the first MAVLink `System` it detects.

The problem I have seen is that in some setups, there are sometimes MAVLink systems that are not an autopilot but that are advertising on the link that `mavsdk_server` is listening to. I had the issue with a Herelink radio, that advertises radio messages with sysid 42. For an app running on Herelink, if `mavsdk_server` picks up the system "42" first, then there is nothing to do other than restarting it and hoping that it picks up the right system.

Now I argue that the "right" system for `mavsdk_server`, given the current situation where it defaults to being a ground station, is a system that has an autopilot (i.e. a drone).

This PR makes `mavsdk_server` connect not to the first system it encounters, but to the first _drone_ (a.k.a. "system with an autopilot") it encounters.

### DIFF
--- a/src/mavsdk/core/lazy_plugin.h
+++ b/src/mavsdk/core/lazy_plugin.h
@@ -4,6 +4,7 @@
 #include <mutex>
 
 #include <mavsdk.h>
+#include <log.h>
 
 namespace mavsdk::mavsdk_server {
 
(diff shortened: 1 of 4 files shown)


julianoes (approved):
Nice!

julianoes (approved):

JonasVautherin:
Tested with SiteScan and the Astro :+1:

PR #2160, opened 2023-10-19
Split mission transfer into client and server

julianoes:
This was weirdly mixed up in the same class before and required another thread in the server plugin due to locking problems.

Here I go again :smile: 

Currently on top of #2060.


JonasVautherin (approved):
A long one again :see_no_evil: :laughing:.

If mission transfer still works, that's fine for me!

PR #2169, opened 2023-10-30
Consolidate components

julianoes:
This should make dealing with the configuration and components a bit clearer.

This is the last step towards v2 :crossed_fingers:.

Closes #2125.

### DIFF
--- a/examples/autopilot_server/autopilot_server.cpp
+++ b/examples/autopilot_server/autopilot_server.cpp
@@ -59,8 +59,7 @@ int main(int argc, char** argv)
     // thread as a ground station.
     std::thread autopilotThread([]() {
         mavsdk::Mavsdk mavsdkTester;
-        mavsdk::Mavsdk::Configuration configuration(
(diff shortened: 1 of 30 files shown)


julianoes:
@JonasVautherin and @ThomasDebrunner I'm thinking of changing the API to require a default, so from:

```
Mavsdk mavsdk;
```

to:

```
Mavsdk mavsdk{Mavsdk::ComponentType::GroundStation};
```

What do you think? And should it be a breaking change or deprecation notice if you use it without constructor?

Related to #2125.

JonasVautherin:
> And should it be a breaking change or deprecation notice if you use it without constructor?

For me it is fine to completely remove the default and make the build fail if no value is passed. It's fine because it will be a major update, and it's very easy to fix ("oh, it now requires an argument"). Also it's good to make it explicit since apparently some people got confused by the fact that MAVSDK may behave as a ground station by default.

JonasVautherin (approved):
Nice! :rocket:

PR #2176, opened 2023-11-13
Disable -ffast-math for iOS builds

ThomasDebrunner:
This PR disables the `-ffast-math` option for iOS builds. 

We found that for all MAVSDK 1.4.7+, this option has been breaking `std::isfinite` in the code.

This breaks the mission upload logic, which checks with `std::isfinite` to decide wheter speed or gimbal attitude has changed. 

We do not enable `-ffast-math` on any other platform. 

Specifically this line breaks: https://github.com/mavlink/MAVSDK/blob/aa7087dff395334ff86403236866cdfa3af265f5/src/mavsdk/plugins/mission/mission_impl.cpp#L436
With `-ffast-math` enabled on iOS, this check will return true, despite the value being NaN. T

This is allowed, since `-ffast-math` implies `-ffinite-math-only`, which is explicitly allowed to do this. However, apple clang must only recently have started doing this optimization, hence the problem not present on CI builds for MAVSDK <= 1.4.7.


MaEtUgR (approved):
I didn't see a specific requirement for this to be enabled from the history :eyes:

ThomasDebrunner:
I believe it was enabled by accident for iOS because the iOS toolchain file was copied over from another project

julianoes:
Thanks for the patch! Nice catch. We should probably fix CI first, so that we know the changes don't break more than they fix.

The toolchain comes from https://github.com/leetal/ios-cmake/blob/master/ios.toolchain.cmake and it looks like fast-math was removed there as well, see: https://github.com/leetal/ios-cmake/pull/143/files

I suggest we update the whole file to the latest upstream instead of just patching this.

JonasVautherin (changes requested):
I agree with Julian that the toolchain should be updated from upstream instead of patched, unless there is a good reason not to do it.

The problem with patching it is that next time we update the toolchain from upstream, we may just lose those patches. Unless somebody wants to maintain a fork of the toolchain, but I would really prefer to keep as close as possible to upstream :innocent:.

julianoes:
Fixed in v1.4 but for v2, we'll try to update this properly with #2184.

PR #2177, opened 2023-11-13
[RELEASE v1.4] Backport disable -ffast-math for iOS

ThomasDebrunner:
Backport of https://github.com/mavlink/MAVSDK/pull/2176

CI seems to fail, since macOS runners come with Python 3.12, which break the mavlink generators


JonasVautherin (changes requested):
Same request as in the `main` PR: I think that the toolchain should be updated from upstream instead of patched, when possible.

julianoes:
@JonasVautherin for v1.4 I would be ok just patching the one option. If we update the full toolchain here, we introduce more risk and churn.

JonasVautherin:
@julianoes: good point, fine for me then :+1:

julianoes:
Rebased and force-pushed.

julianoes (approved):

JonasVautherin (approved):

PR #2182, opened 2023-11-15
[v1.4] Downgrade Python to 3.11 for macOS CI builds

JonasVautherin:
Backporting https://github.com/mavlink/MAVSDK/pull/2179

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -486,6 +486,11 @@ jobs:
           - name: macOS-framework
             build-framework: ON
     steps:
+      - name: Install Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.11'
+          architecture: 'x64'
       - uses: actions/checkout@v4
         with:
           submodules: recursive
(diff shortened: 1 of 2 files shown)


JonasVautherin:
Hmm CMake keeps detecting 3.12:

```
-- Found Python3: /usr/local/bin/python3.12 (found version "3.12.0") found components: Interpreter
```

JonasVautherin:
Oh, v1.4 [uses CMake to detect Python](https://github.com/mavlink/MAVSDK/blob/v1.4/third_party/mavlink/CMakeLists.txt#L17):

```
find_package(Python3 COMPONENTS Interpreter REQUIRED)
```

`main` [does not](https://github.com/mavlink/MAVSDK/blob/main/third_party/mavlink/CMakeLists.txt).

JonasVautherin:
Ok let's see. In the current CI run, iOS has the older runner image, and macOS has the new one. If your latest suggestion fixes it, both should work :+1:.

JonasVautherin:
Seems better! Good catch @julianoes!

JonasVautherin:
Now Windows breaks :see_no_evil:. Retrying with `set(Python_FIND_REGISTRY "NEVER")` for Windows, which was the old value.

PR #2185, opened 2023-11-23
Include header <atomic> in connection.h to avoid compiler erros with msvc2019 16.4

chehrlic:
Even though newer MSVC 2019 version compile perfectly fine it's worth including this header since it's used in 'Connection' class.


julianoes:
Thanks @chehrlic!

julianoes (approved):

PR #2189, opened 2023-12-02
Update to MavlinkPassthrough unsubscribe_message

c-devine:
This updates MavlinkPassthrough to require a message id in unsubscribe_message.  It fixes the bug where unsubscribing from a passthrough message also unsubscribed from messages with the same handle id but different message ids.


JonasVautherin (approved):
Nice, thanks!

julianoes:
Thanks for the follow up @c-devine! Let's see if it passes now.

julianoes (approved):

PR #2195, opened 2023-12-17
camera: remove cached camera definition files

julianoes:
As of MAVSDK v2 we will only support camera definition files available via http(s) or mavlink-ftp and no longer carry these hard-coded definition files which are unlikely still relevant.


JonasVautherin (approved):
Nice cleanup!

PR #2200, opened 2023-12-27
ci: fix tinyxml2 build for alpine 3.19

JonasVautherin:
Applying the patch from alpine (see https://git.alpinelinux.org/aports/tree/main/tinyxml2/no-lfs64.patch). I slightly modified the patch for Android.


JonasVautherin:
Android is not happy with that patch, I'll have another look later.

julianoes (approved):

PR #2205, opened 2024-01-05
Update some more dependencies

JonasVautherin:
* cpptrace
* curl
* jsoncpp


JonasVautherin:
@julianoes: Updating curl to v8 somehow breaks the build, so I updated to the latest v7 for now.

julianoes (approved):

PR #2206, opened 2024-01-07
Update grpc and protobuf

JonasVautherin:


JonasVautherin:
Oh the patch does not apply anymore for cross-compilation, I need to fix that :+1:.

julianoes (approved):
Thanks!

Issue #2207, opened 2024-01-10
heap-use-after-free after heartbeats timed out when using UDP connection


keith-bennett-airmap:
# tl;dr

This appears to be a timing issue and the application typically crashes at intermittent intervals. Shared or overused compute clusters can cause delays to network traffic (especially with a cloud-based simulation). Without directly observing a packet capture, this is a proposed sequence of events to reproduce in a regression test:
- connect over udp
- stop sending heartbeats
- get notified of disconnect
- immediately send gimbal events
- crash

# Description
This is observed with [mavsdk 7fe1b6083545a0760dc306388ea946ded78316da](https://github.com/mavlink/MAVSDK/tree/7fe1b6083545a0760dc306388ea946ded78316da) when compiled with Address Sanitizer.

mavsdk library is used by a DroneUp application, the `mavlink_shim` and the whole application is built in a Docker-based workflow. We see a heap-use-after-free error which we believe is related to mavsdk and UDP network connectivity. The issue's reproducibility is intermittent -- we don't see it frequently nor for every heartbeats timed out event. Debugging an intermittent issue is difficult :rofl: 

We observe these events:
- our `mavlink_client` object is using mavsdk. It's configured to use a udp connection and has successfully connected.
- At some point, mavsdk reports a connection loss (eg, missed heartbeats)
- We then observe a crash in the gimbal manager inside of mavsdk
- The crash appears to be within mavsdk's internal components.

The crash provides the following stack trace information.

# Address Sanitizer Defect Report
```
=================================================================
==1==ERROR: AddressSanitizer: heap-use-after-free on address 0x6080000103e0 at pc 0x563c693f2add bp 0x7f6c3d6f0d50 sp 0x7f6c3d6f0d40
READ of size 8 at 0x6080000103e0 thread T17
    #0 0x563c693f2adc in std::_Function_base::_M_empty() const /usr/include/c++/11/bits/std_function.h:247
    #1 0x563c69467a55 in std::function<void (mavsdk::Gimbal::ControlStatus)>::operator bool() const /usr/include/c++/11/bits/std_function.h:573
    #2 0x563c698e4a32 in mavsdk::GimbalProtocolV2::process_gimbal_manager_status(__mavlink_message const&) (/usr/src/app/install/bin/mavlink_shim+0x6cea32)
[...]

PR #2208, opened 2024-01-11
gimbal: fix unregister message subscription

julianoes:
We should unregister anything we have subscribed to when being destructed.

Related to #2207

### DIFF
--- a/src/mavsdk/plugins/gimbal/gimbal_protocol_v2.cpp
+++ b/src/mavsdk/plugins/gimbal/gimbal_protocol_v2.cpp
@@ -18,6 +18,15 @@ GimbalProtocolV2::GimbalProtocolV2(
     _gimbal_manager_compid(gimbal_manager_compid)
 {}
 
+GimbalProtocolV2::~GimbalProtocolV2()
+{
+    if (_is_mavlink_manager_status_registered) {
+        _is_mavlink_manager_status_registered = false;
+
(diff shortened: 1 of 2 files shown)


src/mavsdk/plugins/gimbal/gimbal_protocol_v2.cpp
```diff
@@ -18,6 +18,15 @@ GimbalProtocolV2::GimbalProtocolV2(
     _gimbal_manager_compid(gimbal_manager_compid)
 {}
 
+GimbalProtocolV2::~GimbalProtocolV2()
+{
+    if (_is_mavlink_manager_status_registered) {
+        _is_mavlink_manager_status_registered = false;
+
+        _system_impl.unregister_mavlink_message_handler(MAVLINK_MSG_ID_GIMBAL_MANAGER_STATUS, this);
```
JonasVautherin, line 26: Not sure if it matters, but I'm always more comfortable setting the boolean after the fact (i.e. inverting those lines). So that it reads "now that I unregistered, I update the boolean that holds the registration state".

Probably just a nitpick though.

JonasVautherin (approved):

src/mavsdk/plugins/gimbal/gimbal_protocol_v2.cpp
julianoes: The advantage of that way is that the check and set are right after each other and less likely to drift apart from each other and get lost.

Best case, it should not be a two step approach but one. Actually, we don't need this variable. We can just unregister and if we haven't registered it's silently ignored...

PR #2210, opened 2024-01-12
Full gimbal angle control

Daybreakerflint:
Implementation of the full gimbal angle control

This feature is a request as I have to use the gimbal I have in a certain way and the roll and pitch axis are switched.
As the pitch of the payload of the gimbal is vitally important, this feature is needed.

I have tested the new implementation and it fully works.

Proto PR:
https://github.com/mavlink/MAVSDK-Proto/pull/328

Quick note from my side: This is the first time I have done something like this and thus I am completely new to the process or anything. If I have done something wrong, let me know. I like to improve.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -3,4 +3,4 @@
 	url = https://github.com/google/googletest
 [submodule "mavsdk-proto"]
 	path = proto
-	url = https://github.com/mavlink/MAVSDK-Proto.git
+	url = https://github.com/Daybreakerflint/MAVSDK-Proto
--- a/.gitmodules
+++ b/.gitmodules
@@ -3,4 +3,4 @@
 	url = https://github.com/google/googletest
 [submodule "mavsdk-proto"]
 	path = proto
-	url = https://github.com/Daybreakerflint/MAVSDK-Proto
+	url = https://github.com/mavlink/MAVSDK-Proto.git
--- a/src/mavsdk/core/CMakeLists.txt
+++ b/src/mavsdk/core/CMakeLists.txt
@@ -121,7 +121,7 @@ if(ANDROID)
 endif()
 
 if((BUILD_STATIC_MAVSDK_SERVER AND ("${CMAKE_C_COMPILER_ID}" STREQUAL "GNU")) OR
-    (${CMAKE_HOST_SYSTEM_PROCESSOR} MATCHES "(armv6|armv7)"))
+    (${CMAKE_HOST_SYSTEM_PROCESSOR} MATCHES "(armv6|armv7|aarch64)"))
     target_link_libraries(mavsdk PRIVATE atomic)
 endif()
 
--- a/src/mavsdk/plugins/gimbal/gimbal_protocol_v2.cpp
+++ b/src/mavsdk/plugins/gimbal/gimbal_protocol_v2.cpp
@@ -90,7 +90,7 @@ void GimbalProtocolV2::set_angles_async(
     float pitch_deg, float yaw_deg, float roll_deg, Gimbal::ResultCallback callback)
 {
     // Sending the message should be quick and we can just do that straighaway.
-    Gimbal::Result result = set_angles_yaw(pitch_deg, yaw_deg, roll_deg);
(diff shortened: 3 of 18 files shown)


src/mavsdk/core/CMakeLists.txt
```diff
@@ -121,7 +121,7 @@ if(ANDROID)
 endif()
 
 if((BUILD_STATIC_MAVSDK_SERVER AND ("${CMAKE_C_COMPILER_ID}" STREQUAL "GNU")) OR
-    (${CMAKE_HOST_SYSTEM_PROCESSOR} MATCHES "(armv6|armv7)"))
+    (${CMAKE_HOST_SYSTEM_PROCESSOR} MATCHES "(armv6|armv7|aarch64)"))
```
julianoes, line 124: Interesting, is this now suddenly required for aarch64 as well?

src/mavsdk/plugins/gimbal/gimbal_protocol_v2.cpp
```diff
@@ -86,11 +86,27 @@ Gimbal::Result GimbalProtocolV2::set_pitch_and_yaw(float pitch_deg, float yaw_de
...
+{
+    return set_angles(pitch_deg, yaw_deg, 0.0f);
+}
+
 void GimbalProtocolV2::set_pitch_and_yaw_async(
     float pitch_deg, float yaw_deg, Gimbal::ResultCallback callback)
 {
     // Sending the message should be quick and we can just do that straighaway.
-    Gimbal::Result result = set_pitch_and_yaw(pitch_deg, yaw_deg);
+    Gimbal::Result result = set_angles(pitch_deg, yaw_deg, 0.0f);
```
julianoes, line 109: Nice!

julianoes (commented):
See comments in the proto PR regarding ordering, and then the one comment here.

And why did gtest change?

src/mavsdk/core/CMakeLists.txt
Daybreakerflint: This was to make it work on the raspberry pi. Without the statement, it wouldn't include the needed library, if I remember correctly.

src/mavsdk/plugins/gimbal/gimbal_protocol_v2.cpp
julianoes: If you do an interactive rebase you should be able to edit and amend that commit. Or you can just undo it again in a commit.

src/mavsdk/core/CMakeLists.txt
julianoes: Ok, that's possible.

Daybreakerflint:
The gtest should now be back to the version used in the original MAVSDK repository.
As well as the order of the angles should be changed everywhere.

.gitmodules
```diff
@@ -3,4 +3,4 @@
 	url = https://github.com/google/googletest
 [submodule "mavsdk-proto"]
 	path = proto
-	url = https://github.com/mavlink/MAVSDK-Proto.git
+	url = https://github.com/Daybreakerflint/MAVSDK-Proto
```
julianoes, line 6: Let me merge the MAVSDK-Proto PR, then you can update that and we can merge it.

julianoes (approved):

.gitmodules
Daybreakerflint: Done.

JonasVautherin (dismissed):

julianoes (approved):

julianoes:
I've fixed this up, pointing to the right proto hash (after the merge) and running the style fix again.

Daybreakerflint:
No problem. Our project really needs it. Also thank you for guiding me through my first PR!
I was wondering when the next release is planned for MAVSDK? And when it would be in the python library? (estimations are OK)

julianoes:
Let me trigger a new release. No reason not to :). And then Python sometime next week.

julianoes:
Released: https://github.com/mavlink/MAVSDK-Python/releases/tag/2.0.0
Needs testing though.

PR #2212, opened 2024-01-22
ArduPilot: Prevent std::future_error in takeoff, and return attitude in euler angles

lamlaaaam:
This PR addresses 2 issues when using MAVSDK with ArduPilot. These were tested on [ArduPilot SITL Copter-4.4.4](https://github.com/ArduPilot/ardupilot/tree/Copter-4.4.4), using MAVSDK v2.0.1.

### Observation
When ArduPilot SITL starts up, it takes time before the GPS initializes and for the UAV to obtain a position estimate. During this time, if the takeoff command is given, an unhandled std::future_error exception is thrown.

### Cause
Within Action_Impl::takeoff_async_apm, if set_flight_mode_async fails, it triggers the callback to set a failed result. However, it still calls send_takeoff_command, which triggers the same callback, and tries to set another result value for a fulfilled promise.

### Fix
Just don't call the send_takeoff_command if switch to Guided mode fails.

### Observation
The values received for roll, pitch, yaw in the telemetry are NaN.

### Cause
In TelemetryImpl::process_attitude, the values are extracted from the received telemetry message and correctly populated into a Telemetry::EulerAngle struct. However, this struct is then never used again. Instead, it grabs the quaternion version, and does a conversion to euler angles. I'm not sure what the actual reason causing the NaN is, but it's probably somewhere in the conversion.

### Fix
Use the Telemetry::EulerAngle struct that was populated.

### DIFF
--- a/src/mavsdk/plugins/action/action_impl.cpp
+++ b/src/mavsdk/plugins/action/action_impl.cpp
@@ -407,8 +407,7 @@ void ActionImpl::takeoff_async_apm(const Action::ResultCallback& callback) const
                     if (callback) {
                         callback(action_result);
                     }
-                }
-                else {
+                } else {
                     send_takeoff_command();
                 }
             });


JonasVautherin (dismissed):
Nice ones!

JonasVautherin:
OpenSSL is timing out for now, let's try the CI again in a few hours :+1:.

JonasVautherin:
There is a styling issue to fix, though:

```patch
diff --git a/src/mavsdk/plugins/action/action_impl.cpp b/src/mavsdk/plugins/action/action_impl.cpp
index fae1052..e75dc7e 100644
--- a/src/mavsdk/plugins/action/action_impl.cpp
+++ b/src/mavsdk/plugins/action/action_impl.cpp
@@ -407,8 +407,7 @@ void ActionImpl::takeoff_async_apm(const Action::ResultCallback& callback) const
                     if (callback) {
                         callback(action_result);
                     }
-                }
-                else {
+                } else {
                     send_takeoff_command();
                 }
             });
Error: Process completed with exit code 1.
```

JonasVautherin (approved):

julianoes (approved):
Thanks for the fixes, much appreciated!

Issue #329, opened 2018-03-20
Rename "Mavlink" stuffs to that of "MAVLink"


shakthi-prashanth-m:
These classes use lower case for MavLink:
```
class MavlinkCommands {...};
class MavlinkParameters {...};
class MavlinkReceiver {...};
class MavlinkChannels {...};
```
In QGC source and elsewhere the spelling is MAVLink. Should I change these to use the correct case (e.g. `MavlinkCommands` to `MAVLinkCommands`, etc.)

Please suggest: @hamishwillee @julianoes @JonasVautherin .
Thanks.

hamishwillee:
If MAVLink is always implied then it isn't needed. If something is intended to be generic then it shouldn't be prefixed. You'd use a MAVLink where something must always be for working with MAVLink and only makes sense in that context.

shakthi-prashanth-m:
OK. Thanks for rephrasing my question @hamishwillee . :)

shakthi-prashanth-m:
Addressed as part of #310.

julianoes:
To me mavlink is sort of like a word which is why I wrote it that way. We can change it but I don't know if the churn is really warranted.

hamishwillee:
Probably not, but already done :-)

shakthi-prashanth-m:
Addressed in #310 .

PR #2213, opened 2024-01-31
Capability flags in camera information

ddatsko:
This PR goes together with the [#329 PR in MAVSDK-Proto](https://github.com/mavlink/MAVSDK-Proto/pull/329) and adds the implementation for setting [CAMERA_CAP_FLAGS](https://mavlink.io/en/messages/common.html#CAMERA_CAP_FLAGS) in camera information by the user. This allows the user to set additional capabilities that cannot be determined automatically based on callback presence or so. For example, the implementation can not determine whether the [CAMERA_CAP_FLAGS_CAN_CAPTURE_IMAGE_IN_VIDEO_MODE](https://mavlink.io/en/messages/common.html#CAMERA_CAP_FLAGS_CAN_CAPTURE_IMAGE_IN_VIDEO_MODE) should be set, but this can be useful for some users. Another example of such a need is when the component also uses the tracking_server plugin for object tracking and the [CAMERA_CAP_FLAGS_HAS_TRACKING_POINT](https://mavlink.io/en/messages/common.html#CAMERA_CAP_FLAGS_HAS_TRACKING_POINT) or similar should be set in that case.

I didn't remove the functionality of determining four of the flags based on callback presence. This way, setting no flags will behave in exactly the same way as before


JonasVautherin (commented):
Can you fix the style by running `./tools/fix_style.sh .`? That should make the CI happy :innocent:

JonasVautherin:
Oh one more thing: you need to make the mavsdk-proto submodule point to your branch that adds this definition (otherwise the CI will fail).

When both PRs (this one and the mavsdk-proto one) are accepted, we can merge the mavsdk-proto one, you can update the submodule here, and finally we can merge this PR.

Does that make sense?

ddatsko:
Yes, that makes sense. I will then wait for that one to get merged and will change the proto submodule here

JonasVautherin:
You can already point this PR to your MAVSDK-Proto branch, such that the CI passes here. Then it will have to be changed again once the MAVSDK-Proto PR gets merged :+1:

julianoes (approved):

PR #2217, opened 2024-02-08
CI: don't include debug symbols for musl builds

julianoes:
When we add the debug symbols for the musl builds we end up with binaries around 300 to 400 MB which are too big for PyPi.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -91,11 +91,11 @@ jobs:
         if: steps.cache.outputs.cache-hit == 'true'
         run: echo "superbuild=-DSUPERBUILD=OFF" >> $GITHUB_ENV && echo "cmake_prefix_path=-DCMAKE_PREFIX_PATH=$(pwd)/build/release/third_party/install" >> $GITHUB_ENV
       - name: configure
-        run: cmake $superbuild $cmake_prefix_path -DCMAKE_BUILD_TYPE=RelWithDebInfo -DBUILD_MAVSDK_SERVER=ON -DWERROR=OFF -DENABLE_CPPTRACE=On -Bbuild/release -H.
+        run: cmake $superbuild $cmake_prefix_path -DCMAKE_BUILD_TYPE=RelWithDebInfo -DBUILD_MAVSDK_SERVER=ON -DWERROR=OFF -DENABLE_CPPTRACE=On -DCMAKE_INSTALL_PREFIX=install -Bbuild/release -H.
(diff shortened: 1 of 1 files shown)


JonasVautherin (approved):

.github/workflows/main.yml
```diff
@@ -104,7 +104,7 @@ jobs:
       - name: install examples dependencies
         run: sudo apt-get install libsdl2-dev
       - name: configure examples
-        run: cmake -DCMAKE_PREFIX_PATH=$(pwd)/build/release/third_party/install -DCMAKE_BUILD_TYPE=RelWithDebInfo -DWERROR=OFF -Bexamples/build -Hexamples
+        run: cmake -DCMAKE_PREFIX_PATH=$(pwd)/install -DCMAKE_PREFIX_PATH=$(pwd)/build/release/third_party/install -DCMAKE_BUILD_TYPE=RelWithDebInfo -DWERROR=OFF -Bexamples/build -Hexamples
```
JonasVautherin, line 107: Are you sure this works? I think I usually do something more like `-DCMAKE_PREFIX_PATH="$(pwd)/install;$(pwd)/build/release/third_party/install"` :thinking:

julianoes: Oh, I think you're right!

.github/workflows/main.yml
```diff
@@ -104,7 +104,7 @@ jobs:
       - name: install examples dependencies
         run: sudo apt-get install libsdl2-dev
       - name: configure examples
-        run: cmake -DCMAKE_PREFIX_PATH=$(pwd)/build/release/third_party/install -DCMAKE_BUILD_TYPE=RelWithDebInfo -DWERROR=OFF -Bexamples/build -Hexamples
+        run: cmake -DCMAKE_PREFIX_PATH=$(pwd)/install;$(pwd)/build/release/third_party/install -DCMAKE_BUILD_TYPE=RelWithDebInfo -DWERROR=OFF -Bexamples/build -Hexamples
```
JonasVautherin, line 107: I would use double quotes, because otherwise I am not sure how the `;` is interpreted.

JonasVautherin: Yep, seems like that was it :blush:.

JonasVautherin (dismissed):

.github/workflows/main.yml
julianoes: Thanks!

PR #2223, opened 2024-02-16
Fix illegal instruction on RPi 4

julianoes:
The absl library tries to make use of the crypto libraries on arm64 platforms. However, it turns out that the RPi 4 chip does not have the crypto extensions. Hence, we need to patch absl to refrain from using the hardware crypto stuff and fallback to slower software implementations. This is only required for a random back-off timer at the moment, so shouldn't matter too much.

Note that the RPi 5 has the crypto extensions but our build is the same for both.

Fixes https://github.com/mavlink/MAVSDK-Python/issues/654.
Fixes https://github.com/mavlink/MAVSDK-Python/issues/651.

### DIFF
--- a/third_party/absl/CMakeLists.txt
+++ b/third_party/absl/CMakeLists.txt
@@ -31,5 +31,6 @@ ExternalProject_add(
     GIT_REPOSITORY https://github.com/abseil/abseil-cpp.git
     GIT_TAG 20240116.1
     PREFIX absl
+    PATCH_COMMAND git checkout . && git apply ${PROJECT_SOURCE_DIR}/rpi-no-crypto.patch
     CMAKE_ARGS "${CMAKE_ARGS}"
     )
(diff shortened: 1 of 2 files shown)


JonasVautherin:
Isn't it the same as https://github.com/abseil/abseil-cpp/issues/662? Weird that they seem to have fixed it :thinking:.

In any case, if there really is a need for a patch, I would do it only for the relevant platform. But I wonder if that could not be solved somehow with [ABSL_RANDOM_HWAES_ARM64_FLAGS](https://github.com/abseil/abseil-cpp/blob/master/absl/copts/copts.py#L185) :thinking:

akrv:
waiting for this fix 🤞

julianoes:
@JonasVautherin it looks related. I'm not sure. What I found is that hardware crypto is enabled for aarch64, and that might make sense for most platforms, but the Broadcom BCM2711 chip is lacking the crypto extensions.

julianoes:
Of course updating absl was a mistake and it breaks the build. Gosh I hate dependencies.

ilysec:
Seems to work, thanks a lot!

alltilla:
Thank you, we faced the exact same issue in AxoSyslog running in an alpine based container on an RPi4. This PR was a good confirmation for us. :)

P.S.: OpenSSL does runtime checks in init time to determine which CPU features are available on the runner host for certain operations, it would be nice to have something similar in the upstream libabseil, too.

PR #2222, opened 2024-02-16
ArduPilot mission_raw fixes

julianoes:
This fixes an `InvalidSequence` when using QGC plan import, and a `Timeout` when trying to upload, both for ArduPilot.

PR #2224, opened 2024-02-17
Added the ability to get the currently connected autopilot type.

Zvereman:
Sometimes you need to get the type of connected autopilot into your program. For example, in order to correctly compose a mission for Ardupilot and PX4 at the same time.

Here's a small example of use.

```
std::shared_ptr< mavsdk::System > system_;

if( system_->autopilot() == mavsdk::System::Autopilot::ArduPilot ) {
                   // Do something
}
```

### DIFF
--- a/src/mavsdk/plugins/ftp/ftp_impl.cpp
+++ b/src/mavsdk/plugins/ftp/ftp_impl.cpp
@@ -42,9 +42,14 @@ void FtpImpl::download_async(
         use_burst,
         [callback, this](
             MavlinkFtpClient::ClientResult result, MavlinkFtpClient::ProgressData progress_data) {
-            callback(
-                result_from_mavlink_ftp_result(result),
-                progress_data_from_mavlink_ftp_progress_data(progress_data));
+            if (callback) {
+                _system_impl->call_user_callback(
+                    [temp_callback = callback, result, progress_data, this]() {
+                        temp_callback(
+                            result_from_mavlink_ftp_result(result),
+                            progress_data_from_mavlink_ftp_progress_data(progress_data));
+                    });
+            }
         });
 }
 
@@ -58,9 +63,14 @@ void FtpImpl::upload_async(
         remote_folder,
         [callback, this](
             MavlinkFtpClient::ClientResult result, MavlinkFtpClient::ProgressData progress_data) {
-            callback(
-                result_from_mavlink_ftp_result(result),
-                progress_data_from_mavlink_ftp_progress_data(progress_data));
+            if (callback) {
+                _system_impl->call_user_callback(
+                    [temp_callback = callback, result, progress_data, this]() {
+                        temp_callback(
+                            result_from_mavlink_ftp_result(result),
+                            progress_data_from_mavlink_ftp_progress_data(progress_data));
+                    });
+            }
         });
 }
 
@@ -100,7 +110,11 @@ void FtpImpl::create_directory_async(const std::string& path, Ftp::ResultCallbac
 {
     _system_impl->mavlink_ftp_client().create_directory_async(
         path, [callback, this](MavlinkFtpClient::ClientResult result) {
-            callback(result_from_mavlink_ftp_result(result));
+            if (callback) {
+                _system_impl->call_user_callback([temp_callback = callback, result, this]() {
+                    temp_callback(result_from_mavlink_ftp_result(result));
+                });
+            }
         });
 }
 
@@ -118,7 +132,11 @@ void FtpImpl::remove_directory_async(const std::string& path, Ftp::ResultCallbac
 {
     _system_impl->mavlink_ftp_client().remove_directory_async(
         path, [callback, this](MavlinkFtpClient::ClientResult result) {
-            callback(result_from_mavlink_ftp_result(result));
+            if (callback) {
+                _system_impl->call_user_callback([temp_callback = callback, result, this]() {
+                    temp_callback(result_from_mavlink_ftp_result(result));
+                });
+            }
         });
 }
 
@@ -136,7 +154,11 @@ void FtpImpl::remove_file_async(const std::string& path, Ftp::ResultCallback cal
 {
     _system_impl->mavlink_ftp_client().remove_file_async(
         path, [callback, this](MavlinkFtpClient::ClientResult result) {
-            callback(result_from_mavlink_ftp_result(result));
+            if (callback) {
+                _system_impl->call_user_callback([temp_callback = callback, result, this]() {
+                    temp_callback(result_from_mavlink_ftp_result(result));
+                });
+            }
         });
 }
 
@@ -155,7 +177,11 @@ void FtpImpl::rename_async(
 {
     _system_impl->mavlink_ftp_client().rename_async(
         from_path, to_path, [callback, this](MavlinkFtpClient::ClientResult result) {
-            callback(result_from_mavlink_ftp_result(result));
+            if (callback) {
+                _system_impl->call_user_callback([temp_callback = callback, result, this]() {
+                    temp_callback(result_from_mavlink_ftp_result(result));
+                });
+            }
         });
 }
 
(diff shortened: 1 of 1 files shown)


JonasVautherin:
> For example, in order to correctly compose a mission for Ardupilot and PX4 at the same time.

Shouldn't the mission plugin compose the right mission and expose a higher-level interface to the user? :thinking:

Zvereman:
> > For example, in order to correctly compose a mission for Ardupilot and PX4 at the same time.
> 
> Shouldn't the mission plugin compose the right mission and expose a higher-level interface to the user? 🤔

Sorry that I didn't initially indicate what exactly I was using. To load the mission I use **mission_raw**. Without adding one point for the Ardupilot mission, the mission will not load. Here is my method for converting and adding mission points. If there is any simpler implementation of loading a mission simultaneously for Ardupilot and PC4, then I will be happy to hear from you :)

```
MissionItems convertToMissionItems( const FlightMission& mission )
{
	MissionItems items;
	if( mission.entries.empty() )
		return items;
	items.reserve( mission.entries.size() + 1 );

	if( system_->autopilot() == mavsdk::System::Autopilot::ArduPilot ) {
		auto& item = items.emplace_back();
		item.autocontinue = 1;
		item.command = 22;
		item.mission_type = 0;
		item.seq = 0;
		item.frame = 3;
		item.x = 0;
		item.y = 0;
		item.z = 5.0;
	}

	auto addEntry = [&]< typename Entry >( const Entry& entry )
	{
		auto& item = items.emplace_back();
		if constexpr( std::is_same_v< Entry, FlightMission::WaypointEntry > )
		{
			item.autocontinue = 1;
			item.command = MAV_CMD::MAV_CMD_NAV_WAYPOINT;
			item.seq = ( uint32_t )items.size() - 1;
			if( system_->autopilot() == mavsdk::System::Autopilot::ArduPilot ) {
				if (item.seq == 1)
					item.current = 1;
				switch( entry.altType )
				{
				case FlightMission::WaypointEntry::AltType::Absolute:
					item.frame = MAV_FRAME::MAV_FRAME_GLOBAL_INT;
					break;
				case FlightMission::WaypointEntry::AltType::Relative:
					item.frame = MAV_FRAME::MAV_FRAME_GLOBAL_RELATIVE_ALT;
					break;
				case FlightMission::WaypointEntry::AltType::Terrain:
					item.frame = MAV_FRAME::MAV_FRAME_GLOBAL_TERRAIN_ALT;
					break;
				default:
					break;
				}
			} else {
				item.current = items.size() == 1;
				switch( entry.altType )
				{
				case FlightMission::WaypointEntry::AltType::Absolute:
					item.frame = MAV_FRAME::MAV_FRAME_GLOBAL_INT;
					break;
				case FlightMission::WaypointEntry::AltType::Relative:
					item.frame = MAV_FRAME::MAV_FRAME_GLOBAL_RELATIVE_ALT_INT;
					break;
				case FlightMission::WaypointEntry::AltType::Terrain:
					item.frame = MAV_FRAME::MAV_FRAME_GLOBAL_TERRAIN_ALT_INT;
					break;
				default:
					break;
				}
				item.param1 = 0.0f;
				item.param2 = 1.0f;
				item.param3 = 0.0f;
				item.param4 = std::numeric_limits< float >::quiet_NaN();
			}
			item.mission_type = MAV_MISSION_TYPE::MAV_MISSION_TYPE_MISSION;			
			item.x = int32_t( std::round( entry.lat * 1e7 ) );
			item.y = int32_t( std::round( entry.lon * 1e7 ) );
			item.z = entry.alt;
		}
	};
	for( auto& entryVariant : mission.entries )
		std::visit( addEntry, entryVariant );
	if( mission.postAction != FlightMission::PostMissionAction::Nothing )
	{
		auto& item = items.emplace_back();
		item.seq = ( uint32_t )items.size() - 1;		
		switch( mission.postAction )
		{
		case FlightMission::PostMissionAction::ReturnToLaunch:
			item.command = MAV_CMD::MAV_CMD_NAV_RETURN_TO_LAUNCH;
			break;
		case FlightMission::PostMissionAction::Land:
			item.command = MAV_CMD::MAV_CMD_NAV_LAND;
			break;
		default:
			break;
		}
		if( system_->autopilot() == mavsdk::System::Autopilot::ArduPilot ) {
			item.frame = MAV_FRAME::MAV_FRAME_GLOBAL_INT;
		} else {
			item.command = MAV_CMD::MAV_CMD_NAV_RETURN_TO_LAUNCH;
			item.frame = MAV_FRAME::MAV_FRAME_MISSION;
			item.current = 0;
			item.param1 = 0.0f;
			item.param2 = 0.0f;
			item.param3 = 0.0f;
			item.param4 = std::numeric_limits< float >::quiet_NaN();
		}
		item.autocontinue = 1;
		item.x = 0;
		item.y = 0;
		item.z = 0;
		item.mission_type = MAV_MISSION_TYPE::MAV_MISSION_TYPE_MISSION;
	}
	return items;
}
```

julianoes:
I agree with @JonasVautherin that it generally should not be required that different code runs based on the autopilot. The idea is that Mavsdk is abstracting that away. However, I don't mind exposing the information as such about what autopilot is connected.

src/mavsdk/core/include/mavsdk/system.h
```diff
@@ -52,6 +52,21 @@ class System {
...
     bool has_autopilot() const;
 
+    /**
+     * @brief ENUM contains types of basic autopilots..
+     */
+    enum class Autopilot {
+        Unknown,
+        Px4,
+        ArduPilot,
+    };
```
julianoes, line 62: We need to include the existing Autopilot enum instead of duplicating it.

Zvereman:
Thank you for your comments. I will think more abot how i can implement mission upload for Ardupilot and PX4 at the same time. After all, @JonasVautherin said that MAVSDK itself should provide such an opportunity. However, after several tests, I found out that it is still necessary to compose a mission simultaneously for these types of autopilot as indicated in my example of my _convertToMissionItems_ function. Since I tried to use ready-made tools provided in mission_raw. But I haven't been in MAVLINK and MAVSDK for long, so I'll think about it a little more.

Yes. As @julianoes said, it would be nice to know which autopilot is currently connected.
I’ll probably wait for these changes to be implemented and reconsider how this can be done.
https://github.com/mavlink/MAVSDK/pull/2222

JonasVautherin:
> I will think more abot how i can implement mission upload for Ardupilot and PX4 at the same time.

The mission plugin (instead of mission_raw) should do just that. But we need somebody to contribute the Ardupilot side, I think :blush:

Zvereman:
> The mission plugin (instead of mission_raw) should do just that. But we need somebody to contribute the Ardupilot side, I think 😊

I've only been studying MAVLINK for a few months. But since I mostly develop for Ardupilot, I will try to help as much as I can and to the best of my knowledge. :)

julianoes:
@Zvereman I'm happy to merge this if you fix my inline comment.

julianoes:
Huh, what happened here?

Zvereman:
Sorry. Mr. @julianoes Got a lot of things to do :) But I still intend to offer a few of my ideas. As soon as I have enough time, I will open another pull request.
By the way. I tried to simply include autopilot.h, but for some reason I started getting some errors. And since I already said that now, unfortunately, I don’t have as much time as I would like, I will create a new pull request later.

If you're wondering what I mean, here's my fork in the public repository.
https://github.com/Zvereman/MAVSDK/tree/skystream
I'm making some changes in a separate branch. It's called skystream. There are also those that are needed only for my project, but there are also some things in common. For example, connecting via UDP. I reworked it a little.

julianoes:
No worries, thanks for the note. I'll see if I can add it before you.

julianoes:
Done: https://github.com/mavlink/MAVSDK/pull/2236

What you probably missed is that autopilot.h needs to become a public header file and therefore needs to move.

PR #2230, opened 2024-02-22
Ftp: use call_user_callback

chep:
I experienced dead locks while calling ftp functions inside ftp callback. For example, I was trying to start a download while in `are_files_identical_async` callback.
Currently, only `list_directory_async` is using `call_user_callback` since commit 50c23012a65bd5f28b8b19192068a555f582f911


JonasVautherin (approved):
Yeah that makes sense!

julianoes (approved):
Thank you!

PR #2235, opened 2024-02-27
Ftp client: Fix burst download

chep:
I attempted FTP burst download with PX4 and encountered some issues. After a certain period, the burst download would halt, and the remaining data would be downloaded as missed parts. Upon inspecting the PX4 code, I discovered the following comment: `/* perform transfers in 35K chunks - this is determined empirically */`.

When attempting to resume burst download for incomplete files, I observed that missed parts were managed using a file-sized buffer with a binary value (1 or 0) for each byte. This implies that for a 10MB file, it requires 10MB of RAM to download.

With this pull request, instead of storing the entire missed parts in memory, we now only store the offset and size of the missing parts. Burst download is restarted if the file is not complete before checking missed parts.

To test the handling of missed parts, I modified my PX4 to skip 4 packets every 10,000.

### DIFF
--- a/src/mavsdk/core/mavlink_ftp_client.cpp
+++ b/src/mavsdk/core/mavlink_ftp_client.cpp
@@ -512,17 +512,20 @@ bool MavlinkFtpClient::download_burst_continue(
                        << " write: " << std::to_string(payload->size);
         }
 
-        if (payload->offset != item.next_offset) {
-            if (payload->offset < item.next_offset) {
+        if (payload->offset != item.current_offset) {
+            if (payload->offset < item.current_offset) {
                 // Not sure why this would happen but we don't know how to deal with it and ignore
                 // it.
+                LogWarn() << "Got payload offset: " << payload->offset
+                          << ", next offset: " << item.current_offset;
                 return false;
             }
+
             // we missed a part
             item.missing_data.emplace_back(DownloadBurstItem::MissingData{
-                item.next_offset, payload->offset - item.next_offset});
+                item.current_offset, payload->offset - item.current_offset});
             // write some 0 instead
-            std::vector<char> empty(payload->offset - item.next_offset);
+            std::vector<char> empty(payload->offset - item.current_offset);
             item.ofstream.write(empty.data(), empty.size());
             if (!item.ofstream) {
                 LogWarn() << "Write failed";
@@ -542,7 +545,7 @@ bool MavlinkFtpClient::download_burst_continue(
         }
 
         // Keep track of what was written.
-        item.next_offset = payload->offset + payload->size;
+        item.current_offset = payload->offset + payload->size;
 
         if (_debugging) {
             LogDebug() << "Received " << payload->offset << " to "
@@ -617,11 +620,18 @@ bool MavlinkFtpClient::download_burst_continue(
             missing.size -= payload->size;
         }
 
+        // Check if this was the last one
+        if (item.file_size == payload->offset + payload->size) {
+            item.current_offset = item.file_size;
+        }
+
         const size_t bytes_transferred = burst_bytes_transferred(item);
 
-        LogDebug() << "Written " << bytes_transferred << " of " << item.file_size << " bytes";
+        if (_debugging) {
+            LogDebug() << "Written " << bytes_transferred << " of " << item.file_size << " bytes";
+        }
 
-        if (item.missing_data.empty()) {
+        if (item.missing_data.empty() && bytes_transferred == item.file_size) {
             // Final step
             download_burst_end(work);
         } else {
@@ -668,7 +678,7 @@ void MavlinkFtpClient::request_burst(Work& work, DownloadBurstItem& item)
     work.payload.seq_number = work.last_sent_seq_number++;
     work.payload.session = _session;
     work.payload.opcode = work.last_opcode;
-    work.payload.offset = item.next_offset;
+    work.payload.offset = item.current_offset;
 
     // Fill up the whole packet.
     work.payload.size = max_data_length;
@@ -701,13 +711,13 @@ void MavlinkFtpClient::request_next_rest(Work& work, DownloadBurstItem& item)
 
 size_t MavlinkFtpClient::burst_bytes_transferred(DownloadBurstItem& item)
 {
-    return item.next_offset - std::accumulate(
-                                  item.missing_data.begin(),
-                                  item.missing_data.end(),
-                                  size_t(0),
-                                  [](size_t acc, const DownloadBurstItem::MissingData& missing) {
-                                      return acc + missing.size;
-                                  });
+    return item.current_offset - std::accumulate(
+                                     item.missing_data.begin(),
+                                     item.missing_data.end(),
+                                     size_t(0),
+                                     [](size_t acc, const DownloadBurstItem::MissingData& missing) {
+                                         return acc + missing.size;
+                                     });
 }
 
 bool MavlinkFtpClient::upload_start(Work& work, UploadItem& item)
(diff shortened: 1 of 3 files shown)


chep:
Can someone help me for the macOS check ? I have no such hardware and I don't know what is the problem.
` Got signal: Abort trap: 6 (6) (system_tests_runner.cpp:12)
708
<empty trace>`

julianoes:
Neither do I have a Mac. Have you run the tests with address sanitizer? Configure with `-DASAN=On`.

src/mavsdk/core/mavlink_ftp_client.cpp
```diff
@@ -668,48 +674,35 @@ void MavlinkFtpClient::request_burst(Work& work, DownloadBurstItem& item)
...
 {
-    return std::count(
-        item.transferred.begin(), item.transferred.end(), DownloadBurstItem::Transferred::Yes);
+    return item.next_offset - std::accumulate(
+                                  item.missing_data.begin(),
+                                  item.missing_data.end(),
+                                  size_t(0),
+                                  [](size_t acc, const DownloadBurstItem::MissingData& missing) {
+                                      return acc + missing.size;
+                                  });
```
julianoes, line 705: Beautiful :)

src/mavsdk/core/mavlink_ftp_client.h
```diff
@@ -151,12 +152,13 @@ class MavlinkFtpClient {
...
-            No,
-            Yes,
+        uint32_t file_size{0};
+        struct MissingData {
+            size_t offset;
+            size_t size;
         };
-        std::vector<Transferred> transferred{};
-        int last_progress_percentage{-1};
+        std::list<MissingData> missing_data{};
```
julianoes, line 160: I would generally just use a `std::vector` unles you have strong reasons. I guess your thinking is that we are pushing at the back and then popping from the front? Actually, I think that should be a `std::queue`, right?

julianoes (commented):
Very nice, thanks for this optimization! See inline comments.

src/mavsdk/core/mavlink_ftp_client.h
chep: Yes you're right

.github/workflows/main.yml
```diff
@@ -487,7 +487,7 @@ jobs:
           python3 -m pip install wheel
           python3 -m pip install future
       - name: configure
-        run: cmake $superbuild $cmake_prefix_path -DENABLE_STRICT_TRY_COMPILE=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_TOOLCHAIN_FILE=$(pwd)/tools/ios.toolchain.cmake -DENABLE_BITCODE=Off -DPLATFORM=${{ matrix.platform }} -DDEPLOYMENT_TARGET=14.0 -DBUILD_MAVSDK_SERVER=ON -DBUILD_SHARED_LIBS=OFF -DWERROR=OFF -Bbuild/${{ matrix.name }} -H.
+        run: cmake $superbuild $cmake_prefix_path -DENABLE_STRICT_TRY_COMPILE=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_TOOLCHAIN_FILE=$(pwd)/tools/ios.toolchain.cmake -DENABLE_BITCODE=Off -DPLATFORM=${{ matrix.platform }} -DDEPLOYMENT_TARGET=14.0 -DBUILD_MAVSDK_SERVER=ON -DBUILD_SHARED_LIBS=OFF -DWERROR=OFF -Bbuild/${{ matrix.name }} -DENABLE_CPPTRACE=On -H.
```
JonasVautherin, line 490: Nitpick:

```suggestion
        run: cmake $superbuild $cmake_prefix_path -DENABLE_STRICT_TRY_COMPILE=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_TOOLCHAIN_FILE=$(pwd)/tools/ios.toolchain.cmake -DENABLE_BITCODE=Off -DPLATFORM=${{ matrix.platform }} -DDEPLOYMENT_TARGET=14.0 -DBUILD_MAVSDK_SERVER=ON -DBUILD_SHARED_LIBS=OFF -DWERROR=OFF -Bbuild/${{ matrix.name }} -DENABLE_CPPTRACE=ON -H.
```

julianoes:
I rented a Mac server (Intel based) and ran tests in debug and relwithdebinfo build type, and guess what, everything passed fine :sob:.

julianoes:
Finally https://github.com/mavlink/MAVSDK/actions/runs/8133538655/job/22225232773?pr=2235#step:10:725

julianoes:
@chep this looks like a bit of a problematic one:

```
[01:25:36|Debug] allocate 18446744073709543729 empty bytes (mavlink_ftp_client.cpp:520)
```
https://github.com/mavlink/MAVSDK/actions/runs/8134545366/job/22227629255?pr=2235#step:10:718

src/mavsdk/core/mavlink_ftp_client.cpp
```diff
@@ -495,96 +498,88 @@ bool MavlinkFtpClient::download_burst_continue(
...
-                    static_cast<uint32_t>(item.transferred.size())});
+                    static_cast<uint32_t>(burst_bytes_transferred(item)),
+                    static_cast<uint32_t>(item.file_size)});
 
             if (payload->burst_complete) {
-                // The burst is supposedly complete but we still need data, so request next without
-                // burst.
+                // This burst is complete but the file isn't. we need to start a
+                // new one
                 request_next_rest(work, item);
```
chep, line 577: I think we can use burst download with offset (it works with PX4): https://mavlink.io/en/services/ftp.html#reading-a-file-burstreadfile
`Client (i.e. GCS) sends BurstReadFile command specifying the part of the file that it wants to get from an offset to the end, along with the default size of each burst payload.
The payload must specify: session: the current session id, offset = offset in file of start of burst, size= default length of payload in burst responses, data is ignored.`

I changed this because PX4 stop burst after 35KB and then download with standard read is slow (very slow when using serial line).
So we can start bursts until we don't receive the last data of the file and then ask for missing parts with standard reads.

julianoes:
@chep by the way, I can reproduce it like this:

```
cmake --build build -j30 && build/src/system_tests/system_tests_runner --gtest_filter=SystemTest.FtpDownloadBurstBigFileLossy --gtest_repeat=100 --gtest_break_on_failure
```

I think it's an edge cases when the message containing the `burst_end=1` is lost.

chep:
> I think it's an edge cases when the message containing the `burst_end=1` is lost.

Oh…
You're right.
I'm gonna check that.

julianoes:
@chep I've reverted your fix given it didn't fix it, and came up with something that seems to pass the tests.

Would you mind testing this against your PX4 test case again?

src/mavsdk/core/mavlink_ftp_client.cpp
```diff
@@ -1210,8 +1220,35 @@ void MavlinkFtpClient::timeout()
...
+                            // We are done anyway.
+                            item.callback(ClientResult::Success, {});
+                            download_burst_end(*work);
+                            work_queue_guard.pop_front();
+                        } else {
+                            // The burst is supposedly complete but we still need data because
+                            // we missed some, so request next without burst.
+                            // We presumably missed the very last chunk.
+                            if (item.current_offset < item.file_size) {
+                                item.missing_data.emplace_back(DownloadBurstItem::MissingData{
```
chep, line 1238: Maybe we should start a new burst download because if we miss the first burst ack, we would download all the file with standard read.

chep:
Test failed again, remaining byte count was negative. I think I fixed it. I'll try with my PX4 test case today.

julianoes:
You can see how PX4 arbitrary "stop at 35K" which is not according to the spec, opens all sorts of edge cases. :cry:

chep:
It works here with PX4.
I don't know px4 code at all, I just found this comment… 
https://github.com/PX4/PX4-Autopilot/commit/97b8e7a20e1bf64b7677911e6bbab300ff471358

julianoes:
Yep, I've seen this comment. Thanks for figuring this out in the first place. While I would want to change the PX4 implementation, we need to try to be compatible with current and older PX4 versions anyway, so the fixes are good!

Thanks for testing again, I'll go ahead and merge this now!

julianoes (approved):

PR #2246, opened 2024-03-13
Stop list directory job if received size is null

chep:
When we ask PX4 to list an empty directory, it sends back an ack with a empty payload.


julianoes (approved):
I think that makes sense, thanks!

PR #2250, opened 2024-03-18
Ftp server : use a thread to send ftp burst

chep:
Currently, 100 packets are sent per second. With a 255 bytes payload, send rate is 25kB/s. With a thread, we use the line at its maximum speed.

### DIFF
--- a/src/mavsdk/core/mavlink_ftp_server.cpp
+++ b/src/mavsdk/core/mavlink_ftp_server.cpp
@@ -770,22 +770,31 @@ void MavlinkFtpServer::_work_burst(const PayloadHeader& payload)
 
     _session_info.burst_offset = payload.offset;
     _session_info.burst_chunk_size = payload.size;
-
     _burst_seq = payload.seq_number + 1;
 
+    if (_session_info.burst_thread.joinable()) {
+        _session_info.burst_stop = true;
+        _session_info.burst_thread.join();
+    }
+
+    _session_info.burst_stop = false;
+
     // Schedule sending out burst messages.
-    // Use some arbitrary "fast" rate: 100 packets per second
-    _server_component_impl.add_call_every(
-        [this]() { _send_burst_packet(); }, 0.01f, &_burst_call_every_cookie);
+    _session_info.burst_thread = std::thread([this]() {
+        while (!_session_info.burst_stop)
+            if (_send_burst_packet())
+                break;
+    });
 
     // Don't send response as that's done in the call every burst call above.
 }
 
-void MavlinkFtpServer::_send_burst_packet()
+// Returns true if sending is complete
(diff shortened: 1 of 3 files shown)


julianoes (approved):
Nice, thanks!

julianoes:
> FTP server is now faster, we need to increase file length because it was
sometimes already entirely received when the callback was called.

Is this a potential problem? Or just to make sure testing is proper?

chep:
> > FTP server is now faster, we need to increase file length because it was
> > sometimes already entirely received when the callback was called.
> 
> Is this a potential problem? Or just to make sure testing is proper?

Mac OS test fails without this. The file is send in only 5 or 6 packets which are already queued before the callback can change the boolean.

julianoes:
So it's a race condition that needs to be addressed, right?

chep:
> So it's a race condition that needs to be addressed, right?

What do you mean by addressed ?

chep:
```
762 [03:18:30|Debug] Drop message in ? 110 from 1/1 to 245/190 (ftp_download_file_burst.cpp:268)
763 [03:18:30|Debug] Drop message in ? 110 from 1/1 to 245/190 (ftp_download_file_burst.cpp:268)
764 [03:18:30|Debug] Drop message in ? 110 from 1/1 to 245/190 (ftp_download_file_burst.cpp:268)
765 [03:18:30|Debug] Drop message in ? 110 from 1/1 to 245/190 (ftp_download_file_burst.cpp:268)
766 [03:18:30|Debug] Download progress: 239/1000 bytes (ftp_download_file_burst.cpp:307)
767 [03:18:30|Debug] Download progress: 478/1000 bytes (ftp_download_file_burst.cpp:307)
768 [03:18:30|Debug] Download progress: 717/1000 bytes (ftp_download_file_burst.cpp:307)
769 [03:18:30|Debug] Download progress: 956/1000 bytes (ftp_download_file_burst.cpp:307)
```


Messages are checked by the intercept callback (`ftp_download_file_burst.cpp:268`) and then queued in the FTP client for its own callbacks (`ftp_download_file_burst.cpp:307`). If the FTP client callbacks are slow, many messages can be queued (4 in this run). The decision to simulate the connection problem is made in the FTP callback. Thus, potentially, the entire file can already be queued at this point. I'm uncertain if this is problematic; perhaps it's merely an ineffective method to simulate the connection break, and it should be confined to the intercept callback exclusively.

The test is successful this time, but I believe it's solely because I slowed down the processing with `LogDebug`.

chep:
When only intercept callbacks are used, the test works. If this suits you, I will clean up my branch.

julianoes:
Oh, now I understand. It's a test issue. Raising to 5000 makes sense then!

julianoes (approved):

PR #2251, opened 2024-03-20
added winch_reel plugin

jperez-droneup:
Removed winch actions not allowed by Reel winch
Changed `status_flags` to enum value, which is used to represent state on reel winch


julianoes:
Thanks @jperez-droneup! Can you also create the Proto PR, so I can review the API there first?

jperez-droneup:
> Thanks @jperez-droneup! Can you also create the Proto PR, so I can review the API there first?

https://github.com/mavlink/MAVSDK-Proto/pull/332

src/mavsdk/core/mavlink_command_sender.cpp
```diff
@@ -131,7 +131,16 @@ void MavlinkCommandSender::receive_command_ack(mavlink_message_t message)
     mavlink_command_ack_t command_ack;
     mavlink_msg_command_ack_decode(&message, &command_ack);
 
-    if ((command_ack.target_system &&
+    // Accept all broadcast command ack's
+    if(command_ack.target_component == MAV_COMP_ID_ALL)
```
julianoes, line 135: I think I can see why this one is required. Otherwise we filter out component 0.

src/mavsdk/core/mavlink_command_sender.cpp
```diff
@@ -158,6 +169,7 @@ void MavlinkCommandSender::receive_command_ack(mavlink_message_t message)
         // Currently, the gripper/winch sends the ack with source sysid/compid 1/1 instead of 1/169.
         // Until that's fixed, we ignore the component ID for any commands going to the winch.
         const bool compid_exception =
+            (work->identification.target_component_id == MAV_COMP_ID_ALL) ||
```
julianoes, line 172: I doubt this one is required because MAVSDK doesn't send out commands to component ID 0, unless you changed that.

jperez-droneup: Yeah it should not be needed then [c97a703](https://github.com/mavlink/MAVSDK/pull/2251/commits/c97a7036d101cbb0b23ed54ffb979a5b21f1d7f9)

julianoes:
Converted to draft as the PR is only here to see the diff/workaround required for the Watts reel-winch.

As discussed via email, MAVSDK can't carry workarounds for the Watts reel-winch and instead the Watts product should be updated to adhere to the MAVLink spec.

PR #2252, opened 2024-03-21
camera_server: implement zoom

julianoes:
This implements zoom for the camera server, range and continuous mode and adds an example to try it out.

Depends on https://github.com/mavlink/MAVSDK-Proto/pull/333.


julianoes (commented):
Self reviewed with fresh eyes.

PR #2254, opened 2024-03-23
[log_files] remove callback after get_entries

dakejahl:
This fixes an future/promise exception which causes a crash. This would occur if you clicked "Refresh" in the QGC Log Download menu while MAVSDK is running on another system.


julianoes (approved):

julianoes:
Thanks!

PR #2257, opened 2024-03-26
core: another workaround for winch

julianoes:
It looks like the Watts winch reel has either on component 169 or 69.

Also, it sends back acks to 255/0 instead of where it actually came from.


julianoes:
Since the Watts reel-winch is clearly not adhering to the MAVLink spec and MAVSDK can't carry workarounds just for that, I'm closing this.

PR #2260, opened 2024-03-27
Added function to get message ID and name of received Mavlink v1 messages for sniffer example

takamasanumuro:


julianoes (commented):
Nice, thanks!

Now it would be nice to have something like this generated using pymavlink somehow, so we don't have to do it manually.

takamasanumuro:
I have been thinking about it, but I got stuck. I don't know how to proceed with this. Perhaps directly parsing the common.xml file to get the message names and IDs?

julianoes:
By the way, I usually use Wireshark and the Lua MAVLink plugin for that sort of debugging.
https://mavlink.io/en/guide/wireshark.html

julianoes:
My plan is to use [libmav](https://github.com/Auterion/libmav) with v4 and have all the message information available at runtime. Until then, I'm not too thrilled to have a manual list I'm afraid.

PR #2267, opened 2024-04-01
Arm authorization

ddatsko:
This PR adds the implementation of arm authorization server functionality, and is an extension of [this PR in MAVSDK-Proto](https://github.com/mavlink/MAVSDK-Proto/pull/334). It implements a server that allows user to register a custom callback on arm authorization request. 

It also adds an example of plugins usage, which was tested with a PX4 autopilot and it's implementation of arm authorization functionality


julianoes (commented):
Thanks a lot for the contribution! This looks very clean and well done!

One thing I don't fully understand is the `system_id` passed to the subscription.

I assume the `arm_authorizer_server` is meant to filter for the correct system ID that it is responsible for?

ddatsko:
Thanks for the comment! Regarding the `system_id`, I was following the logic documented in MavLink [MAV_CMD_ARM_AUTHORIZATION_REQUEST](https://mavlink.io/en/messages/common.html#MAV_CMD_ARM_AUTHORIZATION_REQUEST). As I understand, `arm_authorizer_server` can be a separate instance, responsible for arm authorization for multiple systems with different ids. So the user callback can make a decision based on the `system_id` and its knowledge about that system (e.g. weather conditions are suitable for one system, but not for another)

julianoes:
I think you're right in terms of system id. A component server doesn't have a notion of "connected system" like a normal system plugin does, by design, so it could potentially decide for any system, and this should indeed work as you describe.

julianoes:
@JonasVautherin I'm puzzled why this is still broken: https://github.com/mavlink/MAVSDK/pull/2267

JonasVautherin:
Oh, I think the subscribe response is wrong?

See e.g. [here](https://github.com/mavlink/MAVSDK-Proto/blob/main/protos/telemetry/telemetry.proto#L16):

```
rpc SubscribePosition(SubscribePositionRequest) returns(stream PositionResponse) {}
```

vs

```
rpc SubscribeArmAuthorization(SubscribeArmAuthorizationRequest) returns(stream SubscribeArmAuthorizationResponse) { option (mavsdk.options.async_type) = ASYNC; }
```

I guess it should be

```
rpc SubscribeArmAuthorization(SubscribeArmAuthorizationRequest) returns(stream ArmAuthorizationResponse) { option (mavsdk.options.async_type) = ASYNC; }
```

i.e. `SubscribeArmAuthorizationResponse` -> `ArmAuthorizationResponse` :thinking:

julianoes:
Ah thanks. I'll check that. However, why does it not show up when I build it locally?

PR #2272, opened 2024-04-05
Implement component metadata protocol

bkueng:
This implements the [MAVLink component metadata](https://mavlink.io/en/services/component_information.html) protocol.
Details:
- Allows users to retrieve the JSON metadata. The idea is to then build on top of that to handle the specific component types (e.g. parameters or events)
- Implementation for both client + server (with system test)
- File caching based on CRC
- Support for mftp + http (if curl is enabled)
- translation is not implemented

Client example usage:
```cpp
auto component_metadata = ComponentMetadata{system.value()};
component_metadata.register_notification_callback([] (ComponentMetadata::MetadataType type, uint32_t compid, const std::string& json_data) {
	std::cout << "Got metadata: type: " << (int)type << " compid: " << compid << std::endl;
});

```

TODO:
- To support https, curl with openssl needs to be enabled (I tested this already)
- For XZ decompression, the [xz-embedded submodule](https://github.com/nobled/xz-embedded) is added. `liblzma` could be used instead.
- Are plugins required to stop ongoing mftp/http downloads when `PluginImplBase::disable` is called?

Requires https://github.com/PX4/PX4-Autopilot/pull/22980

Tested with PX4 (SITL + Pixhawk, mftp + https links)


src/mavsdk/core/locked_queue.h
```diff
@@ -13,13 +13,13 @@ template<class T> class LockedQueue {
 
     void push_back(std::shared_ptr<T> item_ptr)
     {
-        std::lock_guard<std::mutex> lock(_mutex);
+        std::lock_guard<std::recursive_mutex> lock(_mutex);
```
julianoes, line 16: I tried to avoid having to use the recursive mutex. Instead, the user callback needs to be called from the user callback thread pool, e.g.

https://github.com/mavlink/MAVSDK/blob/de90b234f6a8a7ab8aa75e3943de65f2bae32f9c/src/mavsdk/plugins/ftp/ftp_impl.cpp#L93-L95

```
_system_impl->call_user_callback(...);
```

Does that not work for your case?

src/mavsdk/core/mavlink_ftp_client.h
```diff
@@ -71,7 +71,8 @@ class MavlinkFtpClient {
         const std::string& remote_file_path,
         const std::string& local_folder,
         bool use_burst,
-        DownloadCallback callback);
+        DownloadCallback callback,
+        std::optional<uint8_t> target_compid = {});
```
julianoes, line 75: I'm usually calling `std::optional`s `maybe_`, so I would suggest `std::optional<uint8_t> maybe_target_compid = {});`

src/mavsdk/core/mavlink_ftp_client.cpp
```diff
@@ -1234,6 +1234,7 @@ void MavlinkFtpClient::timeout()
                         // In that case start requesting what we missed.
                         if (item.current_offset == item.file_size && item.missing_data.empty()) {
                             // We are done anyway.
+                            item.ofstream.close();
```
julianoes, line 1246: Nice catch!

.gitmodules
```diff
@@ -4,3 +4,6 @@
 [submodule "mavsdk-proto"]
 	path = proto
 	url = https://github.com/mavlink/MAVSDK-Proto.git
+[submodule "src/third_party/xz-embedded/submodule"]
+	path = src/third_party/xz-embedded/submodule
+	url = https://github.com/tukaani-project/xz-embedded.git
```
julianoes, line 9: I'd prefer to get the dependency in the same way like all others (except gtest):

https://github.com/mavlink/MAVSDK/blob/main/third_party/CMakeLists.txt
E.g. https://github.com/mavlink/MAVSDK/blob/main/third_party/jsoncpp/CMakeLists.txt

Which library is up to you.

src/mavsdk/core/fs_utils.cpp
```diff
@@ -0,0 +1,113 @@
...
+#include <pwd.h>
+#include <sys/types.h>
+#include <unistd.h>
+#endif
+
+#ifdef _WIN32
+#include <windows.h>
+#include <shlobj.h>
+
+static std::optional<std::filesystem::path> get_known_windows_path(REFKNOWNFOLDERID folderId)
```
julianoes, line 17: This could be within the `mavsdk` namespace, right?

src/mavsdk/core/fs_utils.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+namespace mavsdk {
+
+std::optional<std::filesystem::path> get_cache_directory()
+{
+    // Windows: %LOCALAPPDATA%
+    // Android: /data/data/<package>/cache
+    // Linux: ~/.cache
+    // macOS: ~/Library/Caches
+#if defined(_WIN32)
```
julianoes, line 52: You can use `#if defined(WINDOWS)`

src/mavsdk/core/fs_utils.cpp
```diff
@@ -0,0 +1,113 @@
...
+    std::ifstream cmdline("/proc/self/cmdline");
+    std::string line;
+    if (std::getline(cmdline, line)) {
+        // line might have a trailing \0
+        if (line.length() > 0 && *line.end() == 0) {
+            line.pop_back();
+        }
+        return "/data/data/" + line + "/mavsdk_cache";
+    }
+#elif defined(__APPLE__) || defined(LINUX)
```
julianoes, line 68: And `#elif defined(APPLE) || defined(LINUX)`.

src/mavsdk/core/fs_utils.cpp
```diff
@@ -0,0 +1,113 @@
...
+    // Windows: %LOCALAPPDATA%
+    // Android: /data/data/<package>/cache
+    // Linux: ~/.cache
+    // macOS: ~/Library/Caches
+#if defined(_WIN32)
+    auto path = get_known_windows_path(FOLDERID_LocalAppData);
+    if (path) {
+        return path->append("mavsdk_cache");
+    }
+#elif defined(__ANDROID__)
```
julianoes, line 57: And `#elif defined(ANDROID)`

src/mavsdk/plugins/component_metadata/file_cache.h
```diff
@@ -0,0 +1,84 @@
...
+
+    struct Meta {
+        uint32_t magic{0x9a9cad0e};
+        uint32_t version{0};
+        AccessCounterType access_counter{0};
+    };
+
+    struct AccessCounters {
+        AccessCounterType next_access_counter{0};
+        std::map<AccessCounterType, std::string> cached_files;
```
julianoes, line 71: Looks like it can be `std::unordered_map`.

src/mavsdk/plugins/component_metadata/file_cache_test.cpp
```diff
@@ -0,0 +1,151 @@
+
+#include <gtest/gtest.h>
+#include <fstream>
+#include "file_cache.h"
+#include "fs_utils.h"
+
+using namespace mavsdk;
+
+class FileCacheTest : public testing::Test {
```
julianoes, line 9: Amazing, I like the tests!

src/mavsdk/plugins/component_metadata/file_cache.cpp
```diff
@@ -0,0 +1,231 @@
...
+    }
+#else
+    mode_t m = umask(0);
+    _fd = open(_path.c_str(), O_RDWR | O_CREAT, 0666);
+    umask(m);
+    if (_fd < 0) {
+        LogErr() << "Cannot open file " << _path;
+    } else {
+        if (flock(_fd, LOCK_EX) == -1) {
+            LogErr() << "Cannot lock file " << strerror(errno);
```
julianoes, line 34: What happens in this case? It doesn't abort the program, right?

src/mavsdk/plugins/component_metadata/component_metadata.cpp
```diff
@@ -0,0 +1,36 @@
+// WARNING: THIS FILE IS AUTOGENERATED! As such, it should not be edited.
+// Edits need to be made to the proto files
+// (see
+// https://github.com/mavlink/MAVSDK-Proto/blob/master/protos/component_metadata/component_metadata.proto)
```
julianoes, line 4: Can you also make the pull request to MAVSDK-Proto?

src/mavsdk/plugins/component_metadata/component_metadata_impl.cpp
```diff
@@ -0,0 +1,559 @@
...
+        _system_impl->request_message().request(
+            MAVLINK_MSG_ID_COMPONENT_METADATA, component, [this](auto&& result, auto&& message) {
+                receive_component_metadata(result, message);
+            });
+    }
+}
+
+void ComponentMetadataImpl::disable()
+{
+    // TODO: stop ongoing downloads?
```
julianoes, line 87: I would think we should do so, if possible.

src/mavsdk/plugins/component_metadata/component_metadata_impl.cpp
```diff
@@ -0,0 +1,559 @@
...
+        buf,
+        sizeof(buf),
+        "compid-%03i_crc-%08x_type-%02i_trans-%i",
+        compid,
+        crc,
+        comp_info_type,
+        (int)is_translation);
+    return buf;
+}
+bool ComponentMetadataImpl::uri_is_MAVLinkFTP(
```
julianoes, line 127: Nitpick: `uri_is_mavlinkftp`.

src/mavsdk/plugins/component_metadata/component_metadata_impl.cpp
```diff
@@ -0,0 +1,559 @@
...
+                    [this, &component, local_path, compid, type, file_cache_tag](
+                        MavlinkFtpClient::ClientResult download_result,
+                        MavlinkFtpClient::ProgressData progress_data) {
+                        if (download_result == MavlinkFtpClient::ClientResult::Next) {
+                            LogDebug()
+                                << "File download progress: " << progress_data.bytes_transferred
+                                << '/' << progress_data.total_bytes;
+                            // TODO: detect slow link (e.g. telemetry), and cancel download
+                            // (fallback to http) e.g. by estimating the remaining download time,
+                            // and cancel if >40s
```
julianoes, line 204: How do you know there is a http fallback?

src/mavsdk/plugins/component_metadata_server/component_metadata_server_impl.cpp
```diff
@@ -0,0 +1,163 @@
...
+{
+    LogDebug() << "MAVLINK_MSG_ID_COMPONENT_METADATA request received";
+
+    const std::lock_guard lg{_mutex};
+
+    if (!_metadata_set) {
+        return MAV_RESULT_TEMPORARILY_REJECTED;
+    }
+
+    char general_metadata_uri[MAVLINK_MSG_COMPONENT_METADATA_FIELD_URI_LEN];
```
julianoes, line 54: Properly done :).

src/mavsdk/plugins/component_metadata_server/component_metadata_server_impl.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+    // Create tmp directory as ftp root directory
+    const auto tmp_option = create_tmp_directory("mavsdk-component-metadata-server");
+    if (!tmp_option) {
+        LogErr() << "Failed to create tmp directory";
+        return;
+    }
+    _tmp_path = *tmp_option;
+
+    LogDebug() << "Storing metadata under " << _tmp_path;
```
julianoes, line 89: By the way, I would probably add a `_debugging` bool like other plugins have it which can be enabled using `export MAVSDK_COMPONENT_METADATA_DEBUGGING=1`, and then you can make it quite verbose in that case.

src/system_tests/component_metadata.cpp
```diff
@@ -0,0 +1,220 @@
...
+                    received_events = true;
+                    break;
+                default:
+                    ASSERT_TRUE(false) << "Unexpected metadata type " << static_cast<int>(type);
+            }
+        });
+
+    for (int i = 0; i < 100 && (!received_events || !received_parameters); ++i) {
+        std::this_thread::sleep_for(std::chrono::milliseconds(100));
+    }
```
julianoes, line 217: You didn't want to do this one with `std::future` and `std::promise`?

src/mavsdk/core/mavlink_ftp_client.cpp
```diff
@@ -204,10 +204,19 @@ void MavlinkFtpClient::process_mavlink_ftp_message(const mavlink_message_t& msg)
...
-                    stop_timer();
-                    item.callback(result_from_nak(payload), {});
-                    work_queue_guard.pop_front();
+                    const ServerResult sr = static_cast<ServerResult>(payload->data[0]);
+                    // In case there's no session available, there's another transfer in progress
+                    // for the given component. Back off and try again later.
+                    if (sr == ERR_NO_SESSIONS_AVAILABLE) {
+                        payload->seq_number = 0; // Ignore this response
+                        start_timer(3.0);
+                        LogDebug() << "No session available, retrying...";
```
julianoes, line 213: That makes sense, nice.

src/mavsdk/core/mavlink_ftp_client.cpp
```diff
@@ -1177,11 +1186,11 @@ void MavlinkFtpClient::send_mavlink_ftp_message(const PayloadHeader& payload, ui
     });
 }
 
-void MavlinkFtpClient::start_timer()
+void MavlinkFtpClient::start_timer(std::optional<double> duration_s)
 {
     _system_impl.unregister_timeout_handler(_timeout_cookie);
     _system_impl.register_timeout_handler(
-        [this]() { timeout(); }, _system_impl.timeout_s(), &_timeout_cookie);
+        [this]() { timeout(); }, duration_s.value_or(_system_impl.timeout_s()), &_timeout_cookie);
```
julianoes, line 1193: Oh, I didn't know `value_or`, that's nice.

julianoes (commented):
Amazing! Thanks so much for the split commits, makes the review much easier.

I left a bunch of comments. I might do a more in-depth review once I try out some of the pieces.

What's missing is the proto pull request. I want to make sure I understand the user interface.

.gitmodules
JonasVautherin: This ^. [Do not add dependencies as submodules](https://www.acarg.ch/posts/cmake-deps/).

src/mavsdk/core/locked_queue.h
bkueng: That works too. I changed it in https://github.com/mavlink/MAVSDK/pull/2272/commits/4297106d92780945363823b3e182c1e787b59d1a

.gitmodules
bkueng: Alright, I changed to the system lib liblzma in https://github.com/mavlink/MAVSDK/pull/2272/commits/51758840389ad5f54b72eb210ec80b28f77ee802

src/mavsdk/core/fs_utils.cpp
bkueng: Done in https://github.com/mavlink/MAVSDK/pull/2272/commits/998ed155ccdcc02aff4dc236cc096e4455dedb93

src/mavsdk/core/fs_utils.cpp
bkueng: https://github.com/mavlink/MAVSDK/pull/2272/commits/998ed155ccdcc02aff4dc236cc096e4455dedb93

src/mavsdk/core/fs_utils.cpp
bkueng: https://github.com/mavlink/MAVSDK/pull/2272/commits/998ed155ccdcc02aff4dc236cc096e4455dedb93

src/mavsdk/plugins/component_metadata/file_cache.h
bkueng: No, the cache eviction depends on the order (least recently used)

src/mavsdk/plugins/component_metadata/file_cache.cpp
bkueng: No, it just continues. I thought that's ok given the chance for a race condition is small, and there is an error printed.

src/mavsdk/plugins/component_metadata/component_metadata.cpp
bkueng: Yes: https://github.com/mavlink/MAVSDK-Proto/pull/338

src/mavsdk/plugins/component_metadata/component_metadata_impl.cpp
bkueng: This is a bit more work I think. I might do it in a follow-up if you're ok.

src/mavsdk/plugins/component_metadata/component_metadata_impl.cpp
bkueng: Updated

src/mavsdk/plugins/component_metadata/component_metadata_impl.cpp
bkueng: You can check if the uri is set. It's not guaranteed, but PX4 generally always sets a http fallback uri.

src/mavsdk/plugins/component_metadata_server/component_metadata_server_impl.cpp
bkueng: Yeah not my first time :)

src/mavsdk/plugins/component_metadata_server/component_metadata_server_impl.cpp
bkueng: Done in https://github.com/mavlink/MAVSDK/pull/2272/commits/a725299afaeb5f2d30815494174226d285d7f814

bkueng:
Thanks for the review @julianoes! Here's the proto changes: https://github.com/mavlink/MAVSDK-Proto/pull/338
I'll amend the commits after things settle.

src/system_tests/component_metadata.cpp
bkueng: I thought it's not worth the extra complexity given it's a test and we need to wait for multiple things.

src/mavsdk/plugins/component_metadata/file_cache.h
julianoes: Oh, I see.

src/mavsdk/plugins/component_metadata/file_cache.cpp
julianoes: Fair enough.

src/mavsdk/plugins/component_metadata/component_metadata.cpp
julianoes: Thanks!

src/mavsdk/plugins/component_metadata_server/component_metadata_server_impl.cpp
julianoes: Very nice!

src/system_tests/component_metadata.cpp
julianoes: Right, I wonder what the canonical way is to wait for two futures.

bkueng:
CI is passing now, except for the proto submodule not being updated yet.
I also had to update the `re2` dependency (https://github.com/mavlink/MAVSDK/pull/2272/commits/c1883f47d9da545b8942b8bd37035213f05e622c). I'm not sure why that problem shows up here, but not on `main`.

src/mavsdk/core/fs_utils.h
```diff
@@ -0,0 +1,20 @@
+#pragma once
```
julianoes, line 1: We should probably merge this file with https://github.com/mavlink/MAVSDK/blob/main/src/system_tests/fs_helpers.h but it can happen after this PR.

julianoes (dismissed):
Fine to be merged once we have proto merged!

julianoes:
Thanks a lot @bkueng!

PR #2274, opened 2024-04-05
proto: forgot to update

julianoes:


julianoes:
Oh dear, this already contains the arm authorizer stuff now.

JonasVautherin:
I wanted to move everything into one repo, and I think it would work with MAVSDK-Java and MAVSDK-Python, but not MAVSDK-Swift. The reason was this: https://forums.swift.org/t/allow-package-swift-to-not-be-at-the-root-of-the-repository/60431

Still, it would work to have everything but MAVSDK-Swift in this repo. And MAVSDK-Swift/third party repos would just have to make a copy of the proto files (instead of using the git submodule). That would probably work :+1:.

julianoes:
Ah, interesting. I forgot about that.

julianoes:
Actually, we could probably just have a shim repo for Swift, and then pull in all files required and then publish whatever needs to be published, right?

JonasVautherin:
Yes, that should work :+1:

PR #2277, opened 2024-04-11
CI: add macOS arm runners and update macOS CI in general 

julianoes:
Let's see what happens.


.github/workflows/main.yml
```diff
@@ -515,7 +533,11 @@ jobs:
...
         with:
           name: mavsdk_server_macos.framework
-          path: ./build/macos/src/mavsdk_server/src/mavsdk_server.framework
+          path: ./build/macos/src/mavsdk_server/src/mavsdk_server_x64.framework
+      - uses: actions/download-artifact@v4
+        with:
+          name: mavsdk_server_macos.framework
+          path: ./build/macos/src/mavsdk_server/src/mavsdk_server_arm64.framework
       - name: Package
         run: bash ./src/mavsdk_server/tools/package_mavsdk_server_framework.bash
```
JonasVautherin, line 542: Note that this script will have to be updated if you want to include the new builds in the final framework :+1:

julianoes: Yes, do you know how? I couldn't really find the docs on how the structure should be.

JonasVautherin: I think what you did (after the comment) makes sense, but I don't really understand the other changes. Was there a need to update the dependencies and toolchain for this?

tools/ios.toolchain.cmake
```diff
@@ -151,10 +151,10 @@
...
 
 # CMake invokes the toolchain file twice during the first build, but only once during subsequent rebuilds.
-#if(DEFINED ENV{_IOS_TOOLCHAIN_HAS_RUN})
-#  return()
-#endif()
-#set(ENV{_IOS_TOOLCHAIN_HAS_RUN} true)
+if(DEFINED ENV{_IOS_TOOLCHAIN_HAS_RUN})
+  return()
+endif()
+set(ENV{_IOS_TOOLCHAIN_HAS_RUN} true)
```
JonasVautherin, line 157: Pretty sure you don't want this. The maintainer of the toolchain file doesn't remember why he added this and refuses to revert it, and it has always been a problem for us. See https://github.com/leetal/ios-cmake/pull/198 and https://github.com/leetal/ios-cmake/issues/146.

julianoes: Oh, right, thanks!

julianoes: So it didn't actually update at all, ok I'm stupid.

.github/workflows/main.yml
julianoes: I'm working on it...

julianoes:
@JonasVautherin I'm lost, any idea?
https://github.com/mavlink/MAVSDK/actions/runs/8903079132/job/24451638181?pr=2277#step:9:1604

julianoes:
I think I need to buy a Mac, just so I can throw it out the window.

julianoes:
I will have to rent a Mac in the cloud again and try it...

julianoes:
Hurray! Thanks for the help @JonasVautherin!

PR #2280, opened 2024-04-12
plugins: add log streaming

julianoes:
This adds log streaming. The stream is transmitted via a base64 string because `bytes` didn't compile using our gRPC infrastructure.

This PR also exposes base64 encoding and decoding functions for C++ users of the API. For Python, this comes with the batteries included.


julianoes:
@bkueng it would be awesome if you could give this a quick check/review. I'm basically just saving the data stream but not doing any reordering or retrying.

src/mavsdk/plugins/log_streaming/log_streaming_impl.cpp
```diff
@@ -0,0 +1,326 @@
...
+                  << std::to_string(logging_data_acked.length)
+                  << ", first message: " << std::to_string(logging_data_acked.first_message_offset)
+                  << ", sequence: " << logging_data_acked.sequence;
+    }
+
+    if (logging_data_acked.first_message_offset == std::numeric_limits<uint8_t>::max()) {
+        _ulog_data.insert(
+            _ulog_data.end(),
+            logging_data_acked.data,
+            logging_data_acked.data + logging_data_acked.length);
```
bkueng, line 220: Just to be on the safe side, you could add a check if `logging_data_acked.length` exceeds the `logging_data_acked.data` size.

src/mavsdk/plugins/log_streaming/log_streaming_impl.cpp
```diff
@@ -0,0 +1,326 @@
...
+    } else {
+        if (logging_data_acked.first_message_offset > 0) {
+            _ulog_data.insert(
+                _ulog_data.end(),
+                logging_data_acked.data,
+                logging_data_acked.data + logging_data_acked.first_message_offset);
+            process_message();
+            _ulog_data.clear();
```
bkueng, line 228: You can call this even if `logging_data_acked.first_message_offset == 0`, which means less data drops in case the next one is lost.

bkueng (commented):
Looks good so far. I would also have a look at the mavlink-router implementation, as it has additional timeout handling.
You might already have thought about it, but also make sure to clear the buffer and `_maybe_current_sequence` after stopping (or when starting).

src/mavsdk/plugins/log_streaming/log_streaming_impl.cpp
julianoes: Good catch. I'll add that. So that's basically for sneaky values between 250 and 254.

src/mavsdk/plugins/log_streaming/log_streaming_impl.cpp
julianoes: Ok, so if we only have part of a message, we would add it anyway?

julianoes (commented):
Thanks @bkueng. I think I have addressed it all, minus the timeout that you suggested.

bkueng:
> Is the timeout to basically restart logging if we haven't received data in a bit?

Yes exactly. It's not strictly required.

julianoes:
> Yes exactly. It's not strictly required.

We'll see if I have to add it once I'm testing it.

julianoes:
Forgot to update proto submodule, duh.

bkueng:
> I'm basically just saving the data stream but not doing any reordering or retrying.

Just keep in mind that without considering the sequence number, a message drop will likely lead to a corrupt ulog stream. Which is very likely going to happen if going through a serial link.

julianoes:
Aha, so what do you do on drop? Remove what's currently in the buffer to be written and start fresh?

bkueng:
Yes, that's what `first_message_offset` is used for. If data is dropped, it tells you when there is a valid message start again.

julianoes:
Ok thanks. I'll try to add that.

julianoes:
Coming with https://github.com/mavlink/MAVSDK/pull/2290.

PR #2290, opened 2024-04-30
log_streaming: make it robust to message drops

julianoes:
When a message is dropped, we need to discard any message part. Otherwise, we end up having corrrupt ulog files.

### DIFF
--- a/src/mavsdk/plugins/log_streaming/log_streaming_impl.cpp
+++ b/src/mavsdk/plugins/log_streaming/log_streaming_impl.cpp
@@ -178,28 +178,45 @@ void LogStreamingImpl::process_logging_data(const mavlink_message_t& message)
     }
 
     std::lock_guard<std::mutex> lock(_mutex);
-    check_sequence(logging_data.sequence);
+    auto drop_state = check_sequence(logging_data.sequence);
 
-    if (logging_data.first_message_offset == std::numeric_limits<uint8_t>::max()) {
-        _ulog_data.insert(
-            _ulog_data.end(), logging_data.data, logging_data.data + logging_data.length);
-    } else {
-        if (logging_data.first_message_offset > sizeof(logging_data.data)) {
-            LogWarn() << "Invalid first_message_offset";
-            return;
-        }
+    switch (drop_state) {
+        case DropState::Ok:
+            if (logging_data.first_message_offset == std::numeric_limits<uint8_t>::max()) {
+                _ulog_data.insert(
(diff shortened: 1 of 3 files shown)


julianoes:
@bkueng would be great if you could check I got this right.

julianoes:
I believe this is better than what's in main, so I'm merging it now before the next release.

src/mavsdk/plugins/log_streaming/log_streaming_impl.cpp
```diff
@@ -254,46 +271,62 @@ void LogStreamingImpl::process_logging_data_acked(const mavlink_message_t& messa
...
+                    logging_data_acked.data + logging_data_acked.length);
+            }
+            break;
+
+        case DropState::Dropped:
+            _ulog_data.clear();
+            _ulog_data.insert(
+                _ulog_data.end(),
+                logging_data_acked.data + logging_data_acked.first_message_offset,
+                logging_data_acked.data + logging_data_acked.length);
```
bkueng, line 308: Here you must only insert if there's a valid offset: `logging_data_acked.first_message_offset != std::numeric_limits<uint8_t>::max()`

src/mavsdk/plugins/log_streaming/log_streaming_impl.cpp
```diff
@@ -254,46 +271,62 @@ void LogStreamingImpl::process_logging_data_acked(const mavlink_message_t& messa
...
-        _ulog_data.insert(
-            _ulog_data.end(),
-            logging_data_acked.data + logging_data_acked.first_message_offset,
-            logging_data_acked.data + logging_data_acked.length);
+    auto drop_state = check_sequence(logging_data_acked.sequence);
+
+    switch (drop_state) {
+        case DropState::Ok:
+            if (logging_data_acked.first_message_offset == std::numeric_limits<uint8_t>::max()) {
+                _ulog_data.insert(
```
bkueng, line 279: After a dropout you must first wait for a valid `logging_data_acked.first_message_offset` before inserting in `_ulog_data`

src/mavsdk/plugins/log_streaming/log_streaming_impl.cpp
julianoes: Oh, right!

src/mavsdk/plugins/log_streaming/log_streaming_impl.cpp
julianoes: I see. So I need to keep track of the previous state.

PR #2291, opened 2024-04-30
Fix warnings

julianoes:
Not sure why CI didn't highlight these warnings @bkueng.

PR #2297, opened 2024-05-08
Fix warnings (take two)

julianoes:
Second attempt at this, replacing #2291.


julianoes:
Note to self: I need to figure out why the hell gtest doesn't have the same static vs. dynamic build flags and thus fails linking.

PR #2298, opened 2024-05-10
gimbal: mutex fixes and added env variable for v2

julianoes:
- Make sure that _gimbal_protocol is also protected by the mutex.
- Add MAVSDK_FORCE_GIMBAL_V2 to force usage of the gimbal v2 protocol and prevent the fallback to gimbal v1. This helps when we don't immediately detect all the components.
- Keep trying to request the GIMBAL_MANAGER_INFORMATION message.

With MAVSDK v3, I want to remove support for gimbal v1, it's just a pain.

### DIFF
--- a/src/mavsdk/plugins/gimbal/gimbal_impl.cpp
+++ b/src/mavsdk/plugins/gimbal/gimbal_impl.cpp
@@ -44,14 +44,17 @@ void GimbalImpl::deinit()
 
 void GimbalImpl::enable()
 {
-    _system_impl->register_timeout_handler(
-        [this]() { receive_protocol_timeout(); }, 1.0, &_protocol_cookie);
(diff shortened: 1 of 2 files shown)


pperpich (approved):

JonasVautherin (commented):
Instead of adding an env variable, why not just defaulting to V2 in case of timeout? I guess your problem now is that sometimes the init times out and it defaults to V1, right? But if V1 is presumably not used by anyone anymore, then V2 seems like a better default, and it would move the problem to those who use V1. What do you think?

julianoes:
@JonasVautherin that's a good idea. It's technically a break though as it changes the behavior somewhat unexpectedly.

JonasVautherin:
Oh, good point :sweat_smile:.

PR #2303, opened 2024-05-17
log_streaming: refactor/fixup drop logic

julianoes:
I think I finally understand how this is supposed to work, and my testing seems to pass, phew.

Thanks to @bkueng for the reviews! I hope this one will be easiest one.


bkueng (approved):
Looks correct now.
(only reordering isn't handled)

julianoes:
Thanks @bkueng!

PR #2304, opened 2024-05-17
add arm64-lts docker workflow

AlexKlimaj:
This is an attempt to add an arm64-lts target which is needed on Jetson Jetpack 5.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -235,6 +235,9 @@ jobs:
     strategy:
       matrix:
         arch_name: [armv6, armv7, arm64, arm64-lts]
+        include:
+          - arch_name: arm64-lts
+            rename_distro: true
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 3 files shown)


JonasVautherin:
Let's see what the CI says :blush:

julianoes:
I usually run this docker script manually and do the push, so that would explain the failure here.

julianoes:
Built and pushed. Now I'm re-running the failed jobs.

@AlexKlimaj have you thought about the naming of the LTS one? Should it be called debian 11 or something instead? We'll have to see what happens I suppose.

AlexKlimaj:
Hm I don't know what the best name for it should be.

julianoes:
It looks like right now it is `debian12-arm-lts` but that doesn't make much sense, I'd say we rename it to debian11 manually because it should be compatible with that.

julianoes:
Here to test: https://github.com/mavlink/MAVSDK/releases/download/v2.11.0/libmavsdk-dev_2.11.0_debian11_arm64-lts.deb

AlexKlimaj:
Hm, installed the debian11 release, our logloader code builds. But then the program segfaults when trying to run it.

```jetson@jetson:~/code/logloader$ ./build/logloader
```

julianoes:
Check with gdb where it segfaults. I bet it's `std::filesystem::path`.

Issue #2215, opened 2024-02-05
Race condition in assigning callbacklist handle


devbharat:
There could be a race condition in assigning handle [here](https://github.com/mavlink/MAVSDK/blob/3131cc9b34b04d6d7affb09d3ad2fdbf81c677af/src/mavsdk/core/callback_list_impl.h#L23)

julianoes:
Hmm. What if we extend the mutex around that part?

Would nested subscriptions work using `std::recursive_mutex` and would that be safe?

julianoes:
I like the deferred subscription and removal! It would be great if you can make a pull request for it.

julianoes:
@devbharat I think this would still be great :)

devbharat:
ah damn i forgot

julianoes:
That was quick, thanks!

julianoes:
Fixed in #2307, thanks @devbharat!

PR #2307, opened 2024-05-20
Add conditional callback, use atomic handle_id and add some tests

devbharat:
Fixes https://github.com/mavlink/MAVSDK/issues/2215

Adds
 - Conditional callbacks for one-shot callback where you want unsubscription to happen automatically without dealing with the Handle object
 - Atomic _last_id to make sure when subscribing in burst from many threads at once each id is unique
 - Deferred subscriptions so you can also add subscriptions within the callback
 - Some tests

I needed to add these to handle to create something similar to the MavlinkMessageHandler class that uses a map of msgid to callbacklists, but crucially it allows adding subscriptions from within callbacks (if you get this mavlink message, send this message and subscribe to this mavlink message kind of chained workflow), which meant (in addiiton to deferred subs in cblist) the execution of the callbacklist from MavlinkMessageHandler needed to be outside the lock (comparison [here](https://gist.github.com/devbharat/17a11229bd7df818f9a7fd65828f4ddd)).


src/mavsdk/core/callback_list_test.cpp
```diff
@@ -98,3 +136,147 @@ TEST(CallbackList, UnsubscribeAllWithNullptr)
...
+        threads.emplace_back([&]() {
+            // Add a burst of subscriptions in a multithreaded env all at once
+            auto handle_ptr = std::make_shared<std::optional<Handle<int, double>>>();
+            addSubCallback(handle_ptr); // this will get deferred sub
+
+            // safely put them in a vec for later comparison
+            std::lock_guard<std::mutex> guard(mutex_);
+            if (handle_ptr && handle_ptr->has_value()) {
+                if (!handle_set.insert(handle_ptr->value()).second) {
+                    EXPECT_TRUE(false);  // handles are not unique
```
devbharat, line 270: To make the check fail on main i had to increase the thread count to over 100k and in subscribe() call comment out the check_removals() - not sure why, maybe it was causing some synchronization as a side effect on the rest of the code. In that setting, i could reproduce the failures on main and not on this PR. Another way I checked it was adding a method to pull the _last_id from the callbacklist class, then comparing it with the threadcount, on main (under the conditions mentioned above) it was usually less then the thread_count and not on this PR.

julianoes: Wow! :exploding_head:

src/mavsdk/core/callback_list_impl.h
```diff
@@ -16,15 +17,25 @@ template<typename... Args> class CallbackListImpl {
     Handle<Args...> subscribe(const std::function<void(Args...)>& callback)
     {
         check_removals();
+        process_subscriptions();
 
         // We need to return a handle, even if the callback is nullptr to
         // unsubscribe. That's fine, the handle just won't remove anything
         // when/if used later.
-        auto handle = Handle<Args...>(_last_id++);
+        auto handleId = _last_id.fetch_add(1, std::memory_order_relaxed);
```
julianoes, line 25: How did this work before without an atomic id?

julianoes (approved):
Amazing, thanks for that!

julianoes:
I wonder if/how I can re-use this for timeout and call every subscription. I basically have similar problems there.

src/mavsdk/core/callback_list_impl.h
devbharat: I think the issue only happens if you add a ton of subscriptions from different threads simultaneously - kind of a rare scenario, so we didn’t see it earlier.

devbharat:
> I wonder if/how I can re-use this for timeout and call every subscription. I basically have similar problems there.

I haven’t seen it for a while, wouldn’t it make sense that all of those also use this callbacklist class? don’t they, I guess they were implemented before we added this class.

julianoes:
I think it might make sense, indeed. They also use the stupid `void *` as the cookie instead of a uint64_t handle which is safer.

julianoes:
I will do that in the future, don't want to hold up this PR.

julianoes:
@devbharat do you mind fixing style?

julianoes:
Oh, thought I had merged it already. Funny. Anyway, it's all good. SonarCloud is just a bit picky.

PR #2311, opened 2024-05-23
mission_raw: fix upload race condition

julianoes:
This fixes the case where a rally points upload right after a geofence upload fails with busy. That's because the work item still exists and hasn't been cleaned up yet.

However, we can ignore that case if the work item has been done.


julianoes:
Fixes https://github.com/mavlink/MAVSDK-Python/issues/692.

JonasVautherin (approved):

PR #2313, opened 2024-05-27
camera: support camera def files via MAVLink FTP

julianoes:
Instead of only supporting camera definition files to be downloaded using cURL, we now also support getting them using MAVLink FTP.

### DIFF
--- a/src/mavsdk/plugins/camera/camera_impl.cpp
+++ b/src/mavsdk/plugins/camera/camera_impl.cpp
@@ -5,10 +5,13 @@
 #include "http_loader.h"
 #include "unused.h"
 #include "callback_list.tpp"
+#include "fs_utils.h"
 
 #include <algorithm>
-#include <functional>
 #include <cmath>
+#include <filesystem>
+#include <fstream>
+#include <functional>
 #include <sstream>
 
 namespace mavsdk {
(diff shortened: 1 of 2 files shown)


src/mavsdk/plugins/camera/camera_impl.cpp
```diff
@@ -1557,17 +1560,34 @@ bool CameraImpl::should_fetch_camera_definition(const std::string& uri) const
...
-
-    return result;
+    auto& uri = camera_information.cam_definition_uri;
+
+    if (starts_with(uri, "http://") || starts_with(uri, "https://")) {
+        LogInfo() << "Download file: " << uri << " using cURL...";
+        auto result = download_definition_file_curl(uri, camera_definition_out);
+        LogInfo() << "Downloaded file, result " << result;
+        return result;
+    } else if (starts_with(uri, "mftp://") || starts_with(uri, "mavlinkftp://")) {
```
JonasVautherin, line 1578: Is `mftp` a thing? I thought it was `mavftp` :grin:. But then I have never seen a drone exposing the camera definition over mavlink ftp, so... :grin:

src/mavsdk/plugins/camera/camera_impl.cpp
```diff
@@ -1557,17 +1560,34 @@ bool CameraImpl::should_fetch_camera_definition(const std::string& uri) const
            !_has_camera_definition_timed_out;
 }
 
+bool CameraImpl::starts_with(const std::string& str, const std::string& prefix)
+{
+    return str.compare(0, prefix.size(), prefix) == 0;
```
JonasVautherin, line 1565: Could [`string::starts_with`](https://en.cppreference.com/w/cpp/string/basic_string/starts_with) be used instead?

src/mavsdk/plugins/camera/camera_impl.cpp
```diff
@@ -1585,6 +1605,77 @@ CameraImpl::download_definition_file(const std::string& uri, std::string& camera
...
+{
+    const std::string prefix1 = "mftp://";
+    const std::string prefix2 = "mavlinkftp://";
+
+    if (str.compare(0, prefix1.size(), prefix1) == 0) {
+        return str.substr(prefix1.size());
+    }
+    if (str.compare(0, prefix2.size(), prefix2) == 0) {
+        return str.substr(prefix2.size());
+    }
```
JonasVautherin, line 1627: If `starts_with` can be used, this would be more readable IMO

JonasVautherin (approved):

src/mavsdk/plugins/camera/camera_impl.cpp
julianoes: Yes, `mftp` is a thing:
https://github.com/mavlink/MAVSDK/blob/68c8041bc911e31f99c8575a042d6b3919583b39/src/mavsdk/plugins/component_metadata/component_metadata_test.cpp#L15

And so is `mavlinkftp` according to [CAMERA_INFORMATION](https://mavlink.io/en/messages/common.html#CAMERA_INFORMATION).

Not sure about `mavftp`, although that would be the nicest one...

src/mavsdk/plugins/camera/camera_impl.cpp
julianoes: It's only in C++20 but we're only on C++17 so far, unfortunately.

PR #2325, opened 2024-06-02
mission_raw: implement mission changed based on opaque ID

### LINKED ISSUE #2324: Support opaque mission ID
This switches the subscription for mission current to the opaque mission ID as specced in https://github.com/mavlink/mavlink/pull/2012.

julianoes:
This uses the new opaque mission ID to check if a mission has changed.

The implementation is backwards compatible. If the mission_id is 0, we
just use the previous way of checking for MISSION_ACK.

Spec: https://github.com/mavlink/mavlink/pull/2012

Closes #2324.


hamishwillee:
Nice one. YOu're busy!

PR #2342, opened 2024-06-23
Change connection CLI, add TCP server connections

### LINKED ISSUE #2315: Switch to clearer API for UDP and TCP
The way that `udp://:port` means "listen" and `udp://IP:port` means "send to IP" often causes confusion and frustration.

I suggest we adopt [the way it's done in pymavlink](https://mavlink.io/en/mavgen_python/#connection_string).

We can leave the current way to be backwards compatible.

Opinions @JonasVautherin ?

julianoes:
This changes the connection CLI:

- `udp://IP:PORT` to `udpout://IP:PORT` 
- `udp://:PORT` to `udpin://0.0.0.0:PORT`

which should make the usage easier.

Also, the baudrate for serial connection is now mandatory, to avoid confusing implicit defaults.

Additionally, this PR implements TCP as a server and not just client.

Closes #2315.

### DIFF
--- a/examples/ftp_server/ftp_server.cpp
+++ b/examples/ftp_server/ftp_server.cpp
@@ -15,21 +15,18 @@ using namespace mavsdk;
 
 void usage(const std::string& bin_name)
 {
-    std::cerr << "Usage : " << bin_name << " <remote_ip> <remote_port> <root_dir>\n"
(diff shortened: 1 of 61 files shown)


julianoes:
@JonasVautherin any objections?

PR #2343, opened 2024-06-25
Improve the CMake organization

JonasVautherin:
@julianoes this is what we discussed recently, related to [this](https://www.acarg.ch/posts/cmake-deps/) and [this](https://www.acarg.ch/posts/cmake-superbuild/). Note that I haven't checked the CI at all yet, I just wanted your opinion first.

This keeps a superbuild, but I am not a big fan of it. As you can see, it really just forwards options, all for the sake of running cmake only once, e.g. with:

```cmake
cmake -DCMAKE_PREFIX_PATH=$(pwd)/build/install -DCMAKE_INSTALL_PREFIX=build/install -DBUILD_MAVSDK_SERVER=ON -Bbuild -S.
cmake --build build
```

An issue with the superbuild being that the "non-superbuild" way tends to rot. Right now I don't think it works to build MAVSDK from `src/`, though it used to work.

Without the superbuild, the user has to explicitly decide to build the dependencies, and it's really just two more lines, for instance:

```
# Build the dependencies
cmake -DCMAKE_PREFIX_PATH=$(pwd)/third_party/build/install -DCMAKE_INSTALL_PREFIX=third_party/build/install -Bthird_party/build -Sthird_party
# Now build MAVSDK, pointing CMAKE_PREFIX_PATH to where the dependencies were installed
cmake -DCMAKE_PREFIX_PATH=$(pwd)/third_party/build/install -DCMAKE_INSTALL_PREFIX=build/install -DBUILD_MAVSDK_SERVER=ON -Bbuild -Ssrc
cmake --build build
```

The nice thing being that once the user understands how CMAKE_PREFIX_PATH and CMAKE_INSTALL_PREFIX work, it's easy for them to build the dependencies once and install them locally. So that a clean build of MAVSDK does not delete the dependencies. I think it's very rare that the dependencies need to be rebuilt, except for protobuf and gRPC (everytime MAVSDK updates them) because we version the generated code. A next step to make the build more robust would be to not version the generated code and let the user run the generation themselves (which would make it easier to use system packages, for instance). But that's another discussion :grin:.

I don't have a strong opinion. We already discussed it a couple years ago and you were not really into it. I just wanted to show it since we had another discussion last week :blush:. Fine for me if you prefer keeping the current way :+1:.


julianoes:
Thanks @JonasVautherin for the PR! I think I generally like it. The risk of breaking the non-superbuild is just a matter of having it in CI. As you say, we already have that now.

> ```
> cmake -DCMAKE_PREFIX_PATH=$(pwd)/build/install -DCMAKE_INSTALL_PREFIX=build/install -DBUILD_MAVSDK_SERVER=ON -Bbuild -S.
>```

So, with this. Can you still install the "final" artifacts to e.g. `/usr/local`?

JonasVautherin:
> So, with this. Can you still install the "final" artifacts to e.g. /usr/local?

If you don't set `CMAKE_INSTALL_PREFIX`, it will install on the system by default. But in the current state, it will install all the dependencies as well, which is almost certainly a bad idea. I would argue that in order to install something on the system, one should make a proper package (such that it can be uninstalled as well). But then it implies making a proper package out of MAVLink :see_no_evil:.

I can probably change this PR in order to make it behave *almost* like the current `main`. But I don't know if it's worth it: it will be slightly different (so some people will be confused and I guess it may break someone's flow) and I am not convinced that it brings anything to the superbuild :thinking:. I believe there is value in handling dependencies separately because they can be cached (or just installed with the system package manager, which may require re-generating the code with protoc). But if nobody uses that, maybe it's not worth the change.

What do you think? Should I try to imitate the current superbuild with this scheme, just to have an idea? I.e. get to a point where the CI passes without requiring any change?

julianoes:
> But in the current state, it will install all the dependencies as well, which is almost certainly a bad idea.

Right, I have to think about that. It would be in `/usr/local`, so at least it doesn't mess with system dependencies.

One thing that we need to keep compatible is to create the library and install it with .deb. However, that one uses the system dependencies and not the superbuild, so I guess it doesn't matter.

JonasVautherin:
Hmm I don't think it works well with the superbuild. I tried it in a separate, smaller project, and it was annoying to mix the dependencies and the main project (not sure if it is worse than the current way, but the current way is known, at least).

Really what works well for me is to do without the superbuild: have the helpers for the dependencies, and use the main project with CMAKE_PREFIX_PATH.

I'll close this, I don't think it will improve the current superbuild.

julianoes:
Yer, I think let's leave it for now. Thanks for trying it though!

PR #2345, opened 2024-07-03
Fix Gazebo simulation connection issue with port adjustments 

serkanMzlm:
- Port number adjusted from 14551 to 14550 for Gazebo simulation connection.
- `agrc` value checked
- Added autonomous flight and landing feature 
- Integrated functionality to autonomously take off, fly for 20 seconds, and land the drone using MAVSDK. The program monitors altitude during flight and waits for the drone to land successfully before exiting.

### DIFF
--- a/examples/autopilot_server/autopilot_server.cpp
+++ b/examples/autopilot_server/autopilot_server.cpp
@@ -59,7 +59,7 @@ int main(int argc, char** argv)
     // thread as a ground station.
 
     if (argc < 2) {
-        std::cerr << "Too few input arguments. usage: $ arm_authorizer_server <connection_url>"
+        std::cerr << "Too few input arguments. usage: argv[0]"
                   << std::endl;
         return 1;
     }
--- a/examples/autopilot_server/autopilot_server.cpp
+++ b/examples/autopilot_server/autopilot_server.cpp
@@ -15,7 +15,7 @@
 
 /*
  This example runs a MAVLink "autopilot" utilising the MAVSDK server plugins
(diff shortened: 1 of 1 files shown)


examples/autopilot_server/autopilot_server.cpp
```diff
@@ -154,7 +161,7 @@ int main(int argc, char** argv)
     // to communicate with the autopilot server plugins.
     mavsdk::Mavsdk mavsdk{Mavsdk::Configuration{Mavsdk::ComponentType::GroundStation}};
 
-    auto result = mavsdk.add_any_connection("udp://:14551");
+    auto result = mavsdk.add_any_connection(argv[1]);
```
julianoes, line 164: So what do you use to run the example? I'm a bit confused in terms of connections. And what do you mean when you say "Gazebo simulatoin connection issue"? How are you connecting this to Gazebo?

julianoes (changes requested):

examples/autopilot_server/autopilot_server.cpp
serkanMzlm: When running PX4 on Gazebo using the command make px4_sitl gazebo_x500, it did not connect because it was using port 14551. Changing it to port 14550 allowed the connection to succeed. `argv[1]  = udp://:14540`

julianoes: But this is not supposed to connect to that case. It is supposed to simulate an autopilot by itself. If you run PX4 Gazebo then you already have an autopilot, and you shouldn't have to use the MAVSDK one. That's why I'm confused.

PR #2348, opened 2024-07-12
Update Readme and fix CI

julianoes:
CI broke itself without any changes from our side, fun fun fun.


third_party/absl/CMakeLists.txt
```diff
@@ -11,6 +11,11 @@ list(APPEND CMAKE_ARGS
     "-DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}"
     "-DCMAKE_CXX_STANDARD=17"
     "-DABSL_PROPAGATE_CXX_STD=ON"
+    "-DABSL_BUILD_TESTING=OFF"
+    "-DBUILD_TESTING=OFF"
+    "-DABSL_BUILD_TEST_HELPERS=ON"
+    "-DABSL_USE_EXTERNAL_GOOGLETEST=ON"
+    "-DABSL_FIND_GOOGLETEST=ON"
```
JonasVautherin, line 18: Hmm shouldn't that be set to OFF, if BUILD_TESTING is set to OFF? It seems like the CI is failing because gtest is not found 🤔. Otherwise you need to move [gtest](https://github.com/mavlink/MAVSDK/blob/main/third_party/CMakeLists.txt#L32) up the list, but it feels unnecessary here.

julianoes: Yes, I agree. And I tried that.

julianoes: Good hint on moving it up though. I'll try that.

JonasVautherin:
I think that you need both BUILD_TESTING=OFF and ABSL_BUILD_TESTING=OFF, as per [this](https://github.com/abseil/abseil-cpp/blob/master/CMake/AbseilHelpers.cmake#L390).

julianoes:
@JonasVautherin that's what I have.

JonasVautherin:
[This](https://github.com/mavlink/MAVSDK/pull/2350/) looks promising. Something has been fixed in absl's master that is not yet released (this `TESTONLY` thingy).

julianoes:
Let's use #2350 instead.

Issue #2354, opened 2024-07-21
Since v2.10.1 notify_on_discover() invoked every second after a system has already been discovered


fibonacci-matrix:
This method should only be called once, so it doesn't work as intended right now

julianoes:
Oh, thanks for the issue. That sounds indeed wrong.

PR #2356, opened 2024-07-23
core: Fixed unnecessary repeated invocations of `notify_on_discover`

fibonacci-matrix:
This problem introduced in v2.10.1

Now it works only if there was a change in the connection status

closes: #2354


JonasVautherin (approved):

JonasVautherin (approved):
Looks fine to me!

julianoes (approved):
Yep, that looks about right! Thanks for fixing this @fibonacci-matrix!

julianoes:
I will have to create a 2.10 branch and backport it there, as main is now already on the way towards v3.

Issue #2353, opened 2024-07-19
TCP connection from MAVSDK breaks app after 16 minutes of idle time due to reconnection


fibonacci-matrix:
I'm developing an app that uses MAVSDK to connect to a device via the built-in MAVSDK TCP connection. MAVSDK has an excellent TCP reconnection mechanism that fits my requirements well, namely the possibility of temporary disconnection. In this case timer is started, which checks every second whether the connection has been restored. But after 16 minutes, this mechanism breaks my app with the message **"socket errorToo many open files"**. 

I did some basic research and the problem is that the MAVSDK in a loop reopens the sockets, but does not close them with the `close` function in case of unsuccessful execution of the `connect` function. 

```
// from tcp_connection.cpp

_socket_fd = socket(AF_INET, SOCK_STREAM, 0);
```

I have tested basic fix on Linux, but I don't know if it will work on other operating systems

JonasVautherin:
Would you like to open a PR with your proposed fix? :blush:

julianoes:
Good catch. Yes a PR would be great.

fibonacci-matrix:
I can do a PR in a few days, but I don't mind if someone else does it. Besides, https://github.com/mavlink/MAVSDK/pull/2342 will conflict with my changes

julianoes:
#2342 is merged.

PR #2357, opened 2024-07-23
core: Correctly close sockets

fibonacci-matrix:
Closes: #2353


src/mavsdk/core/tcp_client_connection.cpp
```diff
@@ -212,7 +199,7 @@ void TcpClientConnection::receive()
         }
 
         if (recv_len < 0) {
-            // This happens on desctruction when close(_socket_fd) is called,
+            // This happens on desctruction when close(_socket_fd.get()) is called,
```
JonasVautherin, line 202:
```suggestion
            // This happens on destruction when close(_socket_fd.get()) is called,
```

julianoes (approved):
Love it, thank you! Keep them coming :smile:

julianoes:
Windows CI isn't happy somehow: https://github.com/mavlink/MAVSDK/actions/runs/10061001774/job/27817265595?pr=2357#step:7:55

src/mavsdk/core/socket_holder.h
```diff
@@ -0,0 +1,38 @@
...
+    ~SocketHolder() noexcept;
+
+    void reset(int fd) noexcept;
+    void close() noexcept;
+
+    bool empty() const noexcept;
+    int get() const noexcept;
+
+#if defined(WINDOWS)
+    static constexpr int invalid_socket_fd = INVALID_SOCKET;
```
julianoes, line 26: Presumably to fix the Windows warning? :shrug: 

```suggestion
    static constexpr int invalid_socket_fd = static_cast<int>(INVALID_SOCKET);
```

fibonacci-matrix:
It turns out that on 64-bit Windows the socket handle is also 64-bit, unlike POSIX systems where it is always `int`. I will work on it. 

For more details see https://stackoverflow.com/questions/1953639/is-it-safe-to-cast-socket-to-int-under-win64

fibonacci-matrix:
@julianoes , @JonasVautherin I don't have much experience with Windows, could you please check if these lines are necessary in the newly created `.cpp` file?
```
#ifdef WINDOWS
#ifndef MINGW
#pragma comment(lib, "Ws2_32.lib") // Without this, Ws2_32.lib is not included in static library.
#endif
#else
```

julianoes:
@fibonacci-matrix I don't know either. If someone that knows Windows complains, we can fix it.

julianoes:
Should be good to be merged, right @fibonacci-matrix?

fibonacci-matrix:
> Should be good to be merged, right @fibonacci-matrix?

Yes of course, @julianoes

PR #2364, opened 2024-08-01
Fix Follow Me example

KarthiAru:
Fix Follow Me example by explicitly setting absolute_altitude_m in target_location.


julianoes (approved):
Oh thanks!

PR #2366, opened 2024-08-05
CI: fix macOS

julianoes:
macOS tooling shifting from under you, it doesn't get much worse than that.
:angry:


julianoes:
Thanks, so now we're chasing two problems. Great.

julianoes:
Don't care about Sonar for this one.

src/mavsdk_server/tools/package_mavsdk_server_framework.bash
```diff
@@ -37,6 +40,12 @@ xcrun codesign --verbose --sign - ${IOS_SIM_FAT_BACKEND_DIR}/mavsdk_server.frame
 echo "Running lipo for macos"
 mkdir -p ${MACOS_FAT_BACKEND_DIR}
 cp -a ${MACOS_X64_BACKEND_DIR}/mavsdk_server.framework ${MACOS_FAT_BACKEND_DIR}
+# Fix up symlinks that seem to break when github artifacts unzips the archives
```
JonasVautherin, line 43: Oh, I think I have seen this in the past... but I can't remember where (there are no other symlinks in this script, so probably somewhere else :thinking:).

Good catch!

PR #2368, opened 2024-08-09
examples: fix manual_control for loop

MaEtUgR:
Just saw this while glancing over trying to understand what's necessary to support MANUAL_CONTROL.aux fields, counterpart to https://github.com/PX4/PX4-Autopilot/pull/22108


JonasVautherin (approved):
Thanks :blush:

julianoes (approved):
Oh dear! Looks like I stuffed this up! Thanks for the fix @MaEtUgR!

PR #2369, opened 2024-08-10
core: don't ignore radio messages and system

### LINKED ISSUE #2302: Why MAVSDK doesnt create a new system for telemetry radios?
Seems very arbitrary, for my application I want to show the radio stats and was wondering why system 3 (wfb-ng mavlink stream stats) wasnt being detected until I found an if statement that seems to just drop these packets.

The if block is in lines 384 to 389 in src/mavsdk/core/mavsdk_impl.cpp.

julianoes:
It is confusing that we just ignore the system for the telemetry radio. We should instead treat it as a proper system. Users should be able to filter for it properly, especially given they now have the nice `wait_for_autopilot` method.

Closes #2302.

### DIFF
--- a/src/mavsdk/core/mavsdk_impl.cpp
+++ b/src/mavsdk/core/mavsdk_impl.cpp
@@ -383,13 +383,6 @@ void MavsdkImpl::receive_message(mavlink_message_t& message, Connection* connect
         }
     }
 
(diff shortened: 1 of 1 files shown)


JonasVautherin (approved):
Agreed :+1:

PR #2370, opened 2024-08-10
telemetry: remove various hacks/workarounds

### LINKED ISSUE #2318: Remove old calibration params
Remove param fallbacks like this one:
https://github.com/mavlink/MAVSDK/blob/68c8041bc911e31f99c8575a042d6b3919583b39/src/mavsdk/plugins/telemetry/telemetry_impl.cpp#L1785-L1794

julianoes:
This removes some of the hacks which were required in the past:
- Calibration health using params as this should be covered by system status now.
- Position validity based on system status instead of GPS magic.
- HITL workarounds that should also be taken care of autopilot internal.

Closes #2318

### DIFF
--- a/src/mavsdk/plugins/telemetry/telemetry_impl.cpp
+++ b/src/mavsdk/plugins/telemetry/telemetry_impl.cpp
@@ -173,9 +173,6 @@ void TelemetryImpl::init()
         [this](const mavlink_message_t& message) { process_altitude(message); },
         this);
 
-    _system_impl->register_param_changed_handler(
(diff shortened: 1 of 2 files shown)


JonasVautherin (approved):
Oh yeah! Those were a bit of an inconvenience!

Does this change require a minimum version of PX4/Ardupilot?

julianoes:
> Does this change require a minimum version of PX4/Ardupilot?

Presumably, no idea...

JonasVautherin:
Fine for me anyway, I was just curious!

PR #2373, opened 2024-08-13
Move to Ubuntu 24.04

julianoes:
This moves from clang-format-12 to clang-format-18 which comes with Ubuntu 24.04.

Also, let's try to add Ubuntu 24.04 to CI as well while I'm at it.


JonasVautherin (approved):

julianoes:
Oh wow, it passed, what a surprise!

PR #2381, opened 2024-08-20
Switch to new REQUEST_MESSAGE but fall back to deprecated request message commands if required

### LINKED ISSUE #1605: Camera plugin uses deprecated messages
The camera plugin currently uses deprecated messages like [MAV_CMD_REQUEST_CAMERA_INFORMATION](https://mavlink.io/en/messages/common.html#MAV_CMD_REQUEST_CAMERA_INFORMATION) instead of [MAV_CMD_REQUEST_MESSAGE](https://mavlink.io/en/messages/common.html#MAV_CMD_REQUEST_MESSAGE).

This would make it unusable with a camera that only supports the "new" request message.

### LINKED ISSUE #2013: REQUEST_MESSAGE vs custom getters like REQUEST_CAMERA_INFORMATION
It's currently part of the [list for MAVSDK v2](https://github.com/mavlink/MAVSDK/issues/2002), and though I don't think I can contribute an implementation, I think I can contribute a (relatively informed) opinion.

TL;DR: I don't think this can go in MAVSDK v2 :see_no_evil:.

#### We cannot detect what the receiver supports

There is no way to detect if the receiver supports `REQUEST_MESSAGE`, because the requested message could have been requested by someone else, or it could be transmitted spontaneously by the receiver. So making statistics to try to guess stuff like "was this `CAMERA_INFORMATION` message an answer to my request, or is it a coincidence?" is probably not worth it by far.

Also each message is independent. So even if we managed to detect that `REQUEST_MESSAGE` is supported for `CAMERA_INFORMATION`, it would not say anything about e.g. `REQUEST_PROTOCOL_VERSION`.

#### What we can do

* Push the receivers to support both messages (i.e. the drone should support both `REQUEST_MESSAGE(<camera_information>)` and `REQUEST_CAMERA_INFORMATION`.
* Keep the old way in MAVSDK until we feel like we can completely drop it.

In other words, the implementation work is not on the MAVSDK side, but on the receiver side (e.g. make sure that PX4/Ardupilot support the newer messages, make sure that cameras do, push proprietary players to do it).

Some day, maybe we will feel comfortable in dropping the old way and telling people "well if your drone does not support the new way, that's your problem".

julianoes:
This implements the logic of requesting messages using the new command as well as the deprecated commands consolidated in the RequestMessage class.

Currently WIP implementing the functionality in REQUEST_MESSAGE. The next step is to use this class consistently everywhere.

Closes #1605, closes #2013.

### DIFF
--- a/src/mavsdk/core/CMakeLists.txt
+++ b/src/mavsdk/core/CMakeLists.txt
@@ -42,6 +42,7 @@ target_sources(mavsdk
     mavlink_parameter_subscription.cpp
     mavlink_parameter_helper.cpp
     mavlink_receiver.cpp
+    mavlink_request_message.cpp
     mavlink_request_message_handler.cpp
     mavlink_statustext_handler.cpp
     mavlink_message_handler.cpp
@@ -60,7 +61,6 @@ target_sources(mavsdk
     log.cpp
     cli_arg.cpp
     geometry.cpp
-    request_message.cpp
     mavsdk_time.cpp
     timesync.cpp
 )
--- a/src/mavsdk/core/mavlink_component_metadata.cpp
+++ b/src/mavsdk/core/mavlink_component_metadata.cpp
@@ -61,7 +61,7 @@ void MavlinkComponentMetadata::request_component(uint32_t compid)
     const std::lock_guard lg{_mavlink_components_mutex};
     if (_mavlink_components.find(compid) == _mavlink_components.end()) {
         _mavlink_components[compid] = MavlinkComponent{};
-        _system_impl.request_message().request(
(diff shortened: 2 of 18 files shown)


JonasVautherin (approved):
Yeah I think this logic makes sense 👍

julianoes:
@JonasVautherin do you mind double checking the last commit please, thanks! Better to get that one right...

JonasVautherin:
It looks good to me! Though I haven't tested it, I only read the commit (carefully, though). Let me know if you need me to test something.

julianoes:
It fixes the system_test failure, so that's enough testing for me :)

PR #2377, opened 2024-08-18
Add a BUILD_TESTING option to superbuild

JonasVautherin:
It has been a problem e.g. [here](https://github.com/mavlink/MAVSDK/issues/2374#issuecomment-2293752977) and [here](https://discuss.px4.io/t/mavsdk-main-branch-fails-to-build-on-rpi4-blueos/40003/9). I think it makes sense to add an option for those who don't need to run the tests.

Also I renamed BUILD_TESTS to BUILD_TESTING, which is more common with CMake (that's actually the CTest way, I believe).

PR #2382, opened 2024-08-21
CI fails because the third_party folder is too big to cache

JonasVautherin:
Trying to reproduce the CI failure from https://github.com/mavlink/MAVSDK/pull/2377

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -96,15 +96,20 @@ jobs:
         run: |
           rm -rf ./build/release/third_party/absl
           rm -rf ./build/release/third_party/cares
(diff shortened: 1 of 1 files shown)


JonasVautherin:
It fails as well... I had not seen before but the summary shows this:

```
The hosted runner: GitHub Actions 12 lost communication with the server. Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error.
```

And it feels like something similar happened in the past, seeing [this](https://github.com/mavlink/MAVSDK/blob/main/.github/workflows/main.yml#L95-L107)...

julianoes (approved):

PR #2383, opened 2024-08-21
Added function to query SYS_STATUS present/enabled/health flags

dakejahl:
Added function to query SYS_STATUS present/enabled/health. This allows applications to look at specific sensor statuses, such as [logloader](https://github.com/ARK-Electronics/logloader) which looks at the [MAV_SYS_STATUS_LOGGING](https://mavlink.io/en/messages/common.html#MAV_SYS_STATUS_LOGGING) flag

### DIFF
--- a/src/mavsdk/plugins/telemetry/include/plugins/telemetry/telemetry.h
+++ b/src/mavsdk/plugins/telemetry/include/plugins/telemetry/telemetry.h
@@ -432,6 +432,15 @@ public:
     friend std::ostream& operator<<(std::ostream& str, Telemetry::Health const& health);
 
     /**
+     * @brief SysStatusSensors type.
+     */
+    struct SysStatusSensors {
+        uint32_t present{0}; /**< Bitmap showing which onboard controllers and sensors are present */
+        uint32_t enabled{0}; /**< Bitmap showing which onboard controllers and sensors are enabled */
+        uint32_t health{0}; /**< Bitmap showing which onboard controllers and sensors have an error */
+    };
+
+    /**
      * @brief Remote control status type.
      */
     struct RcStatus {
@@ -1427,6 +1436,8 @@ public:
      * @return One Health update.
      */
     Health health() const;
+
+    SysStatusSensors sys_status_sensors() const;
 
     /**
      * @brief Callback type for subscribe_rc_status.
--- a/src/mavsdk/plugins/telemetry/include/plugins/telemetry/telemetry.h
+++ b/src/mavsdk/plugins/telemetry/include/plugins/telemetry/telemetry.h
@@ -1440,6 +1440,11 @@ public:
      */
     Health health() const;
 
+    /**
+     * @brief Poll for 'sys_status_sensors' (blocking).
+     *
+     * @return One SysStatusSensors update.
+     */
     SysStatusSensors sys_status_sensors() const;
 
     /**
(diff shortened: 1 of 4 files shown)


dakejahl:
Not sure what the `check for diff` check is doing or why it's failing

JonasVautherin:
Thanks!

I guess the question is whether it should expose an `uint32_t` or some list of booleans for all (or a subset) of the sensors in [MAV_SYS_STATUS_SENSOR](https://mavlink.io/en/messages/common.html#MAV_SYS_STATUS_SENSOR).

Does it always make sense to expose all those values, or are some values fundamentally "internal"? I could imagine for instance that if `SENSOR_BATTERY=false` means "there is no battery", maybe it should go together with the battery update (e.g. an `is_present` field in https://github.com/mavlink/MAVSDK-Proto/blob/main/protos/telemetry/telemetry.proto#L566-L573?). Same for the GPS, maybe if a GPS sensor is not "present", we should have a [FixType](https://github.com/mavlink/MAVSDK-Proto/blob/main/protos/telemetry/telemetry.proto#L556) saying "NO_GPS_SENSOR" (if there is a difference between "no GPS connected" and "no GPS present").

What do you think? 

> Not sure what the check for diff check is doing or why it's failing

I think it would be the style check. Try running `./tools/fix_style.sh` 👍

julianoes:
I usually don't want to expose bitflags like this but make the API clearer. I know that's a pain but otherwise we might as well expose all over MAVLink directly and we're back to square one.

So generally the idea is to have boolean flags that actually mean something instead and can be queried.

For logloader specifically, could you use passthrough instead?

julianoes:
> I did run fix_style.sh, it's only failing on 24.04 🤔

You need to fix the style on 24.04 or use the docker helper.

dakejahl:
I could use passthrough but then my code becomes disorderly. Okay if I changed it to a boolean flag for each of those mentioned sensors, what pattern should I follow?

This isn't doing anything more
```
tools/run-docker-clang-format.sh tools/fix_style.sh .
```

dakejahl:
> if there is a difference between "no GPS connected" and "no GPS present"

also part of why I was avoiding the specific flags.. PX4 makes it hard to use the sys_status_sensor flags properly since they're coupled to the HealthAndArmingChecks Report. I can't remember exact details, but when I implemented the loggerCheck I wasn't able to set the present/enabled flags directly, and as a result they toggle true/false together.

JonasVautherin:
> Okay if I changed it to a boolean flag for each of those mentioned sensors, what pattern should I follow?

I would think something like the winch status flags? https://github.com/mavlink/MAVSDK-Proto/blob/517fef5a3a673e9b50f65a56dff350c6af515564/protos/winch/winch.proto#L77-L101

julianoes:
I think it will either be a Status struct per component:

```
message StatusFlags {
   bool present = 1;
   bool enabled = 2;
   bool healthy = 3;
}
```

Or a flat enum:

```
enum Status {
    STATUS_NA = 0; // Not present, not enabled, not healthy
    STATUS_PRESENT = 1; // Not enabled, not healthy
    STATUS_ENABLED = 2; // Not present, not healthy
    STATUS_PRESENT_ENABLED = 3; // Not healthy
    // STATUS_HEALTHY = 4; // -> invalid
    STATUS_PRESENT_HEALTHY = 5; // Not enabled 
    // STATUS_ENABLED_HEALTHY = 6; -> invalid
    STATUS_PRESENT_ENABLED_HEALTHY = 7; -> All ok!
```

JonasVautherin:
> I think it will either be a Status struct per component:

The StatusFlags struct with 3 booleans is a lot more intuitive to me, I would say :see_no_evil:

julianoes:
I believe the events interface means we don't need this anymore, but in any case it would have to be along the "struct per component" outlined above, so I'll close this PR.

PR #2384, opened 2024-08-28
Update gimbal_full_control.cpp example

KarthiAru:
- Added `gimbal_id` parameter to all gimbal related methods.
- Replaced the non-existing `set_mode` method with the `set_angles` method.
- Revised `set_angle` methods to include the correct parameters

### DIFF
--- a/examples/gimbal_full_control/gimbal_full_control.cpp
+++ b/examples/gimbal_full_control/gimbal_full_control.cpp
@@ -51,20 +51,40 @@ int main(int argc, char** argv)
     // Instantiate plugins.
     auto gimbal = Gimbal{system.value()};
 
+    auto prom = std::promise<int32_t>();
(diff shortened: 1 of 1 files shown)


julianoes (dismissed):
Oh, that's great, thanks!

JonasVautherin (approved):
Thanks!

The style check is failing, though 😕

JonasVautherin:
I think you have to run `./tools/fix_style.sh` from Ubuntu 24.04. It used to work with a container if you are not on Ubuntu 24. Maybe `./tools/tools/run-docker-clang-format.sh ./tools/fix_style.sh`? I haven't used it in a while 🙈

KarthiAru:
@JonasVautherin Thanks. I've fixed the style on Ubuntu 24.04

JonasVautherin (approved):

PR #2387, opened 2024-08-29
Update gimbal.cpp

KarthiAru:
- Added gimbal_id param to all the gimbal methods
- Updated yaw values to point to NESW directions

### DIFF
--- a/examples/gimbal/gimbal.cpp
+++ b/examples/gimbal/gimbal.cpp
@@ -85,81 +85,85 @@ int main(int argc, char** argv)
 
     std::cout << "Start controlling gimbal...\n";
     Gimbal::Result gimbal_result = gimbal.take_control(gimbal_id, Gimbal::ControlMode::Primary);
+    sleep_for(seconds(5));
     if (gimbal_result != Gimbal::Result::Success) {
         std::cerr << "Could not take gimbal control: " << gimbal_result << '\n';
         return 1;
     }
 
+    // Fetch the gimbal attitude
+    auto [result, attitude] = gimbal.get_attitude(gimbal_id);
+    gimbal.
     std::cout << "Use yaw mode to lock to a specific direction...\n";
 
     std::cout << "Look North...\n";
-    gimbal.set_angles(0, 0.0f, 0.0f, 0.0f, Gimbal::GimbalMode::YawLock, Gimbal::SendMode::Once);
-    sleep_for(seconds(2));
+    gimbal.set_angles(gimbal_id, 0.0f, 0.0f, -attitude.euler_angle_north.yaw_deg, Gimbal::GimbalMode::YawLock, Gimbal::SendMode::Once);
+    sleep_for(seconds(5));
 
     std::cout << "Look East...\n";
-    gimbal.set_angles(0, 0.0f, 0.0f, 90.0f, Gimbal::GimbalMode::YawLock, Gimbal::SendMode::Once);
-    sleep_for(seconds(2));
+    gimbal.set_angles(gimbal_id, 0.0f, 0.0f, -attitude.euler_angle_north.yaw_deg + 90.0f, Gimbal::GimbalMode::YawLock, Gimbal::SendMode::Once);
+    sleep_for(seconds(5));
 
     std::cout << "Look South...\n";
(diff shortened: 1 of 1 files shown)


examples/gimbal/gimbal.cpp
```diff
@@ -85,81 +85,85 @@ int main(int argc, char** argv)
...
 
+    // Fetch the gimbal attitude
+    auto [result, attitude] = gimbal.get_attitude(gimbal_id);
+
     std::cout << "Use yaw mode to lock to a specific direction...\n";
 
     std::cout << "Look North...\n";
-    gimbal.set_angles(0, 0.0f, 0.0f, 0.0f, Gimbal::GimbalMode::YawLock, Gimbal::SendMode::Once);
-    sleep_for(seconds(2));
+    gimbal.set_angles(gimbal_id, 0.0f, 0.0f, -attitude.euler_angle_north.yaw_deg, Gimbal::GimbalMode::YawLock, Gimbal::SendMode::Once);
```
JonasVautherin, line 100: I don't understand the `-attitude.euler_angle_north.yaw_deg` added in these lines. Why is that necessary?

KarthiAru: @JonasVautherin The initial implementation for Look North/East/South/West was originally designed as Look Forward/Left/Back/Right. Therefore, I had to adjust the yaw based on its current orientation relative to North. If this adjustment isn't necessary, we could consider renaming it to simply Look Forward/Left/Back/Right. However, this might be redundant since this functionality is already included in the code.

JonasVautherin: Right, makes sense!

JonasVautherin (dismissed):

JonasVautherin (dismissed):

examples/gimbal/gimbal.cpp
```diff
@@ -85,81 +85,116 @@ int main(int argc, char** argv)
...
     std::cout << "Start controlling gimbal...\n";
     Gimbal::Result gimbal_result = gimbal.take_control(gimbal_id, Gimbal::ControlMode::Primary);
+    sleep_for(seconds(5));
     if (gimbal_result != Gimbal::Result::Success) {
         std::cerr << "Could not take gimbal control: " << gimbal_result << '\n';
         return 1;
     }
 
+    // Fetch the gimbal attitude
+    auto [result, attitude] = gimbal.get_attitude(gimbal_id);
```
julianoes, line 95: Nice :smile:

examples/gimbal/gimbal.cpp
```diff
@@ -85,81 +85,116 @@ int main(int argc, char** argv)
 
     std::cout << "Start controlling gimbal...\n";
     Gimbal::Result gimbal_result = gimbal.take_control(gimbal_id, Gimbal::ControlMode::Primary);
+    sleep_for(seconds(5));
```
julianoes, line 88: Is this sleep required?

examples/gimbal/gimbal.cpp
```diff
@@ -85,81 +85,116 @@ int main(int argc, char** argv)
...
     gimbal.set_angles(0, 0.0f, 0.0f, 0.0f, Gimbal::GimbalMode::YawFollow, Gimbal::SendMode::Once);
-    sleep_for(seconds(2));
+    sleep_for(seconds(5));
 
     std::cout << "Stop controlling gimbal...\n";
-    gimbal_result = gimbal.release_control(0);
+    gimbal_result = gimbal.release_control(gimbal_id);
     if (gimbal_result != Gimbal::Result::Success) {
-        std::cerr << "Could not take gimbal control: " << gimbal_result << '\n';
+        std::cerr << "Could not release gimbal control: " << gimbal_result << '\n';
```
julianoes, line 197: Good catch. Thanks for cleaning this up.

julianoes (dismissed):
:+1:

examples/gimbal/gimbal.cpp
KarthiAru: @julianoes No, the sleep is not required here. I've removed it.
Initially, I added it because I was encountering issues with taking control, and I thought introducing a delay might help. But it doesn't make sense as it is not an async method.

examples/gimbal/gimbal.cpp
KarthiAru: Thanks :)

julianoes (approved):

PR #2388, opened 2024-09-02
[v2.12] mavsdk_server: fix autopilot discovery

julianoes:
Instead of using the subscribe_on_new_systems callback, we just use a while loop here. That's because the subscribe_on_new_system callback won't tell us if an autopilot is discovered when another component of the same system is discovered first, thus triggering the callback early when no autopilot is around.

With a stupid old while loop we can just keep checking this until we have an autopilot and then exit.


JonasVautherin (commented):
But then what happens when MAVSDK does not connect to an Autopilot? E.g. when running as an autopilot itself?

This assumes that there is always an autopilot on the other side, doesn't it?

julianoes:
Yes, but that was the assumption before as well:
https://github.com/mavlink/MAVSDK/pull/2388/files#diff-7b7dbd5a57f1ca31f7681654158233825a19e674d05e6a7877785538fc808976L61

julianoes:
Thanks @JonasVautherin. I'm going to deploy and test this now :crossed_fingers:.

JonasVautherin:
> Yes, but that was the assumption before as well:

Oh, right :man_facepalming:

PR #2400, opened 2024-09-12
core: Fix reading int8/16 params from ArduPilot

### LINKED ISSUE #2396: Getting UINT16 param from ArduPilot doesn't work
Cannot read any param from ArduPilot other than `float` and `int32` - say `RC1_MIN` (`int16`).
Aftere debugging I can see that the problem is in `mavlink_parameter_client::get_param_async_typesafe` function.
Before it parameter is succesfully read. But checking its type (`int16_t` vs `int32_t`) doesn't pass here
https://github.com/mavlink/MAVSDK/blob/b4160bb0ceab38db13113d24f0ede087b8804f1f/src/mavsdk/core/mavlink_parameter_client.cpp#L280

pavloblindnology:
Fixes #2396

### DIFF
--- a/src/mavsdk/core/mavlink_parameter_client.cpp
+++ b/src/mavsdk/core/mavlink_parameter_client.cpp
@@ -289,6 +289,29 @@ void MavlinkParameterClient::get_param_async_typesafe(
     get_param_async(name, callback_future_result, cookie);
 }
 
+template<>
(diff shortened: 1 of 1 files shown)


julianoes (approved):
Nice, thanks for that! Let's check CI...

pavloblindnology:
@julianoes CI seems fine...

JonasVautherin (approved):

PR #2413, opened 2024-10-04
[core] Added ComponentType::RemoteID

dakejahl:
- ~Removed unused duplicate enum class in system.h~ Can we consolidate `System::ComponentType` with `Mavsdk::ComponentType`?
- Added `ComponentType::RemoteID` -- I'm transitioning [RemoteIDTransmitter](https://github.com/ARK-Electronics/RemoteIDTransmitter) to MAVSDK
- Removed `ComponentType::Custom` to reduce confusion since it was defaulting to MAV_TYPE_GENERIC
- Added `Mavsdk::Configuration` constructor for `ComponentType::RemoteId`

### DIFF
--- a/src/mavsdk/core/include/mavsdk/mavsdk.h
+++ b/src/mavsdk/core/include/mavsdk/mavsdk.h
@@ -357,14 +357,18 @@ public:
     void intercept_outgoing_messages_async(std::function<bool(mavlink_message_t&)> callback);
 
 private:
+    static constexpr int DEFAULT_SYSTEM_ID_AUTOPILOT = 1;
+    static constexpr int DEFAULT_COMPONENT_ID_AUTOPILOT = MAV_COMP_ID_AUTOPILOT1;
     static constexpr int DEFAULT_SYSTEM_ID_GCS = 245;
     static constexpr int DEFAULT_COMPONENT_ID_GCS = MAV_COMP_ID_MISSIONPLANNER;
     static constexpr int DEFAULT_SYSTEM_ID_CC = 1;
     static constexpr int DEFAULT_COMPONENT_ID_CC = MAV_COMP_ID_PATHPLANNER;
-    static constexpr int DEFAULT_SYSTEM_ID_AUTOPILOT = 1;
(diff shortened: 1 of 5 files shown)


dakejahl:
Nevermind it looks like it didn't do a clean build and `ComponentType` in `System` is indeed used. But now I'm confused why there are 2 definitions in different namespaces that are not in the same order. Can we just use the Mavsdk::ComponentType?

dakejahl:
This is hard to do without introducing circular dependencies. Can we have some kind of **global_include.hpp** where we can define things common to the core classes?

src/mavsdk/core/include/mavsdk/mavsdk.h
```diff
@@ -153,8 +153,7 @@ class Mavsdk {
         GroundStation, /**< @brief SDK is used as a ground station. */
         CompanionComputer, /**< @brief SDK is used as a companion computer on board the MAV. */
         Camera, /** < @brief SDK is used as a camera. */
-        Custom /**< @brief the SDK is used in a custom configuration, no automatic ID will be
-                  provided */
+        RemoteId /**< @brief the SDK is used as a RemoteId system. */
```
julianoes, line 156: Hang on, what happens to `Custom`? We always need `Custom`.

julianoes (changes requested):
Thanks for this.

I absolutely don't want a global_include.h file, however, I also don't want circular dependencies, so we'll have to clean it up.

I also don't want multiple component enums. I thought I had already cleaned that up but it looks like I haven't, so I should do that.

src/mavsdk/core/include/mavsdk/mavsdk.h
dakejahl: `Custom` just defaults to MAV_TYPE_GENERIC so it's misleading.

julianoes:
#2416

dakejahl:
Rebased off of your branch

julianoes (approved):
Thanks for cleaning my changes up and adding yours :)

PR #2417, opened 2024-10-06
mavsdk_server: fix crash on stop/destruction

julianoes:
This fixes a segfault which happens when the mavsdk_server is stopped and destroyed before every having discovered any autopilot.

What happened is that:
1. ConnectionInitiator would be cancelled.
2. ConnectionInitiator would be destructed.
3. And only now the connect thread would wake up from sleeping and read the _should_exit flag which at this point has been destroyed. At this point the connect function would try to access Mavsdk::systems() which is also destructed by that point and cause a segfault.

The fix is to wrap the ConnectionInitiator class in a shared ptr and keep that one alive in the connect function to avoid destruction until it has returned.

### DIFF
--- a/src/mavsdk_server/src/connection_initiator.h
+++ b/src/mavsdk_server/src/connection_initiator.h
@@ -1,25 +1,26 @@
 #pragma once
 
+#include <atomic>
 #include <chrono>
 #include <future>
+#include <memory>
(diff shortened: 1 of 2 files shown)


JonasVautherin (approved):
Nice catch! :rocket:

PR #2423, opened 2024-10-13
Autopilot server and command server fixes

julianoes:
See commits.

The command fix gets rid of these confusing warnings:
```
[02:27:01|Debug] Received ack from 1/1 for not-existing command: 512! Ignoring... (mavlink_command_sender.cpp:285)
```

Finally...

### DIFF
--- a/src/mavsdk/plugins/telemetry_server/telemetry_server_impl.cpp
+++ b/src/mavsdk/plugins/telemetry_server/telemetry_server_impl.cpp
@@ -83,20 +83,25 @@ void TelemetryServerImpl::init()
                 case MAVLINK_MSG_ID_HOME_POSITION:
                     if (_maybe_home) {
                         if (_send_home()) {
-                            return _server_component_impl->make_command_ack_message(
-                                command, MAV_RESULT::MAV_RESULT_ACCEPTED);
+                            return std::optional<mavlink_command_ack_t>{
+                                _server_component_impl->make_command_ack_message(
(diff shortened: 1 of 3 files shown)


src/mavsdk/plugins/telemetry_server/telemetry_server_impl.cpp
```diff
@@ -83,20 +83,25 @@ void TelemetryServerImpl::init()
...
+                                command, MAV_RESULT::MAV_RESULT_DENIED)};
                     }
 
                 default:
-                    return _server_component_impl->make_command_ack_message(
-                        command, MAV_RESULT::MAV_RESULT_DENIED);
+                    // Let's hope someone else answers and keep silent. In an ideal world we would
+                    // explicitly deny all the ones that we ought to answer but haven't implemented
+                    // yet.
+                    return std::optional<mavlink_command_ack_t>{};
```
JonasVautherin, line 104: Is the issue that multiple components (e.g. autopilot_server or telemetry_server) may process the request, and they can't send `MAV_RESULT_DENIED` because they don't know if another component will answer? Or do I misunderstand it?

JonasVautherin (approved):
LGTM!

src/mavsdk/plugins/telemetry_server/telemetry_server_impl.cpp
julianoes: Exactly!

PR #2425, opened 2024-10-15
core: make udpin parsing stricter

julianoes:
We don't want to accept udpin://:PORT but we always need the network interface, e.g.:
- udpin://192.168.X.Y:PORT, or
- udpin://0.0.0.0:PORT for all network interfaces

Without this we got an error further down with inet_pton.

### DIFF
--- a/src/mavsdk/core/cli_arg.cpp
+++ b/src/mavsdk/core/cli_arg.cpp
@@ -103,8 +103,11 @@ bool CliArg::parse_udpin(const std::string_view rest)
     auto& p = std::get<Udp>(protocol);
     p.host = rest.substr(0, pos);
     p.mode = Udp::Mode::In;
-    p.host = rest.substr(0, pos);
-    p.mode = Udp::Mode::In;
+
+    if (p.host.empty()) {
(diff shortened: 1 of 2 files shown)


src/mavsdk/core/cli_arg.cpp
```diff
@@ -103,8 +103,11 @@ bool CliArg::parse_udpin(const std::string_view rest)
     auto& p = std::get<Udp>(protocol);
     p.host = rest.substr(0, pos);
     p.mode = Udp::Mode::In;
-    p.host = rest.substr(0, pos);
-    p.mode = Udp::Mode::In;
+
+    if (p.host.empty()) {
+        LogErr() << "No network interface (or 0.0.0.0 for all network interfaces) supplied";
```
JonasVautherin, line 108: Small suggestion:

```suggestion
        LogErr() << "No network interface supplied (use 0.0.0.0 for all network interfaces)";
```

JonasVautherin (dismissed):

JonasVautherin (approved):

PR #2441, opened 2024-11-14
cmake: set RPATH of installed mavsdk_server

julianoes:
This allows to install the mavsdk_server binary in any install prefix and run it.

### DIFF
--- a/src/mavsdk_server/src/CMakeLists.txt
+++ b/src/mavsdk_server/src/CMakeLists.txt
@@ -7,6 +7,8 @@ list(APPEND MAVSDK_SERVER_SOURCES
     grpc_server.cpp
 )
 
+set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
+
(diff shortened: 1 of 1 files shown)


JonasVautherin (commented):
Won't this always inject an rpath, even when the intent is actually to use system dependencies?

Also does it inject relative paths or absolute paths? It feels like `CMAKE_INSTALL_PREFIX` may be absolute, right?

julianoes:
I don't think you ever want to use system dependencies with mavsdk_server, because you build mavsdk, mavsdk_server, and mavsdk_server_bin at the same time and install it together, locally or system wide.

Yes, `CMAKE_INSTALL_PREFIX` is absolute, but it does work, and without it, it doesn't work.

JonasVautherin:
> and without it, it doesn't work.

Yeah without it you probably have to set LD_LIBRARY_PATH, which is inconvenient. I was just concerned that this may inject paths into some system packages people may distribute.

julianoes:
Let's merge it and check the release once done.

PR #2451, opened 2024-11-25
Fixes MAVSDK-Java#155 camera::possibleSettingOptions crashes as a result of erroneous string format

rayw-dronesense:
Corrects the conversion to numeric string representation

### DIFF
--- a/src/mavsdk/plugins/camera/camera_impl.cpp
+++ b/src/mavsdk/plugins/camera/camera_impl.cpp
@@ -1848,10 +1848,8 @@ bool CameraImpl::get_possible_options(
     }
 
     for (const auto& value : values) {
-        std::stringstream ss{};
-        ss << value.get_string();
         Camera::Option option{};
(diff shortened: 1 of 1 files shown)


julianoes (commented):
:thinking: wouldn't that then not require the whole `stringstream` logic at all and you could just directly set the option_id?

```
option.option_id = value.get_string()
```

I'm not sure whether we're missing something here.

rayw-dronesense:
> 🤔 wouldn't that then not require the whole `stringstream` logic at all and you could just directly set the option_id?
> 
> ```
> option.option_id = value.get_string()
> ```
> 
> I'm not sure whether we're missing something here.

You're right, it's not needed anymore. I was originally basing the fix on existing logic.

julianoes:
I guess it doesn't break any tests, so I'm going to trust you that this is right.

julianoes (approved):

PR #2453, opened 2024-11-28
Enable ssl in curl

JonasVautherin:
This should resolve https://github.com/mavlink/MAVSDK-Java/issues/192.

Note: when using openssl, it seems like curl needs `dlopen`: https://github.com/curl/curl/blame/a72b479decd555c9e9c2c50ded34e317123174f2/packages/vms/report_openssl_version.c#L54. We already add `${CMAKE_DL_LIBS}` to the `mavsdk_server` and `unit_tests_mavsdk_server` targets but not `integration_tests`, `system_tests_runner` and `unit_tests_runner`. I think it makes sense to add it to libmavsdk as PUBLIC so that it gets propagated everywhere.


third_party/curl/CMakeLists.txt
```diff
@@ -9,8 +9,9 @@ list(APPEND CMAKE_ARGS
     "-DCMAKE_TOOLCHAIN_FILE:PATH=${CMAKE_TOOLCHAIN_FILE}"
     "-DCMAKE_POSITION_INDEPENDENT_CODE=ON"
     "-DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}"
-    "-DCURL_USE_OPENSSL=OFF"
+    "-DCURL_USE_OPENSSL=ON"
     "-DCURL_USE_LIBSSH2=OFF"
+    "-DBUILD_CURL_EXE=OFF"
```
JonasVautherin, line 14: I don't see a need to disable that only for Android and iOS, and disabling it pushes the need for dlopen to the build step, which is convenient.

JonasVautherin:
Hmm actually, instead of adding it everywhere, I wonder if it could not be propagated from libmavsdk :thinking:

julianoes (approved):
Wow you're good!

PR #2454, opened 2024-11-28
[BACKPORT v2.12] Enable ssl in curl

JonasVautherin:
Let's try to backport this


JonasVautherin:
Actually this will go in v3.

PR #2459, opened 2024-12-03
curl: point to system CA at build time on Android

JonasVautherin:
Trying to solve https://github.com/mavlink/MAVSDK-Java/issues/192


JonasVautherin:
I don't think this will do it. It seems like openssl3 just doesn't support Android system certificates (because they use an old format that openssl3 doesn't support).

PR #2460, opened 2024-12-04
Fix HTTPS downloads using curl

JonasVautherin:
Resolves https://github.com/mavlink/MAVSDK-Java/issues/192.

In a previous PR, we have built curl with ssl support. But that is not enough for Android. First, curl does not auto-detect the system certificates on Android (at least not on all versions of Android). Second, OpenSSL3 cannot parse the Android system certificates (again at least on the versions of Android we tested) and it does not seem like this will change. BoringSSL, on the other hand, works with the Android system certificates.

I don't think that we should move MAVSDK entirely to BoringSSL for different reasons, one of them being that the default ssl library on many systems is OpenSSL. However, if OpenSSL3 does not really support Android, it seems like it makes sense to move to BoringSSL. Just for Android.

@julianoes: In order for boringssl to build, I had to update grpc, which in turn forced me to update protobuf, and hence to regenerate the code :see_no_evil:
@rayw-dronesense: would you mind testing this on your setup? :pray:


rayw-dronesense:
@JonasVautherin  It works!

JonasVautherin:
Awesome! I just need to fix the iOS/macOS build with the updated gRPC now!

JonasVautherin:
Hmm with the latest version of gRPC I had to update protobuf, which means that I need to re-generate the code... I don't know how I feel about that. Trying with an older version of gRPC, see if that works.

The PR for v3 does update protobuf, though: https://github.com/mavlink/MAVSDK/pull/2460

julianoes:
Closing this as I opted for simple hack for v2: https://github.com/mavlink/MAVSDK/pull/2471

PR #2467, opened 2024-12-11
Add Example of MAVLink Proxy Using MavlinkPassthrough Plugin

flaviol-souza:
This pull request introduces an example demonstrating how to implement a MAVLink proxy using the MAVSDK `MavlinkPassthrough` plugin. The proxy facilitates communication between two MAVLink components (e.g., PX4 and QGroundControl) by intercepting and forwarding messages.

#### **Key Features**
- **Message Interception and Forwarding**: Captures MAVLink messages from one component and routes them to another.
- **Multi-Threaded Architecture**: Ensures efficient processing of messages in parallel threads.
- **Queue Management**: Handles message queuing with size limits and thread-safe operations.
- **Graceful Shutdown**: Implements signal handling for controlled termination.
- **Error Handling**: Provides robust logging and checks for connection integrity.

#### **Use Cases**
- Simulating a communication intermediary for testing MAVLink behavior.
- Debugging or monitoring MAVLink messages between systems.
- Building advanced communication relays for MAVLink networks.

#### **Execution Command**
To run the example, use the following command:

```bash
./build/qgc_proxy_px4 udp://127.0.0.1:14550 udp://:14050
```

**Important Note:** To avoid conflicts, the PX4 SITL instance must not run on the default port `14550`. Adjust the PX4 configuration to use a different port (e.g., `14050` in this example).


examples/qgc_proxy_px4/qgc_proxy_px4.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+    if (px4_conn_result != ConnectionResult::Success || qgc_conn_result != ConnectionResult::Success) {
+        log_message("ERROR", "Failed to connect to the provided URLs.");
+        return 1;
+    }
+
+    log_message("INFO", "Waiting for systems to connect...");
+
+    intercept_outgoing_messages(mavsdk_px4, PROXY_SYSID_PX4, PROXY_SYSID_QGC);
+    //intercept_outgoing_messages(mavsdk_qgc, PROXY_SYSID_PX4, PROXY_SYSID_QGC);
```
JonasVautherin, line 138: Why is this commented out?

examples/qgc_proxy_px4/qgc_proxy_px4.cpp
```diff
@@ -0,0 +1,183 @@
...
+    message_queues[PX4_NAME] = std::queue<mavlink_message_t>();
+    message_queues[QGC_NAME] = std::queue<mavlink_message_t>();
+    queue_mutexes[PX4_NAME] = std::make_unique<std::mutex>();
+    queue_mutexes[QGC_NAME] = std::make_unique<std::mutex>();
+    queue_cvs[PX4_NAME] = std::make_unique<std::condition_variable>();
+    queue_cvs[QGC_NAME] = std::make_unique<std::condition_variable>();
+
+    for (uint16_t msg_id = 0; msg_id < 500; ++msg_id) {
+        passthrough_map[PX4_NAME]->subscribe_message(msg_id, [&](const mavlink_message_t& message) {
+            //std::thread([=]() { enqueue_message(PX4_NAME, message); }).detach();
```
JonasVautherin, line 163: Same question here 😇

examples/qgc_proxy_px4/qgc_proxy_px4.cpp
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    message_queues[PX4_NAME] = std::queue<mavlink_message_t>();
+    message_queues[QGC_NAME] = std::queue<mavlink_message_t>();
+    queue_mutexes[PX4_NAME] = std::make_unique<std::mutex>();
+    queue_mutexes[QGC_NAME] = std::make_unique<std::mutex>();
+    queue_cvs[PX4_NAME] = std::make_unique<std::condition_variable>();
+    queue_cvs[QGC_NAME] = std::make_unique<std::condition_variable>();
+
+    for (uint16_t msg_id = 0; msg_id < 500; ++msg_id) {
```
JonasVautherin, line 161: Are you sure that you want to subscribe 500 times? Shouldn't you subscribe just once for each? 🤔

JonasVautherin (commented):
It's a nice idea! I have a few questions on the implementation but overall it looks pretty good!

examples/qgc_proxy_px4/qgc_proxy_px4.cpp
```diff
@@ -0,0 +1,183 @@
...
+std::map<std::string, std::unique_ptr<std::condition_variable>> queue_cvs;
+std::map<std::string, std::shared_ptr<MavlinkPassthrough>> passthrough_map;
+std::mutex passthrough_map_mutex;
+
+// Thread-safe logging
+void log_message(const std::string& level, const std::string& message) {
+    static std::mutex log_mutex;
+    std::lock_guard<std::mutex> lock(log_mutex);
+    std::cout << "[" << level << "] " << message << std::endl;
+}
```
julianoes, line 37: Maybe we should actually do this within MAVSDK, right?

examples/qgc_proxy_px4/qgc_proxy_px4.cpp
julianoes: Hm, and why 500? That seems somewhat arbitrary, as message IDs go up to 43000, with mostly gaps in-between.

julianoes (commented):
Thanks for this!

I would maybe call it MAVLink proxy? Or is it QGC specific?

And it would be nice to have a comment at the top of the file for anyone that finds it to know more about what it is and what it does.

examples/qgc_proxy_px4/qgc_proxy_px4.cpp
JonasVautherin: If we want to enable this, should we maybe have something like a `subscribe_all_messages` in passthrough?

examples/qgc_proxy_px4/qgc_proxy_px4.cpp
flaviol-souza: it'll be removed!

examples/qgc_proxy_px4/qgc_proxy_px4.cpp
flaviol-souza: it'll be removed!

examples/qgc_proxy_px4/qgc_proxy_px4.cpp
flaviol-souza: As a Proxy, the correct thing to do is to subscribe to all message IDs. So a `subscribe_all_messages` might help. Another option is to create a list of message ID constants. I even went as far as manually importing the ID constants into a list. However, I'm open to suggestions.

examples/qgc_proxy_px4/qgc_proxy_px4.cpp
flaviol-souza: How can to do this with MAVSDK ? Is there a function to this ?

examples/qgc_proxy_px4/qgc_proxy_px4.cpp
julianoes: Oh, I see the problem.

I see that `Mavsdk::intercept_messages_async` offers that but MavlinkPassthrough does not. It would probably be a good idea to add that.

julianoes:
Thanks for the changes and comment on the top.

Out of curiosity, what is the use case for queuing? Like why do you need archived messages? Are the momentary ones not enough?

julianoes:
@flaviol-souza bumping my question again, before merging this:

> Out of curiosity, what is the use case for queuing? Like why do you need archived messages? Are the momentary ones not enough?

julianoes:
@flaviol-souza are you still around? See my question above. :arrow_up: I would like to understand this before I merge this and have to maintain it.

julianoes:
I'm closing this until @flaviol-souza is back.

PR #2472, opened 2024-12-15
core: only use http redirects

julianoes:
This is an attempt to prevent cURL from being redirected from http to https since we ship without SSL in MAVSDK v2.


julianoes:
Fascinating :(, thanks for testing @rayw-dronesense.

PR #2473, opened 2024-12-17
examples: Add example on how to use mavsdk_server lib

julianoes:
This just shows how to use the C API to instantiate the gRPC server.

### DIFF
--- a/examples/start_stop_server/start_stop_server.cpp
+++ b/examples/start_stop_server/start_stop_server.cpp
@@ -13,7 +13,7 @@ static struct MavsdkServer* mavsdk_server;
 
 void signal_handler(int sig)
 {
-    std::cout << "Received SIGUSR1 signal" << std::endl;
+    std::cout << "Received signal " << sig << std::endl;
 
     mavsdk_server_stop(mavsdk_server);
 }
@@ -25,7 +25,7 @@ int main(int argc, char* argv[])
         return 1;
     }
 
-    signal(SIGUSR1, signal_handler);
+    signal(SIGINT, signal_handler);
 
     mavsdk_server_init(&mavsdk_server);
     mavsdk_server_run(mavsdk_server, argv[1], 50051);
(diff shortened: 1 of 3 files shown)


examples/start_stop_server/start_stop_server.cpp
```diff
@@ -0,0 +1,36 @@
...
+        std::cout << "Connection string required as argument" << std::endl;
+        return 1;
+    }
+
+    signal(SIGUSR1, signal_handler);
+
+    mavsdk_server_init(&mavsdk_server);
+    mavsdk_server_run(mavsdk_server, argv[1], 50051);
+
+    mavsdk_server_destroy(mavsdk_server);
```
JonasVautherin, line 33: Shouldn't you `mavsdk_server_stop` before?

examples/start_stop_server/start_stop_server.cpp
```diff
@@ -0,0 +1,36 @@
...
+}
+
+int main(int argc, char* argv[])
+{
+    if (argc != 2) {
+        std::cout << "Connection string required as argument" << std::endl;
+        return 1;
+    }
+
+    signal(SIGUSR1, signal_handler);
```
JonasVautherin, line 28: How is `SIGUSR1` triggered? I thought it would have to come programmatically, but in this example nothing sends it 🤔

examples/start_stop_server/start_stop_server.cpp
julianoes: `mavsdk_server_stop` is done in the signal handler.

examples/start_stop_server/start_stop_server.cpp
julianoes: Actually, I should change it to the signal you get with Ctrl+C and catch that nicely.

examples/start_stop_server/start_stop_server.cpp
JonasVautherin: But then shouldn't you call `mavsdk_server_attach()` before? Maybe I'm misremembering it, but here it feels like it will destroy as soon as the autopilot is detected and `mavsdk_server_run` returns 🤔

julianoes:
CI is unrelated, merging.

PR #2487, opened 2025-01-01
Add Option to set the MAV_TYPE manually

cramke:
### Description

- Add MAV_TYPE to Configuration
- Set mav_type from component by default or by setter method
- Move deriving of mav type from component type to Configuration

### Context
I want to use MAVSDK (MAVLink) as an API for a Rover type vehicle and want a more specific type than MAV_TYPE_GENERIC in the heartbeat. By default the mav type is derived from the component type or id, using the already existing implementation. It is implemented in a way that does not break the exsiting Constructors but only adds getter and setters.


julianoes (approved):
Thanks, I think that's ok to change/add with v3.

PR #2490, opened 2025-01-03
Vitepress build

hamishwillee:
This is vitepress build from https://github.com/mavlink/MAVSDK-docs/pull/292 adjusted to deploy from MAVSDK repo. It adds the docs to build into `/docs/` tree. 

Still to do
- [ ] Fix up notes, tips, warnings
- [x] Check/Fix the script for generating references so it doesn't have any cases of HTML element tags - looks OK
- [x] Add access token for CI - done
- [ ] Test workflow action for building.
- [ ] Disable old repo

Not doing

- [ ] Make api ref docs in CI build into github docs tree (Julian to sort)

This can be tested locally by cloning, then from root doing:

```
yarn install
yarn start
```

Then view in browser.


.gitignore
```diff
@@ -26,3 +26,9 @@ cmake-build-*
 # mavlink repo to generate mavlink-headers
 mavlink
 mavlink-headers
+
+
+# vitepress - Docs guide
+node_modules/
```
julianoes, line 32: Would be nice if we can move the `node_modules` into docs.

docs/.vitepress/config.mjs
```diff
@@ -0,0 +1,204 @@
...
+        text: "Version",
+        items: [
+          {
+            text: "main",
+            link: "https://mavsdk.mavlink.io/main/en/cpp/api_changes.html",
+          },
+          {
+            text: "v2",
+            link: "https://mavsdk.mavlink.io/v2.0/en/cpp/api_changes.html",
+          },
```
julianoes, line 154: I will create v3 and we should add it here, because main won't be v3 forever.

julianoes (commented):
A few small comments. I think this will be good, thanks for that!

docs/.vitepress/config.mjs
hamishwillee: We should create main, then branch it to create v3. At that point we'd create this link

.gitignore
hamishwillee: Done. We just have to build from docs rather than root.

.github/workflows/docs_deploy.yml
```diff
@@ -0,0 +1,75 @@
...
+        uses: actions/checkout@v4
+
+      - name: Setup Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
+          cache: npm
+
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile
```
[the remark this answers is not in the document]
hamishwillee: Yes. Done.

docs/.vitepress/config.mjs
hamishwillee: Though first we'd fix up main to have the right API reference - that being your job after this merges.

julianoes: Right, that's fine.

julianoes (approved):
Let's try it!

Issue #2496, opened 2025-01-18
Throw invalid argument exception  in camera plugin


persuader72:
This piece of code in camera_definition.cpp at row 745 (in main)

```
`  for (const auto& option : _parameter_map[setting_name]->options) {
        if (option->value == option_name) {
            description = option->name;
            return true;
        }
    }
    LogWarn() << "Option " << option_name << " not found";`
```

Is throwing an  invalid argument exception because **option_name** can't be converted to an unsigned integer.

![Image](https://github.com/user-attachments/assets/47584f3d-6c22-4078-80c5-b366b530fafb)

Maybe is a typo?

I changed 

```
        if (option->value == option_name) {
```

with

```
        if (option->name == option_name) {
```

and now the library i working normally.

The following is the xml file that is throwing the exception when is parsed.

```xml
<?xml version="1.0" encoding="UTF-8" ?>
<mavlinkcamera>
    <definition version="11">
        <model>GaeboEO</model>
        <vendor>XP4</vendor>
    </definition>
    <parameters>
        <!-- control = 0 tells us this should not create an automatic UI control -->
        <parameter name="CAM_MODE" type="uint32" default="1" control="0">
            <description>Camera Mode</description>
            <options>
                <option name="Photo" value="0"/>
                <option name="Video" value="1"/>
            </options>
        </parameter>
        <parameter name="CAM_ZOOM" type="uint32" default="1" control="1">
            <description>Optical Zoom</description>
            <options>
                <option name="1x" value="1"/>
                <option name="2x" value="2"/>
                <option name="3x" value="3"/>
                <option name="4x" value="4"/>
                <option name="5x" value="5"/>
                <option name="6x" value="6"/>
                <option name="7x" value="7"/>
                <option name="8x" value="8"/>
                <option name="9x" value="9"/>
                <option name="10x" value="10"/>
                <option name="11x" value="11"/>
                <option name="12x" value="12"/>
                <option name="13x" value="13"/>
                <option name="14x" value="14"/>
                <option name="15x" value="15"/>
                <option name="16x" value="16"/>
                <option name="17x" value="17"/>
                <option name="18x" value="18"/>
                <option name="19x" value="19"/>
                <option name="20x" value="20"/>
                <option name="21x" value="21"/>
                <option name="22x" value="22"/>
                <option name="23x" value="23"/>
                <option name="24x" value="24"/>
                <option name="25x" value="25"/>
                <option name="26x" value="26"/>
[...]

PR #2498, opened 2025-01-20
Fix for Invalid argument exception in camera plugin

persuader72:
Fix for Issue [2495](https://github.com/mavlink/MAVSDK/issues/2496)

### DIFF
--- a/src/mavsdk/plugins/camera/camera_definition.cpp
+++ b/src/mavsdk/plugins/camera/camera_definition.cpp
@@ -742,7 +742,7 @@ bool CameraDefinition::get_option_str(
     }
 
     for (const auto& option : _parameter_map[setting_name]->options) {
-        if (option->value == option_name) {
+        if (option->name == option_name) {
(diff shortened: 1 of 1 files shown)


persuader72:
I see that some test are failing

I can't understand the real purpose of get_option_str method.

Reading from source code it would seem that the function will return the option name and return true if the **option name** is present in the list,

But the tests are expecting that the function will return the option name and return true if the **option value** is present in the list.

Do you know what is the correct purpose?

julianoes (approved):
Thanks!

julianoes:
Can you look into the unit tests? Somehow it makes them fail...

persuader72:
Hi 

I tried to understand better the flow of **get_option_str** function i now i think that function is correct, maybe the variable names inside it are a little misleading but it is correct.

The function take in input an option value and return the option name is it exists.

I close di PR because i have to look for the cause of the Invalid argument exception somewhere else.

PR #2501, opened 2025-01-29
Fix notes tips warnings in docs

hamishwillee:
These use a different style than gitbook. Note, I plan to self-merge if tests pass


julianoes:
Nice!

PR #2506, opened 2025-02-10
update to working geofence

ZacharyLowell:
Overview

Currently there is a bug in MAVSDK server that will only allow geofence upload for UAV's with sysID of 1. This has been replaced with a function to return the sys id of the device assigned.

Tested IRL and in Gazebo with quad's assigned sys ID's from 1 - 6.


julianoes (approved):
Thanks!

JonasVautherin (approved):

PR #2512, opened 2025-02-18
ActionServer: allow direct setting of arm/disarm state and vehicle flight mode

jonathanreeves:
# Summary
SDK changes to implement the interface provided in the corresponding [proto PR](https://github.com/mavlink/MAVSDK-Proto/pull/370). Testing continues to be done using QGroundControl to provide visibility into vehicle mode changes.

It's worth noting that there is probably a broad class of state changes like this that could use public APIs. I've opted to keep the scope constrained and focus only on the most essential information for now.

### DIFF
--- a/src/mavsdk/plugins/action_server/action_server_impl.h
+++ b/src/mavsdk/plugins/action_server/action_server_impl.h
@@ -51,7 +51,14 @@ public:
 
     ActionServer::AllowableFlightModes get_allowable_flight_modes();
 
+    ActionServer::Result set_armed_state(bool armed);
+
+    ActionServer::Result set_flight_mode(ActionServer::FlightMode flight_mode);
+
 private:
+    static ActionServer::FlightMode telemetry_flight_mode_from_flight_mode(FlightMode flight_mode);
+    static uint32_t to_px4_mode_from_flight_mode(ActionServer::FlightMode flight_mode);
+
     void set_base_mode(uint8_t base_mode);
     uint8_t get_base_mode() const;
 
@@ -74,16 +81,6 @@ private:
     std::atomic<bool> _disarmable = false;
     std::atomic<bool> _force_disarmable = false;
     std::atomic<bool> _allow_takeoff = false;
-
-    union px4_custom_mode {
-        struct {
-            uint16_t reserved;
-            uint8_t main_mode;
-            uint8_t sub_mode;
-        };
-        uint32_t data;
-        float data_float;
-    };
 
     std::mutex _flight_mode_mutex;
     ActionServer::AllowableFlightModes _allowed_flight_modes{};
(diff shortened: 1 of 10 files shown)


src/mavsdk/plugins/action_server/action_server_impl.h
```diff
@@ -75,16 +82,6 @@ class ActionServerImpl : public ServerPluginImplBase {
     std::atomic<bool> _force_disarmable = false;
     std::atomic<bool> _allow_takeoff = false;
 
-    union px4_custom_mode {
```
jonathanreeves, line 78: Opportunistic cleanup, this declaration is a redundant copy of the one in `px4_custom_mode.h`

julianoes:
There is now a new step that needs doing:

```
sudo apt install doxygen
tools/generate_docs.sh --overwrite --skip-checks
```

jonathanreeves:
> There is now a new step that needs doing:
> 
> ```
> sudo apt install doxygen
> tools/generate_docs.sh --overwrite --skip-checks
> ```

Thanks @julianoes, I did notice that. Quick question though: the github workflow shows it running on ubuntu 24.04, I assume the doxygen version is important here? I tried doing it on a container running 22.04 and it gave me a much larger diff than I was expecting. Wonder if it's worth having a dedicated container to do this similar to the `run-docker-clang-format.sh` script that's already there?

jonathanreeves:
> > There is now a new step that needs doing:
> > ```
> > sudo apt install doxygen
> > tools/generate_docs.sh --overwrite --skip-checks
> > ```
> 
> Thanks @julianoes, I did notice that. Quick question though: the github workflow shows it running on ubuntu 24.04, I assume the doxygen version is important here? I tried doing it on a container running 22.04 and it gave me a much larger diff than I was expecting. Wonder if it's worth having a dedicated container to do this similar to the `run-docker-clang-format.sh` script that's already there?

I tried building an ubuntu 24.04 container and running the commands as described. Unfortunately the output comes out quite different, you can check the commit history but I ended up reverting most of it. It seems that the doxygen version available to the github runner is different in a way that I don't totally understand. I'm open to suggestions but as of right now I'm not sure how to force the output to match.

jonathanreeves:
> > There is now a new step that needs doing:
> > ```
> > sudo apt install doxygen
> > tools/generate_docs.sh --overwrite --skip-checks
> > ```
> 
> Thanks @julianoes, I did notice that. Quick question though: the github workflow shows it running on ubuntu 24.04, I assume the doxygen version is important here? I tried doing it on a container running 22.04 and it gave me a much larger diff than I was expecting. Wonder if it's worth having a dedicated container to do this similar to the `run-docker-clang-format.sh` script that's already there?

Nevermind, sorry for the churn. It looks like running with a 24.04 container does in fact make the docs consistent.

julianoes:
Good point adding oxygen to the clang-format docker image, adding a doxygen version check and using that image for both. I'll do that.

jonathanreeves:
Hey @julianoes, just want to check in and see if you need anything from me on this PR? I know I created some noise while I was sorting out the doxygen stuff, figured it was worth just confirming I'm not dropping the ball on something.

julianoes:
@jonathanreeves sorry, I took a break. Back now.

julianoes:
Force pushed your commit. Hopefully I didn't do anything stupid :smile:

julianoes (approved):

PR #2526, opened 2025-03-18
Remove clunky iterator_invalidated flags

julianoes:
This is an improvement to avoid having to invalidate the iterator. Instead we copy out all callbacks, release the lock, and then call them.


julianoes:
Integrated into https://github.com/mavlink/MAVSDK/pull/2542.

PR #2528, opened 2025-03-18
CI: fix package generation

julianoes:
We changed the path for the build but not package generation.


julianoes:
This was the error:
https://github.com/mavlink/MAVSDK/actions/runs/13861584250/job/38791176169

PR #2530, opened 2025-03-29
telem: remove periodic request HOME_POSITION

dakejahl:
PX4 already streams the HOME_POSITION message at 0.5Hz. When MAVSDK is used with a system that does not have a GPS the home position is never set, and therefore this periodic message request results in log spam.

This is basically just reverting https://github.com/mavlink/MAVSDK/pull/1816

### DIFF
--- a/src/mavsdk/plugins/telemetry/telemetry_impl.cpp
+++ b/src/mavsdk/plugins/telemetry/telemetry_impl.cpp
@@ -191,35 +191,9 @@ void TelemetryImpl::deinit()
     _system_impl->unregister_all_mavlink_message_handlers(this);
 }
 
-void TelemetryImpl::enable()
-{
-    // We're going to retry until we have the Home Position.
-    _homepos_cookie =
-        _system_impl->add_call_every([this]() { request_home_position_again(); }, 2.0f);
-}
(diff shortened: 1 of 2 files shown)


JonasVautherin:
This should fix the CI: https://github.com/mavlink/MAVSDK/pull/2532

julianoes (commented):
I wonder if we should still try to request it every now and then. Maybe with exponential back off. And then disable the warning if we don't get it .

dakejahl:
> I wonder if we should still try to request it every now and then. Maybe with exponential back off. And then disable the warning if we don't get it .

I think that would be good. Let's say request it up to 5 times with increasing delay between the requests. Maybe covering the span of the first couple of minutes?

julianoes:
Right, for now we can just disable the warning, and then we can add a class that does the exponential backoff logic. I'm sure we can re-use it in other places as well.

dakejahl:
Okay. I'm not sure where/how to disable the warning. I think it's a generic warning when the REQUEST_MESSAGE command fails? Let me know if a fix goes in and I'll close this.

julianoes:
First step:
https://github.com/mavlink/MAVSDK/pull/2534

PR #2537, opened 2025-04-08
core: another potential mutex fix

julianoes:
An attempt to fix the invalid argument mutex error on macOS.


julianoes:
Thinking more about this, it would probably make sense to run the system tests using thread sanitizer. And if that doesn't show anything, write a wrapper around std::mutex and do some printing around what's happening.

julianoes:
Nope, no change.

PR #2538, opened 2025-04-10
mocap: Add vision speed estimate

Luc-Meunier:


julianoes:
I've pushed a commit that updates the proto submodule and updates docs.

PR #2539, opened 2025-04-11
takeoff and land prolblem while running has solved

salarshekari:
Issue:
When running the takeoff_and_land example while connected to the simulator, the code froze at line 68 and failed to issue the takeoff command. Only the message "Vehicle is getting ready to arm" was printed repeatedly.

Fix:
Replacing the while loop with an if condition resolved the issue.

Key Technical Details:
Original Problem:

The blocking while loop likely waited indefinitely for a state change that never occurred (e.g., stuck in PREFLIGHT mode).

Common causes:

Simulator (jMAVSim)

Solution Rationale:

if checks the condition once, avoiding deadlocks.

Better practice: Add a timeout or state machine for robustness.

Code Example (Before/After):

// Before (Problematic)
// Check until vehicle is ready to arm
    while (telemetry.health_all_ok() != true) {
        std::cout << "Vehicle is getting ready to arm\n";
        sleep_for(seconds(1));
    }

// After (Fixed)
// Check until vehicle is ready to arm
    if (telemetry.health_all_ok() != true) {
        std::cout << "Vehicle is getting ready to arm\n";
        sleep_for(seconds(1));
    }

### DIFF
--- a/examples/takeoff_and_land/takeoff_and_land.cpp
+++ b/examples/takeoff_and_land/takeoff_and_land.cpp
@@ -66,7 +66,7 @@ int main(int argc, char** argv)
     });
 
     // Check until vehicle is ready to arm
-    while (telemetry.health_all_ok() != true) {
+    if (telemetry.health_all_ok() != true) {
         std::cout << "Vehicle is getting ready to arm\n";
         sleep_for(seconds(1));
     }


JonasVautherin:
The root cause seems to be that "health" is not okay. Do you know why?

julianoes:
Is it this health issue?
https://github.com/PX4/PX4-Autopilot/pull/24494
https://github.com/mavlink/MAVSDK/pull/2521

salarshekari:
I think this might be a GPS or accelerometer error. Normally, the flight controller (or PX4 code) should block arming in this case. But weirdly, in the simulator, the drone does arm and even takes off!

If health_all_ok() isn’t actually... okay, why are we even seeing the message 'Vehicle is getting ready to arm'?

Is this a simulator quirk or did I miss a check somewhere? Thanks for your patience!"

julianoes:
The important thing is actually the `health.is_armable()`, so that's what you can go with always.

julianoes:
The change in itself doesn't make sense to me, so I'm closing this.

PR #2544, opened 2025-04-17
Add fence and rally missions to raw server

tborglum:
I'm working with @jonathanreeves to update the mission server to support more mission types.

In this branch, I have done the following:
- Added fence and rally item vectors to the raw server so they can be sent and received independently from the mission vector.
- Added getter for mission items based on the mission type; if an unsupported mission type is used, an empty vector is returned.
- Added mission type parameter to ResultAndItemsCallback so server callbacks can use the relevant item vector.
- Updated raw server calls to mission transfer server calls to use the mission type of messages instead of being hard coded mission.
- Added raw server logic to only reset mission completion if mission items are received for a mission type mission.
- Added raw server logic to clear fence and rally missions if needed.
- Updated raw server test helper function to create mission items relevant to the mission type they are for.
- Updated raw server tests to be parameterized with supported mission types.

### DIFF
--- a/src/mavsdk/core/mavlink_mission_transfer_server.cpp
+++ b/src/mavsdk/core/mavlink_mission_transfer_server.cpp
@@ -311,7 +311,7 @@ void MavlinkMissionTransferServer::ReceiveIncomingMission::process_timeout()
 void MavlinkMissionTransferServer::ReceiveIncomingMission::callback_and_reset(Result result)
 {
     if (_callback) {
-        _callback(result, _items);
+        _callback(result, _type, _items);
     }
     _callback = nullptr;
     _done = true;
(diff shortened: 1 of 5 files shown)


jonathanreeves:
@julianoes @JonasVautherin just wanted to make a quick intro, as noted in the PR description, @tborglum is indeed a fellow team member at Elroy, we're working together to build out our MAVSDK integration. We'll both keep an eye on this PR in case something comes up. Thanks again for all your work on this project!

JonasVautherin (approved):

julianoes:
Thanks, looks like I need to fixup CI!

julianoes (approved):
Including tests! Very nice, thank you!

tborglum:
Thanks to you all for what you're doing with this project!

PR #2552, opened 2025-04-23
Update dockcross and fix CMAKE_PREFIX_PATH in workflows

JonasVautherin:
I was only setting CMAKE_PREFIX_PATH on cache hit, which I believe is wrong: it should always be set.

It may explain issues with "OpenSSL not found" by CMake over the years. It has been made clearly visible with the recent update of dockcross which systematically fails. This PR fixes it.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -32,11 +32,11 @@ jobs:
           key: ${{ github.job }}-${{ hashFiles('./third_party/**') }}
       - name: disable superbuild on cache hit
         if: steps.cache.outputs.cache-hit == 'true'
-        run: echo "superbuild=-DSUPERBUILD=OFF" >> $GITHUB_ENV && echo "cmake_prefix_path=-DCMAKE_PREFIX_PATH=$(pwd)/build/third_party/install" >> $GITHUB_ENV
+        run: echo "superbuild=-DSUPERBUILD=OFF" >> $GITHUB_ENV
       - name: Install dependencies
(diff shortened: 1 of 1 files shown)


julianoes (approved):
Nice catch!

julianoes:
Reading this again, I'm confused. Why should this be required? Wouldn't this be automatic when doing a superbuild?

JonasVautherin:
Maybe you're right. It's already done [here](https://github.com/mavlink/MAVSDK/blob/main/third_party/CMakeLists.txt#L7-L8), and I think before this PR the dependencies were building fine. It's the main build that was failing. So something was missing, but maybe it could go in the CMakeLists.

Feels like it should be done in the superbuild CMakeLists instead, and it should append to whatever the user set for CMAKE_PREFIX_PATH (instead of setting like it is now). And then maybe it should still do something in the dependencies CMakeLists if we want it to build standalone (e.g. if someone wants to build the dependencies only).

julianoes:
Maybe it was not using our FindOpenSSL.cmake for gRPC?
https://github.com/mavlink/MAVSDK/pull/2568/commits/23a7e653a2993869c8f6f17b26c1705d5d5a1d7f

JonasVautherin:
> Maybe it was not using our FindOpenSSL.cmake for gRPC?

But then why would it work by setting CMAKE_PREFIX_PATH? Because I think it helped 🤔

julianoes:
It must have... really odd.

PR #2562, opened 2025-05-06
[PATCH] Fix goto_location sending radians instead of degrees

RoseBeaupreARA:
As per the MAVLink documentation, `param4` is the desired heading in degrees, not radians:

https://mavlink.io/en/messages/common.html#MAV_CMD_DO_REPOSITION


julianoes (approved):
Whoops. You're right!

JonasVautherin (approved):

PR #2564, opened 2025-05-06
Fix manual control inverted r (yaw) axis

RoseBeaupreARA:
Protobuf message definition is defined as positive +CW, -CCW: https://github.com/mavlink/MAVSDK-Proto/blob/main/protos/manual_control/manual_control.proto#L50

MAVLink message is defined as +CCW, -CW: https://mavlink.io/en/messages/common.html#MANUAL_CONTROL

Therefore, to respect both specifications, sign must be reversed.

### DIFF
--- a/src/mavsdk_server/src/plugins/manual_control/manual_control_service_impl.h
+++ b/src/mavsdk_server/src/plugins/manual_control/manual_control_service_impl.h
@@ -165,7 +165,7 @@ public:
         }
 
         auto result = _lazy_plugin.maybe_plugin()->set_manual_control_input(
-            request->x(), request->y(), request->z(), request->r());
+            request->x(), request->y(), request->z(), -request->r());
 
         if (response != nullptr) {
             fillResponseWithResult(response, result);


hamishwillee:
@julianoes FWIW reminder the spec is a bit broken in respect to this message, so even if this is correct (I have not looked at it) we might have to ignore the spec. See https://github.com/mavlink/mavlink/pull/1922#issuecomment-2824145685

julianoes:
Oh dear. You're right. However, I don't think I can change the API without a major version increase, as it would be a breaking and very surprising change!

I think we will have to rename the fields to how they are actually used in reality and then merge that before v4.

julianoes:
@hamishwillee can we talk about this on the next MAVLink dev call? I would like to get to a conclusion. 

Edit: Please, thanks! :)

hamishwillee:
Of course - added to next meeting https://github.com/mavlink/mavlink/wiki/20251217-Dev-Meeting - feel free to just add whatever you like to that agenda yourself.

julianoes:
@RoseBeaupreARA so what happened is that PX4, ArduCopter, and QGC implement r/yaw inverted from the description, and hence that's what I tested and implemented in MAVSDK.

I am trying to "fix" the spec in https://github.com/mavlink/mavlink/pull/2398.

hamishwillee:
@julianoes FYI https://github.com/mavlink/mavlink/pull/2398 went in. So MAVSDK is correct to spec now and IMO this can be closed.

RoseBeaupreARA:
Thanks for taking care of this!

PR #2566, opened 2025-05-07
CI: try to add .deb package for Debian x64

julianoes:
Because Ubuntu 20.04 is no longer available via GitHub runners.


julianoes:
Too messy, I'm giving up. Stupid GitHub runners disappearing is such a pain.

PR #2568, opened 2025-05-07
CI: move to docker containers

julianoes:
The recent deprecation of the Ubuntu 20.04 image as a GitHub action runner showed us that we can't rely on the GitHub action images and should instead maintain our own docker images.


JonasVautherin:
But isn't Ubuntu 20.04 getting EOL?

PR #2573, opened 2025-05-14
Add incoming and outgoing message queue, fixing threading/locking issues

julianoes:
This is a bigger refactor where we no longer do work from the receiver threads but instead copy messages into a queue and work on them subsequently.

This is an attempt to prevent deadlocks where locks are acquired in different orders.

Additional, there are various small fixes to make all system and unit tests pass in address sanitizer and thread sanitizer.

I need to review this with fresh eyes again tomorrow.

Follow up to #2542

### DIFF
--- a/src/mavsdk/core/call_every_handler.cpp
+++ b/src/mavsdk/core/call_every_handler.cpp
@@ -61,24 +61,27 @@ void CallEveryHandler::remove(Cookie cookie)
 
 void CallEveryHandler::run_once()
 {
-    std::lock_guard<std::recursive_mutex> lock(_mutex);
-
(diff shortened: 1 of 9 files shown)


julianoes:
I already might have found one issue with this. When you unsubscribe within a callback to avoid getting called again and fulfilling your promise twice, it doesn't work because more than one messages have arrived in the meantime.

julianoes:
> I already might have found one issue with this. When you unsubscribe within a callback to avoid getting called again and fulfilling your promise twice, it doesn't work because more than one messages have arrived in the meantime.

I believe I mostly fixed this with the last commit where we signal a new message arriving.

PR #2582, opened 2025-05-23
More Windows fixes

julianoes:
And splitting the CI Windows build into mavsdk_server and library.


third_party/cares/CMakeLists.txt
```diff
@@ -12,6 +12,7 @@ list(APPEND CMAKE_ARGS
     "-DCARES_SHARED=OFF"
     "-DCARES_STATIC_PIC=ON"
     "-DCARES_BUILD_TOOLS=OFF"
+    "-DBUILD_SHARED_LIBS=OFF"
```
JonasVautherin, line 15: If this is used, shouldn't we remove `-DCARES_STATIC=ON` and `-DCARES_SHARED=OFF` above?

third_party/protobuf/CMakeLists.txt
```diff
@@ -8,6 +8,7 @@ list(APPEND CMAKE_ARGS
     "-DCMAKE_INSTALL_PREFIX:PATH=${CMAKE_INSTALL_PREFIX}"
     "-DCMAKE_TOOLCHAIN_FILE:PATH=${CMAKE_TOOLCHAIN_FILE}"
     "-DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}"
+    "-DBUILD_SHARED_LIBS=OFF"
```
JonasVautherin, line 11: Is that used, versus `-Dprotobuf_BUILD_SHARED_LIBS=OFF` below?

third_party/cares/CMakeLists.txt
julianoes: Good point :thinking:

PR #2587, opened 2025-05-27
CI: add way to cleanup docker images tagged with pr-*

julianoes:
In pull requests we upload docker images with tag pr-* (rather than latest). Every week, we should clean these up.


julianoes:
First manual dispatch worked: https://github.com/mavlink/MAVSDK/actions/runs/15266266779/job/42932392283

PR #2591, opened 2025-06-01
Fix camera_serve lockup

julianoes:
We tried to lock twice which doesn't work well.


jnomikos:
@julianoes I just tested this on my end on PX4 sitl and it fixed the issue for me

jnomikos (approved):

julianoes:
@jnomikos great, thanks for testing!

PR #2594, opened 2025-06-06
gripper: send command to autopilot by default

julianoes:
It's not obvious that this would go to a separate winch, so I'm removing this. Hopefully, this makes it work with PX4 like it does for QGC.

Ok with this @dakejahl?

### DIFF
--- a/src/mavsdk/plugins/gripper/gripper_impl.cpp
+++ b/src/mavsdk/plugins/gripper/gripper_impl.cpp
@@ -33,8 +33,6 @@ void GripperImpl::grab_async(uint32_t instance, const Gripper::ResultCallback ca
     command.params.maybe_param1 = static_cast<float>(instance);
     command.params.maybe_param2 = static_cast<float>(Gripper::GripperAction::Grab);
 
-    command.target_component_id = MAV_COMPONENT::MAV_COMP_ID_WINCH; // TODO
-
     _system_impl->send_command_async(
         command, [this, callback](MavlinkCommandSender::Result result, float) {
             command_result_callback(result, callback);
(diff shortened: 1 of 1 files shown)


dakejahl:
I don't have any skin in this game

hamishwillee:
FWIW I think this is correct.

A gripper is not a winch, but even if it was, a gripper might be attached to the flight stack. We shouldn't assume that any component is assigned to a particular ID.

I don't think this can be done properly in MAVLink right now, based on our learnings from cameras and gimbals.
What should probably happen:
- A new MAV_TYPE for grippers
- A new protocol bit for gripper support for use by Autopilots
- A thing that wants to send commands to grippers should monitor for that MAV_TYPE and also the protocol bit.
- The gripper instance number should be reinterpretted as id 1-6 is an autopilot connected gripper and any other value is a component id that is used when in missions if specified.

@julianoes WOrth us doing a mavlink update?

julianoes:
@hamishwillee I appreciate you bringing this up.

I think it's the same old problem and we don't have a solution for it.

> What should probably happen:
>   - A new MAV_TYPE for grippers

That's easy enough.

>  - A new protocol bit for gripper support for use by Autopilots

Ok, also easy.

>  - A thing that wants to send commands to grippers should monitor for that MAV_TYPE and also the protocol bit.

Hmm, ok. That needs a bit of logic, but that's ok.

>  - The gripper instance number should be reinterpretted as id 1-6 is an autopilot connected gripper and any other value is a component id that is used when in missions if specified.

Yeah, I suppose that makes sense.

---

I was going into this trying to disprove you that this is not easy and we shouldn't try but I think you've laid out an ok path.

hamishwillee:
@julianoes  
> I was going into this trying to disprove you that this is not easy and we shouldn't try but I think you've laid out an ok path.

This is why I like design patterns. This is the camera and gimbal story turned into a gimbal story.

PR #2602, opened 2025-06-26
Docs: Fix info box general_usage.md

matthieu-db:


matthieu-db:
Not sure about the syntax, but anyway it's wrong now so I guess this wouldn't make it much worse :upside_down_face:

hamishwillee (approved):
Yes, that's right. Thank you.

PR #2605, opened 2025-07-04
core: change Android tmp dir, make it configurable

julianoes:
When Android tries to access the default `/tmp` temporary directory it crashes due to missing permissions.

As a workaround, let's try to use `/data/local/tmp` instead, and if that doesn't work, make it configurable within JNI for the user.

Hopefully fixes https://github.com/mavlink/MAVSDK-Java/issues/206.

### DIFF
--- a/src/mavsdk/core/fs_utils.cpp
+++ b/src/mavsdk/core/fs_utils.cpp
@@ -20,6 +20,13 @@
 
 namespace mavsdk {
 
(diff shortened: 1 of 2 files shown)


julianoes:
@JonasVautherin should be configurable as mentioned in the PR description :arrow_up:, at link time.

julianoes:
Going to merge it and then try it out.

PR #2607, opened 2025-07-07
core: add a few potential safety fixes

julianoes:
These are some attempts at safe-guarding within the FTP client, according to Claude anyway.


julianoes:
Nevermind, some of these checks are redundant, and others might not actually be correct because data doesn't necessarily have to be 0 terminated, pretty sure.

PR #2609, opened 2025-07-08
Switch from tinyxml2 to rapidxml

julianoes:
The plan is to incorporate libmav which requires rapidxml. So the hope is that we can consolidate on one xml library.


third_party/rapidxml/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+endforeach()
+
+ExternalProject_add(
+    rapidxml
+    GIT_REPOSITORY https://github.com/timniederhausen/rapidxml
+    GIT_TAG master
+    PREFIX rapidxml
+    CMAKE_ARGS ${CMAKE_ARGS}
+        -DRAPIDXML_BUILD_TESTS=OFF
+        -DRAPIDXML_INSTALL=ON
```
JonasVautherin: Why not setting them with all the CMAKE_ARGS above? :thinking:

third_party/rapidxml/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+    # Custom install step to ensure headers are properly installed
+    INSTALL_COMMAND ${CMAKE_COMMAND} --build <BINARY_DIR> --target install
+        COMMAND ${CMAKE_COMMAND} -E copy_if_different <SOURCE_DIR>/rapidxml.hpp ${CMAKE_INSTALL_PREFIX}/include/rapidxml/rapidxml.hpp
+        COMMAND ${CMAKE_COMMAND} -E copy_if_different <SOURCE_DIR>/rapidxml_iterators.hpp ${CMAKE_INSTALL_PREFIX}/include/rapidxml/rapidxml_iterators.hpp
+        COMMAND ${CMAKE_COMMAND} -E copy_if_different <SOURCE_DIR>/rapidxml_print.hpp ${CMAKE_INSTALL_PREFIX}/include/rapidxml/rapidxml_print.hpp
+        COMMAND ${CMAKE_COMMAND} -E copy_if_different <SOURCE_DIR>/rapidxml_utils.hpp ${CMAKE_INSTALL_PREFIX}/include/rapidxml/rapidxml_utils.hpp
+    )
```
JonasVautherin: Given that rapidxml is a header-only library, isn't the whole point of the CMakeLists to install the headers somewhere? Here it seems like we use a fork of rapidxml just because it adds a CMakeLists, but then we don't need that CMakeLists because anyway we copy the headers manually :thinking:.

At that point, can't we just go for the "most official" rapidxml?

JonasVautherin (commented):
I have two concerns regarding libmav:

1. It doesn't handle dependencies nicely. It claims that it has "no dependencies" but in fact includes rapidxml and picosha in a way that is not exactly cooperative (What if you want to use the system packages for those dependencies? What if you already depend on rapidxml/picosha transitively?).
2. It loads the MAVLink message definitions at runtime. Meaning that those files will have to be made available to MAVSDK somehow. Not sure how we do that on Android/iOS in a practical way. My guess is that the easiest will be to embed the files into the C++ binaries just like we were doing for the camera definitions. Which kinda defeats the purpose of libmav making it easier to develop with custom mavlink messages.

And then won't it require massive changes in the MAVSDK codebase? And doesn't it mean that users including "unexpected" mavlink definitions will make MAVSDK crash at runtime (e.g. if the definition of the takeoff message is incompatible with what MAVSDK expects, then MAVSDK won't know about it until runtime, where libmav actually complains, right?).

Anyway, not my call :man_shrugging:, but it feels risky and I am not sure what it will bring :innocent:.

third_party/rapidxml/CMakeLists.txt
julianoes: Because that's what Claude did :expressionless:. Thanks.

third_party/rapidxml/CMakeLists.txt
julianoes: The cmake stuff clearly isn't quite right, so maybe we shouldn't go for a fork like this indeed.

julianoes:
@JonasVautherin regarding your concerns.

1. Yes. I don't like that rapidxml is embedded within libmav. I'm considering making a fork that removes the dependencies from tree, so we don't end up with duplicate headers. I have to check what to do about picosha.
2. I haven't thought about how/when it loads the definitions. That's an interesting point. It's certainly possible to embed it within the file but it's one of those steps that's potentially a pain with cross compilation as you have to generate stuff before the build.
3. Absolutely agree that we don't want to switch over the MAVSDK internals and plugins to the runtime defined one, unless we could have the runtime defined ones run at compile time, so basically `constexpr` if used at compile time and otherwise runtime. This would be quite neat but I'm not sure how or whether it would be possible.
For now, my plan was to use it only for the dynamic MavlinkPassthrough which can then be used in language wrappers, and allows custom MAVLink to be used easily without having to fork and mess with the internal mavlink includes. And those two points would actually be very beneficial for many users (or potential users).

One more comment:
My attempt here was to figure out if/how this was feasible. I'd also consider just rewriting (with inspiration and crediting) the part of libmav that we need.

julianoes:
It turns out rapidxml is not a good idea when used without exceptions!

https://github.com/timniederhausen/rapidxml/blob/ce0fb9a6cb42270ec5deb77042c823d5e2361322/rapidxml.hpp#L56-L71

PR #2610, opened 2025-07-12
Integrate parts of libmav into MAVSDK and add MavlinkDirect plugin

julianoes:
This is work towards MavlinkPassthrough for language wrappers called MavlinkDirect.

The work is based on a [hard fork](https://github.com/julianoes/libmav) of [libmav](https://github.com/Auterion/libmav). The fork is required because:
- MAVSDK doesn't work with C++ exceptions
- rapidxml is replaced with tinyxml2 as rapidxml is not the right choice without exceptions
- The connection classes are not required.
- Header only is a poor way to manage dependencies and unnecessarily increases compile time for incremental builds.

The MavlinkDirect allows you to:
- [x] Send and receive MAVLink messages based on just the message name and fields/data as JSON (strings which are easily forwarded to and parsed in language wrappers).
- [x] Add custom MAvlink xml to send and receive messages

While this seems to be working end to end in system tests, most of the work is Claude Code generated slop, so I will have to review this properly myself.

### DIFF
--- a/third_party/libmavlike/CMakeLists.txt
+++ b/third_party/libmavlike/CMakeLists.txt
@@ -28,8 +28,8 @@ endforeach()
 
 ExternalProject_add(
     libmavlike
-    GIT_REPOSITORY https://github.com/julianoes/libmav
-    GIT_TAG 6566a060814d06942c563db8d8c7c846a7089fc0
+    GIT_REPOSITORY https://github.com/julianoes/libmavlike
+    GIT_TAG ccc3addc341ae907de7a2937ebf51c6ff2cd30fa
     PREFIX libmavlike
     CMAKE_ARGS ${CMAKE_ARGS}
     )


src/mavsdk/core/CMakeLists.txt
```diff
@@ -12,6 +12,29 @@ get_target_property(
 configure_file(version.h.in version.h)
 configure_file(include/mavsdk/mavlink_include.h.in include/mavsdk/mavlink_include.h)
 
+# Generate embedded XML header for libmav
+set(MAVLINK_MINIMAL_XML "${DEPS_INSTALL_PATH}/include/mavlink/message_definitions/v1.0/minimal.xml")
+set(MAVLINK_STANDARD_XML "${DEPS_INSTALL_PATH}/include/mavlink/message_definitions/v1.0/standard.xml")
+set(MAVLINK_COMMON_XML "${DEPS_INSTALL_PATH}/include/mavlink/message_definitions/v1.0/common.xml")
+set(MAVLINK_ARDUPILOTMEGA_XML "${DEPS_INSTALL_PATH}/include/mavlink/message_definitions/v1.0/ardupilotmega.xml")
```
JonasVautherin, line 19: I think this creates a hard dependency on the superbuild. Maybe it would be worth having the path as an option, e.g. `MAVLINK_MSG_DEFINITIONS_PATH`, and the superbuild would by default set it to `${DEPS_INSTALL_PATH}/include/mavlink` or something?

third_party/libmavlike/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
+cmake_minimum_required(VERSION 3.13)
+
+project(external-libmavlike)
```
JonasVautherin, line 3: I like the name, `libmavlike` :grimacing:

src/mavsdk/core/CMakeLists.txt
```diff
@@ -12,6 +12,29 @@ get_target_property(
...
+set(EMBEDDED_XML_HEADER "${CMAKE_CURRENT_BINARY_DIR}/embedded_mavlink_xml.h")
+
+add_custom_command(
```
JonasVautherin, line 22: I am always nervous with `add_custom_command`. CMake is really not a super nice scripting language, and those always look cryptic to me. It took me way longer than I'm comfortable with to realise that this really just creates **one header file** called `embedded_mavlink_xml.h` that gets included later and that doesn't change unless we update the mavlink dependency.

The risk I see is going the road of the PX4 build system, which is so custom that I don't consider it to be CMake anymore, and it is always painful for me to read/use.

What about doing it like we do for the proto files, i.e. generate it once and version it? Just like you only need to run `generate_from_protos.sh` when you modify the protos, you would only have to run `generate_embedded_mavlink.sh` when you modify mavlink. And nobody should ever have to do it except those who update libmavlink in MAVSDK (so... pretty much the maintainers).

So that it keeps the CMakeLists readable, and I'm pretty sure that a shell script would be a lot more readable than this custom command :see_no_evil:.

What do you think?

JonasVautherin (commented):
Very nice to have this feature! :rocket: :rocket:

src/mavsdk/core/CMakeLists.txt
julianoes: I thought that too but then it worked in CI...

third_party/libmavlike/CMakeLists.txt
julianoes: I might rename my fork as well.

src/mavsdk/core/CMakeLists.txt
julianoes: I agree in terms of readability but would you then check the generated header file in, or generate it each time?

JonasVautherin: I would check it in, like we do with the protobuf files. Maybe in a `generated/` folder, to be consistent?

julianoes: We could do that but I want to make it easy for users to swap this out as well. I guess we can make it, so you need to load "common" manually before using it.

JonasVautherin: I thought, a script like `./tools/generate_mavlink_direct.sh` that would generate the file and put it in the right place (`something/generated/mavlink_embedded_headers.h`). And that file is versioned. So that it doesn't require a change in the CMakeLists (except for installing the header), users don't have to care about it but they can always update `libmavlink` and regenerate the file with the `./tools/generate_mavlink_direct.sh` script.

Exactly like we do now with the proto files.

Or does that sound to complicated?

julianoes: I'm somewhat tempted to leave it as is because it has the advantage that the MAVLink version included will match the one from the headers, and if someone changes the C headers, both change accordingly, and don't get out of sync.

JonasVautherin: Sure, your call. I'm just always nervous when I see `add_custom_command` in CMake, but I guess it won't go as far as in did in PX4 anyway 🙈.

PR #2614, opened 2025-07-15
CI: remove hunter support

julianoes:
I don't actually like if dependency management has to be embedded within the build. Imagine if we had to put conan, vcpkg, hunter, brew, pacman, etc. you name it, all within cmake. It would be a mess.

We strive for just simple plain cmake instead.

The reason this is removed now is that I need to add the dependency picosha2 and I can't find it in hunter. Also, I haven't seen any signs that hunter support is actually (still) used out there in the wild.

@bazfp are you relying on hunter support? If so, I'm happy to leave this in but I'd need help how to add picosha2 (or something like it).


julianoes:
Actually, it turns out we just need sha256 from picosha2, but since we already have openssl (or boringssl) in the build, we can just use that.

bazfp:
We have actually moved to Conan builds for our newer repos as hunter development seems to have ceased.

I may look to add a PR with Conan support in future from my internal fork

JonasVautherin (approved):
:tada: I'm all for removing it, especially given that @bazfp doesn't need it anymore and that apparently Hunter is not maintained anymore.

> I may look to add a PR with Conan support in future from my internal fork

@bazfp: but Conan already works with `find_package`, doesn't it? There shouldn't be anything to change in MAVSDK, you can just have Conan install the dependencies somewhere and point CMAKE_PREFIX_PATH to that location when you build MAVSDK, right?

bazfp:
> @bazfp: but Conan already works with `find_package`, doesn't it? There shouldn't be anything to change in MAVSDK, you can just have Conan install the dependencies somewhere and point CMAKE_PREFIX_PATH to that location when you build MAVSDK, right?

We just add a conan recipe (conanfile.py) for the dependencies that works with our internal conan tooling

``` ["boost/1.86.0", "tinyxml2/[~9]","jsoncpp/[~1]", "cmake/[~3]","libcurl/[~8]", "zlib/[~1]"],``` etc

julianoes:
Hi @bazfp :wave:
I think adding a conanfile.py would be fine.

PR #2622, opened 2025-07-23
Build error windows without JOM

JohanOlsson99:
Fixed a build error on windows when not using JOM and updated the windows build documentation.


JonasVautherin (approved):
We had to do it for the BUILD_COMMAND, so I guess it makes sense 😁

PR #2628, opened 2025-07-26
CI: actually set CC and CXX vars

julianoes:
This didn't actually seem to work.


julianoes:
Now it's correct, yay.

PR #2637, opened 2025-08-03
Enable forwarding unknown messages

julianoes:
This fixes the problem that MAVSDK did not forward messages which it had not seen before. This could happen when:
- A message is added in a later version of MAVLink whchi MAVSDK hasn't shipped with,
- A custom message needs to be forwarded that MAVSDK doesn't know about.

### DIFF
--- a/src/mavsdk/core/serial_connection.cpp
+++ b/src/mavsdk/core/serial_connection.cpp
@@ -263,52 +263,11 @@ ConnectionResult SerialConnection::stop()
 
 std::pair<bool, std::string> SerialConnection::send_message(const mavlink_message_t& message)
 {
-    std::pair<bool, std::string> result;
-
-    if (_serial_node.empty()) {
-        LogErr() << "Dev Path unknown";
-        result.first = false;
-        result.second = "Dev Path unknown";
-        return result;
-    }
-
-    if (_baudrate == 0) {
-        result.first = false;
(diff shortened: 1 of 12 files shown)


src/mavsdk/core/serial_connection.cpp
```diff
@@ -410,4 +369,48 @@ int SerialConnection::define_from_baudrate(int baudrate)
...
+        result.first = false;
+        result.second = "Baudrate unknown";
+        return result;
+    }
+
+    int send_len;
+#if defined(LINUX) || defined(APPLE)
+    send_len = static_cast<int>(write(_fd, bytes, length));
+#else
+    if (!WriteFile(_handle, bytes, static_cast<DWORD>(length), LPDWORD(&send_len), NULL)) {
```
JonasVautherin, line 393: Isn't this a Windows function? 🤔

julianoes: I believe so. It was there before, it just moved.

JonasVautherin: Oh you're right. So I guess we don't use this function on iOS or Android 🤷

julianoes: No idea.

hamishwillee:
This is cool!

PR #2640, opened 2025-08-13
cmake: only link stdc++fs for GCC < 9

grembeter:
Newer GCC versions (>= 9) include the C++17 filesystem library in libstdc++ by default, so linking stdc++fs explicitly is unnecessary and may cause build errors.

Update the CMake condition to link stdc++fs only when using GCC < 9, while keeping the existing exclusions for Apple, Android, and MSVC.

### DIFF
--- a/src/mavsdk/CMakeLists.txt
+++ b/src/mavsdk/CMakeLists.txt
@@ -23,7 +23,8 @@ target_link_libraries(mavsdk
     mav::mav
 )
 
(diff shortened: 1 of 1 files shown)


julianoes (approved):
Thanks. I wonder if we actually support GCC < 9 at this point.

reedev:
> Thanks. I wonder if we actually support GCC < 9 at this point.

Fyi on the Jetson Nano we have gcc 8 when using via the repo.

PR #2644, opened 2025-08-19
CI: move from coveralls to codecov

julianoes:
Coveralls.io has been very flaky recently.


julianoes:
Alright, aborting this as coveralls.io seems back and codecov seems to be already enabled for mavlink/qgroundcontrol and I can't enable it for this repo, nevermind.

PR #2646, opened 2025-08-21
mavlink_direct: handle NaN correctly

julianoes:
JSON doesn't support NaN, so we need to use null instead.

Also, we need to assume arrays have all the same types, and not check the type of each array field. This way we can correctly use 0 for int and null for float/double.

### DIFF
--- a/src/mavsdk/core/libmav_receiver.cpp
+++ b/src/mavsdk/core/libmav_receiver.cpp
@@ -6,6 +6,7 @@
 #include <variant>
 #include <cstring>
 #include <sstream>
+#include <cmath>
 #include "log.h"
 
 namespace mavsdk {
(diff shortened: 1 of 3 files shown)


julianoes:
I have to review this with fresh eyes tomorrow. I might still switch to using jsoncpp instead of this manual JSON encoding and decoding madness.

JonasVautherin:
Yeah it feels like jsoncpp would make sense here 👍

julianoes:
The problem I initially had using jsoncpp was that it doesn't let you order the fields and so the JSON output failed the test. I didn't want to have to use jsoncpp in the test as well (although maybe that would be fine?) and so Claude just came up with this manual generation.

Issue #2663, opened 2025-09-25
Invalid index not handled properly in set_current_mission_item_async, ultimately causing segmentation fault


irajkovic:
This is a compound issue that ultimately leads to segmentation fault.

**Main issue**:
If `MissionRawImpl::set_current_mission_item` (synchronous) is called with an invalid index, the call will be forwarded to the `set_current_mission_item_async`, which checks for index validity, **but it doesn't handle it**. Instead, it only calls the provided callback (lambda from `set_current_mission_item` with `InvalidArgument`) and **proceeds to call `set_current_item_async`** which will ultimately call the same user callback a second time. 

```
if (index < 0 || index >= _mission_progress.last.total) {
        _system_impl->call_user_callback([callback]() {
            if (callback) {
                callback(MissionRaw::Result::InvalidArgument); // <-- we called the callback first time
                return; // <-- this return applies to the inner lambda
            }
        });
    // <-- we probably want to return here instead, otherwise, function execution continues below
    }
    _system_impl->mission_transfer_client().set_current_item_async(
        index,
        _system_impl->get_system_id(),
        [this, callback](MavlinkMissionTransferClient::Result result) {
            auto converted_result = convert_result(result);
            _system_impl->call_user_callback([callback, converted_result]() {
                if (callback) {
                    callback(converted_result); // <-- and now we call the same callback again!
                }
            });
        });
```

**Second issue**
User callback defined in `MissionRawImpl::set_current_mission_item` works properly for the first invocation. However, after that, `set_current_mission_item` will return. After that, the callback lambda is invoked the second time. By this time, `prom` local variable was already dropped from stack, but lambda holds a reference to it anyway and tries to set it. This causes segmentation fault.

```
MissionRaw::Result MissionRawImpl::set_current_mission_item(int index)
{
    auto prom = std::promise<MissionRaw::Result>();
    auto fut = prom.get_future();

    set_current_mission_item_async(
        index, [&prom](MissionRaw::Result result) { prom.set_value(result); }); // <-- Lambda will use local variables from stack, which is fine as long as function didn't return
    return fut.get(); // <-- This will only block until the first invocation of the provided lambda callback
}
```

**Third issue**
Since we are invoking callback multiple times, it will try to `set_value` on the promise a second time, which is undefined behavior. Although, the code would crash due to second issue already.

irajkovic:
[...]

PR #2664, opened 2025-09-25
Fix: Only set the current item if index is validated

irajkovic:
Fix for #2663.

Makes sure `set_current_mission_item_async` returns early to avoid calling callback a second time.


JonasVautherin (approved):
LGTM, thanks!

julianoes (approved):
Ooh, yes. Thanks!

irajkovic:
Thank you for merging.

PR #2666, opened 2025-09-26
Added checks for the mavlink definitions

### LINKED ISSUE #2665: Proper way to build with superbuild off?
Hi,

When I try to build MAVSDK with superbuild off, from v3.8.0 I get this message:
gmake[2]: *** No rule to make target 'third_party/install/include/mavlink/message_definitions/v1.0/ardupilotmega.xml', needed by 'src/mavsdk/core/embedded_mavlink_xml.h'.  Stop.
It makes sense since this one, and the others, are not there.

I read your discussion @julianoes and @JonasVautherin when the MavlinkDirect plugin was first added, but could not find a way how to point to a different location. All I see are the hard-coded set(MAVLINK_MINIMAL_XML... in the CMakeLists.txt and when I change these to point to my location, MAVSDK is built properly.

If, like @JonasVautherin suggested, there is a way to specify this location via a cmake flag, it would be fine I guess. Not sure how that should be done so I would like to ask you... thanks!

Update: ok, that wasn't so hard, created a PR to set the variables only when not specified

reedev:
Resolves #2665

### DIFF
--- a/src/mavsdk/plugins/mission_raw/mission_raw_impl.cpp
+++ b/src/mavsdk/plugins/mission_raw/mission_raw_impl.cpp
@@ -558,9 +558,9 @@ void MissionRawImpl::set_current_mission_item_async(
         _system_impl->call_user_callback([callback]() {
             if (callback) {
                 callback(MissionRaw::Result::InvalidArgument);
-                return;
             }
         });
+        return;
     }
 
     _system_impl->mission_transfer_client().set_current_item_async(


JonasVautherin:
Thanks!

When are those defined? By the superbuild? Just to understand 😊. I guess I'm wondering if those should not be exposed as CMake options?

reedev:
Hi,

I am not sure what it means 'by the superbuild' but the url to the 4 xml files are set in src/mavsdk/core/CMakeLists.txt and now I added a check to set them only when they are not already set (via cmake). Of course, it could be nicer to use a single path to the files. We can easily change that.

JonasVautherin:
Right, sorry it was late an I wasn't clear 😅.

I was wondering if we should not take this as an opportunity to add them as options, next to `MAVLINK_DIALECT` here: https://github.com/mavlink/MAVSDK/blob/main/CMakeLists.txt#L7. So that it makes it clear that they can be set manually

reedev:
Sure, good suggestion to keep things consistent. I will try it out and make another PR (not right now, at a later moment).

PR #2667, opened 2025-09-30
Fix updating mission item total

irajkovic:
Total number of items in a mission should only be updated when mission is downloaded. Downloading geofence or rally points should not affect this total.

### DIFF
--- a/src/mavsdk/plugins/mission_raw/mission_raw_impl.cpp
+++ b/src/mavsdk/plugins/mission_raw/mission_raw_impl.cpp
@@ -172,6 +172,12 @@ void MissionRawImpl::upload_mission_items_async(
         [this, callback, int_items](MavlinkMissionTransferClient::Result result) {
             auto converted_result = convert_result(result);
             auto converted_items = convert_items(int_items);
+
+            if (converted_result == MissionRaw::Result::Success) {
+                std::lock_guard<std::mutex> lock(_mission_progress.mutex);
+                _mission_progress.last.total = int_items.size();
+            }
+
             _system_impl->call_user_callback([callback, converted_result, converted_items]() {
                 if (callback) {
                     callback(converted_result);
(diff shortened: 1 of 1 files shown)


irajkovic:
Just as a clarification, function `convert_items` is also called when downloading geofence, so number of points in geofence would be used to set the number of items in a mission.

JonasVautherin:
Thanks! Would you mind fixing the style with `./tools/fix_style.sh`?

irajkovic:
Pushed forced after reformatting the code. I only formatted my changes though, clang-format also reformatted unrelated file: ./`tools/fix_style.sh` that I didn't want to include in this commit as it's unrelated. Let me know if it's ok now.

src/mavsdk/plugins/mission_raw/mission_raw_impl.cpp
```diff
@@ -172,6 +172,12 @@ void MissionRawImpl::upload_mission_items_async(
         [this, callback, int_items](MavlinkMissionTransferClient::Result result) {
             auto converted_result = convert_result(result);
             auto converted_items = convert_items(int_items);
+
+            {
```
julianoes:
```suggestion
            if (converted_result == MissionRaw::Result::Success) {
```

src/mavsdk/plugins/mission_raw/mission_raw_impl.cpp
```diff
@@ -290,6 +296,12 @@ void MissionRawImpl::download_mission_async(const MissionRaw::DownloadMissionCal
             std::vector<MavlinkMissionTransferClient::ItemInt> items) {
             auto converted_result = convert_result(result);
             auto converted_items = convert_items(items);
+
+            {
```
julianoes:
```suggestion
            if (converted_result == MissionRaw::Result::Success) {
```

src/mavsdk/plugins/mission_raw/mission_raw_impl.cpp
irajkovic: Thanks for suggestion, added to the original commit.

src/mavsdk/plugins/mission_raw/mission_raw_impl.cpp
irajkovic: Same as above.

julianoes:
Thanks @irajkovic.

PR #2670, opened 2025-09-30
Added capability to specify the path to the mavlink xml files

reedev:


reedev:
This one replaces the previous change for specifying the xml filepaths to the mavlink xml files. This change shows up as cmake optional path.

I am sorry for mixing up with the other change, lack of experience with PRs here...

third_party/tinyxml2/CMakeLists.txt
```diff
@@ -10,7 +10,7 @@ list(APPEND CMAKE_ARGS
     "-DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}"
     "-DCMAKE_POSITION_INDEPENDENT_CODE=ON"
     "-DBUILD_SHARED_LIBS=OFF"
-    "-DBUILD_TESTS=OFF"
+    "-DBUILD_TESTING=OFF"
```
JonasVautherin: This should stay in your other PR about tinyxml2 😇

CMakeLists.txt
```diff
@@ -39,6 +39,8 @@ message(STATUS "Version patch: ${MAVSDK_VERSION_PATCH}")
 set(MAVSDK_SOVERSION_STRING ${MAVSDK_VERSION_MAJOR})
 set(MAVSDK_VERSION_STRING ${MAVSDK_VERSION_MAJOR}.${MAVSDK_VERSION_MINOR}.${MAVSDK_VERSION_PATCH})
 
+set(MAVLINK_XML_PATH "" CACHE PATH "Where the mavlink xml files are located.")
```
JonasVautherin, line 42: Yeah that good to me, I don't think that people would want to set all 4 mavlink paths 👍

src/mavsdk/core/CMakeLists.txt
```diff
@@ -13,10 +13,17 @@ configure_file(version.h.in version.h)
...
+    set(MAVLINK_MINIMAL_XML "${MAVLINK_XML_PATH}/minimal.xml")
+    set(MAVLINK_STANDARD_XML "${MAVLINK_XML_PATH}/standard.xml")
+    set(MAVLINK_COMMON_XML "${MAVLINK_XML_PATH}/common.xml")
+    set(MAVLINK_ARDUPILOTMEGA_XML "${MAVLINK_XML_PATH}/ardupilotmega.xml")
+else()
+    set(MAVLINK_MINIMAL_XML "${DEPS_INSTALL_PATH}/include/mavlink/message_definitions/v1.0/minimal.xml")
+    set(MAVLINK_STANDARD_XML "${DEPS_INSTALL_PATH}/include/mavlink/message_definitions/v1.0/standard.xml")
+    set(MAVLINK_COMMON_XML "${DEPS_INSTALL_PATH}/include/mavlink/message_definitions/v1.0/common.xml")
+    set(MAVLINK_ARDUPILOTMEGA_XML "${DEPS_INSTALL_PATH}/include/mavlink/message_definitions/v1.0/ardupilotmega.xml")
+endif()
```
JonasVautherin: I'm nitpicking a bit, but wouldn't it be a bit easier if we did something like:

```CMake
if(NOT MAVLINK_XML_PATH)
    set(MAVLINK_XML_PATH "${DEPS_INSTALL_PATH}/include/mavlink/message_definitions/v1.0")
endif()

set(MAVLINK_MINIMAL_XML "${MAVLINK_XML_PATH}/minimal.xml")
set(MAVLINK_STANDARD_XML "${MAVLINK_XML_PATH}/standard.xml")
set(MAVLINK_COMMON_XML "${MAVLINK_XML_PATH}/common.xml")
set(MAVLINK_ARDUPILOTMEGA_XML "${MAVLINK_XML_PATH}/ardupilotmega.xml")
```

What do you think?

reedev: yes, good point indeed. I've gone ahead and made the PR. Hope that this is the right way to do it (still unfamiliar with github PRs)

reedev:
Ok, thanks for your comments. I have cleaned up the PR and I think it's now ready to go. Please review.

reedev:
Perfect, glad to be able to contribute something. Great work you are doing.

src/mavsdk/core/CMakeLists.txt
julianoes: Hang on, we should not have different xmls, actually! That's because if you choose common.xml it will include minimal.xml and standard.xml.

And if you select ardupilotmega.xml, it includes common.xml. I guess we don't follow these includes and that's why we need to set it manually. But probably something to improve in the future.

reedev: Indeed, xml files include 'each other'. I am confused now, does that have to do with my changes? I guess not, just double checking...

JonasVautherin: @julianoes: is that blocking for this PR, or should we merge it?

julianoes: Assuming the PR doesn't change the behaviour, not a blocker.

reedev: Right, that's the case, the PR does not change the behaviour.

JonasVautherin: Yeah I think it's fine 👍

PR #2673, opened 2025-10-01
dependencies: libevents should be built with the superbuild

JonasVautherin:
I was skimming through the dependencies, and this seems wrong.

Why would libevent be built outside the superbuild? 🤔


julianoes:
Looks like we need to fix macOS CI.

julianoes (dismissed):
I think that's right!

julianoes:
Hm, this means our non-superbuild gets more and more complicated. You now need to build libevents, and mavlink manually, right?

JonasVautherin:
Yes, but that's the cost of dependencies... there is no way around that, right?

julianoes:
I suppose so. We should add the manual non superbuild build back to CI.

JonasVautherin:
Should we merge this or do you think it breaks something? IMO the current version is wrong.

julianoes:
Sorry, missed it.

PR #2674, opened 2025-10-01
Update openssl, boringssl and curl

JonasVautherin:
While I'm at it, I may as well update a few dependencies 😇


JonasVautherin:
Hmm this test consistently fails here, I'm not sure how it is related to openssl or curl 🤔

JonasVautherin:
Seems unrelated, happens in #2667 as well 🙈

PR #2680, opened 2025-10-07
mavlink_direct: handle (u)int > 2^32

### LINKED ISSUE #2679: MavlinkDirect API doesn't seem to support uint64_t fields when they have large number
MavlinkDirect API doesn't seem to support uint64_t fields when they have large number, for example GPS2_RAW message:
```
JSON: {"alt_ellipsoid":45358,"lat":201849905,"lon":108122620,"time_usec":1759841673570912}
[03:54:33|Warn ] Field time_usec is not a float or double field (mavlink_direct_impl.cpp:292)
```

julianoes:
It turns out jsoncpp uses isUInt for numbers below 2^32 and isUInt64 for numbers above 2^32.

Fixes #2679.

### DIFF
--- a/src/mavsdk/plugins/mavlink_direct/mavlink_direct_impl.cpp
+++ b/src/mavsdk/plugins/mavlink_direct/mavlink_direct_impl.cpp
@@ -233,69 +233,20 @@ bool MavlinkDirectImpl::json_to_libmav_message(
         const Json::Value& field_value = json[field_name];
 
         // Convert JSON values to appropriate types and set in message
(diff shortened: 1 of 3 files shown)


reedev:
Nice cleanup. Did you test with a small number in a uin64 for instance? That's how I encountered the issue.

julianoes:
> Nice cleanup. Did you test with a small number in a uin64 for instance? That's how I encountered the issue.

I'm pretty sure a small number should still work, I just tested it using a system_test and it seemed to pass.

PR #2681, opened 2025-10-07
Fix unix epoch async rate setter conflict with Ardupilot

JoC2000:
Corrects an issue where `set_rate_unix_epoch_time_async` requested the wrong MAVLINK message. Now it uses `SYSTEM_TIME` to march the message handler.
Tested with SITL

### DIFF
--- a/src/mavsdk/plugins/telemetry/telemetry_impl.cpp
+++ b/src/mavsdk/plugins/telemetry/telemetry_impl.cpp
@@ -571,7 +571,7 @@ void TelemetryImpl::set_rate_unix_epoch_time_async(
     double rate_hz, Telemetry::ResultCallback callback)
 {
     _system_impl->set_msg_rate_async(
-        MAVLINK_MSG_ID_UTM_GLOBAL_POSITION,
+        MAVLINK_MSG_ID_SYSTEM_TIME,
         rate_hz,
         [callback](MavlinkCommandSender::Result command_result, float) {
             command_result_callback(command_result, callback);


JonasVautherin (commented):
Is it requesting the wrong message, or is it a difference between PX4 and Ardupilot?

JoC2000:
I think it's the wrong message, `set_rate_unix_epoch_time_async` is asking for UTM_GLOBAL_POSITION (430) but the message handler expects SYSTEM_TIME (2). 
I just tested with both px4 and Ardupilot SITL, and the rate setter fails on both. 
However, for PX4 `subscribe_unix_epoch_time` works and give a correct timestamp. 
On Ardupilot no timestamp is received, likely because there is an warning of no ap message for ID 430.

After this modification, the rate setter works for both PX4 and Ardupilot SITL.

julianoes:
@JonasVautherin we used to use UTM_GLOBAL_POSITION for UTC time but changed it without fixing the set_rate part.

julianoes (approved):

reedev:
Question/remark: I know that I had an issue with this functionality in the past. Getting the time via this subscription did not produce the correct time, if I remember correctly. It was a subtle thing, that's what I remember. I can/will test this on a real setup once you @julianoes create a new release.

reedev:
@julianoes  fyi I tested with a real setup with the main branch and I do still get the correct unix epoch time.

julianoes:
Nice, thanks for testing @reedev!

PR #2682, opened 2025-10-08
core: add raw connection

julianoes:
This adds a "raw connection" which accepts and delivers raw bytes. This can be handy when a custom transport other than the included UDP/TCP/serial needs to be used.

### DIFF
--- a/src/system_tests/raw_bytes.cpp
+++ b/src/system_tests/raw_bytes.cpp
@@ -1,5 +1,6 @@
 #include "log.h"
 #include "mavsdk.h"
+#include "plugins/telemetry_server/telemetry_server.h"
 #include "mavlink_include.h"
 #include <atomic>
 #include <chrono>
@@ -9,39 +10,36 @@
 
 using namespace mavsdk;
 
-TEST(SystemTest, RawBytesHeartbeat)
+TEST(SystemTest, RawBytesSendReceive)
 {
-    // Create a MAVSDK instance that will receive and respond to raw bytes
-    Mavsdk mavsdk_responder{Mavsdk::Configuration{ComponentType::Autopilot}};
+    // Test 1: Verify subscribe_raw_bytes_to_be_sent captures outgoing messages
+    Mavsdk mavsdk_autopilot{Mavsdk::Configuration{ComponentType::Autopilot}};
+    ASSERT_EQ(mavsdk_autopilot.add_any_connection("raw://"), ConnectionResult::Success);
 
-    // Add a raw connection - this will automatically enable heartbeats
-    ASSERT_EQ(mavsdk_responder.add_any_connection("raw://"), ConnectionResult::Success);
+    auto telemetry_server = std::make_shared<TelemetryServer>(mavsdk_autopilot.server_component());
 
-    // Set up a promise/future to capture the heartbeat response
+    // Set up to capture outgoing raw bytes
     auto prom = std::promise<void>();
     auto fut = prom.get_future();
-    bool heartbeat_received = false;
+    std::atomic<bool> message_captured{false};
 
-    // Subscribe to raw bytes to capture the heartbeat response
-    auto handle = mavsdk_responder.subscribe_raw_bytes_to_be_sent(
-        [&prom, &heartbeat_received](const char* bytes, size_t length) {
-            LogInfo() << "Received " << length << " raw bytes";
-
-            // Parse to check if this is a heartbeat
+    // Subscribe to outgoing raw bytes
+    auto handle = mavsdk_autopilot.subscribe_raw_bytes_to_be_sent(
+        [&prom, &message_captured](const char* bytes, size_t length) {
+            // Parse the bytes
             mavlink_message_t msg = {};
             mavlink_status_t status = {};
 
             for (size_t i = 0; i < length; ++i) {
                 if (mavlink_parse_char(
                         MAVLINK_COMM_0, static_cast<uint8_t>(bytes[i]), &msg, &status)) {
-                    if (msg.msgid == MAVLINK_MSG_ID_HEARTBEAT) {
-                        LogInfo() << "Captured HEARTBEAT message from system " << (int)msg.sysid
-                                  << ", component " << (int)msg.compid;
-                        if (!heartbeat_received) {
-                            heartbeat_received = true;
-                            prom.set_value();
-                        }
-                        return;
+                    LogInfo() << "Captured outgoing message ID " << msg.msgid;
+
+                    // Look for GLOBAL_POSITION_INT message
+                    if (msg.msgid == MAVLINK_MSG_ID_GLOBAL_POSITION_INT &&
+                        !message_captured.exchange(true)) {
+                        LogInfo() << "Captured GLOBAL_POSITION_INT being sent";
+                        prom.set_value();
                     }
                 }
             }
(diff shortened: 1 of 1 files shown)


JonasVautherin (commented):
I am a bit confused by the API 😕. Is the intent to "observe" all the connections (in the manner of running wireshark for debugging) and to "inject" arbitrary bytes into all the connections?

If the goal is to enable a custom `Connection` implementation, isn't it possible to just expose that? `add_connection(Connection connection)`, and let the user implement their own variant?

But maybe that cannot be exposed easily in the includes without a lot of refactoring? Right now, every connection manipulates mavlink in its own implementation. I guess ideally this would be separate: a `Connection` would only send and receive bytes over a transport, and there would be one unified "MavlinkHandler" that would use the Connection and do the MAVLink stuff. That MavlinkHandler could take a UdpConnection or a TcpConnection or a user-implemented QuicConnection all the same.

So a user would use something like `add_connection(Connection connection)`, e.g. `add_connection(my_quic_connection)`, and ideally we would only have to expose `connection.h` in the public API.

What do you think?

julianoes:
@JonasVautherin what you describe could be a bit nicer indeed. It's just that so far we have not exposed connection.h, and we have some specific stuff in connections, e.g. for UDP we have to keep track of the remotes and let them time out, etc.

The aim is exactly that you can just feed it bytes from your own connection, and get a callback for any bytes that need sending.

src/mavsdk/core/include/mavsdk/mavsdk.h
```diff
@@ -463,6 +463,50 @@ class Mavsdk {
...
+
+    /**
+     * @brief Handle type for raw bytes subscriptions.
+     */
+    using RawBytesHandle = Handle<const char*, size_t>;
+
+    /**
+     * @brief Send raw MAVLink bytes.
+     *
+     * This allows sending raw MAVLink message bytes through all connections.
```
JonasVautherin: I don't get that line. What does "through all connections" mean? I understand this function as this: say I am writing a GCS with MAVSDK and communicating with a remote drone. I want to use Quic and want to use my own custom implementation. So outside of MAVSDK I create a connection to the drone.

* When I receive bytes from the drone, I need to feed them to MAVSDK. This is this `send_raw_bytes`, right?
* When MAVSDK sends bytes to the drone, it needs to feed them to my custom Quic implementation. This is `subscribe_raw_bytes`, I think? So I call `send` to *receive*, and `subscribe` to *send*, because I am outside of MAVSDK so it's inverted.

But then what does it have to do with the other connections? Say there is a UDP connection, if I receive bytes from the drone and feed them to MAVSDK, I don't want MAVSDK to send them to the remote of the UDP connection 🤔??

src/mavsdk/core/include/mavsdk/mavsdk.h
```diff
@@ -463,6 +463,50 @@ class Mavsdk {
...
+     * @param bytes Pointer to raw MAVLink message bytes.
+     * @param length Number of bytes to send.
+     * @return true if bytes were successfully sent, false otherwise.
+     */
+    bool send_raw_bytes(const char* bytes, size_t length);
+
+    /**
+     * @brief Subscribe to all outgoing raw bytes.
+     *
+     * This allows monitoring all MAVLink bytes that are sent out through any connection.
```
JonasVautherin: Same here, though I guess we always send the same MAVLink messages over all the active connections.

Is that function meant to "monitor" stuff happening with other connections? My feeling is that it's meant for MAVSDK to send MAVLink bytes over the raw connection (e.g. my custom Quic implementation). So I'm not monitoring anything, I'm just providing a transport implementation to MAVSDK, right?

src/mavsdk/core/mavsdk_impl.cpp
```diff
@@ -929,6 +933,44 @@ std::pair<ConnectionResult, Mavsdk::ConnectionHandle> MavsdkImpl::add_serial_con
...
+    if (ret != ConnectionResult::Success) {
+        return {ret, Mavsdk::ConnectionHandle{}};
+    }
+
+    auto handle = add_connection(std::move(new_conn));
+
+    // Enable heartbeats for raw connection
+    auto new_configuration = get_configuration();
+    new_configuration.set_always_send_heartbeats(true);
+    set_configuration(new_configuration);
```
JonasVautherin, line 970: I don't understand why this is required. That means that I cannot write my own custom UDP implementation that would behave like the internal UDP implementation, because for some reason I'm forced to enable heartbeats?

src/mavsdk/core/mavsdk_impl.cpp
```diff
@@ -1299,6 +1341,57 @@ void MavsdkImpl::unsubscribe_outgoing_messages_json(Mavsdk::InterceptJsonHandle
...
+        if (raw_conn != nullptr) {
+            return raw_conn;
+        }
+    }
+    return nullptr;
+}
+
+bool MavsdkImpl::send_raw_bytes(const char* bytes, size_t length)
+{
+    auto* raw_conn = find_raw_connection();
```
JonasVautherin, line 1378: I guess that's the reason for the maximum one raw connection? Because for every single message MAVSDK wants to send, it has to go through all the connections, `dynamic_cast` each of them and try to find the `RawConnection`?

I guess it could just loop through all of them, but right now there is no need for multiple raw connections (maybe there will never be one) so it's simpler like this?

My next concern is the overhead of casting all the connections, but again I guess in practice there is only one connection so it's okay?

src/mavsdk/core/cli_arg.cpp
```diff
@@ -59,6 +59,11 @@ bool CliArg::parse(const std::string& uri)
             std::string_view(uri).substr(serial_flowcontrol.size() + delimiter.size()), true);
     }
 
+    const std::string raw = "raw";
+    if (uri.find(raw + delimiter) == 0) {
+        return parse_raw(std::string_view(uri).substr(raw.size() + delimiter.size()));
+    }
+
```
JonasVautherin, line 66: Do we need to expose this at all? Does that mean that one can run `mavsdk_server raw://`? When is it useful for a CLI to do that? 🤔

src/mavsdk/core/include/mavsdk/mavsdk.h
julianoes: Oh dear. Yes, Claude was confused how it worked, hence the comments. I'll have to fix those.

julianoes:
Thanks for the comments!

The reason I hadn't gone for the more object oriented approach of "creating your own connection and adding it" is that it's not straightforward to support that in language wrappers. However, we already have the API to add any connection and use that in all language wrappers. And we can expose the new mavsdk API via the C wrapper that we have. As an example, this would allow serial on Android without having to go via UDP.

src/mavsdk/core/mavsdk_impl.cpp
julianoes: That's a valid concern. I'm not sure what the overhead is.

The reason I wanted to have it as a "real connction" was so that forwarding could be used with it.

src/mavsdk/core/cli_arg.cpp
julianoes: The point of having the `add_any_connection("raw://")` call is that you can also use `add_any_connection("raw://", ForwardingOption::ForwardingOn);`

src/mavsdk/core/include/mavsdk/mavsdk.h
julianoes: Yes, again Claude confused. Let me fix it.

src/mavsdk/core/mavsdk_impl.cpp
julianoes: The idea here is that we should send out heartbeats on a custom/raw connection in order to start MAVLink flowing. The only place where we don't send out heartbeats is UDP connections that are just listening and don't have a remote to actually send messages to. In fact, we could just remote the whole "always send heartbeats" logic and have it enabled for all connections and just drop messages as we don't have a remote yet.

src/mavsdk/core/include/mavsdk/mavsdk.h
julianoes: And you're right naming is confusing. I should call it `receive_raw_bytes` and `subscribe_to_bytes_to_send`.

src/mavsdk/core/mavsdk_impl.cpp
JonasVautherin: Got it, makes sense!

docs/en/cpp/api_reference/classmavsdk_1_1_mavsdk.md
```diff
@@ -584,4 +609,64 @@ This functionality is provided primarily for testing in order to simulate packet
...
+
+### pass_received_raw_bytes() {#classmavsdk_1_1_mavsdk_1a65329315ac07bae110839d9e054fbc05}
+```cpp
+void mavsdk::Mavsdk::pass_received_raw_bytes(const char *bytes, size_t length)
+```
+
+
+Pass received raw MAVLink bytes.
+
+This allows passing raw MAVLink message bytes into MAVSDK to be processel. The bytes can contain one or more MAVLink messages.
```
JonasVautherin:
```suggestion
This allows passing raw MAVLink message bytes into MAVSDK to be processed. The bytes can contain one or more MAVLink messages.
```

src/mavsdk/core/include/mavsdk/mavsdk.h
```diff
@@ -463,6 +463,53 @@ class Mavsdk {
...
+
+    /**
+     * @brief Handle type for raw bytes subscriptions.
+     */
+    using RawBytesHandle = Handle<const char*, size_t>;
+
+    /**
+     * @brief Pass received raw MAVLink bytes.
+     *
+     * This allows passing raw MAVLink message bytes into MAVSDK to be processel.
```
JonasVautherin:
```suggestion
     * This allows passing raw MAVLink message bytes into MAVSDK to be processed.
```

src/mavsdk/core/include/mavsdk/mavsdk.h
```diff
@@ -463,6 +463,53 @@ class Mavsdk {
...
+     * The bytes can contain one or more MAVLink messages.
+     *
+     * @note Before using this, run add_any_connection("raw://")
+     *
+     * This goes together with subscribe_to_bytes_to_send.
```
JonasVautherin:
```suggestion
     * This goes together with subscribe_raw_bytes_to_be_sent.
```

docs/en/cpp/api_reference/classmavsdk_1_1_mavsdk.md
```diff
@@ -584,4 +609,64 @@ This functionality is provided primarily for testing in order to simulate packet
...
+
+
+::: info
+Before using this, run add_any_connection("raw://")
+:::
+
+This goes together with subscribe_to_bytes_to_send.
```
JonasVautherin:
```suggestion
This goes together with subscribe_raw_bytes_to_be_sent.
```

julianoes:
Thanks for the proof reading. I'm clearly a bit distracted :expressionless:

PR #2698, opened 2025-10-30
Update mavlink and add PARAM_ERROR

julianoes:
This updates mavlink including a patch containing https://github.com/ArduPilot/pymavlink/pull/1143.

Then, we add PARAM_ERROR for better param error handling as added in https://github.com/mavlink/mavlink/pull/2344.

Depends on https://github.com/mavlink/MAVSDK-Proto/pull/384.

FYI @hamishwillee


julianoes:
Pending: https://github.com/mavlink/mavlink/pull/2373
Once that's in I can remove the patch and this will pass Windows again.

JonasVautherin (approved):

hamishwillee:
@julianoes Thanks very much! 

FYI I am tidying up the param xml in  https://github.com/mavlink/mavlink/pull/2365 - in particular to make it clear when index/id are returned in the PARAM_VALUE and PARAM_ERROR (I'll get you to look when fully agree this with Peter).

Anyhow, the point is that there will be another error for "no parameters at all" returned by the request for all parameters.
Note that I haven't got to the point of thinking about actually sending that though.

PR #2703, opened 2025-11-04
Add rpath

### LINKED ISSUE #2699: CMake local install RPATH missing
I tried consuming a locally installed version of MAVSDK in another package's install, and MAVSDK's CMake target fails to correctly export the shared library destination on an installed target.
I suspected this is an rpath issue, and spent 2 hours trying various recommendations, but have not found a solution.
Here is a conference talk from a CMake maintainer specifically about this use case with rpath:
https://youtu.be/m0DwB4OvDXk?si=3TaOFA6OdxY4z5F6&t=2452

I tried his recommendation on MAVSDK, and that did not fix the issue.

To reproduce, here's a Dockerfile:

```docker
FROM ubuntu:22.04

WORKDIR /ws

# https://mavsdk.mavlink.io/main/en/cpp/guide/build_linux.html

RUN --mount=target=/var/lib/apt/lists,type=cache,sharing=locked \
    --mount=target=/var/cache/apt,type=cache,sharing=locked \
    apt-get update && \
    DEBIAN_FRONTEND=noninteractive apt-get -y install \
        build-essential \
        cmake \
        git \
        python3 \
        python3-pip

RUN git clone https://github.com/mavlink/MAVSDK.git
WORKDIR /ws/MAVSDK
RUN git submodule update --init --recursive

RUN cmake -DCMAKE_BUILD_TYPE=Debug -Bbuild -S.
RUN cmake --build build -j8
# Install is slightly different not using the install target, but it's the same idea.
RUN cmake --install build --prefix install


RUN --mount=target=/var/lib/apt/lists,type=cache,sharing=locked \
    --mount=target=/var/cache/apt,type=cache,sharing=locked \
    apt-get update && \
    DEBIAN_FRONTEND=noninteractive apt-get -y install \
        pkg-config \
        astyle \
        bluez \
        bluez-tools \
        libbluetooth-dev

WORKDIR /ws
RUN git clone --recurse-submodules --branch cmake-install https://github.com/ryanf55/RemoteIDTransmitter.git
WORKDIR /ws/RemoteIDTransmitter
RUN CMAKE_PREFIX_PATH=/ws/MAVSDK/install cmake -B build
RUN cmake --build build
RUN cmake --install build --prefix install
RUN ./install/bin/rid-transmitter
```

Just run `docker build . --progress=plain` and you will eventually get 
```
 > [stage-0 17/17] RUN ./install/bin/rid-transmitter:
0.378 ./install/bin/rid-transmitter: error while loading shared libraries: libmavsdk.so.3: cannot open shared object file: No such file or directory
```

The example here is when I tried using an installed executable that links to MAVSDK.
The linkage to MAVSDK is using the standard target_link_libraries approach:
https://github.com/Ryanf55/RemoteIDTransmitter/blob/cmake-install/CMakeLists.txt#L37

FYI @dakejahl

JonasVautherin:
@Ryanf55 what do you think?

Resolves #2699.


JonasVautherin:
Hmm now I'm thinking that maybe we should set that only for the Linux builds? :thinking:

julianoes (commented):
I don't know much about it, but sure.

julianoes:
This is what Claude suggests:

```
# Build with local paths for development
set_target_properties(mavsdk PROPERTIES
    BUILD_RPATH "${CMAKE_BINARY_DIR}/lib"
    BUILD_WITH_INSTALL_RPATH FALSE
)

# For installation:
if(APPLE)
    set(ORIGIN_PREFIX "@loader_path")
else()
    set(ORIGIN_PREFIX "$ORIGIN")
endif()

set_target_properties(mavsdk PROPERTIES
    INSTALL_RPATH "${ORIGIN_PREFIX};${CMAKE_INSTALL_FULL_LIBDIR}"
    INSTALL_RPATH_USE_LINK_PATH TRUE
)
```

JonasVautherin:
Now that I think about it, I don't think that we want to set the runpath... it really only makes sense in some use-cases. If we build for Android or iOS we don't want it. But then even if we build for Linux, we may not want it. The only time we want it is if we install MAVSDK locally and want to use it from there.

In all other cases, it will just leak the path where it was installed, which is the reason why CMake erases the runpath in the first place, I think.

So maybe we should just not set it and let the packager care about it (e.g. by setting the runpath manually with `patchelf --set-rpath` or by using LD_LIBRARY_PATH or by installing MAVSDK in a global location).

PR #2709, opened 2025-11-18
third_party: update mavlink

julianoes:
This is required for figure eight messages for PX4 MAVSDK tests.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -751,6 +751,9 @@ jobs:
         with:
           submodules: recursive
           fetch-depth: 0
+      - name: Free Disk Space
+        uses: jlumbroso/free-disk-space@v1.3.1
+        continue-on-error: true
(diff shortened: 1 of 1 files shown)


julianoes:
@JonasVautherin any idea what is happening here in CI? It seems to run out of space.

JonasVautherin:
I can't access the logs, it says "Failed to generate URL to download logs." :thinking:.

EDIT: actually I could download a log archive, but there is no complaint in what's there... it just appears to hang. Since it's marked as "failed" by github, I would expect github to tell us something...

julianoes:
I think it runs out of space, but it's a mystery. It's literally due to the mavlink change. I'm going to try to replicate it locally.

julianoes:
Oh, it runs out of space, so I added the action to save some space but I added it to arm64 instead of android-arm64!

PR #2712, opened 2025-11-23
CI: bump macOS and Windows runners

julianoes:


julianoes:
Looks like we now run into https://github.com/actions/runner/issues/449.

PR #2716, opened 2025-11-25
core: speed up deserialization using std::to_chars

julianoes:
When parsing messages, profiling shows that most time is spent in ostream to format the values to string.

This adds a test and a faster implementation based on std::to_chars.


julianoes:
Alternative would be to pull in https://github.com/fmtlib/fmt.

julianoes:
@JonasVautherin This reduces CPU usage from 8% to 7% in a typical scenario:

- Application on RPi5 using MavlinkDirect subscriber from 8% to 7%
- PX4 sending MAVLink messages at a rate of 2.2kbit/s

The baseline using the C library and MavlinkPassthrough is 6%.

I'm not sure it's worth the complexity at this stage. I'd be more tempted to pull the fmt lib and use that.

julianoes:
Some day...

Issue #2685, opened 2025-10-14
sender/receiver based on MavlinkDirect working too slow


pedramsafi:
I have created a sender/receiver pair to exchange custom mavlink messages over SiK Radio.

ideally, I would like to minimize the amount of data being sent over the link and keep it as minimal as possible to conserve bandwidth.

Below is an example of each of the sender and receiver modules:

**Sender**

```
int main(argc, char* argv[])
{

    uint8_t drone_system_id = 2;
    uint8_t drone_component_id = 1;
    bool always_send_heartbeat = true;

    Mavsdk::Configuration config = Mavsdk::Configuration(drone_system_id, drone_component_id, always_send_heartbeat);
    Mavsdk mavsdk_drone{config};
    std::shared_ptr<mavsdk::System> peer_system_;

    //connect to the local serial radio and obtain the peer_system_
    auto mavlink_direct = MavlinkDirect{peer_system_};

    //define and load custom_xml

    //create an instance of the message
    MavlinkDirect::MavlinkMessage msg_name{};

    while(1){
        msg_name.fields_json = R"({...});
        auto result = mavlink_direct.send_message(msg_name);
    }

    std::this_thread::sleep_for(std::chrono::milliseconds(1000)); // 1Hz
    return 0;

}
```

**Receiver**

```
int main(argc, char* argv[])
{
    uint8_t gcs_system_id = 250;
    uint8_t gcs_component_id = 1;
    bool always_send_heartbeat = true;

    Mavsdk::Configuration config = Mavsdk::Configuration(gcs_system_id, gcs_component_id, always_send_heartbeat);
    Mavsdk mavsdk_drone{config};
    std::shared_ptr<mavsdk::System> peer_system_;

    //connect to the local serial radio and obtain the peer_system_
    auto mavlink_direct = MavlinkDirect{peer_system_};

    //define and load custom_xml

    //create an instance of the message
    MavlinkDirect::MavlinkMessage msg_name{};

    auto result = mavlink_direct.subscribe_message("msg_name", [](const mavsdk::MavlinkDirect::MavlinkMessage &message){
        std::cout << message.message_name << " was received!" <<  std::endl;
    });

    while(1){
         std::this_thread::sleep_for(std::chrono::milliseconds(1000)); // 1Hz
    }
return 0;

}
```

The problem is the receiver is not receiving all the messages. It is dropping some. Here is a sample of the output log. the lat/lon alt values are dummy values and show "the number of seconds since start". 

```
[...]

PR #2722, opened 2025-12-01
Fix dropped messages by libmav_receiver.cpp

rantalaine:
PR to receive all MAVLink messages when using `libmav_receiver.cpp`

Background: we noticed significant amount of messages being dropped by MAVSDK server. This was verified by comparing received message count between `pymavlink` and `mavsdk_server`, while using discretely sent msg type (`CAMERA_IMAGE_CAPTURED`) for easier counting. On pymavlink, all messages came through. MAVSDK dropped roughly a third of messages. Missing messages was double checked with other streams.

In `mavlink_receiver.cpp`, there is a comment:
`// Note that one datagram can contain multiple mavlink messages.`

But `libmav_receiver.cpp` always set `datagram` to `nullptr` after parsing one message - essentially discarding any remaining messages in the `datagram`.

Changes tested with PX4-based flightcontroller connected over serial to aarch64/Ubuntu 22.04-based machine running mavsdk_server binary. No message drops observed with patched code.

### DIFF
--- a/src/mavsdk/core/libmav_receiver.cpp
+++ b/src/mavsdk/core/libmav_receiver.cpp
@@ -86,9 +86,14 @@ bool LibmavReceiver::parse_libmav_message_from_buffer(const uint8_t* buffer, siz
 
     _last_message.fields_json = json;
 
-    // Clear the original datagram since we processed it
-    _datagram = nullptr;
-    _datagram_len = 0;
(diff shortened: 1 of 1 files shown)


rantalaine:
Based on issue description probably fixes https://github.com/mavlink/MAVSDK/issues/2685

PR #2737, opened 2025-12-11
Add RemoteId plugin

julianoes:
This is Claude's attempt at implementing what @persuader72 has started in https://github.com/mavlink/MAVSDK-Proto/pull/327.

Requires https://github.com/mavlink/MAVSDK-Proto/pull/387 to be merged first.


julianoes:
I just re-read this that this PR is not really what I expected it to be:

> The goal of the plugin to use the companion computer to work with [open drone id](https://www.opendroneid.org/) compatible hardware like [this](https://github.com/ArduPilot/ArduRemoteID) even is the autopilot is not compatible with such protocol.

So, this isn't really the user/ground station side of remote id. I'm closing it for now and might start fresh with that goal.

julianoes:
This is supposed to be the drone side, rather than the ground side, so this would have to be a remote_id_server, but even then it's not so clear how it would work.

At some point I'll probably work on the ground/user side of things but I'm closing this for now.

PR #2755, opened 2026-01-28
Added support ardupilot mission and land start command for ardupilot

scropions:
Missions in ArduPilot aren't being read because some commands aren't being processed. Missions differ between ArduPilot and PX4.
image

image

### DIFF
--- a/src/mavsdk/plugins/mission/include/plugins/mission/mission.h
+++ b/src/mavsdk/plugins/mission/include/plugins/mission/mission.h
@@ -95,6 +95,7 @@ public:
         enum class VehicleAction {
             None, /**< @brief No action. */
             Takeoff, /**< @brief Vehicle will takeoff and go to defined waypoint. */
+            LandStart, /**< @brief Upon reaching the checkpoint, the vehicle begins to land.. */
             Land, /**< @brief When a waypoint is reached vehicle will land at current position. */
             TransitionToFw, /**< @brief When a waypoint is reached vehicle will transition to
                                fixed-wing mode. */
--- a/src/mavsdk/plugins/mission/mission_impl.cpp
+++ b/src/mavsdk/plugins/mission/mission_impl.cpp
@@ -655,8 +655,11 @@ std::pair<Mission::Result, Mission::MissionPlan> MissionImpl::convert_to_result_
             LogDebug() << "Assembling Message: " << int(int_item.seq);
 
             if (int_item.command == MAV_CMD_NAV_WAYPOINT ||
-                int_item.command == MAV_CMD_NAV_TAKEOFF) {
-                if (int_item.frame != MAV_FRAME_GLOBAL_RELATIVE_ALT_INT) {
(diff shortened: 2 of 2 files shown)


julianoes:
A couple of comments/questions:

1. The change would have to go into https://github.com/mavlink/MAVSDK-Proto first and then things get generated there, see https://mavsdk.mavlink.io/main/en/cpp/contributing/autogen.html
2. The point of `Mission` is to enable simple/easy multicopter missions out of the box. Are you using this with ArduCopter or ArduPlane, because I'm not sure we can support that properly without creating a bigger and more confusing API.

scropions:
1. okay
2. Arduplane. From my observations, this is the only difference in missions. I can test it on an Ardocopter.

JonasVautherin:
> Arduplane. From my observations, this is the only difference in missions. I can test it on an Ardocopter.

It would be great if `Mission` worked for Ardupilot! Right now I have seen people using `MissionRaw`.

Thanks a lot for looking into that, it's really valuable to get reports from such testing!

scropions:
There's a command called MAV_CMD_NAV_LOITER_TO_ALT that isn't processed. I haven't been able to figure out how to process it correctly yet. 
Do I need to make changes to MAVSDK-PROTO?

JonasVautherin:
`mission.h` gets autogenerated from MAVSDK-Proto, so you should add your `LandStart` there. The `_impl` files are the ones you can edit manually, they contain the implementation.

Not sure what `MAV_CMD_NAV_LOITER_TO_ALT` does, but if it is not a new feature and there is no need to modify the `Mission` API, then this would be handled on the implementation side, I think?

scropions:
This command has different parameters, so I hope to leave this decision up to the developers. Sorry for the English. I use Google Translate. If I find a way to handle this command correct, I will make a separate pull request for it.

scropions:
To prevent this message from appearing, you need to uncheck the box.
image

Here are the parameters
image

julianoes:
I don't think we can easily make Mission work well with plane/fixedwing. It is mostly aimed at Multicopter and VTOL(hybrid). I think for plane/fixedwing, we should recommend to use MissionRaw or we could create a MissionFixedwing plugin for that.

scropions:
I did this on ArduPlane. The difference isn't that big. I don't see the point in creating a separate plugin if we want to unify everything. I'll test it on Arducopter later.
Mission should make things easier. MissionRaw can be replaced with Mavlink Direct. So I don't see much point in it.
In the next commit, I'll also add mission loading. They have the same problem.

julianoes:
The problem is that I want the Mission API to be as easy as possible, and having `Land` and `LandStart` as options is ambiguous. How does a user know which one to use?

The only thing that we could consider is that Mission chooses the correct command under the hood based on what is connected but that would require a bit more magic and logic, but that tends to make things brittle. Hence, my suggestion for a separate fixedwing plugin.

julianoes:
I'm sorry but I'm closing this. While I appreciate the effort, I want to have the Mission API small and simple, without ambiguity.

PR #2761, opened 2026-02-02
Expanded capabilities for setting flight mode

jonathanreeves:
# Summary Of Changes
This is the companion PR to the proto changes made here:
https://github.com/mavlink/MAVSDK-Proto/pull/391

The PR implements the changes associated with the proto updates, and also adds support for the [MAVLINK SET_MODE command](https://mavlink.io/en/messages/common.html#SET_MODE)

# Justification For Changes
In addition to the notes provided in the [proto PR](https://github.com/mavlink/MAVSDK-Proto/pull/391), the main reason for adding the SET_MODE support is that this command is still the primary means by which QGroundControl sets flight mode on an autopilot, and in particular PX4. Since flight modes in MAVSDK are forced to be PX4 custom, in order to be interoperable with QGC in a straightforward way, we need to handle SET_MODE instead of DO_SET_MODE.

As a note about some of the stickier nuance here, QGC expects all custom flight modes to be represented as a single 32-bit value, which is indeed how the SET_MODE command treats it. DO_SET_MODE breaks this into two separate fields. QGroundControl could also be updated to expand this (and arguably should be), but there's also no reason that MAVSDK can't support this very standard MAVLINK message.

As always please don't hesitate to reach out with questions.


jonathanreeves:
Just FYI I'm working on the mavsdk_server fix, I thought I would be able to re-use datatypes in the proto but I was wrong. Will push fix soon

jonathanreeves:
Should be good to go now (I think)

src/mavsdk/plugins/action_server/action_server_impl.cpp
```diff
@@ -249,6 +258,9 @@ void ActionServerImpl::init()
             if (allow_mode) {
                 px4_mode.main_mode = custom_mode;
                 px4_mode.sub_mode = sub_custom_mode;
+                uint8_t system_base_mode = get_base_mode();
```
jonathanreeves, line 261: Just noting: I added this because the CUSTOM_MODE_ENABLED flag is required to correctly interpret the custom mode, but it's not set here. You can get into a situation where the default state is "non custom", but the ground station attempts to set a custom mode and it isn't actually recorded as custom.

julianoes:
We'll have to wait until GitHub Actions are happy again.

jonathanreeves:
> We'll have to wait until GitHub Actions are happy again.

Not the first time I've opened a PR on a github "down day". Starting to wonder if it's me :sweat_smile:

jonathanreeves:
Looking better now

PR #2763, opened 2026-02-05
Move to cpp folder

JonasVautherin:
Let's see what the CI says...

Makes it look like this:

Screenshot 2026-02-06 at 00 34 27

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -225,9 +225,9 @@ jobs:
           cd /d %GITHUB_WORKSPACE%
           cmake -G"Visual Studio 17 2022" -A ${{ matrix.platform }} ^
             -DCMAKE_BUILD_TYPE=Debug ^
-            -DCMAKE_PREFIX_PATH=%GITHUB_WORKSPACE%\combined ^
+            -DCMAKE_PREFIX_PATH=%GITHUB_WORKSPACE%\cpp\combined ^
             -Bbuild-examples-debug ^
-            -Scpp/examples
+            -Scpp\examples
 
           cmake --build build-examples-debug --config Debug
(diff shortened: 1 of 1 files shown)


julianoes:
@JonasVautherin this will probably be MAVSDK v4 when it merges, right?

I mean technically it might not have to, but it feels like it could.

julianoes (approved):
Nice! I like folders.

JonasVautherin:
> this will probably be MAVSDK v4 when it merges, right?

IMO the major version of a library should indicate if it will break the build or not. This just moves the projects into a subfolder, but the resulting compiled library should be exactly the same; it should not break the build of anyone.

By bumping to v4, my concern would be that some people may not update because they don't want to fix compilation issues...

.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "mavsdk-proto"]
 	path = proto
-	url = https://github.com/ryanf55/MAVSDK-Proto.git
+	url = https://github.com/mavlink/MAVSDK-Proto.git
```
julianoes, line 3: Whoops. Thanks.

caiannello:
This change breaks the build instructions at https://mavsdk.mavlink.io/main/en/cpp/guide/build_linux.html .
I had to change `-S.` to `-Scpp`  ...  and I still cant figure out how to build the fuzzer. :(    OH, I was supposed to know to `cd cpp` before attempting the build.

julianoes:
Thanks for the comment @caiannello. We should do a pass through the docs @JonasVautherin.

julianoes:
@caiannello done: https://github.com/mavlink/MAVSDK/pull/2774.

PR #2768, opened 2026-02-17
Shell scripts cleanup

JoC2000:
This PR refactors some shell scripts in the `cpp/tools/` folder to follow best practices or eliminate unnecessary commands.

While refactoring, I noticed that `fix_style.sh` is deprecated and being replaced by the python version. May I suggest removing it?


julianoes:
> While refactoring, I noticed that fix_style.sh is deprecated and being replaced by the python version. May I suggest removing it?

It's just telling you about the new one. We can leave it a bit longer.

JoC2000:
Looks like word splitting was making the fpm run go through. That's why the run tests failed before. I've submitted the fix for it.

julianoes:
The failure with the auto-generation looks real, it's a bit garbled.

JoC2000:
@julianoes The python script was relying on word splitting to parse each plugin and the quotes prevented that. I switched to bash array to better handle it.

PR #2771, opened 2026-02-18
Fix: setting current mission item

jazzvaz:
Setting current mission item when the argument outside mission items range or there is no mission downloaded, would result in a crash. This is faced when using mavsdk server: it exists with error code.

### DIFF
--- a/cpp/src/mavsdk/plugins/mission/mission_impl.cpp
+++ b/cpp/src/mavsdk/plugins/mission/mission_impl.cpp
@@ -890,11 +890,11 @@ void MissionImpl::clear_mission_async(const Mission::ResultCallback& callback)
 
 Mission::Result MissionImpl::set_current_mission_item(int current)
 {
-    auto prom = std::make_shared<std::promise<Mission::Result>>();
-    auto fut = prom->get_future();
+    auto prom = std::promise<Mission::Result>();
+    auto fut = prom.get_future();
 
     set_current_mission_item_async(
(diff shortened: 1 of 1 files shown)


cpp/src/mavsdk/plugins/mission/mission_impl.cpp
```diff
@@ -890,11 +890,11 @@ void MissionImpl::clear_mission_async(const Mission::ResultCallback& callback)
...
 {
-    auto prom = std::promise<Mission::Result>();
-    auto fut = prom.get_future();
+    auto prom = std::make_shared<std::promise<Mission::Result>>();
+    auto fut = prom->get_future();
 
     set_current_mission_item_async(
-        current, [&prom](Mission::Result result) { prom.set_value(result); });
+        current, [prom](Mission::Result result) { prom->set_value(result); });
     return fut.get();
```
julianoes: This change is not needed. We don't need to create a `std::shared_ptr` as we are waiting for the `std::future`, so the scope remains.

cpp/src/mavsdk/plugins/mission/mission_impl.cpp
```diff
@@ -924,9 +924,9 @@ void MissionImpl::set_current_mission_item_async(
             if (callback) {
                 // FIXME: come up with better error code.
                 callback(Mission::Result::InvalidArgument);
-                return;
             }
         });
+        return;
```
julianoes, line 929: This one is correct though!

PR #2773, opened 2026-02-19
Add mavsdk server docker with compose

jazzvaz:
This is a minimal Docker image (~190MB) with Compose that builds and runs the MAVSDK server. In the future, it would be good idea to publish it to any public docker registry.


docker/example/Dockerfile
```diff
@@ -0,0 +1,37 @@
+FROM debian:bookworm-slim AS builder
```
julianoes, line 1: The alternative would be to grab a minimal alpine image and download the musl binary into it, instead of building.

julianoes:
I appreciate your pull request. What I'm worried about is that this opens one more surface that needs to be maintained.

What is your use case that needs it run in docker? Why doesn't the simple musl binary work? And isn't the mavsdk_server_bin usually bundled with whatever you're running, so you don't need to run it separately?

jazzvaz:
The main idea behind was to build and run mavsdk server with custom changes and not to worry about architecture or dependencies. I am running it in a separate container and communicate with gRPC.

julianoes:
Right, so it would be worthwhile to have as an example for that maybe. But I don't think we need/want to support an official image on dockerhub for it, if there isn't a clear recommended use case.

jazzvaz:
Totally understandable, dockerhub was just an suggestion. What do you think about keeping at least this dockerfile and compose, should it be moved to examples?

julianoes:
Maybe as an example in here: https://github.com/mavlink/MAVSDK/tree/main/docker ?

docker/example/Dockerfile
jazzvaz: Agree, however, this is specific to build mavsdk with server from sources (useful especially if there are custom changes).

julianoes (approved):
Yeah, I think that's nice like that.

PR #2775, opened 2026-02-23
Add devcontainer

jazzvaz:
Useful for those who wants to do development without installing all dependencies in their system.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,8 @@
+FROM mcr.microsoft.com/devcontainers/cpp:1-debian-12
+
+RUN apt-get update \
+    && apt-get -y --no-install-recommends install \
+        python3 \
+        python3-pip \
+        python3-future \
+        python3-tqdm
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,19 @@
+{
+    "name": "mavsdk",
+    "build": {
+        "dockerfile": "Dockerfile"
+    },
+    "runArgs": [
+        "--network=host"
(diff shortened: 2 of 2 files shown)


.devcontainer/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM mcr.microsoft.com/devcontainers/cpp:1-debian-12
```
julianoes: Why not debian-13?

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,8 @@
...
+
+RUN apt-get update \
+    && apt-get -y --no-install-recommends install \
+        python3 \
+        python3-pip \
+        python3-future \
+        python3-tqdm
```
julianoes, line 8: Don't we have a requirements.txt file for this somewhere?

jazzvaz: The requirements.txt is a part of mavlink itself; also if i remember correctly, there is a bug: future is needed but not included in the file.

.devcontainer/Dockerfile
jazzvaz: Debian-13 is not yet there. [Link to available cpp devcontainers](https://hub.docker.com/r/microsoft/devcontainers-cpp)

.devcontainer/Dockerfile
julianoes: Ah, nevermind, you're right.

.devcontainer/Dockerfile
julianoes: Oh, ok.

julianoes (approved):

PR #2777, opened 2026-02-26
c: unsubscribe properly before destroy

JonasVautherin:
This makes the C wrapper unsubscribe properly before destroying MAVSDK.

### DIFF
--- a/c/src/cmavsdk/plugins/camera/camera.cpp
+++ b/c/src/cmavsdk/plugins/camera/camera.cpp
@@ -5,7 +5,9 @@
 #include "camera.h"
 
 #include <mavsdk/plugins/camera/camera.h>
+#include <algorithm>
 #include <cstring>
+#include <mutex>
 #include <vector>
 
 // ===== C++ to C Type Conversions =====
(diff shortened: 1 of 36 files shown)


c/src/cmavsdk/plugins/camera/camera.cpp
```diff
@@ -1097,6 +1107,48 @@ void mavsdk_camera_destroy(mavsdk_camera_t camera) {
...
+        }
+        wrapper->camera_list_handles.clear();
+        for (auto* h : wrapper->mode_handles) {
+            wrapper->cpp_plugin->unsubscribe_mode(std::move(*h));
+            delete h;
+        }
+        wrapper->mode_handles.clear();
+        for (auto* h : wrapper->video_stream_info_handles) {
+            wrapper->cpp_plugin->unsubscribe_video_stream_info(std::move(*h));
+            delete h;
```
julianoes, line 1127: My first thought is whether this could be a `std::unique_ptr` so we don't have raw `delete` but presumably not because it has to live in a C header, right?

c/src/cmavsdk/plugins/mission/mission.cpp
```diff
@@ -5,7 +5,9 @@
 #include "mission.h"
 
 #include <mavsdk/plugins/mission/mission.h>
+#include <algorithm>
```
julianoes, line 8: What's `algorithm` for? Is that for `std::remove`?

JonasVautherin: Yes, for std::remove :+1:

c/src/cmavsdk/plugins/camera/camera.cpp
JonasVautherin: I think we could change it to have unique_ptr, but:

1. That would require changes elsewhere in the file (changing `mavsdk_camera_wrapper` to use `std::unique_ptr` and propagating that everywhere).
2. I am not sure it adds a lot of clarity, given that the whole file is an interface between C and C++.
3. I have been testing this version, if I change it I will have to re-test (or hope for the best :sweat_smile:), and I am not sure if it is worth it if it doesn't improve clarity :thinking:.

I mean if you look at the rest of the file, it's very much dealing with C, e.g.:

```c
void mavsdk_camera_option_array_destroy(
    mavsdk_camera_option_t** array,
    size_t size) {
    if (!array || !*array) return;

    for (size_t i = 0; i < size; i++) {
        mavsdk_camera_option_destroy(&(*array)[i]);
    }

    delete[] *array;
    *array = nullptr;
}
```

What do you think?

PR #2779, opened 2026-02-26
Run proto check for C wrappers, too

JonasVautherin:
This adds a "global" `generate_from_protos.bash` script that calls the generation scripts for the different languages (for now cpp and c).

The "proto check" in the CI runs it and verifies that the generated files are in sync with the proto submodule, for all the languages (not only cpp).

So now, whenever the proto submodule is modified, the global `./tools/generate_from_protos.bash` must be called to update all the languages in the repo.

### DIFF
--- a/c/src/cmavsdk/plugins/action_server/action_server.h
+++ b/c/src/cmavsdk/plugins/action_server/action_server.h
@@ -36,7 +36,7 @@ typedef struct mavsdk_action_server_terminate_handle_s *mavsdk_action_server_ter
  * @brief Flight modes.
  *
  * For more information about flight modes, check out
- * https://docs.px4.io/master/en/config/flight_mode.html.
+ * https://docs.px4.io/main/en/config/flight_mode.html.
  */
 typedef enum {
     /**  Mode not known. */
--- a/c/src/cmavsdk/plugins/camera/camera.h
+++ b/c/src/cmavsdk/plugins/camera/camera.h
@@ -16,7 +16,7 @@ extern "C" {
 
 /**
  * @brief Can be used to manage cameras that implement the MAVLink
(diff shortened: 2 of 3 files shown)


.github/workflows/linux.yml
```diff
@@ -470,7 +474,8 @@ jobs:
       - name: build necessary protoc tooling
         run: cmake $superbuild $cmake_prefix_path -DCMAKE_BUILD_TYPE=Debug -DBUILD_MAVSDK_SERVER=ON -Bbuild/default -H.
       - name: generate code from protos
-        run: PATH="$PATH:$HOME/.local/bin" tools/generate_from_protos.sh
+        working-directory: .
+        run: PATH="$PATH:$HOME/.local/bin:$(pwd)/cpp/build/default/third_party/install/bin" tools/generate_from_protos.bash
```
JonasVautherin, line 478: I add `$(pwd)/cpp/build/default/third_party/install/bin` because we need the exact `protoc` that is built by this version of MAVSDK-C++. This is *exclusively* for gRPC/protobuf, i.e. `mavsdk_server`.

`protoc-gen-mavsdk` does not care about the version of `protoc`, which is convenient.

tools/generate_from_protos.bash
```diff
@@ -0,0 +1,71 @@
...
+    #echo "------------------"
+    #pushd py/mavsdk
+    #"${project_root}/py/mavsdk/tools/generate_from_protos.bash"
+    #popd
+
+    #echo "------------------"
+    #echo "Generating Python (asyncio)"
+    #echo "------------------"
+    #pushd py/aiomavsdk
+    #"${project_root}/py/aiomavsdk/tools/generate_from_protos.bash"
```
JonasVautherin, line 64: Spoiler 🙈

PR #2781, opened 2026-02-26
ci: build cmavsdk with musl

JonasVautherin:
This builds libcmavsdk.so with musl. For now it makes sure that it builds, but later it will be used to create a musl python wheel.

### DIFF
--- a/c/example/camera-server.c
+++ b/c/example/camera-server.c
@@ -1,3 +1,4 @@
+#define _POSIX_C_SOURCE 200809L
 #include <stdbool.h>
 #include <stdio.h>
 #include <time.h>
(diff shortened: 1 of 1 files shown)


.github/workflows/linux.yml
```diff
@@ -620,12 +620,23 @@ jobs:
...
         with:
           name: mavsdk_server_musl_x86_64
           path: ./cpp/install/bin/mavsdk_server
           retention-days: 2
+      - name: Upload C wrapper as artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: libcmavsdk_musl_x86_64
+          path: ./c/build/install/lib/
+          retention-days: 2
```
JonasVautherin, line 639: This will be used later by a python job to create a musl wheel.

julianoes (approved):

PR #2784, opened 2026-02-27
ci: add manylinux jobs

JonasVautherin:
With python ctypes, we can't make a static executable with musl like we did for mavsdk_server, because we now link a library. So we need to build the `.so` with manylinux :see_no_evil:.

### DIFF
--- a/c/CMakeLists.txt
+++ b/c/CMakeLists.txt
@@ -15,6 +15,11 @@ set(CMAKE_C_STANDARD 11)
 set(CMAKE_C_STANDARD_REQUIRED ON)
 set(CMAKE_C_EXTENSIONS OFF)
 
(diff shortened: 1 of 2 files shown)


JonasVautherin:
Failing because `libevents` was not properly configured. I have a fix in another PR, I'll rebase on this one when it gets merged :+1:.

c/CMakeLists.txt
```diff
@@ -15,6 +15,11 @@ set(CMAKE_C_STANDARD 11)
 set(CMAKE_C_STANDARD_REQUIRED ON)
 set(CMAKE_C_EXTENSIONS OFF)
 
+option(BUILD_EXAMPLES "Build examples" ON)
```
julianoes, line 18: 👍

julianoes (approved):

PR #2787, opened 2026-03-08
Add pymavsdk

JonasVautherin:
This adds:

* `mavsdk`, which wraps the C MAVSDK library (which itself wraps MAVSDK-C++). This is a "classical" Python API (no asyncio).
* `aiomavsdk`, which wraps `mavsdk` above and exposes an asyncio API that tries to be close to MAVSDK-Python.

The big difference (and really the point here) is that this does not depend on gRPC at all: it's direct FFIs to the C MAVSDK library. Making it lighter and easier to install (we have had issues with gRPC-Python on some systems).

### DIFF
--- a/py/mavsdk/README.md
+++ b/py/mavsdk/README.md
@@ -26,4 +26,4 @@ In order to generate the plugins from the proto submodule, run:
 ## Running it
 
 This project relies on the C MAVSDK library, which it wraps. The library must be
-built separately from the [C wrapper](../../c) and copied into [mavsdk/lib/libcmavsdk.so](mavsdk/lib).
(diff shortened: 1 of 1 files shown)


julianoes (approved):
Wow!

julianoes:
We'll have to think about documentation and whether to replace the awkward http://mavsdk-python-docs.s3-website.eu-central-1.amazonaws.com/ with something on https://mavsdk.mavlink.io/ instead. I can throw that at Claude though later. I wouldn't want to hold this PR up!

JonasVautherin:
Do people still read documentation now that they have LLMs? Because the whole docs is in the source files already :see_no_evil:

julianoes:
Some do, but most importantly LLMs will read docs if it's in source.

PR #2790, opened 2026-03-09
c/py: add raw connection

julianoes:
This adds the raw connection API to C and Python.

~Don't merge yet, currently contains https://github.com/mavlink/MAVSDK/pull/2824.~


julianoes:
Haven't tested it yet... :thinking:

JonasVautherin:
You can test with the wheel produced by the CI, but note that it keeps it only for 2 days. So you may have to re-run the manylinux jobs if you haven't downloaded them yet 😇

julianoes:
Right, that's good to know.

PR #2791, opened 2026-03-10
mocap: wire up reset_counter, estimator_type, quality fields

alireza787b:
## Problem

The Mocap plugin hardcodes several fields when packing MAVLink messages,
ignoring values that PX4's EKF2 can use for better sensor fusion:

**send_vision_position_estimate:**
```cpp
0); // FIXME: reset_counter not set   ← line 119
```

**send_vision_speed_estimate:**
```cpp
0); // FIXME: reset_counter not set   ← line 170
```

**send_odometry:**
```cpp
0,                        // reset_counter hardcoded
MAV_ESTIMATOR_TYPE_MOCAP, // estimator_type hardcoded (wrong for VIO users)
0);                       // quality hardcoded
```

Additionally, `static_cast<uint8_t>(odometry.frame_id)` sends the proto enum
value (0) instead of the MAVLink value (14 = `MAV_FRAME_MOCAP_NED`), because
the proto enum uses sequential values rather than MAVLink values.

## Changes

1. **Wire up `reset_counter`** in `send_vision_position_estimate` — resolves
   the FIXME comment

2. **Wire up `reset_counter`** in `send_vision_speed_estimate` — resolves
   the FIXME comment

3. **Wire up `reset_counter`, `estimator_type`, `quality`** in `send_odometry`
   — replaces all three hardcoded values with the proto fields

4. **Fix frame_id mapping** — add switch to map proto `MavFrame` enum values
   to correct MAVLink `MAV_FRAME` constants (14 for MOCAP_NED, 20 for LOCAL_FRD)

5. **Backward compatibility** — when `estimator_type` is `UNKNOWN` (proto3 default),
   fall back to `MAV_ESTIMATOR_TYPE_MOCAP` to preserve existing behavior

## Dependencies

- Requires mavlink/MAVSDK-Proto#395 (adds the new proto fields)
- Proto submodule update needed after that PR merges, then run
  `tools/generate_from_protos.sh` to regenerate auto-generated code

## Related PRs

- **Proto fields:** mavlink/MAVSDK-Proto#395 — adds reset_counter, estimator_type, quality to proto
- **MavFrame fix:** mavlink/MAVSDK-Proto#396 — fixes proto enum values to match MAVLink (once merged, the switch mapping here can be simplified to a direct cast)
- **Python example:** mavlink/MAVSDK-Python#820 — vision odometry example

## Testing

After proto submodule update:
```bash
cmake -Bbuild/default -S. -DBUILD_MAVSDK_SERVER=ON -DWERROR=ON
cmake --build build/default -j$(nproc)
```

Verified: all changes maintain backward compatibility — existing users who
don't set the new fields get identical MAVLink output (reset_counter=0,
estimator_type=MOCAP, quality=0, frame_id=MOCAP_NED).

### DIFF
--- a/c/src/cmavsdk/plugins/mocap/mocap.cpp
+++ b/c/src/cmavsdk/plugins/mocap/mocap.cpp
@@ -305,6 +305,7 @@ translate_vision_position_estimate_from_c(const mavsdk_mocap_vision_position_est
     cpp_struct.position_body = translate_position_body_from_c(c_struct.position_body);
     cpp_struct.angle_body = translate_angle_body_from_c(c_struct.angle_body);
     cpp_struct.pose_covariance = translate_covariance_from_c(c_struct.pose_covariance);
+    cpp_struct.reset_counter = c_struct.reset_counter;
     return cpp_struct;
 }
 
@@ -315,6 +316,7 @@ translate_vision_position_estimate_to_c(const mavsdk::Mocap::VisionPositionEstim
     c_struct.position_body = translate_position_body_to_c(cpp_struct.position_body);
     c_struct.angle_body = translate_angle_body_to_c(cpp_struct.angle_body);
     c_struct.pose_covariance = translate_covariance_to_c(cpp_struct.pose_covariance);
+    c_struct.reset_counter = cpp_struct.reset_counter;
     return c_struct;
 }
 
@@ -343,6 +345,7 @@ translate_vision_speed_estimate_from_c(const mavsdk_mocap_vision_speed_estimate_
     cpp_struct.time_usec = c_struct.time_usec;
     cpp_struct.speed_ned = translate_speed_ned_from_c(c_struct.speed_ned);
     cpp_struct.speed_covariance = translate_covariance_from_c(c_struct.speed_covariance);
+    cpp_struct.reset_counter = c_struct.reset_counter;
     return cpp_struct;
 }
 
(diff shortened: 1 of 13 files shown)


alireza787b:
Hi — just a note that the CI failures here are expected: this PR references `reset_counter`, `estimator_type`, and `quality_percent` fields that were added in MAVSDK-Proto PR [#395](https://github.com/mavlink/MAVSDK-Proto/pull/395), which hasn't been merged yet.

Once #395 is merged and the proto submodule pointer is updated, the build should pass.

**Dependency chain:**
1. MAVSDK-Proto #395 (merge) → update proto submodule in MAVSDK → this PR's CI passes
2. Then MAVSDK-Python #820 can use the new fields

Happy to rebase or address anything else in the meantime.

alireza787b:
Hey Juliano, you are right. This branch is stale in its current form and the old dependency assumption was not enough. I re-checked the proto/C++/generation path end-to-end and I would rather replace this with clean current-main PRs than keep patching stale history here. I will prepare a small fresh MAVSDK-Proto PR first, then a regenerated MAVSDK follow-up that keeps the frame mapping in C++. Very happy to align with any suggestion on that split if you prefer a different sequencing.

julianoes:
I've merged proto, you can now fix this up.

julianoes:
@alireza787b I've rebased this and fixed it up. Hopefully that works for you like that. I want to get this into the next v3 release, so I went ahead.

PR #2800, opened 2026-03-16
core: do not set disconnected state on heartbeat timeout

PavelGuzenfeld:
## Summary
- Prevent `set_disconnected()` from being called in `heartbeats_timed_out()`, which can incorrectly mark the system as disconnected during transient heartbeat gaps

## Motivation
When using UDP connections, heartbeat timeouts can occur transiently without the system actually being disconnected. Calling `set_disconnected()` on every heartbeat timeout causes unnecessary state transitions and triggers reconnection logic that disrupts ongoing operations.

### DIFF
--- a/src/mavsdk/core/system_impl.cpp
+++ b/src/mavsdk/core/system_impl.cpp
@@ -262,7 +262,7 @@ void SystemImpl::process_autopilot_version(const mavlink_message_t& message)
 void SystemImpl::heartbeats_timed_out()
 {
     LogInfo() << "heartbeats timed out";
-    set_disconnected();
+    // set_disconnected();  due to the way we handle heartbeats, we don't want to set disconnected here
 }
 
 void SystemImpl::system_thread()


PavelGuzenfeld:
Closing this — disabling `set_disconnected()` breaks MAVSDK's core disconnection detection (is_connected stays true forever, disconnect callbacks never fire, plugins never disable). The proper fix for transient UDP heartbeat gaps is `Mavsdk::set_heartbeat_timeout_s()` with a larger value.

JonasVautherin:
What about [`set_heartbeat_timeout`](https://github.com/mavlink/MAVSDK/blob/1af9f4f39357caf14018de225805e7822218ee7e/cpp/src/mavsdk/core/include/mavsdk/mavsdk.h#L358-L366)?

```cpp
    /**
     * @brief Set heartbeat timeout.
     *
     * The default heartbeat timeout is 3 seconds. If no heartbeat is received
     * within this time, the system is considered disconnected.
     *
     * @param timeout_s Timeout in seconds.
     */
    void set_heartbeat_timeout_s(double timeout_s);
```

PavelGuzenfeld:
Thanks @JonasVautherin — you're right, `set_heartbeat_timeout_s()` is the proper way to handle this. Closing the PR.

Issue #399, opened 2018-05-18
fly_mission and fly_qgc_mission examples freezes on mission accepted


hamishwillee:
The future does not appear to be returning with mission uploaded.

```
hamish@hamish-ThinkPad-T470:~/github/dronecore/2test/DroneCore/example/fly_mission/build$ ./fly_mission udp://:14540
Waiting to discover system...
[04:48:15|Info ] New device on: 127.0.0.1:14557 (udp_connection.cpp:208)
[04:48:15|Debug] New: System ID: 1 Comp ID: 1 (dronecore_impl.cpp:286)
[04:48:15|Debug] Component Autopilot added. (mavlink_system.cpp:330)
[04:48:15|Debug] Found 1 component(s). (mavlink_system.cpp:462)
[04:48:15|Debug] Discovered 4294967298 (mavlink_system.cpp:464)
Discovered system with UUID: 4294967298
Waiting for system to be ready
[04:48:15|Debug] We got an ack: 520 (mavlink_commands.cpp:140)
[04:48:16|Debug] We got an ack: 511 (mavlink_commands.cpp:140)
System ready
Creating and uploading mission
Uploading mission...
[04:48:16|Debug] Send mission item 0 (mission_impl.cpp:904)
[04:48:16|Debug] Send mission item 1 (mission_impl.cpp:904)
[04:48:16|Debug] Send mission item 2 (mission_impl.cpp:904)
[04:48:16|Debug] Send mission item 3 (mission_impl.cpp:904)
[04:48:16|Debug] Send mission item 4 (mission_impl.cpp:904)
[04:48:16|Debug] Send mission item 5 (mission_impl.cpp:904)
[04:48:16|Debug] Send mission item 6 (mission_impl.cpp:904)
[04:48:16|Debug] Send mission item 7 (mission_impl.cpp:904)
[04:48:16|Debug] Send mission item 8 (mission_impl.cpp:904)
[04:48:16|Debug] Send mission item 9 (mission_impl.cpp:904)
[04:48:16|Debug] Send mission item 10 (mission_impl.cpp:904)
[04:48:16|Debug] Send mission item 11 (mission_impl.cpp:904)
[04:48:16|Debug] Send mission item 12 (mission_impl.cpp:904)
[04:48:16|Debug] Send mission item 13 (mission_impl.cpp:904)
[04:48:16|Debug] Send mission item 14 (mission_impl.cpp:904)
[...]

PR #2802, opened 2026-03-21
Expose full HOME_POSITION fields in subscribe_home

### LINKED ISSUE #2310: Adding Home Position Mavlink Message to Telemetry
Currently, the [`subcribe_home` function](https://mavsdk.mavlink.io/main/en/cpp/api_reference/classmavsdk_1_1_telemetry.html#classmavsdk_1_1_telemetry_1ae57e4c8741ba7ae05961cadfabc82875) in Telemetry API  returns a HomeHandle that uses the position struct which is in terms of global GPS position.

However, there is a mavlink message that gives more information about the [home position ](https://mavlink.io/en/messages/common.html#HOME_POSITION)

In future releases, it would be much better to implement this mavlink message into mavsdk

PavelGuzenfeld:
## Summary
- Add new `HomePosition` struct with all fields from MAVLink HOME_POSITION message (#242)
- `subscribe_home` now returns `HomePosition` instead of `Position` (breaking API change)
- New fields: local NED (x/y/z), surface quaternion, approach vector, timestamp
- Add system test verifying all fields propagate correctly end-to-end

## Motivation
Fixes #2310 — `subscribe_home` only returns global GPS position but HOME_POSITION also carries local NED position, attitude quaternion, approach vector, and timestamp that were being silently discarded.

## Changes
- **Proto**: Add `HomePosition` message, change `HomeResponse` to use it
- **Header**: Add `HomePosition` struct, update `HomeCallback`/`HomeHandle` types
- **Impl**: Populate all HOME_POSITION fields in `process_home_position()`
- **gRPC**: Add `translateToRpcHomePosition`/`translateFromRpcHomePosition`
- **Test**: System test `HomePositionFullFields` validates all fields

## Breaking Change
`subscribe_home` callback type changes from `std::function<void(Position)>` to `std::function<void(HomePosition)>`. `HomePosition` is a superset of `Position` — the same lat/lon/alt fields are present at the same offsets.

## Companion PR
Proto changes: https://github.com/mavlink/MAVSDK-Proto/pull/399

## Test plan
- [x] Unit tests pass (289/289)
- [x] System test `HomePositionFullFields` passes — validates GPS, local NED, quaternion, approach vector, and timestamp
- [x] Build clean

### DIFF
--- a/cpp/src/mavsdk/plugins/telemetry/include/plugins/telemetry/telemetry.h
+++ b/cpp/src/mavsdk/plugins/telemetry/include/plugins/telemetry/telemetry.h
@@ -269,6 +269,44 @@ public:
     friend std::ostream& operator<<(std::ostream& str, Telemetry::Quaternion const& quaternion);
 
     /**
+     * @brief Home position type.
+     *
+     * Includes the global GPS position, local NED position, surface quaternion,
+     * and approach vector from the MAVLink HOME_POSITION message.
+     */
+    struct HomePosition {
+        double latitude_deg{double(NAN)}; /**< @brief Latitude in degrees (range: -90 to +90) */
+        double longitude_deg{double(NAN)}; /**< @brief Longitude in degrees (range: -180 to +180) */
+        float absolute_altitude_m{
+            float(NAN)}; /**< @brief Altitude AMSL (above mean sea level) in metres */
+        float relative_altitude_m{
+            float(NAN)}; /**< @brief Altitude relative to takeoff altitude in metres */
+        float local_x_m{float(NAN)}; /**< @brief Local X position in NED frame (m) */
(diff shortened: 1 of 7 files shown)


PavelGuzenfeld:
Closing — this depends on proto PR #399 which was parked for v4 due to the API-breaking change to the Home/Position message. Will revisit when v4 work begins.

julianoes:
Let's get an issue going with upcoming v4 breaks that we want to do: https://github.com/mavlink/MAVSDK/issues/2830

julianoes:
@PavelGuzenfeld you can go ahead with the home position change. I have branched off v3.

PavelGuzenfeld:
Reopened as #2839 — rebased on current main (picks up the Asio migration and other v4 changes). Proto PR #399 also rebased.

PR #2803, opened 2026-03-21
Implement set_rate_rc for ArduPilot via SYS_STATUS

### LINKED ISSUE #2516: Telemetry rates: set_rate_rc unsupported
On Ardupilot, I need to set the telemetry rates myself or I don't get any updates. This works for battery and GPS, for instance.

set_rate_rc is not implemented, I guess because it is already sent with system status. However I don't get any updates there unless I open QGroundControl and then come back to my app.

I guess this is more of a list of questions:

- How does a GCS make the rc connection status start working?
- How do I get the connection strength percentage? It is always NaN
- I don't really want to change the rates that are configured in parameters on the autopilot. Is there a way to set the rate to the value configured on the autopilot? Or do I need to fetch the parameter and then set it to that value with set_rate_* ? In that case, how do I know which parameter to check (i.e. which telemetry port: for ArduPilot x in SRx_*)?

PavelGuzenfeld:
## Summary
- Implement `set_rate_rc_status()` which previously returned `Unsupported` unconditionally
- ArduPilot: request `SYS_STATUS` at the desired rate (RC info comes via SYS_STATUS)
- PX4: request `RC_CHANNELS` at the desired rate, plus `SYS_STATUS`
- Handle shared `SYS_STATUS` rate between health and RC by taking the max of both

## Motivation
Fixes #2516 — On ArduPilot, RC link status is delivered via `SYS_STATUS` rather than a dedicated stream. `set_rate_rc` was not implemented because of this, so users got no RC updates unless another GCS opened first.

## Changes
- **telemetry_impl.h**: Add `_health_rate_hz`, `_rc_status_rate_hz` members and `set_rate_sys_status()` helper
- **telemetry_impl.cpp**: 
  - `set_rate_rc_status()`: ArduPilot requests SYS_STATUS, PX4 requests RC_CHANNELS + SYS_STATUS
  - `set_rate_health()`: Now uses shared rate logic via `set_rate_sys_status()`
  - Async versions updated with same logic

## Test Results
- [x] Build clean (Debug, zero errors/warnings)
- [x] Unit tests 289/289 pass
- [x] System tests (3/3 pass): RcStatusViaSysStatus, RcStatusViaRcChannels, RcStatusUnavailableViaSysStatus
- [x] PX4 SITL: `set_rate_rc_status(1.0)` returns **Success**, 3 RC updates received at ~1 Hz
- [x] ArduPilot path covered by RcStatusViaSysStatus test (SYS_STATUS with RC sensor flags)

### DIFF
--- a/cpp/src/system_tests/telemetry_rc_status.cpp
+++ b/cpp/src/system_tests/telemetry_rc_status.cpp
@@ -22,8 +22,7 @@ TEST(SystemTest, RcStatusViaSysStatus)
         mavsdk_groundstation.add_any_connection("udpin://0.0.0.0:15220"),
         ConnectionResult::Success);
     ASSERT_EQ(
-        mavsdk_autopilot.add_any_connection("udpout://127.0.0.1:15220"),
-        ConnectionResult::Success);
+        mavsdk_autopilot.add_any_connection("udpout://127.0.0.1:15220"), ConnectionResult::Success);
 
     auto maybe_system = mavsdk_groundstation.first_autopilot(10.0);
     ASSERT_TRUE(maybe_system);
@@ -90,8 +89,7 @@ TEST(SystemTest, RcStatusViaRcChannels)
         mavsdk_groundstation.add_any_connection("udpin://0.0.0.0:15221"),
         ConnectionResult::Success);
     ASSERT_EQ(
-        mavsdk_autopilot.add_any_connection("udpout://127.0.0.1:15221"),
-        ConnectionResult::Success);
+        mavsdk_autopilot.add_any_connection("udpout://127.0.0.1:15221"), ConnectionResult::Success);
 
     auto maybe_system = mavsdk_groundstation.first_autopilot(10.0);
     ASSERT_TRUE(maybe_system);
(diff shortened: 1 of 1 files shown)


cpp/src/system_tests/telemetry_rc_rate_sitl.cpp
```diff
@@ -0,0 +1,64 @@
...
+#include <atomic>
+#include <chrono>
+#include <thread>
+#include <gtest/gtest.h>
+
+using namespace mavsdk;
+
+// Integration test against PX4 SITL.
+// Requires PX4 SITL running on udp://:14540
+// Run with: --gtest_filter='SitlTest.*'
```
julianoes: I appreciate adding a test. However, system_tests are run without external dependencies such as SITL. Instead, system_tests generally use the server (e.g. autopilot_server and telemetry_server) as the counterpart and test things "internally".

Therefore, either test against the server equivalent, or otherwise move to an example (or just remove).

PavelGuzenfeld: Thanks for the feedback. Let me take care of that.

PavelGuzenfeld: Dropped the SITL test. The three system tests using MavlinkDirect already cover the SYS_STATUS and RC_CHANNELS paths.

julianoes (approved):

PR #2810, opened 2026-03-27
core: migrate all async I/O and work queues to Asio

bansiesta:
## Motivation

MAVSDK historically grew a patchwork of per-component threads: one OS thread per UDP/TCP/serial connection, a `work_thread` polling queues every 10 ms, a `system_thread` running a periodic loop, and a separate callback thread for user-facing notifications. Each hand-off between threads required locks, condition variables, or atomic flags, and the interactions between them were the source of recurring deadlocks, data races, and hard-to-reproduce crashes — including several that required `cancel_all_operations()` and `cancel_all_param()` to be commented out with FIXMEs because they would crash on destruction.

This PR replaces all of that with a single [Asio](https://think-async.com/Asio/) `io_context` running on one dedicated `_io_thread`. The rule is simple: all MAVSDK-internal state is touched only on that thread. No locks needed for internal queues. No polling loops. The only remaining separate thread is the user-callback thread — intentional, to keep slow user code off the I/O path.

## What changed

**`third_party/asio/`** — new superbuild entry  
Fetches `chriskohlhoff/asio@asio-1-30-2` (header-only, BSL-1.0) via `ExternalProject_Add`. Also adds a `Findasio.cmake` module for non-superbuild builds. No compiled artifact.

**`UdpConnection`** — async receive  
- `SocketHolder`, `_recv_thread`, `_should_exit`, `SO_RCVTIMEO` hack removed  
- Socket is now `asio::ip::udp::socket` bound to the shared `io_context`  
- `do_receive()` posts `async_receive_from`; completion handler re-arms itself  
- `stop()` just calls `_socket.close()` — cancels pending op cleanly

**`TcpClientConnection` / `TcpServerConnection`** — async receive  
- Dedicated receive threads removed; both use `async_read_some` on the shared `io_context`  
- `stop()` closes sockets on the io_context thread (double-fence pattern) to eliminate TSan races with `do_accept()` / `do_receive()`  
- `send_raw_bytes()` uses `asio::dispatch()` (not `post()`) so calls from the io_context thread execute inline, preventing `future.get()` deadlock

**`SerialConnection`** — async receive  
- Blocking receive thread removed; uses `async_read_some` on the shared `io_context`

**`MavsdkImpl`** — io_context, queues, and threads  
- Single `asio::io_context` shared by all connections, running on a dedicated `_io_thread`  
- `receive_message()` / `receive_libmav_message()` post directly to io_context — received-message queues, mutexes, and condition variables removed  
- `send_message()` posts directly to io_context — outgoing message queue and mutex removed  
- `TimeoutHandler` / `CallEveryHandler` driven by a recurring 5 ms `asio::steady_timer` (`schedule_timers_poll`) — no more polling in work_thread  
- `do_work()` on server components driven by a recurring 10 ms `asio::steady_timer` (`schedule_do_work`)  
- `work_thread` and `system_thread` eliminated entirely

**`CallEveryHandler`** — TSan fix  
- `_next_cookie++` moved inside the mutex in `add()` — fixes a data race when `start_sending_heartbeats()` is called from the user-callback thread while the io_context thread is also in `add()`

**Protocol-handler work queues** — thread-safety fixes  
- All six protocol handlers (`MavlinkCommandSender`, `MavlinkFtpClient`, `MavlinkMissionTransferClient/Server`, `MavlinkParameterClient/Server`) replaced `LockedQueue` with plain `std::deque`; all enqueue calls from user threads are now wrapped in `asio::post()` so queue access is exclusively on the io_context thread  
- `MavlinkFtpClient::cancel_all_operations()` and `MavlinkParameterClient::cancel_all_param()` were previously racing with the io_context thread (one had a FIXME and was commented out); both now post the queue mutation to the io_context and block via `std::promise/future` until complete  
- `SystemImpl::cancel_all_param()` re-enabled (was disabled with a crash FIXME)

**CI robustness**  
- `concurrency` groups added to all workflow files to cancel stale runs on new pushes  
- `timeout-minutes` added to every `system_tests_runner` step (10 min macOS, 15 min Linux) so a hung job cannot block CI indefinitely  
- All `while (systems().size() == 0) { sleep(100ms); }` loops in system tests now have a 10 s deadline assertion — tests fail fast and clearly instead of hanging forever if system discovery is delayed

## Architecture after this PR

All async I/O completions, message dispatch, timer callbacks, protocol-handler ticks, and work-queue mutations run on a single `io_context` thread. The only remaining separate thread is the user-callback thread (intentional — keeps slow user code off the I/O path).

| Before | After |
|--------|-------|
| 1 OS thread per UDP connection | 0 extra threads for UDP receive |
| 1 OS thread per TCP connection | 0 extra threads for TCP receive |
| 1 OS thread for serial receive | 0 extra threads for serial receive |
| `work_thread` polling queues every 10 ms | io_context wakes immediately on events |
| `system_thread` running a periodic loop | `asio::steady_timer` on io_context |
| Received-message queue + mutex + CV | Direct `asio::post()` dispatch |
| Outgoing message queue + mutex | Direct `asio::post()` dispatch |
| `LockedQueue` + mutex in 6 protocol handlers | Plain `std::deque`, io_context-exclusive |
| `cancel_all_operations` / `cancel_all_param` racy or disabled | Posted to io_context, safe from any thread |

## Test plan

- [x] Linux superbuild compiles
- [x] Linux system tests pass (all 72)
- [x] Unit tests pass (290)
- [x] TSan clean (one warning from `CallEveryHandler::add()` pre-fix, resolved in `a325045d`)
- [x] System-test hang fixed — all unbounded discovery loops now have 10 s deadline guards
- [x] macOS CI passes

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -383,7 +383,8 @@ jobs:
       - name: Install tools
         run: |
           sudo apt-get update && sudo apt-get install -y \
-            python3-venv
+            python3-venv \
+            libasio-dev
       - name: git permission workaround
         run: |
           git config --global --add safe.directory "$GITHUB_WORKSPACE"
@@ -494,6 +495,9 @@ jobs:
         run: |
           echo "superbuild=-DSUPERBUILD=OFF" >> $GITHUB_ENV
           echo "cmake_prefix_path=-DCMAKE_PREFIX_PATH=$(pwd)/build/release/third_party/install" >> $GITHUB_ENV
+      - name: install asio (non-superbuild fallback)
+        if: steps.cache.outputs.cache-hit == 'true'
+        run: sudo apt-get update && sudo apt-get install -y libasio-dev
       - name: configure
         run: cmake $superbuild $cmake_prefix_path -DCMAKE_BUILD_TYPE=RelWithDebInfo -DBUILD_MAVSDK_SERVER=OFF -DBUILD_SHARED_LIBS=ON -DCMAKE_INSTALL_PREFIX=install -DWERROR=ON -Bbuild/release -H.
       - name: build


cpp/src/mavsdk/CMakeLists.txt
```diff
@@ -21,6 +22,7 @@ target_link_libraries(mavsdk
     tinyxml2::tinyxml2
     LibEvents::LibEvents
     mav::mav
+    $<BUILD_INTERFACE:asio::asio>
```
julianoes, line 25: Why is this required? Can't it just be `asio::asio`?

cpp/src/mavsdk/core/mavlink_command_sender.h
```diff
@@ -2,10 +2,12 @@
 
 #include "timeout_handler.h"
 #include "mavlink_include.h"
-#include "locked_queue.h"
```
julianoes, line 5: Can we remove the locked_queue.h file or is it still required somewhere?

cpp/src/mavsdk/core/mavlink_ftp_client.cpp
```diff
@@ -56,47 +57,74 @@ void MavlinkFtpClient::do_work()
         overloaded{
             [&](DownloadItem& item) {
                 if (!download_start(*work, item)) {
-                    work_queue_guard.pop_front();
+                    _work_queue.pop_front();
+                    if (!_work_queue.empty()) {
+                        asio::post(_io_context, [this] { do_work(); });
```
julianoes, line 62: I don't understand this, maybe you can explain. Why do we call `do_work()` within `do_work`, doesn't this call itself?

cpp/src/mavsdk/CMakeLists.txt
bansiesta: The `$<BUILD_INTERFACE:asio::asio>` generator expression ensures Asio is only visible during the build and is **not** propagated to consumers of the installed `mavsdk` library. Without it, anyone calling `find_package(mavsdk)` would also need to have Asio available on their system, because it would show up as a transitive link requirement. Since Asio is purely an implementation detail (header-only, baked into our compilation units), wrapping it in `BUILD_INTERFACE` keeps it from leaking into the installed targets.

cpp/src/mavsdk/core/mavlink_command_sender.h
bansiesta: Not yet — `locked_queue.h` is still included in `mavsdk_impl.h`, `http_loader.h`, and `mavlink_command_receiver.h`. The change in this PR only removed it from `mavlink_command_sender.h` (where the work queue is now driven by Asio's strand/post instead of the custom locked queue). Removing it entirely would require migrating those remaining users too, which is probably a follow-up once the Asio migration is more complete.

cpp/src/mavsdk/core/mavlink_ftp_client.cpp
bansiesta: It's not a direct recursive call — `asio::post` **schedules** `do_work()` as a new task on the io_context's event queue and returns immediately. The current invocation of `do_work()` then continues and returns normally. The io_context will dequeue and call `do_work()` again later, as a fresh call frame, once the current one has finished. This is a standard Asio pattern for processing a queue sequentially without blocking: instead of looping, each iteration posts the next one. The key distinction is that `asio::post` is deferred — it never calls the function right now, so there's no stack growth or recursion.

cpp/src/mavsdk/core/mavlink_command_sender.h
julianoes: What's stopping us from converting the rest?

cpp/src/mavsdk/core/mavlink_ftp_client.cpp
julianoes: Oh, I see, it's after having done work and seeing that there is still work to do.

cpp/src/mavsdk/core/mavlink_command_sender.h
bansiesta: The main blockers are the threading models of the two remaining users:

**`http_loader`** runs a dedicated blocking thread (`_work_thread`) that calls curl synchronously and relies on `LockedQueue`'s condition variable to sleep-wait for new work. To migrate this to Asio we'd need to switch to curl's multi/socket interface integrated with Asio's event loop (so curl I/O can be driven by the same io_context without blocking a thread). That's a fair-sized change on its own.

**`mavsdk_impl`'s user callback queue** (`_user_callback_queue` / `_process_user_callbacks_thread`) is the externally-visible callback dispatcher. It needs to keep guarantees around ordering and not calling back into the SDK from user callback context. Migrating it to Asio is straightforward in principle (post callbacks onto a strand), but it touches the user-facing threading contract so it warrants a bit more thought.

`mavlink_command_receiver.h` only includes the header; the actual `LockedQueue` usage is in `mavlink_command_receiver.cpp` and is a smaller lift.

None of it is fundamentally hard — I just wanted to keep this PR scoped to the core async I/O path (timers, work queues, FTP) and leave the HTTP and callback queue as follow-up once this lands and we can see if there are any surprises.

julianoes:
@JonasVautherin objections?

julianoes:
Some benchmarking results (increased the file size to 100 MB):

`main` (reference):

```
cmake --build build-relwithdebinfo && time build-relwithdebinfo/src/system_tests/system_tests_runner --gtest_filter="SystemTest.FtpUploadBigFile" --gtest_repeat=10 | rg "FTP upload: "

[08:51:39|Info ] FTP upload: 100000 KB in 127 ms  (787401 KB/s) (ftp_upload_file.cpp:156)
[08:51:41|Info ] FTP upload: 100000 KB in 35 ms  (2857142 KB/s) (ftp_upload_file.cpp:156)
[08:51:43|Info ] FTP upload: 100000 KB in 130 ms  (769230 KB/s) (ftp_upload_file.cpp:156)
[08:51:45|Info ] FTP upload: 100000 KB in 35 ms  (2857142 KB/s) (ftp_upload_file.cpp:156)
[08:51:47|Info ] FTP upload: 100000 KB in 34 ms  (2941176 KB/s) (ftp_upload_file.cpp:156)
[08:51:49|Info ] FTP upload: 100000 KB in 34 ms  (2941176 KB/s) (ftp_upload_file.cpp:156)
[08:51:51|Info ] FTP upload: 100000 KB in 45 ms  (2222222 KB/s) (ftp_upload_file.cpp:156)
[08:51:53|Info ] FTP upload: 100000 KB in 35 ms  (2857142 KB/s) (ftp_upload_file.cpp:156)
[08:51:55|Info ] FTP upload: 100000 KB in 34 ms  (2941176 KB/s) (ftp_upload_file.cpp:156)
[08:51:58|Info ] FTP upload: 100000 KB in 34 ms  (2941176 KB/s) (ftp_upload_file.cpp:156)

0.47s user 0.18s system 3% cpu 20.764 total
```

All system tests:
```
build-relwithdebinfo/src/system_tests/system_tests_runner
2.24s user 1.09s system 1% cpu 4:04.73 total
```


`asio-poc` (this branch):

```
[08:52:36|Info ] FTP upload: 100000 KB in 31 ms  (3225806 KB/s) (ftp_upload_file.cpp:156)
[08:52:36|Info ] FTP upload: 100000 KB in 31 ms  (3225806 KB/s) (ftp_upload_file.cpp:156)
[08:52:36|Info ] FTP upload: 100000 KB in 31 ms  (3225806 KB/s) (ftp_upload_file.cpp:156)
[08:52:36|Info ] FTP upload: 100000 KB in 31 ms  (3225806 KB/s) (ftp_upload_file.cpp:156)
[08:52:37|Info ] FTP upload: 100000 KB in 31 ms  (3225806 KB/s) (ftp_upload_file.cpp:156)
[08:52:37|Info ] FTP upload: 100000 KB in 31 ms  (3225806 KB/s) (ftp_upload_file.cpp:156)
[08:52:37|Info ] FTP upload: 100000 KB in 31 ms  (3225806 KB/s) (ftp_upload_file.cpp:156)
[08:52:37|Info ] FTP upload: 100000 KB in 34 ms  (2941176 KB/s) (ftp_upload_file.cpp:156)
[08:52:37|Info ] FTP upload: 100000 KB in 31 ms  (3225806 KB/s) (ftp_upload_file.cpp:156)
[08:52:37|Info ] FTP upload: 100000 KB in 31 ms  (3225806 KB/s) (ftp_upload_file.cpp:156)

0.41s user 0.14s system 35% cpu 1.550 total
```

All system tests:
```
build-relwithdebinfo/src/system_tests/system_tests_runner
2.19s user 0.97s system 2% cpu 1:49.79 total
```

So, the takeaway is that the FTP transfer is slightly faster and more consistent, but overall the tests run a lot faster!

bansiesta:
Those numbers look great — the `2×` system-test speedup and the more consistent FTP throughput line up with what I was hoping for from the Asio migration. The reduced wall-clock time is likely because timers and deferred work now execute on the io_context thread pool without the sleep overhead from the old LockedQueue condition-variable waits, so sequential tests don't each pay a fixed latency floor.

Happy to address anything else or rebase once Jonas has had a look.

PR #2823, opened 2026-03-29
telemetry: Use REQUEST_DATA_STREAM for ArduPilot set_rate

### LINKED ISSUE #2742: Set rate position and attitude_euler not working with ArduPilot
Apparently the set_rate command isn't working as it ought to against ArduPilot. This came up in https://discord.com/channels/1022170275984457759/1022185801615933460/threads/1453074335584485566.

Code sample to check:

```
_telemetry->set_rate_position_async(10.0, [](mavsdk::Telemetry::Result result) {
        if (result != mavsdk::Telemetry::Result::Success) {
            std::cerr << "Failed to set telemetry position rate" << std::endl;
        }
    });

    std::this_thread::sleep_for(std::chrono::milliseconds(100));

    _telemetry->set_rate_attitude_euler_async(10.0, [](mavsdk::Telemetry::Result result) {
        if (result != mavsdk::Telemetry::Result::Success) {
            std::cerr << "Failed to set drone euler rate" << std::endl;
        }
    });

    std::this_thread::sleep_for(std::chrono::milliseconds(100));
```

PavelGuzenfeld:
ArduPilot doesn't reliably ACK MAV_CMD_SET_MESSAGE_INTERVAL on serial links, so set_rate_position and set_rate_attitude_euler fail.

For ArduPilot, use the legacy REQUEST_DATA_STREAM message (fire-and-forget) instead. Maps position to MAV_DATA_STREAM_POSITION and attitude to MAV_DATA_STREAM_EXTRA1. Also fixes set_rate_velocity_ned which shares the same GLOBAL_POSITION_INT message.

Fixes #2742

### DIFF
--- a/cpp/src/mavsdk/plugins/telemetry/telemetry_impl.cpp
+++ b/cpp/src/mavsdk/plugins/telemetry/telemetry_impl.cpp
@@ -237,6 +237,11 @@ Telemetry::Result TelemetryImpl::set_rate_position(double rate_hz)
     _position_rate_hz = rate_hz;
     double max_rate_hz = std::max(_position_rate_hz, _velocity_ned_rate_hz);
 
+    if (_system_impl->effective_autopilot() == Autopilot::ArduPilot) {
+        send_request_data_stream(MAV_DATA_STREAM_POSITION, max_rate_hz);
+        return Telemetry::Result::Success;
+    }
+
     return telemetry_result_from_command_result(
         _system_impl->set_msg_rate(MAVLINK_MSG_ID_GLOBAL_POSITION_INT, max_rate_hz));
 }
@@ -271,6 +276,11 @@ Telemetry::Result TelemetryImpl::set_rate_attitude_quaternion(double rate_hz)
 
 Telemetry::Result TelemetryImpl::set_rate_attitude_euler(double rate_hz)
 {
+    if (_system_impl->effective_autopilot() == Autopilot::ArduPilot) {
+        send_request_data_stream(MAV_DATA_STREAM_EXTRA1, rate_hz);
+        return Telemetry::Result::Success;
+    }
+
     return telemetry_result_from_command_result(
         _system_impl->set_msg_rate(MAVLINK_MSG_ID_ATTITUDE, rate_hz));
 }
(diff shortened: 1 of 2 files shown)


julianoes:
> ArduPilot doesn't reliably ACK MAV_CMD_SET_MESSAGE_INTERVAL on serial links, so set_rate_position and set_rate_attitude_euler fail.

What does that mean "not reliably"? I know that's what @lakhmanisahil claims but that's not a very clear statement. It either supports it or it doesn't, right?

PavelGuzenfeld:
You're right to push back on "not reliably" — I investigated the ArduPilot source and the claim is wrong.

**ArduPilot fully supports `MAV_CMD_SET_MESSAGE_INTERVAL` for both `ATTITUDE` (30) and `GLOBAL_POSITION_INT` (33).** The mapping table in `GCS_Common.cpp` (`mavlink_id_to_ap_message_id()`) explicitly includes both. The command is processed, ACKed, and the interval is applied.

**The real issue is GCS stream rate override.** ArduPilot has two competing rate-setting systems:

1. `MAV_CMD_SET_MESSAGE_INTERVAL` — individual message rates (what MAVSDK uses)
2. `REQUEST_DATA_STREAM` — legacy, sets rates for entire message groups

When any GCS on the same link (MAVProxy, QGC, Mission Planner) sends `REQUEST_DATA_STREAM`, ArduPilot calls `initialise_message_intervals_for_stream()` which **overwrites** the individual intervals previously set via `SET_MESSAGE_INTERVAL`. So the rate setting works, but gets clobbered seconds later.

This is confirmed by multiple ArduPilot issues:
- https://github.com/ArduPilot/ardupilot/issues/17046
- https://github.com/ArduPilot/ardupilot/issues/19761
- https://github.com/ArduPilot/ardupilot/issues/28550

**Fix is on the ArduPilot side:** set `MAVn_OPTIONS` bit 2 (NOSTREAMOVERRIDE) on the port MAVSDK connects to, or ensure no other GCS is sending `REQUEST_DATA_STREAM` on the same link.

Given this, this PR is wrong — falling back to `REQUEST_DATA_STREAM` doesn't solve the root cause. I'll close this. The issue should be resolved with documentation pointing users to the `MAVn_OPTIONS` setting.

PR #2825, opened 2026-03-30
camera: add xz decompression support for camera definition files

bansiesta:
Camera definition files can be served compressed (`.xml.xz`) to reduce bandwidth, but MAVSDK was passing the compressed bytes directly to the XML parser causing a silent load failure.

The MAVLink camera definition spec only specifies `.xz` (not `.lzma`), so this PR adds support for `.xz` only.

This uses the existing `InflateLZMA` utility (which handles both xz/LZMA2 and raw LZMA streams via liblzma) and applies the same decompression pattern to the camera plugin for both the HTTP and MAVLink FTP download paths: after a successful download, check for the `.xz` extension, decompress to a temp file, then cache and load the decompressed XML as normal.

Only `camera_impl.cpp` is changed — no modifications to the internal inflate_lzma, CMakeLists, or mavlink_component_metadata files.

### DIFF
--- a/cpp/src/mavsdk/core/mavlink_parameter_client.cpp
+++ b/cpp/src/mavsdk/core/mavlink_parameter_client.cpp
@@ -423,16 +423,9 @@ void MavlinkParameterClient::cancel_all_param(const void* cookie)
 
     // We don't call any callbacks before erasing them as this is just used on destruction
     // where we don't care anymore.
-    // Use an explicit loop rather than erase(remove_if(...)) because LockedQueue::erase
-    // only takes a single iterator, so the erase-remove idiom would only remove the first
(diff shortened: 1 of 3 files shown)


bansiesta:
Done — removed .lzma support, keeping only .xz per the MAVLink camera definition spec.

bansiesta:
Done — removed `.lzma` support, keeping only `.xz` per the MAVLink camera definition spec. Renamed the implementation class to `InflateXZ` for clarity.

bansiesta:
Good point, thanks\! I've simplified the PR to only change `camera_impl.cpp` — it now checks for the `.xz` extension and decompresses using the existing `InflateLZMA` utility. All the internal core files (inflate_lzma, CMakeLists, mavlink_component_metadata) are left untouched.

PR #2827, opened 2026-03-30
mocap: Fix MavFrame enum sending wrong values on the wire

### LINKED ISSUE #2001: MavFrame enum is not correct?
In the file https://github.com/mavlink/MAVSDK/blob/main/src/mavsdk/plugins/mocap/include/plugins/mocap/mocap.h

I see MavFrame is defined like the following, which is used to fill up the ODOMETRY message for MAVLINK. I assume they are arrange so `MocapNed` is coded as 0 and `LocalFrd` is coded as 1

```
        enum class MavFrame {
            MocapNed, /**< @brief MAVLink number: 14. Odometry local coordinate frame of data given
                         by a motion capture system, Z-down (x: north, y: east, z: down).. */
            LocalFrd, /**< @brief MAVLink number: 20. Forward, Right, Down coordinate frame. This is
                         a local frame with Z-down and arbitrary F/R alignment (i.e. not aligned
                         with NED/earth frame). Replacement for MAV_FRAME_MOCAP_NED,
                         MAV_FRAME_VISION_NED, MAV_FRAME_ESTIM_NED.. */
        };
```

However, in MAVLINK they are coded both at 20. Is this something wrong here? Shouldn't we just expose the API as the native MAVLink enum, instead of MAVSDK custom ones?

https://github.com/mavlink/MAVSDK/blob/main/src/mavsdk/plugins/mocap/include/plugins/mocap/mocap.h#L293

PavelGuzenfeld:
The Odometry MavFrame proto enum uses values 0/1 but the actual MAVLink frame IDs are 14 (MAV_FRAME_MOCAP_NED) and 20 (MAV_FRAME_LOCAL_FRD). The static_cast sent proto values directly. Map to correct MAVLink constants instead.

Fixes #2001

### DIFF
--- a/cpp/src/mavsdk/plugins/mocap/mocap_impl.cpp
+++ b/cpp/src/mavsdk/plugins/mocap/mocap_impl.cpp
@@ -290,7 +290,8 @@ Mocap::Result MocapImpl::send_odometry(const Mocap::Odometry& odometry)
             channel,
             &message,
             autopilot_time_usec,
-            static_cast<uint8_t>(odometry.frame_id),
+            odometry.frame_id == Mocap::Odometry::MavFrame::LocalFrd ? MAV_FRAME_LOCAL_FRD :
+                                                                       MAV_FRAME_RESERVED_14,
             static_cast<uint8_t>(MAV_FRAME_BODY_FRD),
             odometry.position_body.x_m,
             odometry.position_body.y_m,


julianoes:
@alireza787b was working on this. I'll check his changes first I think.

cpp/src/mavsdk/plugins/mocap/mocap_impl.cpp
```diff
@@ -290,7 +290,8 @@ Mocap::Result MocapImpl::send_odometry(const Mocap::Odometry& odometry)
             channel,
             &message,
             autopilot_time_usec,
-            static_cast<uint8_t>(odometry.frame_id),
+            odometry.frame_id == Mocap::Odometry::MavFrame::LocalFrd ? MAV_FRAME_LOCAL_FRD :
```
julianoes, line 293: That is correct indeed.

PR #2834, opened 2026-04-01
core: fix MavlinkCommandReceiver calling callbacks while holding mutex

bansiesta:
## Problem

`receive_command_int` and `receive_command_long` had two bugs:

### 1. Callbacks invoked while holding `_mavlink_command_handler_table_mutex`

The mutex was locked for the entire loop including the `handler.callback(cmd)` call and the subsequent `queue_message()` call. If any callback (or `queue_message`) tries to register/unregister a command handler — which also acquires the same mutex — the result is a deadlock.

### 2. Lambda capturing local variable by reference (UB)

The lambda passed to `queue_message()` used `[&, this]`, capturing `maybe_command_ack` by reference. Since `maybe_command_ack` is a stack-local variable, it is destroyed before the lambda executes, resulting in undefined behaviour.

## Fix

- Copy matching handlers into a local `handlers_copy` vector under the lock, then release the lock before invoking callbacks and `queue_message()`.
- Change the lambda capture to `[ack = maybe_command_ack.value()]` (capture by value).
- Remove vestigial `#include "locked_queue.h"` — `LockedQueue` is not used in this file.

### DIFF
--- a/cpp/src/mavsdk/core/mavlink_command_receiver.cpp
+++ b/cpp/src/mavsdk/core/mavlink_command_receiver.cpp
@@ -65,14 +65,15 @@ void MavlinkCommandReceiver::receive_command_int(const mavlink_message_t& messag
             auto maybe_command_ack = handler.callback(cmd);
             if (maybe_command_ack) {
                 _server_component_impl.queue_message(
-                    [&, this](MavlinkAddress mavlink_address, uint8_t channel) {
+                    [ack = maybe_command_ack.value()](
+                        MavlinkAddress mavlink_address, uint8_t channel) {
                         mavlink_message_t response_message;
                         mavlink_msg_command_ack_encode_chan(
                             mavlink_address.system_id,
                             mavlink_address.component_id,
                             channel,
                             &response_message,
-                            &maybe_command_ack.value());
+                            &ack);
                         return response_message;
                     });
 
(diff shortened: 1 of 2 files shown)


julianoes:
I don't think we add or remove handlers within the callback, so I don't think we need to make this copy.

bansiesta:
You're right that callbacks don't call `register`/`unregister` — so the deadlock scenario via the handler table itself isn't realistic in practice.

The copy is still doing one useful job though: it ensures that `queue_message()` is also called outside the lock. The old code held `_mavlink_command_handler_table_mutex` all the way through the `queue_message()` call, which is a bit broader than needed (even if not a current deadlock trigger).

That said, the genuinely critical fix in this PR is the lambda capture bug — `[&, this]` capturing `maybe_command_ack` by reference where it goes out of scope before the lambda runs. That's real undefined behaviour regardless of the mutex question.

If you'd prefer, I can simplify to avoid the copy vector entirely: grab a `callback` value from the table under the lock, release the lock, then call the callback. That gives us the same "lock not held during callback" property with less machinery. Want me to rework it that way?

julianoes:
Let's keep the critical fix only.

bansiesta:
Done — simplified to only the critical lambda capture fix. The copy vector approach is gone; the lock now wraps the entire iteration as before. The only change is `[&, this]` → `[ack = maybe_command_ack.value()]` in both `receive_command_int` and `receive_command_long`.

julianoes (approved):

PR #2839, opened 2026-04-02
Expose full HOME_POSITION fields in subscribe_home

PavelGuzenfeld:
## Summary
- Add new `HomePosition` struct with all fields from MAVLink HOME_POSITION message
- `subscribe_home` now returns `HomePosition` instead of `Position` (breaking API change)
- New fields: local NED (x/y/z), surface quaternion, approach vector, timestamp
- Add system test verifying all fields propagate correctly end-to-end

## Motivation
Fixes #2310 — `subscribe_home` only returns global GPS position but HOME_POSITION also carries local NED position, attitude quaternion, approach vector, and timestamp that were being silently discarded.

## Changes
- **Proto**: Add `HomePosition` message, change `HomeResponse` to use it
- **Header**: Add `HomePosition` struct, update `HomeCallback`/`HomeHandle` types
- **Impl**: Populate all HOME_POSITION fields in `process_home_position()`
- **gRPC**: Add `translateToRpcHomePosition`/`translateFromRpcHomePosition`
- **Test**: System test `HomePositionFullFields` validates all fields

## Breaking Change
`subscribe_home` callback type changes from `std::function<void(Position)>` to `std::function<void(HomePosition)>`. `HomePosition` is a superset of `Position` — the same lat/lon/alt fields are present.

Part of v4 plan: #2830
Supersedes #2802

## Companion PR
Proto changes: https://github.com/mavlink/MAVSDK-Proto/pull/399

## Test plan
- [x] Unit tests pass
- [x] System test `HomePositionFullFields` validates GPS, local NED, quaternion, approach vector, and timestamp
- [x] Build clean

PR #2848, opened 2026-04-06
Fix data race in MavlinkDirectImpl callback teardown

PavelGuzenfeld:
## Summary
- Fix ThreadSanitizer-detected data race in `MavlinkDirectImpl` between callback dispatch and destructor

## Motivation
`deinit()` unregisters the system message handler, but an already-dispatched callback on another thread can still be executing when the destructor destroys `_callbacks`. TSan reports this as a race between `CallbackList::operator()` (reading `_impl` unique_ptr) and `~CallbackList()` (deleting it).

Triggered by `Telemetry_AltitudeTimestampUpdates` system test under ThreadSanitizer.

## Changes
- **`mavlink_direct_impl.h`**: Add `_callback_mutex` member
- **`mavlink_direct_impl.cpp`**: Lock `_callback_mutex` in the handler lambda (protects `_callbacks()` call) and in `deinit()` (waits for in-flight callback to complete)

Lock order is always `_callback_mutex` → CallbackList internal mutex, so no deadlock.

## Test plan
- [x] All 14 MavlinkDirect and Timestamp system tests pass
- [x] Full library build succeeds

### DIFF
--- a/cpp/src/mavsdk/plugins/mavlink_direct/mavlink_direct_impl.cpp
+++ b/cpp/src/mavsdk/plugins/mavlink_direct/mavlink_direct_impl.cpp
@@ -55,9 +55,9 @@ void MavlinkDirectImpl::init()
             mavlink_direct_message.target_component_id = message.target_component_id;
             mavlink_direct_message.fields_json = message.fields_json;
 
-            _callbacks.queue(
-                mavlink_direct_message,
-                [this](const auto& func) { _system_impl->call_user_callback(func); });
+            _callbacks.queue(mavlink_direct_message, [this](const auto& func) {
+                _system_impl->call_user_callback(func);
+            });
         });
 }
 


julianoes:
So, which one is the right fix, #2846 or this one? :smile:

PavelGuzenfeld:
Different races — both are needed:

- **#2846**: `UdpConnection` socket race — `async_receive_from()` on the io_context thread vs `_socket.close()` in the destructor on the main thread (`udp_connection.cpp`)
- **#2848**: `MavlinkDirectImpl` callback race — `CallbackList::operator()` on a worker thread vs `~CallbackList()` in the destructor on the main thread (`mavlink_direct_impl.cpp`)

Our TSan report (from the #2839 CI run):
```
WARNING: ThreadSanitizer: data race (pid=51444)
  Write of size 8 at 0x724000016300 by main thread:
    #3 CallbackList<MavlinkDirect::MavlinkMessage>::~CallbackList()
    #4 MavlinkDirectImpl::~MavlinkDirectImpl()  mavlink_direct_impl.cpp:41
  Previous atomic read by thread T205:
    #4 CallbackListImpl<MavlinkDirect::MavlinkMessage>::exec()
    #6 MavlinkDirectImpl::init()::$_0::operator()  mavlink_direct_impl.cpp:59
```

julianoes:
I see, awesome.

cpp/src/mavsdk/plugins/mavlink_direct/mavlink_direct_impl.cpp
```diff
@@ -55,19 +55,20 @@ void MavlinkDirectImpl::init()
             mavlink_direct_message.target_component_id = message.target_component_id;
             mavlink_direct_message.fields_json = message.fields_json;
 
-            // Call callbacks directly, queueing needs to happen in the user APIs
+            std::lock_guard<std::mutex> lock(_callback_mutex);
             _callbacks(mavlink_direct_message);
```
julianoes: Shouldn't we be using the user_callback queue for this?

PavelGuzenfeld:
Updated — replaced the mutex with the standard `call_user_callback` queue pattern that all other plugins use:

```cpp
_callbacks.queue(
    mavlink_direct_message,
    [this](const auto& func) { _system_impl->call_user_callback(func); });
```

The user callback queue serializes with plugin teardown, so the race is eliminated without any extra synchronization.

julianoes:
Before we merge this I need to rethink where the queueing onto the user queue should happen and check we don't do it twice by mistake.

julianoes:
I think this is correct. I don't see queuing onto the user callback queue within system_impl, so this should happen in the plugin, as done here.

PR #2851, opened 2026-04-07
Camera FTP download from component id

jazzvaz:
Camera should use component id for download extention files.


julianoes (approved):
Thanks!

Issue #2852, opened 2026-04-07
Mavsdk and Ros2 causes nondescript segfault


nbingham1:
Minimal example of node that tries to use Mavsdk. This example works if mavsdk is commented out or ros2 is not used. If Mavsdk is initialized at all in a Ros2 node, it causes a segfault.

```bash
~/workspace/px4_mission_service$ find *
CMakeLists.txt
src
src/node.cpp
~/workspace/px4_mission_service$ cat src/node.cpp 
#include <rclcpp/rclcpp.hpp>
#include <mavsdk/mavsdk.h>

struct Px4MissionService : public rclcpp::Node {
	mavsdk::Mavsdk sdk;

	Px4MissionService() 
		: rclcpp::Node("px4_mission_service"),
		  sdk(mavsdk::Mavsdk::Configuration{mavsdk::ComponentType::GroundStation}) 
	{}
};

int main(int argc, char **argv) {
	rclcpp::init(argc, argv);
	rclcpp::spin(std::make_shared<Px4MissionService>());
	rclcpp::shutdown();
	return 0;
}

~/workspace/px4_mission_service$ cat CMakeLists.txt 
cmake_minimum_required(VERSION 3.22)
project(px4_mission_service)

set(CMAKE_CXX_STANDARD 17)
set(CMAKE_CXX_STANDARD_REQUIRED ON)

find_package(rclcpp REQUIRED)
find_package(MAVSDK REQUIRED)

add_executable(mission_service src/node.cpp)
target_link_libraries(mission_service rclcpp::rclcpp MAVSDK::mavsdk)
~/workspace/px4_mission_service$ mkdir build
~/workspace/px4_mission_service$ cd build/
~/workspace/px4_mission_service/build$ CMAKE_BUILD_TYPE=Debug cmake ..
-- The C compiler identification is GNU 13.3.0
-- The CXX compiler identification is GNU 13.3.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found rclcpp: 28.1.16 (/opt/ros/jazzy/share/rclcpp/cmake)
[...]

PR #2856, opened 2026-04-08
fix: hide third-party symbols from libmavsdk.so to prevent clashes

julianoes:
Related to #2852 but on main.

libmavsdk.so statically links OpenSSL, tinyxml2, and curl but their symbols remained globally visible. When loaded alongside libraries that use the same dependencies (e.g. ROS2 FastRTPS), the dynamic linker would resolve calls to MAVSDK's bundled copies, causing ABI mismatches and segfaults.

Add linker symbol export lists (GNU ld version script for Linux, ld64 exported_symbols_list for macOS) that restrict exports to the mavsdk:: namespace plus a few test-support symbols. This drops exported symbols from ~20,000 to ~2,900.

### DIFF
--- a/cpp/src/mavsdk/core/CMakeLists.txt
+++ b/cpp/src/mavsdk/core/CMakeLists.txt
@@ -186,6 +186,22 @@ set_target_properties(mavsdk
         VISIBILITY_INLINES_HIDDEN ON
     )
 
+# Use a linker symbol export list to hide symbols from statically-linked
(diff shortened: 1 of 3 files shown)


JonasVautherin (approved):
I didn't know `-fvisibility=hidden` was not affecting the libraries linked statically!

julianoes:
@JonasVautherin, yes, neither.

PR #2858, opened 2026-04-09
Gimbal manager

jazzvaz:
There are 2 problems gimbal manager has:

1. On each heartbeat it continuously requesting components for gimbal info, even if the component rejects it. 
2. Gimbal vector is being used as potential and actual gimbals storage.

Solved by dummy requesting only 5 times and the dropping the component from requests. Only append gimbal vector on receiving gimbal manager info message. Also some cleanup for getting gimbal info.

### DIFF
--- a/cpp/src/mavsdk/plugins/gimbal/gimbal_impl.cpp
+++ b/cpp/src/mavsdk/plugins/gimbal/gimbal_impl.cpp
@@ -108,14 +108,14 @@ void GimbalImpl::process_heartbeat(const mavlink_message_t& message)
 {
     std::lock_guard<std::mutex> lock(_mutex);
 
-    auto maybe_gimbal = std::find_if(_gimbals.begin(), _gimbals.end(), [&](const GimbalItem& item) {
+    auto gimbal = std::find_if(_gimbals.begin(), _gimbals.end(), [&](const GimbalItem& item) {
         return item.gimbal_manager_compid == message.compid;
     });
 
     auto& discovery = _discovery[message.compid];
 
-    if (maybe_gimbal == _gimbals.end()) {
-        // Maybe new gimbal, request information.
+    if (gimbal == _gimbals.end()) {
+        // Maybe new gimbal, request information first.
         if (discovery.manager_info_requests_left > 0) {
             --discovery.manager_info_requests_left;
             request_gimbal_manager_information(message.compid);
@@ -123,31 +123,30 @@ void GimbalImpl::process_heartbeat(const mavlink_message_t& message)
         return;
     }
 
-    // Try to get device info, and notify the user with it. If exhausted, notify anyways.
-    const bool was_already_exhausted = (discovery.device_info_requests_left == 0);
-
-    if (discovery.device_info_requests_left > 0 && !discovery.device_info_received) {
-        --discovery.device_info_requests_left;
-
-        const uint8_t target_component_id =
-            (maybe_gimbal->gimbal_device_id > 0 && maybe_gimbal->gimbal_device_id <= 6) ?
-                message.compid :
-                maybe_gimbal->gimbal_device_id;
-
-        if (target_component_id != 0) {
-            request_gimbal_device_information(target_component_id);
+    if (discovery.device_info_received || discovery.device_info_requests_left == 0) {
+        // We already have the information or have given up, so we notify the user with what we
+        // have.
+        if (!discovery.notified) {
+            if (!discovery.device_info_received) {
+                LogWarn(
+                    "Continuing without GIMBAL_DEVICE_INFORMATION for compid {}", message.compid);
+            }
+            discovery.notified = true;
+            _gimbal_list_subscriptions.queue(gimbal_list_with_lock(), [this](const auto& func) {
+                _system_impl->call_user_callback(func);
+            });
         }
-
-        if (discovery.device_info_requests_left == 0 && !discovery.device_info_received) {
-            LogWarn("Continuing without GIMBAL_DEVICE_INFORMATION for compid {}", message.compid);
-        }
+        return;
     }
 
-    if (!discovery.notified && (discovery.device_info_received || was_already_exhausted)) {
-        discovery.notified = true;
-        _gimbal_list_subscriptions.queue(gimbal_list_with_lock(), [this](const auto& func) {
-            _system_impl->call_user_callback(func);
-        });
+    // Request gimbal device info
+    --discovery.device_info_requests_left;
+    const uint8_t target_component_id =
+        (gimbal->gimbal_device_id > 0 && gimbal->gimbal_device_id <= 6) ? message.compid :
+                                                                          gimbal->gimbal_device_id;
+
+    if (target_component_id != 0) {
+        request_gimbal_device_information(target_component_id);
     }
 }
 
@@ -166,7 +165,7 @@ void GimbalImpl::process_gimbal_manager_information(const mavlink_message_t& mes
 
     std::lock_guard<std::mutex> lock(_mutex);
 
-    auto maybe_gimbal = std::find_if(_gimbals.begin(), _gimbals.end(), [&](const GimbalItem& item) {
+    auto gimbal = std::find_if(_gimbals.begin(), _gimbals.end(), [&](const GimbalItem& item) {
         return item.gimbal_manager_compid == message.compid;
     });
 
@@ -179,7 +178,7 @@ void GimbalImpl::process_gimbal_manager_information(const mavlink_message_t& mes
         discovery.notified = false;
     };
 
-    if (maybe_gimbal == _gimbals.end()) {
+    if (gimbal == _gimbals.end()) {
         // Register new gimbal manager.
         GimbalItem new_item{};
         new_item.gimbal_manager_compid = message.compid;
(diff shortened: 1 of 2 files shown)


jazzvaz:
I also think a better design would be components that are, for example, gimbal managers send the gimbal manager info with some low frequency, instead of us requesting it.

bansiesta (commented):
The overall direction is good — separating discovery state from confirmed gimbal state is cleaner, and fixing the infinite re-request loop is the right call. A few things worth thinking through:

**1. `_discovery` key mismatch (likely bug)**

`device_info_requests_left` is *set* in `_discovery[manager_compid]` (in `process_gimbal_manager_information`) and *decremented* via the manager's heartbeat (`_discovery[message.compid]` in `process_heartbeat`). But it's *reset to 0* using `_discovery[message.compid]` in `process_gimbal_device_information`, where `message.compid` is the **device's** compid — which can differ from the manager's. When manager ≠ device compid, the reset lands in a different map entry and has no effect, so device info ends up being requested all 5 times even after it's already received. Not catastrophic (just redundant requests), but logically wrong.

**2. Double subscriber notification for new gimbals**

Previously, subscribers got one notification once the gimbal was fully confirmed. Now they get one when manager info arrives (gimbal added with empty vendor/model names), and another when device info arrives. Worth calling out in case callers aren't expecting to see a gimbal with blank names in the list.

**3. Device ID change doesn't re-trigger device info**

If `process_gimbal_manager_information` sees a known gimbal with a changed `gimbal_device_id`, it updates the field and returns — but doesn't reset `device_info_requests_left` to re-fetch device info from the new device, and doesn't notify subscribers. Probably an edge case that rarely matters, but the old `check_is_gimbal_valid` path would have handled it.

**4. `_discovery` grows unboundedly**

Every component that ever sends a heartbeat gets an entry that's never removed. Minor, but worth noting.

jazzvaz:
1. Done.
2. Done.
3. Done.
4. I cannot find better design without storing component has seen and not spamming every component with requests. That's why I suggested to remove asking component for info and let component sends it so we register them.

julianoes:
> That's why I suggested to remove asking component for info and let component sends it so we register them.

Right. That would require a spec change and all gimbal devices would have to support it, so we can't really easily change that.

bansiesta (commented):
The second commit addresses all four points well — the `notified` + `device_info_received` flags cleanly solve the single-notification and key-mismatch problems, and `reset_device_discovery()` handles the device-ID-change case.

One small remaining thing: in `process_heartbeat`, once `device_info_received` is true the counter may still be > 0, so redundant requests keep going out until it drains. Easy fix:

```cpp
if (discovery.device_info_requests_left > 0 && !discovery.device_info_received) {
```

Also the second commit message (`stage`) should be cleaned up / squashed before merge.

cpp/src/mavsdk/plugins/gimbal/gimbal_impl.cpp
```diff
@@ -112,24 +112,43 @@ void GimbalImpl::process_heartbeat(const mavlink_message_t& message)
...
+        if (discovery.manager_info_requests_left > 0) {
+            --discovery.manager_info_requests_left;
+            request_gimbal_manager_information(message.compid);
         }
-    }();
+        return;
+    }
+
+    // Try to get device info, and notify the user with it. If exhausted, notify anyways.
+    const bool was_already_exhausted = (discovery.device_info_requests_left == 0);
```
julianoes: I don't understand this. Does this mean that we will notify if we found a MAVLink component that never actually sent a gimbal device status or information?

jazzvaz: The basic logic is: for each component in heartbeat, we request gimbal manager info, if it is, then we ask for gimbal device info. If there is no gimbal device info, we still notify user with empty info because there is a gimbal manager. It is same logic as old code.

bansiesta (dismissed):
All concerns from the previous reviews are addressed: the  key now correctly uses  in , the  guard stops redundant requests once device info is received, the  flag prevents double subscriber notifications,  handles device ID changes, and the commits are squashed. LGTM.

cpp/src/mavsdk/plugins/gimbal/gimbal_impl.cpp
julianoes: Hm, I guess I'm not following the logic. We are in here when we process a heartbeat. And if we were "already exhausted", so we requested device_info several times, then we will notify... right?

bansiesta: Yes, exactly — that's intentional. The flow is:
1. Heartbeats trigger gimbal manager info requests (up to 5 times) and add the gimbal to `_gimbals` once the response arrives.
2. Subsequent heartbeats then request gimbal device info (again up to 5 times). `was_already_exhausted` is true once `device_info_requests_left` has drained to 0.
3. The notify block fires on the first heartbeat where either device info was received *or* requests are exhausted — whichever comes first. The `notified` flag ensures it only fires once.

So yes: if all 5 device-info requests go unanswered, the subscriber is notified on the next heartbeat with just the manager-level info and empty vendor/model strings. The warning log on exhaustion ("Continuing without GIMBAL_DEVICE_INFORMATION...") covers that case. The idea is to surface the gimbal even when the device-info message never arrives, rather than silently dropping it.

julianoes: What about gimbal_device_status? I don't want to get notified if a gimbal device does neither send device info nor status.

bansiesta: Fair point — the `was_already_exhausted` fallback was too eager. I've pushed a fix on [bansiesta/MAVSDK@fix/gimbal-device-status-check](https://github.com/bansiesta/MAVSDK/tree/fix/gimbal-device-status-check) (commit 2402d2b) that addresses this.

The changes:
- Add a `device_status_received` flag to `GimbalDiscovery`.
- In `process_heartbeat`, change the notification guard from `was_already_exhausted` to `was_already_exhausted && device_status_received` — so exhausting retries alone is no longer enough.
- In `process_gimbal_device_attitude_status`, on the first status message, set `device_status_received = true` and trigger the list notification right away if requests were already exhausted.

Result: a gimbal component that only sends heartbeats (but neither device info nor device status) is silently ignored.

jazzvaz: We request gimbal device_info only when we receive gimbal manager message. After if we dont get gimbal device message, we notify in any case. Heartbeat function is just kind of state machine tick for that logic.

julianoes: And do we notify the user about a gimbal device when we have only received a heartbeat but not a device status or info?

bansiesta:
With the current PR branch (jazzvaz's commits), yes — once `device_info_requests_left` drains to zero, the next heartbeat tick notifies regardless of whether any `GIMBAL_DEVICE_ATTITUDE_STATUS` was ever received.

The fix is in [bansiesta/fix/gimbal-device-status-check](https://github.com/bansiesta/MAVSDK/tree/fix/gimbal-device-status-check) (commit 2402d2b): adds a `device_status_received` flag to `GimbalDiscovery` and changes the exhausted-fallback guard to `was_already_exhausted && device_status_received`. A status arrival while requests are exhausted also triggers the list notification immediately. That way a component that only heartbeats is silently ignored.

cpp/src/mavsdk/plugins/gimbal/gimbal_impl.cpp
bansiesta: With the current PR branch (jazzvaz's commits), yes — after exhausting the 5 requests the next heartbeat notifies regardless of whether any `GIMBAL_DEVICE_ATTITUDE_STATUS` was received.

The stricter behaviour you're describing is in [bansiesta/fix/gimbal-device-status-check](https://github.com/bansiesta/MAVSDK/tree/fix/gimbal-device-status-check): notification only fires when `was_already_exhausted && device_status_received`, so a component that only heartbeats is silently dropped. Happy to rebase that onto this PR if you'd like that behaviour included here.

jazzvaz: I have refactored the if statements inside heartbeat function for clarity.

julianoes:
@jazzvaz ok, so what's the behaviour now? Does it notify when no gimbal device info or status has been received, ever?

bansiesta:
Yes — after jazzvaz's refactor the logic is the same as before: once `device_info_requests_left` drains to 0, the next heartbeat notifies regardless of whether any `GIMBAL_DEVICE_INFORMATION` or `GIMBAL_DEVICE_ATTITUDE_STATUS` was ever received (lines 126–137). A component that only heartbeats will silently exhaust its 5 retries, then trigger the subscriber list with whatever (empty) device data is available.

My [fix/gimbal-device-status-check](https://github.com/bansiesta/MAVSDK/tree/fix/gimbal-device-status-check) branch adds a `device_status_received` flag so the exhausted-fallback guard becomes `device_info_requests_left == 0 && device_status_received` — making heartbeat-only components silently ignored. Happy to rebase it onto this PR if you want that stricter behaviour included.

jazzvaz:
@julianoes yes it does, and it is same logic as original code. If you want me to change it, I am okay with that. The current logic: heartbeat -> ask component for gimbal MANAGER status  (5 attempts) -> if it is, add add to gimbal vector and ask manager for gimbal DEVICE info (5 attempts)-> if received/not gimbal info notify the user because gimbal is pushed into vector.

julianoes:
Oh, I see. I wasn't aware of that behavior. I wonder if that's what we want.

bansiesta:
My take: **no**, we should not notify for a component that sent only heartbeats and nothing else. A heartbeat alone is not enough to confirm a working gimbal — we'd be surfacing incomplete, potentially misleading data to the subscriber.

The two reasonable options are:

1. **Require at least a gimbal manager info response** (already ensured by the existing  push guard) **plus at least one ** before notifying. This is what my [fix/gimbal-device-status-check](https://github.com/bansiesta/MAVSDK/tree/fix/gimbal-device-status-check) branch implements — trivial flag, no protocol changes needed.

2. **Require device info or device status, not both.** Notify once either arrives (or both), but never on exhausted-retries alone. Slightly looser, but still prevents heartbeat-only ghosts.

Option 1 is the stricter/safer choice and the one I already have tested. Happy to rebase it onto this PR if you both agree that's the right direction.

jazzvaz:
The logic itself can be done in another PR, as this one focuses on fixing discovery spamming.

julianoes:
Thanks @jazzvaz, let's get this in and then iterate on the logic change next.

PR #2844, opened 2026-04-04
refactor: rename all public C++ headers from .h to .hpp

bansiesta:
Part of v4 plan: #2830.

Renames all 55 public headers under `include/mavsdk/` and `include/plugins/*/` from `.h` to `.hpp` to follow the C++ convention of using `.hpp` for C++-only headers.

Updated across 333 files:
- Internal mavsdk core and plugin sources
- mavsdk_server gRPC service implementations
- Examples
- Jinja2 code generation templates
- `CMakeLists.txt` `install(FILES ...)` rules

**Breaking change for v4**: downstream code including `<mavsdk/mavsdk.h>`, `<mavsdk/plugins/action/action.h>`, etc. must update to `.hpp`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2861, opened 2026-04-13
refactor: rename mavsdk_export.hpp to mavsdk_export.h for C compatibility

bansiesta:
## Summary

`mavsdk_server_api.h` is a C-compatible public header (uses `extern "C"`) but was `#include`-ing `mavsdk_export.hpp`. Including a `.hpp` file from C code is problematic — `.hpp` signals C++ to build systems and tooling, and any future C++ additions to that file would silently break C consumers.

`mavsdk_export.hpp` contains only preprocessor macros and is already C-compatible in content. This PR simply renames it to `mavsdk_export.h` and updates all include sites.

### Changes

- Rename `cpp/src/mavsdk/core/include/mavsdk/mavsdk_export.hpp` → `mavsdk_export.h`
- Update all 90+ include sites across `cpp/` (source files, headers, CMakeLists.txt)
- No content changes — purely a rename

### Why

`mavsdk_server_api.h` (a C-compatible header) includes this file. A `.hpp` extension must not appear in headers that may be consumed by C code.


julianoes:
@bansiesta note that we need to cherry-pick this for v3 once merged here and make a separate backport PR.

bansiesta:
@julianoes — investigated the v3 backport as requested.

**Result: v3 is already in the correct state — no backport needed.**

After cherry-picking both commits from this PR onto a branch off `upstream/v3`, both applied as **empty** (no-op):

- The v3 branch uses `src/` as its root (not `cpp/src/` like main)
- v3 **never had** `mavsdk_export.hpp` — the rename in #2844 only touched the `cpp/` tree which doesn't exist in v3
- `src/mavsdk/core/include/mavsdk/mavsdk_export.h` already exists in v3 with the `.h` extension
- `src/mavsdk_server/src/mavsdk_server_api.h` already `#include`s `"mavsdk_export.h"` correctly  
- All include sites in `src/` already reference `mavsdk_export.h`
- `templates/plugin_h/file.j2` already references `mavsdk_export.h`

```
$ grep -r "mavsdk_export\.hpp" src/
# (no output)
$ ls src/mavsdk/core/include/mavsdk/mavsdk_export*
src/mavsdk/core/include/mavsdk/mavsdk_export.h
```

Since there are no actual changes to make, GitHub won't allow opening a PR (no commits between the branches). The v3 branch is clean.

JonasVautherin:
I'll try to rebuild MAVSDK-Java with those changes in the next release 👍

PR #2868, opened 2026-04-20
fix: only announce gimbal after device info or attitude status received

bansiesta:
Extracted from the review discussion on #2858.

## Problem

`check_is_gimbal_valid()` marks a gimbal valid and notifies subscribers as soon as `GIMBAL_DEVICE_INFORMATION` requests are exhausted — even when the component never sent any device-level message. A component that sends only a heartbeat + `GIMBAL_MANAGER_INFORMATION` can trigger a spurious gimbal-found notification.

Julian's question in the #2858 review:
> And do we notify the user about a gimbal device when we have only received a heartbeat but not a device status or info?

## Fix

Three-case decision in `check_is_gimbal_valid()`:

| Condition | Action |
|-----------|--------|
| `GIMBAL_DEVICE_INFORMATION` received | Announce immediately (unchanged) |
| Requests exhausted + `GIMBAL_DEVICE_ATTITUDE_STATUS` received | Announce with a warning |
| Requests exhausted, no device info, no status yet | Wait silently |

A new `gimbal_device_attitude_status_received` flag is added to `GimbalItem`. `process_gimbal_device_attitude_status()` sets it on the first status message and calls `check_is_gimbal_valid()` to trigger any deferred notification.

## Verified

Builds cleanly, style-checked.

Co-authored-by: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/cpp/src/mavsdk/plugins/gimbal/gimbal_impl.hpp
+++ b/cpp/src/mavsdk/plugins/gimbal/gimbal_impl.hpp
@@ -128,7 +128,7 @@ private:
     void process_gimbal_device_information(const mavlink_message_t& message);
     void process_gimbal_device_attitude_status(const mavlink_message_t& message);
     void process_attitude(const mavlink_message_t& message);
-    void check_is_gimbal_valid(GimbalItem* gimbal);
+    void check_is_gimbal_valid(GimbalItem& gimbal);
 
     void set_angles_async_internal(
         int32_t gimbal_id,
--- a/cpp/src/mavsdk/plugins/gimbal/gimbal_impl.hpp
+++ b/cpp/src/mavsdk/plugins/gimbal/gimbal_impl.hpp
@@ -101,21 +101,23 @@ private:
         std::string vendor_name;
         std::string model_name;
         std::string custom_name;
-        bool gimbal_device_attitude_status_received{false};
         Gimbal::ControlStatus control_status{0, Gimbal::ControlMode::None, 0, 0, 0, 0};
         Gimbal::Attitude attitude{};
     };
 
+    // Per-compid discovery state. A compid is first seen via heartbeat; we request manager info,
+    // then device info. The gimbal is only promoted to _gimbals (and announced) once we receive
+    // GIMBAL_DEVICE_INFORMATION or GIMBAL_DEVICE_ATTITUDE_STATUS.
     struct GimbalDiscovery {
         unsigned manager_info_requests_left{5};
+        // Set to true once GIMBAL_MANAGER_INFORMATION is received.
+        bool has_manager_info{false};
+        // Copied from GIMBAL_MANAGER_INFORMATION.gimbal_device_id once known.
+        uint8_t gimbal_device_id{0};
+        // Remaining device info requests for the pending gimbal.
         unsigned device_info_requests_left{0};
-        bool device_info_received{false};
-        bool notified{false};
-        // Counts heartbeats received after device-info requests are exhausted but before
-        // GIMBAL_DEVICE_ATTITUDE_STATUS arrives. Used as a fallback so a gimbal that never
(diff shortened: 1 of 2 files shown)


julianoes:
@bansiesta make sure to fix the conflicts.

bansiesta:
Conflicts resolved — the branch is rebased onto the latest `upstream/main` (d3d771837). All CI checks are green. Ready for review.

cpp/src/mavsdk/plugins/gimbal/gimbal_impl.hpp
```diff
@@ -127,6 +128,7 @@ class GimbalImpl : public PluginImplBase {
     void process_gimbal_device_information(const mavlink_message_t& message);
     void process_gimbal_device_attitude_status(const mavlink_message_t& message);
     void process_attitude(const mavlink_message_t& message);
+    void check_is_gimbal_valid(GimbalItem* gimbal);
```
julianoes: Can we use a reference instead of a ptr?

julianoes (approved):

julianoes:
@jazzvaz are you ok with this change?

jazzvaz:
@julianoes let me review and I will let you know. Thanks

jazzvaz:
The only issue I see: when we receive a gimbal manager message, we append it to the gimbals vector and assign it a gimbal ID. Therefore, it must be announced at some point, as we cannot wait indefinitely.

julianoes:
@jazzvaz any objections now? Otherwise I'm tempted to get this in.

jazzvaz:
IMHO, the cleanest logic would be to add a gimbal to the vector and announce it immediately. That way, MAVSDK never has gimbal IDs that are known internally but not yet announced or still pending.

The main question is: at which stage should we add it to the gimbal vector?

For example:
- A manager message tells us that a controller exists.
- Device info tells us that a gimbal exists.
- An attitude message is feedback from the gimbal.

So which of these should be the trigger for adding the gimbal to the vector?

But overall I don't mind of getting this in.

julianoes:
True, that is the question. I think we need to discover at least a gimbal manager, because we will send commands o the gimbal manager, so without discovering it, we can't actually control a gimbal.

jazzvaz:
If we need to discover at least gimbal manager, then this PR is redundant. Originally these are the steps: receive gimbal manager message, append to gimbal vector -> wait for 5 messages of gimbal device info -> with or without device info notify about new gimbal. With this PR if there is no gimbal device info, we wait for attitude from gimbal, and then announce.

julianoes:
Oh, I see, actually if there is no gimbal device, and the gimbal manager _is_ the gimbal device as well, then it ought to send out the attitude status. If we don't have an attitude status, we really don't have a gimbal, do we?

jazzvaz:
That's true but then we still need to announce gimbal as we append to gimbal vector based on gimbal manager message. We can also follow [mavlink gimbal protocol v2](https://mavlink.io/en/services/gimbal_v2.html).

julianoes:
> We can also follow [mavlink gimbal protocol v2](https://mavlink.io/en/services/gimbal_v2.html).

> The autopilot would directly control attached gimbals in response to gimbal manager commands. In addition to sending gimbal manager messages when appropriate, it should also send/stream required gimbal device messages, such as [GIMBAL_DEVICE_ATTITUDE_STATUS](https://mavlink.io/en/services/gimbal_v2.html#GIMBAL_DEVICE_ATTITUDE_STATUS).

Right, so we should only add it to the gimbal list when we either receive a gimbal device information or a gimbal attitude status. If we don't get either, it's not a valid gimbal and we shouldn't expose it.

julianoes (approved):
I'm happy with this. @jazzvaz, any objections?

jazzvaz:
@julianoes no objections, i am good with it.

PR #2867, opened 2026-04-20
example: add tlog_recorder to capture MAVLink traffic as .tlog

bansiesta:
Closes #2142.

## What

Adds a new `tlog_recorder` example that records all incoming MAVLink messages to a `.tlog` telemetry log file — the format used by QGroundControl, Mission Planner, and pymavlink.

## How it works

Each record in a `.tlog` is:
- **8 bytes**: big-endian timestamp in microseconds since the Unix epoch
- **N bytes**: raw MAVLink wire packet

The example uses `intercept_incoming_messages_async()` (rather than MavlinkDirect) because that API delivers `mavlink_message_t` directly, which `mavlink_msg_to_send_buffer()` can re-serialise to wire bytes losslessly. MavlinkDirect only exposes decoded JSON, so the raw wire format cannot be reconstructed from it.

## Usage

```sh
./tlog_recorder udpin://0.0.0.0:14540          # auto-named YYYY-MM-DDTHH-MM-SS.tlog
./tlog_recorder serial:///dev/ttyUSB0:57600 flight.tlog
```

The file can be opened directly in QGC (*File › Open Log File*) or inspected with pymavlink:

```sh
python3 -m pymavlink.tools.mavlogdump output.tlog
```

## Verified

- Builds cleanly against the debug install (`-Wall -Wextra`, no warnings)
- Style-checked with `tools/fix_style.py`

Co-authored-by: Claude Sonnet 4.6 <noreply@anthropic.com>

PR #2869, opened 2026-04-21
fix: eliminate TSAN races in MavlinkDirectImpl subscribe/deinit

bansiesta:
## Summary

Two TSAN data races in `MavlinkDirectImpl`:

• **`subscribe_message` use-after-free**: The `filtering_callback` captured `this`, but the wrapper is enqueued into the user callback queue before `deinit()` runs and can fire after `MavlinkDirectImpl` is destroyed. Fix: remove `this` capture; invoke `callback(message)` directly. The callback is already dispatched through the right execution context by the `CallbackList` machinery.

• **`deinit()` teardown race**: A message arriving in the narrow window between the last heartbeat handler and `unregister_libmav_message_handler()` could still enqueue a wrapper that fires after destruction. Fix: call `_callbacks.clear()` at the top of `deinit()` so no new wrappers are enqueued from that point forward.

(Extracted from #2867 per Julian's request.)

### DIFF
--- a/cpp/src/mavsdk/plugins/mavlink_direct/mavlink_direct_impl.cpp
+++ b/cpp/src/mavsdk/plugins/mavlink_direct/mavlink_direct_impl.cpp
@@ -55,14 +55,19 @@ void MavlinkDirectImpl::init()
             mavlink_direct_message.target_component_id = message.target_component_id;
             mavlink_direct_message.fields_json = message.fields_json;
 
-            _callbacks.queue(mavlink_direct_message, [this](const auto& func) {
-                _system_impl->call_user_callback(func);
-            });
+            _callbacks(mavlink_direct_message);
         });
 }
 
 void MavlinkDirectImpl::deinit()
 {
+    // Synchronise with any in-flight I/O-thread dispatch: _callbacks()/exec() holds
+    // the CallbackList's internal mutex, and clear() acquires the same mutex, so
+    // clear() blocks until the current exec() finishes.  After clear() the list is
+    // empty; subsequent exec() calls are no-ops, making it safe to destroy the
+    // CallbackList.  This also prevents any further user callbacks from being enqueued.
+    _callbacks.clear();
+
     if (_system_subscription.valid()) {
         _system_impl->unregister_libmav_message_handler(_system_subscription);
         _system_subscription = {};
@@ -153,15 +158,16 @@ MavlinkDirect::Result MavlinkDirectImpl::send_message(MavlinkDirect::MavlinkMess
 MavlinkDirect::MessageHandle MavlinkDirectImpl::subscribe_message(
     std::string message_name, const MavlinkDirect::MessageCallback& callback)
 {
-    // Add filtering callback to internal CallbackList
+    // filtering_callback is called synchronously on the I/O thread (via _callbacks()/exec()
+    // in init()).  It captures 'this' only to reach call_user_callback, and is never
+    // enqueued anywhere — so there is no lifetime issue.  deinit() calls _callbacks.clear()
+    // before unregistering the system handler, which blocks until any in-flight exec()
+    // completes; after that, filtering_callback will never be called again.
     auto filtering_callback =
         [this, message_name, callback](const MavlinkDirect::MavlinkMessage& message) {
-            // Apply message filtering (empty string means all messages)
             if (!message_name.empty() && message_name != message.message_name) {
                 return;
             }
-
-            // Queue user callback to user callback thread
             _system_impl->call_user_callback([callback, message]() { callback(message); });
         };
 


cpp/src/mavsdk/plugins/mavlink_direct/mavlink_direct_impl.cpp
```diff
@@ -153,17 +157,19 @@ MavlinkDirect::Result MavlinkDirectImpl::send_message(MavlinkDirect::MavlinkMess
...
+    // MavlinkDirectImpl.  Capturing only value types that own their data is safe.
+    // deinit() clears _callbacks before unregistering the system subscription, so no
+    // new wrappers are enqueued after deinit(); any wrapper already in the queue at that
+    // point is self-contained and safe to execute after the object is gone.
+    auto filtering_callback = [message_name,
+                               callback](const MavlinkDirect::MavlinkMessage& message) {
+        if (!message_name.empty() && message_name != message.message_name) {
+            return;
+        }
+        callback(message);
```
julianoes: Aren't we supposed to use `call_user_callback`?

bansiesta: Good catch questioning this — let me explain the dispatch chain here.

In `init()`, the `queue_func` passed to `_callbacks.queue()` is:
```cpp
[this](const auto& func) { _system_impl->call_user_callback(func); }
```
`CallbackList::queue()` calls this for each subscriber, passing a wrapper `[filtering_callback, message]() { filtering_callback(message); }`. So `call_user_callback` is already invoked here — `filtering_callback` is dispatched onto the user callback thread (or `_callback_executor`) before it even runs.

By the time `filtering_callback(message)` executes, it is **already on the user callback thread** (or the executor). Calling `call_user_callback` again inside it would double-dispatch: `callback` would be queued a second time and run in a later slot. The direct `callback(message)` call is equivalent and avoids that extra hop.

The TSAN fix here is: the original code had `filtering_callback` capture `this`, which was a use-after-free if the wrapper sat in the user callback queue past `MavlinkDirectImpl`'s destruction. By not capturing `this`, any wrapper already in the queue when `deinit()` fires is self-contained and safe to execute. The `_callbacks.clear()` in `deinit()` prevents new wrappers from being enqueued.

bansiesta:
Yes, closely related! Here's how they chain:

**#2848** fixed a race between `_callbacks(mavlink_direct_message)` (called synchronously on the I/O thread) and `~CallbackList()` (destroying `_impl`). The fix was to switch to `_callbacks.queue(..., [this](const auto& func) { _system_impl->call_user_callback(func); })`, which dispatches each `filtering_callback` onto the user callback thread rather than calling it synchronously on the I/O thread.

**But that introduced a new race.** Once `filtering_callback` is dispatched to the user callback queue (rather than called inline), it can sit in that queue past the point where `MavlinkDirectImpl` is destroyed. `filtering_callback` captures `this` and calls `_system_impl->call_user_callback()` — UAF when the wrapper finally executes.

**#2869 fixes that.** Two parts:
1. `deinit()` now calls `_callbacks.clear()` before unregistering — proper unsubscription so no new wrappers are enqueued after teardown begins.
2. `filtering_callback` no longer captures `this`. It doesn't need to: the `queue_func` added by #2848 already routes it through `call_user_callback`, so `filtering_callback` executes on the user callback thread. The second `call_user_callback` inside `filtering_callback` was a leftover from the pre-#2848 era when `filtering_callback` ran synchronously on the I/O thread and had to do its own routing. After #2848 it became redundant — and the source of the new race.

bansiesta:
You're right, it was in the wrong place. Just pushed an updated fix.

**Root cause of the confusion:** #2848 put `call_user_callback` in `init()`'s `queue_func` (dispatching `filtering_callback` onto the user callback thread) rather than in `filtering_callback` itself (where the user callback is actually invoked). That created the new TSAN race: `filtering_callback` — with `this` captured — sitting in the user callback queue past `MavlinkDirectImpl`'s destruction.

**The correct fix (now pushed):**
1. Restore `_callbacks(mavlink_direct_message)` in `init()` — `filtering_callback` runs synchronously on the I/O thread, never enqueued, no lifetime issue with `this`.
2. Keep `_system_impl->call_user_callback(...)` in `filtering_callback` — that's where the user's callback is actually dispatched, exactly as Julian pointed out.
3. Add `_callbacks.clear()` in `deinit()` — `clear()` and `exec()` share the CallbackList's internal mutex, so `clear()` blocks until any in-flight I/O-thread `exec()` finishes, then empties the list. Subsequent `exec()` calls are no-ops; `~CallbackList()` runs safely afterwards. This is the fix for the original #2848 race, done properly.

julianoes (approved):

julianoes (changes requested):
Hang on, you made a bit of a mess. Gimbal stuff shouldn't be in this PR.

julianoes (approved):

PR #2873, opened 2026-04-25
feat(core): expose math_utils helpers in public headers

### LINKED ISSUE #2693: Add euler quaternion conversion functions
When we remove MavlinkPassthrough which pulls in the Mavlink headers, users of the library no longer have access to [mavlink_conversions.h](https://github.com/ArduPilot/pymavlink/blob/master/generator/C/include_v2.0/mavlink_conversions.h), so we probably need to add them somehow in the headers.

g-ampo:
Closes #2693.

Move `math_utils.hpp` to the public include directory so users can convert between `Quaternion` and `EulerAngle` without `MavlinkPassthrough` (and hence without `mavlink_conversions.h`) .

Also clamp the `asinf` input in `to_euler_angle_from_quaternion` to `[-1, 1]` to prevent NaN when a quaternion approaches gimbal-lock ; same fix landing in ArduPilot/pymavlink#1208 for the upstream C library.


julianoes (changes requested):
Could you please add a test to math_utils_test.cpp that tests the constrain you added?

julianoes:
Also note that you will have to make sure there are docstrings and you run `generate_docs.sh` to update the API refernence docs.

g-ampo:
@julianoes  I've addressed both: added the gimbal-lock constrain test in math_utils_test.cpp and added docstrings + ran generate_docs.sh. Happy to adjust anything !

cpp/src/mavsdk/core/include/mavsdk/math_utils.hpp
```diff
@@ -0,0 +1,116 @@
...
+ */
+MAVSDK_TEST_EXPORT EulerAngle to_euler_angle_from_quaternion(Quaternion quaternion);
+
+/**
+ * @brief Convert an EulerAngle (degrees) to a Quaternion (ZYX Tait-Bryan).
+ *
+ * @param euler_angle Input EulerAngle in degrees.
+ * @return Normalised Quaternion.
+ */
+MAVSDK_TEST_EXPORT Quaternion to_quaternion_from_euler_angle(EulerAngle euler_angle);
```
julianoes: I think these should now have the `MAVSDK_PUBLIC` macro rather than `MAVSDK_TEST_EXPORT`.

julianoes (changes requested):
Just the macro, then this is good to go in. Thanks.

julianoes (dismissed):

g-ampo:
@julianoes  Macro swap done in `1ee6f2e9`. Also noticed `operator==(const EulerAngle&, const EulerAngle&)` had no export macro at all and was emitted with hidden visibility (lowercase `t` in `nm`), unreachable from outside `libmavsdk.so`. Added `MAVSDK_PUBLIC` there too in `b4780cf0` :)

g-ampo:
@julianoes `fac48f46` is just the regenerated `cpp/docs/en/cpp/api_reference/namespacemavsdk.md` that the docs CI step diff-checks after the macro change. No source changes. Could you approve the pending workflow runs and re-review once they're green? Sorry for the noise...

PR #2875, opened 2026-04-27
Add telemetry fields from api ref to overview

irowebbn:
The telemetry overview page in the docs only had a partial listing of the type of information available. This adds the rest.


julianoes (approved):
Thanks!

PR #2881, opened 2026-05-16
fix(mavlink_passthrough): export operator<< for Result with MAVSDK_PUBLIC

### LINKED ISSUE #2880: operator<< for MavlinkPassthrough::Result missing MAVSDK_PUBLIC
Title pretty much says it. Hit a linker error trying to stream a MavlinkPassthrough::Result to std::cout. Looks like the operator<< overload in the passthrough header is missing the MAVSDK_PUBLIC specifier that the equivalent MavlinkDirect one has, so the symbol doesn’t get exported.

In mavlink_passthrough.hpp (line 94):
`friend std::ostream& operator<<(std::ostream& str, MavlinkPassthrough::Result const& result);`

In mavlink_direct.hpp (line 126):
`friend MAVSDK_PUBLIC std::ostream& operator<<(std::ostream& str, MavlinkDirect::Result const& result);`

Adding MAVSDK_PUBLIC to the passthrough declaration should fix it I would think.
Building latest main on Rocky 8 using g++ 8.5.0

evanofficial:
The `friend std::ostream& operator<<` for `MavlinkPassthrough::Result` in `mavlink_passthrough.hpp` is missing the `MAVSDK_PUBLIC` visibility specifier, so the symbol is hidden in `libmavsdk.so` and consumers hit a linker error when they try to stream a `Result` (e.g. `std::cout << result;`).

The equivalent `operator<<` for `MavlinkDirect::Result` in `mavlink_direct.hpp` already uses `MAVSDK_PUBLIC` — this PR just mirrors that pattern for consistency.

Fixes #2880.

### DIFF
--- a/cpp/src/mavsdk/plugins/mavlink_passthrough/include/plugins/mavlink_passthrough/mavlink_passthrough.hpp
+++ b/cpp/src/mavsdk/plugins/mavlink_passthrough/include/plugins/mavlink_passthrough/mavlink_passthrough.hpp
@@ -91,7 +91,7 @@ public:
      *
      * @return A reference to the stream.
      */
-    friend std::ostream& operator<<(std::ostream& str, MavlinkPassthrough::Result const& result);
+    friend MAVSDK_PUBLIC std::ostream& operator<<(std::ostream& str, MavlinkPassthrough::Result const& result);
 
     /**
      * @brief Send message (deprecated).


julianoes (approved):

julianoes:
Wondering if we need to backport this one for v3? Can you check @bansiesta ?

PR #2884, opened 2026-05-22
Action: add set_home

### LINKED ISSUE #2799: Add action to be able to set the home
Hi! I would like to add an action to allow setting the current home using the _MAV_CMD_DO_SET_HOME (179)_ command.
See https://mavlink.io/en/messages/common.html#MAV_CMD_DO_SET_HOME

bansiesta:
Implements `set_home` action as requested in #2799.

## What it does

Adds `Action::set_home(use_current_location, latitude_deg, longitude_deg, absolute_altitude_m)` which sends `MAV_CMD_DO_SET_HOME` (179).

- Uses `CommandInt` (not `CommandLong`) for full int32 lat/lon precision, following the `do_orbit`/`goto_location` pattern
- When `use_current_location=true`, only param1=1 is sent (no coordinates needed)
- When `use_current_location=false`, lat/lon are encoded as int32 ×1e7

## Files changed

- **Generated** (from updated proto): `action.hpp`, `action.cpp`, `action_service_impl.hpp`, `action.pb.h/.cc`, `action.grpc.pb.h/.cc`
- **Handwritten**: `action_impl.hpp` (declaration), `action_impl.cpp` (implementation)

## Dependencies

Depends on MAVSDK-Proto PR mavlink/MAVSDK-Proto#397 (submodule updated to that commit).

Closes #2799

### DIFF
--- a/c/src/cmavsdk/plugins/action/action.cpp
+++ b/c/src/cmavsdk/plugins/action/action.cpp
@@ -570,7 +570,7 @@ void mavsdk_action_do_orbit_async(
     mavsdk_action_t action,
     float radius_m,
     float velocity_ms,
-    mavsdk_action_orbit_yaw_behavior_t yaw_behavior,
+    mavsdk_action__t yaw_behavior,
     double latitude_deg,
     double longitude_deg,
     double absolute_altitude_m,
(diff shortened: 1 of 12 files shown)


bansiesta:
Closing in favor of #2886 which is based on current main and includes updated generated files for all plugins.

PR #2885, opened 2026-05-22
Telemetry: use ESTIMATOR_STATUS for is_global_position_ok

### LINKED ISSUE #1852: Telemetry::Health::is_global_position_ok does not gurarantee that global position estimate is ok
## Background
In the PX4's MAVSDK tests, the function `wait_until_ready` relies on the `health_all_ok` function of MAVSDK's telemetry returning true, to determine if the system is ready (sensor calibration, global position, local position estimate, etc).

```c++
void AutopilotTester::wait_until_ready()
{
	std::cout << time_str() << "Waiting for system to be ready" << std::endl;
	CHECK(poll_condition_with_timeout(
	[this]() { return _telemetry->health_all_ok(); }, std::chrono::seconds(30)));

	// FIXME: workaround to prevent race between PX4 switching to Hold mode
	// and us trying to arm and take off. If PX4 is not in Hold mode yet,
	// our arming presumably triggers a failsafe in manual mode.
}
```
https://github.com/PX4/PX4-Autopilot/blob/de1fa11e968ce96a939020ae9194fd2ffe850411/test/mavsdk_tests/autopilot_tester.cpp#L81-L91

And this in turn relies on the `Health` message defined in MAVSDK-Proto, under the Telemetry:

```
    bool is_global_position_ok = 6 [(mavsdk.options.default_value)="false"]; // True if the global position estimate is good enough to fly in 'position control' mode
```
https://github.com/mavlink/MAVSDK-Proto/blob/b9a1c1a51fc3344ee914d0d625f1eafc4189660f/protos/telemetry/telemetry.proto#L604

## Problem
As indicated in the proto file, the `is_global_position_ok` flag should indicate if global position estimate is good enough to fly in position mode. However, this is not the case as currently this flag is set to true simply if we have a 'healthy' GPS sensor:

```
    const bool global_position_ok =
        sys_status_present_enabled_health(sys_status, MAV_SYS_STATUS_SENSOR_GPS);
```
https://github.com/mavlink/MAVSDK/blob/main/src/mavsdk/plugins/telemetry/telemetry_impl.cpp#L1112-L1113

This is a problem since in the commander of PX4, there is a hysteresis that requires at least 3 seconds of valid GPS data to consider 'global position estimate' as valid. As shown here:
image

Therefore, this can lead to CI tests failing simply because the tester script thinks the vehicle is ready for position mode flight as the `health_all_ok` returned true, but in fact the vehicle can't arm since the 3 second hysteresis time hasn't passed and although we have a valid GPS data, our global position estimate is not valid yet.

## Solutions
I think we need to redefine and change the **telemetry_impl** in MAVSDK to actually get the global position validity flag from the Commander, instead of pulling in the healthy flag from the SYS_STATUS message, as that simply gives the status of a 'GPS sensor'.

Any thoughts on this?

bansiesta:
Fixes #1852.

## Problem

`is_global_position_ok` was derived from GPS sensor health in `SYS_STATUS`, which doesn't reflect EKF fusion state. A drone can have GPS signal but the EKF may not yet have a valid global position estimate (e.g. during initialization or after a GPS glitch).

## Fix

Register a handler for `MAVLINK_MSG_ID_ESTIMATOR_STATUS`. When received, use the `ESTIMATOR_PRED_POS_HORIZ_ABS` flag (EKF predicts absolute horizontal position) to set global position health.

Falls back to GPS sensor health from `SYS_STATUS` if no `ESTIMATOR_STATUS` messages are received (older firmware / non-EKF setups).

## Files changed

- `telemetry_impl.hpp`: added `process_estimator_status()` declaration, `_has_estimator_status` flag
- `telemetry_impl.cpp`: register ESTIMATOR_STATUS handler in `init()`, modify `process_sys_status()` to skip global position update when estimator status is available, add `process_estimator_status()` implementation

Closes #1852

### DIFF
--- a/cpp/src/mavsdk/plugins/telemetry/telemetry_impl.cpp
+++ b/cpp/src/mavsdk/plugins/telemetry/telemetry_impl.cpp
@@ -106,6 +106,11 @@ void TelemetryImpl::init()
         this);
 
     _system_impl->register_mavlink_message_handler(
+        MAVLINK_MSG_ID_ESTIMATOR_STATUS,
+        [this](const mavlink_message_t& message) { process_estimator_status(message); },
+        this);
+
+    _system_impl->register_mavlink_message_handler(
         MAVLINK_MSG_ID_HEARTBEAT,
         [this](const mavlink_message_t& message) { process_heartbeat(message); },
         this);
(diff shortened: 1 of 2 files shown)


julianoes:
I disagree with this one, see reasoning in https://github.com/mavlink/MAVSDK/issues/1852#issuecomment-4531437860.

PR #2888, opened 2026-05-29
Time sync API in Python wrappers

TeisybeLT:
C function `mavsdk_system_enable_timesync()` was defined in `mavsdk/system.py`'s, but was not exposed through the `System` class. Exposed this function as `enable_timesync()` and mirrored functionality in `aiomavsdk` to allow time synchronization to be used from Python scripts.


JonasVautherin (approved):
Thank you! 🚀

PR #2890, opened 2026-05-31
Expose MavLink forwarding options in Python wrappers

TeisybeLT:
Python wrappers do not expose options for MavLink forwarding and only contain unused definitions for `ForwardingOption`s. Extended `py/mavsdk` to import C forwarding-related functions and made use of them in `Mavsdk` class.

`aiomavsdk/mavsdk.py` was also modified to achieve feature parity with `py/mavsdk`. Took some inspiration from MavSDK's C++ API and made use of `forwarding_option` parameter with a default value.

### DIFF
--- a/py/aiomavsdk/aiomavsdk/mavsdk.py
+++ b/py/aiomavsdk/aiomavsdk/mavsdk.py
@@ -8,6 +8,7 @@ from mavsdk.mavsdk import Mavsdk as _Mavsdk, Configuration
 from mavsdk.component_type import ComponentType
 from mavsdk.server_component import ServerComponent
 from .system import System
+from .enums import ForwardingOption
 from .exceptions import MavsdkConnectionError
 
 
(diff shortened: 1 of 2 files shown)


JonasVautherin (approved):
Thanks! 🚀

Looks good to me, though I haven't tested it. I assume it works for you?

TeisybeLT:
> Thanks! 🚀
> 
> Looks good to me, though I haven't tested it. I assume it works for you?

For my application, which consists of PX4, QGroundControl and some code running on a companion computer and some custom scripts running on ground station PC, it does indeed work, but I have not tested different permutations of MavLink-capable projects (e.g. Ardupilot and whatnot).

I tried my best to re-use logic present in main C++ library exposed through C wrapper, so it `should work™` :crossed_fingers:.

I suspect having some unit/integration test setup for Python wrappers could be of use to test this properly, but as far as I am aware, automated testing is not even scaffolded for `mavsdk` and `aiomavsdk` so we can't really do that right now. On that note, are unit/integration tests planned for Python wrappers? If time permits, I may be able to implement a basic form of it down the line.

JonasVautherin:
> For my application, [...] it does indeed work

That's great!

> I suspect having some unit/integration test setup for Python wrappers could be of use to test this properly

We don't have any infrastructure for that yet. One thing is that most of the python code is auto-generated, and there isn't a big need to test that part (that's my feeling from the last 8 years of auto-generating the wrappers). The question is for the parts that are manually written (typically what you did here in mavsdk.py), but there again they typically don't change once they are implemented 😅.

PR #2894, opened 2026-06-09
core: fix deleting of callback on rapid requests of messages

gisleandre:
When calling `MavlinkRequestMessage::request()` two times in rapid succession, the second call intermittently ends with a timeout, even though the requested message seems to get delivered. What appears to happen is:
1. Request 1 adds one entry in the `_work_items` list and another in the `_message_handler._table` list
2. On message received, ` MavlinkRequestMessage::handle_any_message()` triggers the user callback
3. `_work_items` entry is removed, and `_message_handler._table` entry is marked as **to-be-removed**
4. Request 2 comes in, and cleanup from previous message is started
5. The old `_message_handler._table` entry is attempted to be removed with calling `_message_handler.unregister_one()`
6. :red_circle: The entry is now **assumed** to be gone, but unregistration was deferred due to handler being busy with other message, and a new entry is added with the same ID, and two identical entries are in `_message_handler._table`
7. On second message received, clean up the handler by removing all entries from `_message_handler._table` which matches `_message_handler._unregister_later_table` (deleting both entries)
8. Look through `_message_handler._table` for callback. Nothing found. Timeout

This change attempts to fix this issue by checking if the requested message is of the same type as an existing entry marked for removal, and if it is, skip the unregister-register step, which is vulnerable to being deleted.

### DIFF
--- a/cpp/src/mavsdk/core/mavlink_request_message.cpp
+++ b/cpp/src/mavsdk/core/mavlink_request_message.cpp
@@ -34,10 +34,16 @@ void MavlinkRequestMessage::request(
     std::unique_lock<std::mutex> lock(_mutex);
 
     // Cleanup previous requests.
-    for (const auto id : _deferred_message_cleanup) {
-        _message_handler.unregister_one(id, this);
+    bool handler_still_registered = false;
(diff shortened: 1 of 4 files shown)


julianoes:
Thanks! I'm just adding some system tests and potentially more broader fixes to - hopefully - fix this for good.

Issue #2896, opened 2026-06-09
Android: camera definition fails to load, Camera.getCurrentSettings() returns UNAVAILABLE


rkganeri:
## Description
I'm using MAVSDK Java / Android with `mavsdk_server` on an Android controller. The vehicle has a MAVLink camera component that advertises a camera definition (`camera_definition-ILX-LR1_Sony-40847.xml`). The camera is discovered correctly, but the MAVSDK Camera plugin never seems to load the camera definition, and `Camera.getCurrentSettings(componentId)` returns `UNAVAILABLE`.

This prevents reading the current camera settings through the Camera plugin. Setting camera params through other paths can still affect the camera, but MAVSDK cannot read back the settings.

After camera discovery, I can see the failure logs:
```
D/Mavsdk  : Component Camera 1 (100) added.
E/Mavsdk  : Cannot open file "/data/data/com.dronebase.af.app.px4
D/Mavsdk  : Cache hit for camera_definition-ILX-LR1_Sony-40847.xml
E/Mavsdk  : Failed to open "/data/data/com.dronebase.af.app.px4
I/Mavsdk  : Using cached file "/data/data/com.dronebase.af.app.px4
E/Mavsdk  : tinyxml2::LoadFile failed: Error=XML_ERROR_FILE_READ_ERROR ErrorID=5 (0x5) Line number=0
E/Mavsdk  : Failed to load camera definition: "/data/data/com.dronebase.af.app.px4
...
io.mavsdk.camera.Camera$CameraException: UNAVAILABLE: Unavailable
	at io.mavsdk.camera.Camera$19.onNext(Camera.java:2006)
	at io.mavsdk.camera.Camera$19.onNext(Camera.java:1999)
	at io.grpc.stub.ClientCalls$StreamObserverToCallListenerAdapter.onMessage(ClientCalls.java:468)
	at io.grpc.internal.ClientCallImpl$ClientStreamListenerImpl$1MessagesAvailable.runInternal(ClientCallImpl.java:667)
	at io.grpc.internal.ClientCallImpl$ClientStreamListenerImpl$1MessagesAvailable.runInContext(ClientCallImpl.java:654)
	at io.grpc.internal.ContextRunnable.run(ContextRunnable.java:37)
	at io.grpc.internal.SerializingExecutor.run(SerializingExecutor.java:133)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1100)
[...]

PR #2899, opened 2026-06-12
core: fix Android cache directory path null byte handling

rkganeri:
## Description
On Android, `get_cache_directory()` reads `/proc/self/cmdline` to build the MAVSDK cache path under `/data/data/<package>/mavsdk_cache`.  `/proc/self/cmdline` is NUL-separated.  The current code attempts to strip out the null byte, but does so by de-referencing `line.end()`, which is one byte past the end and undefined. 

This changes the Android path lookup to read only until the first NUL-delimited entry from `/proc/self/cmdline`, so the cache path is built without the embedded null byte.

## Testing
Tested with `mavsdk_server` on Android and verified camera definitions are cached under `/data/data/<package>/mavsdk_cache/...`, and MAVSDK Camera `getCurrentSettings()` no longer returns `UNAVAILABLE`.

## Addresses
https://github.com/mavlink/MAVSDK/issues/2896

### DIFF
--- a/cpp/src/mavsdk/core/fs_utils.cpp
+++ b/cpp/src/mavsdk/core/fs_utils.cpp
@@ -71,11 +71,8 @@ std::optional<std::filesystem::path> get_cache_directory()
     // Read /proc/self/cmdline
     std::ifstream cmdline("/proc/self/cmdline");
     std::string line;
-    if (std::getline(cmdline, line)) {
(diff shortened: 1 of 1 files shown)


julianoes (approved):
Sounds reasonable, thanks.

julianoes:
Looks like someone needs to fix Windows CI...

PR #2902, opened 2026-06-12
core: apply deferred message handler ops in order

julianoes:
MavlinkMessageHandler deferred register/unregister operations (issued while _mutex was held, e.g. from inside a callback) into two separate tables and always applied all registers before all unregisters. That reordering meant a deferred unregister could delete a handler that was registered after it, since unregister_one removes all entries for a message id and cookie.

Replace the two tables with a single FIFO queue of operations applied in issue order, so an unregister followed by a register of the same message id no longer clobbers the register. Behaviour is unchanged for all existing callers; only the pathological reordering case is fixed. A shared erase_from_table helper removes the duplicated removal logic.

### DIFF
--- a/cpp/src/mavsdk/core/mavlink_message_handler.cpp
+++ b/cpp/src/mavsdk/core/mavlink_message_handler.cpp
@@ -5,6 +5,8 @@
 
 namespace mavsdk {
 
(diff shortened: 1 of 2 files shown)


julianoes:
I think we will soon fix this properly instead.

PR #2903, opened 2026-06-15
core: move hand-rolled thread-safety to the asio executor model

julianoes:
Replace core's hand-rolled thread-safety (a mutex plus deferred side-tables and try_lock-or-defer) with the executor model already used elsewhere: the shared state is only ever touched on the io_context thread. Reads run there directly (inline when already on the io thread); mutations `asio::post()` onto the io_context, so they're serialized and safe to call from inside a callback.

Converted:
- `MavlinkParameterSubscription` — subscription list confined to the io_context.
- `MavlinkMessageHandler` — handler table confined to the io_context.
- `CallbackList` — subscription list confined to the io_context; `empty()` is an atomic. The dedicated mutexes that only guarded a subscription list are removed.

Two things the strand model has to get right:

- **Lifetime.** Owners post onto the io_context during their own destruction (e.g. `~ServerComponentImpl`), so the io_context must outlive them. `_io_context`/`_io_work_guard` are now the first members of `MavsdkImpl`, so it's destroyed last; otherwise those posts hit a freed io_context (use-after-free + a leaked op). A `CallbackList` also drains the io_context in its destructor, since plugins can be destroyed mid-run while posts are still queued.

- **No blocking onto the io thread.** subscribe/unsubscribe/clear and `CallbackList::queue()` post without waiting, so they can't deadlock when called from inside a callback or while holding a lock the io thread also needs (`queue()` is used that way from `subscribe_on_new_system` and `subscribe_component_discovered`).

### DIFF
--- a/cpp/src/mavsdk/core/callback_list.hpp
+++ b/cpp/src/mavsdk/core/callback_list.hpp
@@ -9,13 +9,19 @@
 #include "handle.hpp"
 #include "mavsdk_export.h"
 
+// Forward-declared so this header (the public face of the CallbackList pair) does not pull
+// asio in; the full include lives in callback_list_impl.hpp.
+namespace asio {
+class io_context;
+}
+
 namespace mavsdk {
 
 template<typename... Args> class CallbackListImpl;
 
 template<typename... Args> class MAVSDK_PUBLIC CallbackList {
 public:
-    CallbackList();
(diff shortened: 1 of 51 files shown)


julianoes (commented):
Reviewed what Claude did. I think it might work.

PR #2904, opened 2026-06-17
core: confine MavlinkMessageHandler to the io_context

julianoes:
Replace MavlinkMessageHandler's hand-rolled thread-safety (a mutex plus two deferred side-tables and try_lock-or-defer, applying registers before unregisters) with the executor model the rest of core now uses: the handler table is only ever touched on the io_context thread.

- process_message() already runs on the io thread (it is driven by the connection receive path, which posts onto the io_context), so it iterates the table directly with no lock.
- register/unregister asio::post() their mutation onto the io_context, so they are safe from any thread and from inside a callback (the mutation runs after the current dispatch), with ordering preserved by post FIFO.

Teardown is split by where the owner is destroyed, which the caller always knows by construction, so no run-time "am I on the io thread" detection is needed (asio's running_in_this_thread() is unreliable during destructor cleanup anyway):

- unregister_all_on_io_thread(): direct synchronous erase, for owners that live and die ON the io thread (mission-transfer WorkItems). They are only ever destroyed from do_work(), never mid process_message() iteration, so the direct erase is race-free and avoids a fire-after-free without a round-trip.
- unregister_all_blocking(): posts the removal and waits, for owners destroyed OFF the io thread (plugin destructors on the user thread). Plugins are documented to be destroyed before Mavsdk, so the io thread is alive and the wait completes.

Shutdown is covered by the existing ~MavsdkImpl top-level stop()+join(): the io thread is gone before systems/core objects destruct, so their unregister_all() posts are harmless no-ops and no callback can fire on freed memory.

MavlinkMessageHandler now takes the io_context by reference; wire it through MavsdkImpl and SystemImpl. The mission-transfer unit tests now run the io_context on a dedicated thread (as in production) and route their interactions through an on_io() helper, instead of manual poll()-gating.

Full unit (291) and system (85) suites pass.

### DIFF
--- a/cpp/src/mavsdk/core/mavlink_message_handler.cpp
+++ b/cpp/src/mavsdk/core/mavlink_message_handler.cpp
@@ -1,11 +1,13 @@
 #include <algorithm>
-#include <mutex>
+#include <future>
+#include <utility>
+#include <asio/post.hpp>
 #include "mavlink_message_handler.hpp"
 #include "log.hpp"
 
 namespace mavsdk {
 
-MavlinkMessageHandler::MavlinkMessageHandler()
+MavlinkMessageHandler::MavlinkMessageHandler(asio::io_context& io_context) : _io_context(io_context)
 {
     if (const char* env_p = std::getenv("MAVSDK_MESSAGE_HANDLER_DEBUGGING")) {
         if (std::string(env_p) == "1") {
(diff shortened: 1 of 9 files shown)


julianoes:
Superseding this with a single combined branch: `confine-to-io-context`.

The message-handler change and the parameter-subscription change are the same refactor (confining shared state to the io_context via `asio::post`) and there's no clean point to split them, so they belong in one PR.

That branch also fixes the CI failures seen here. All four red checks (ASan/TSan/UBSan + the fuzzer's LeakSanitizer) had the same root cause: during teardown `~ServerComponentImpl` posts an `unregister_all()` onto the io_context, but `_io_context` was declared after `_server_components` in `MavsdkImpl`, so it was destroyed first — the post then landed on a freed io_context (heap-use-after-free + leaked op). The combined branch declares `_io_context`/`_io_work_guard` as the first members so the io_context is destroyed last. Verified clean under ASan + LeakSanitizer (291 unit tests).

A new PR will be opened from `confine-to-io-context`.

PR #2905, opened 2026-06-19
feat(core): add heartbeat watchdog (deadman timer) (#2895)

### LINKED ISSUE #2895: Add support for disabling native heartbeat sending and triggering HEARTBEAT over gRPC
I would like to contribute a feature that lets MAVSDK users disable internally generated HEARTBEAT sending and instead trigger transmission through gRPC.

The gRPC API would be a fail-closed, one-shot request to send a single MAVSDK-configured HEARTBEAT, not a toggle or persistent enable state. If the client stops calling, disconnects, or fails, MAVSDK does not continue sending heartbeats through that path. The client is responsible for calling at the desired rate (e.g. ~1 Hz).

The gRPC call would not set HEARTBEAT fields; MAVSDK continues to own message content (sysid, compid, mav_type, etc.). The call only requests that MAVSDK send its configured heartbeat.

Some integrations need an external process to control when MAVSDK sends heartbeats, while MAVSDK still constructs the actual MAVLink HEARTBEAT message. This avoids duplicating heartbeat construction outside MAVSDK.

**Proposed Solution**
- Add a configuration option (e.g. automatic_heartbeats_enabled) to disable native heartbeat sending.
- Add a gRPC API to trigger one configured HEARTBEAT per request.
- Optionally expose via mavsdk_server CLI (e.g. --no-automatic-heartbeats).

Default behavior stays unchanged.

tpayne-censystech:
Implements external control of MAVSDK heartbeat emission, per the API added in mavlink/MAVSDK-Proto#415.

When enabled, this disables MAVSDK's internally generated HEARTBEAT sending and drives emission from the client via a runtime-configurable deadman timeout. The client refreshes within the timeout to keep heartbeats flowing; if it crashes, disconnects, or stops servicing the path, MAVSDK stops sending, so PX4 sees the GCS/companion component go offline and triggers the configured lost-link failsafe (e.g. RTL). MAVSDK continues to own heartbeat content (sysid, compid, mav_type, etc.); the client only controls whether emission continues. Default behavior is unchanged.

Includes the proto submodule bump to the merged commit, regenerated mavsdk_server code, and the core implementation with tests.

Depends on mavlink/MAVSDK-Proto#415 (merge first).

Closes mavlink/MAVSDK#2895

Contributed by Tyler Payne on behalf of Censys Technologies, Inc.

### DIFF
--- a/cpp/src/mavsdk/core/include/mavsdk/mavsdk.hpp
+++ b/cpp/src/mavsdk/core/include/mavsdk/mavsdk.hpp
@@ -21,6 +21,17 @@ namespace mavsdk {
 
 class ServerPluginImplBase;
 
+/** @brief Minimum heartbeat watchdog timeout when enabled, in seconds. */
+constexpr double heartbeat_watchdog_min_timeout_s = 1.0;
+
+/**
+ * @brief Check whether a heartbeat watchdog timeout is valid.
+ *
+ * @return true if timeout_s is 0 (disabled), or finite and at least
+ *         heartbeat_watchdog_min_timeout_s.
+ */
+MAVSDK_PUBLIC bool is_valid_heartbeat_watchdog_timeout_s(double timeout_s);
+
 /**
  * @brief ForwardingOption for Connection, used to set message forwarding option.
  */
@@ -241,8 +252,56 @@ public:
 
         /**
          * @brief Set whether to send heartbeats by default.
+         *
+         * Note: when a heartbeat watchdog is configured
+         * (set_heartbeat_watchdog_timeout_s()) and has expired, the watchdog
+         * latch takes precedence: heartbeats stay off until
+         * Mavsdk::feed_heartbeat_watchdog() is called again, even if
+         * always_send_heartbeats is set.
          */
         void set_always_send_heartbeats(bool always_send_heartbeats);
+
+        /**
+         * @brief Get the heartbeat watchdog timeout.
+         * @return Timeout in seconds, 0 if the watchdog is disabled.
+         */
+        double get_heartbeat_watchdog_timeout_s() const;
+
+        /**
+         * @brief Set the heartbeat watchdog (deadman timer) timeout.
+         *
+         * When set to a value greater than 0, the periodic heartbeats sent
+         * by MAVSDK are only sent as long as Mavsdk::feed_heartbeat_watchdog()
+         * keeps being called at least once per timeout period. Heartbeats
+         * never start (and any already-running heartbeats are stopped) until
+         * the watchdog has been fed - including when the watchdog is first
+         * enabled or its timeout is changed. If the watchdog times out,
+         * heartbeats are latched off - including across reconnects and new
+         * system discovery - until the watchdog is fed again.
+         *
+         * Heartbeats are also latched off whenever they stop for any other
+         * reason while the watchdog is configured (e.g. when the connected
+         * system disconnects): after a reconnect, heartbeats only resume
+         * once the watchdog has been fed again.
+         *
+         * This is useful when MAVSDK's heartbeats should reflect the liveness
+         * of the application: if the application hangs or dies, heartbeats
+         * stop.
+         *
+         * While the watchdog is expired, the latch takes precedence over
+         * set_always_send_heartbeats(): heartbeats stay off until the watchdog
+         * is fed again.
+         *
+         * When set to 0, the watchdog is disabled and heartbeats follow the
+         * usual policy (always_send_heartbeats or a connected system).
+         *
+         * Default: 0 (disabled)
+         *
+         * @param timeout_s Timeout in seconds: 0 (disabled) or at least 1.
+         * @return true if the value was accepted, false if it was rejected
+         *         (invalid values are ignored and the previous value kept).
+         */
+        bool set_heartbeat_watchdog_timeout_s(double timeout_s);
 
         /** @brief Component type of this configuration, used for automatic ID set */
         ComponentType get_component_type() const;
@@ -302,6 +361,7 @@ public:
         uint8_t _system_id;
         uint8_t _component_id;
         bool _always_send_heartbeats;
+        double _heartbeat_watchdog_timeout_s{0.0};
         ComponentType _component_type;
         MAV_TYPE _mav_type;
         Autopilot _autopilot{Autopilot::Unknown};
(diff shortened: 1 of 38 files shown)


julianoes:
Thanks @tpayne-censystech, let me check CI, once that passes I'll merge the proto PR, and then update this PR and then merge it.

tpayne-censystech:
@julianoes , I tried to fix the CI but I can't get it to run.

julianoes:
Yes, sorry, I have to click the button to run it.

cpp/src/mavsdk/core/include/mavsdk/mavsdk.hpp
```diff
@@ -21,6 +21,17 @@ namespace mavsdk {
...
+/** @brief Minimum heartbeat watchdog timeout when enabled, in seconds. */
+constexpr double heartbeat_watchdog_min_timeout_s = 1.0;
+
+/**
+ * @brief Check whether a heartbeat watchdog timeout is valid.
+ *
+ * @return true if timeout_s is 0 (disabled), or finite and at least
+ *         heartbeat_watchdog_min_timeout_s.
+ */
+MAVSDK_PUBLIC bool is_valid_heartbeat_watchdog_timeout_s(double timeout_s);
```
julianoes: This is a bit odd. Wouldn't this be inside of the mavsdk class?

cpp/src/mavsdk/core/mavlink_channels.hpp
```diff
@@ -8,13 +8,11 @@ namespace mavsdk {
...
-        // This should be thread-safe in C++11.
-        static MavlinkChannels instance;
-
-        return instance;
-    }
+    // Defined in the .cpp so shared-library builds have one allocator. An
+    // inline Meyers singleton in the header would give the library and each
+    // test binary their own channel map while still sharing
+    // mavlink_get_channel_status(), which races packing on channel 0.
+    static MavlinkChannels& Instance();
```
julianoes: Interesting.

cpp/src/mavsdk_server/src/mavsdk_server_api.h
```diff
@@ -52,6 +52,21 @@ MAVSDK_PUBLIC int mavsdk_server_run_with_mavlink_ids(
     const uint8_t system_id,
     const uint8_t component_id);
 
+/*
+ * Run MavsdkServer with additional options.
+ *
+ * @param heartbeat_watchdog_timeout_s When greater than 0, MAVSDK's periodic
+ *        heartbeats are only sent as long as the FeedHeartbeatWatchdog RPC
+ *        keeps being called at least once per timeout period.
```
julianoes: I think you need to add docstring comments for all params.

cpp/src/mavsdk/core/mavsdk_time.hpp
```diff
@@ -45,8 +46,11 @@ class MAVSDK_TEST_EXPORT FakeTime : public Time {
     void sleep_for(std::chrono::nanoseconds ns) override;
 
 private:
-    std::chrono::time_point<std::chrono::steady_clock, std::chrono::nanoseconds> _current{};
-    void add_overhead();
+    // Atomic so that one thread can advance the fake time (sleep_for) while
+    // other threads (e.g. the io thread polling TimeoutHandler and
+    // CallEveryHandler) concurrently read it via steady_time().
```
julianoes: Ok.

cpp/src/mavsdk/core/mavsdk_impl.hpp
```diff
@@ -141,7 +147,14 @@ class MavsdkImpl {
...
 
-    Time time{};
+private:
+    // Declared before `time` so it is initialized first.
+    Time _own_time{};
+
+public:
+    // References _own_time by default, or the Time passed into the
+    // test-only constructor.
+    Time& time;
```
julianoes: Is this complication required?

cpp/src/mavsdk/core/mavsdk_impl.hpp
```diff
@@ -251,13 +278,38 @@ class MavsdkImpl {
...
     CallbackList<> _new_system_callbacks{_io_context};
 
+    // Serializes configuration writers (set_configuration() and
+    // update_configuration()) so that read-modify-write updates cannot lose
+    // each other's changes. Always the outermost lock: it is never taken by
+    // the io thread, and set_configuration() takes _mutex,
+    // _server_components_mutex and _heartbeat_mutex only individually while
+    // holding it, so it cannot participate in a lock-order cycle. Recursive
+    // because update_configuration() calls set_configuration().
+    std::recursive_mutex _configuration_update_mutex;
```
julianoes: I don't think we should introduce a recursive mutex for this. There has to be a cleaner way.

cpp/src/mavsdk/core/mavsdk_impl.cpp
```diff
@@ -1550,24 +1642,192 @@ void MavsdkImpl::start_sending_heartbeats()
...
+        // restarted right now: the client proved it is alive, so heartbeats
+        // may start again whenever the policy allows it (e.g. on discovery).
+        _heartbeat_watchdog_expired = false;
+    }
+
+    // Only restart heartbeats if they are supposed to be running. If they are
+    // off for another reason (never started, disconnected, policy off), a
+    // feed must not start them. start_sending_heartbeats() re-arms the
+    // watchdog.
+    if (needs_start) {
```
julianoes: I'm trying to understand why this is so complex and requires as much commentary and logic.

Why can't we just have a simple:

- Call every call for sending heartbeats. In there, it checks if the conditions are met and we are actually meant to send out the heartbeat. That way we don't need all those start and stop logic.

julianoes (commented):
Just wondering if we can simplify a few things.

julianoes:
@tpayne-censystech hello :wave:, seen my review?

cpp/src/mavsdk/core/include/mavsdk/mavsdk.hpp
tpayne-censystech: Agreed. Moved onto Mavsdk as static bool is_valid_heartbeat_watchdog_timeout_s(...) (and static constexpr heartbeat_watchdog_min_timeout_s). Still usable without an instance for CLI / gRPC / Configuration.

cpp/src/mavsdk/core/mavlink_channels.hpp
tpayne-censystech: Kept Instance() in the .cpp. Let me know if you don't agree.

cpp/src/mavsdk_server/src/mavsdk_server_api.h
tpayne-censystech: Done. Full param docs (and @return) on mavsdk_server_run_with_mavlink_ids_and_options.

cpp/src/mavsdk/core/mavsdk_time.hpp
tpayne-censystech: Thanks. Left as-is so one thread can advance fake time while the io thread reads it via TimeoutHandler / CallEveryHandler.

cpp/src/mavsdk/core/mavsdk_impl.hpp
tpayne-censystech: Simplified: MavsdkImpl owns one Time (std::unique_ptr<Time>), exposed as Time& time. Production constructs with a normal Time; tests may pass std::make_unique<FakeTime>().

We need FakeTime for the watchdog unit tests. Without it, every timeout/heartbeat period has to wait on the wall clock, which makes those tests slow and flaky. Same idea as the existing TimeoutHandler / CallEveryHandler tests.

cpp/src/mavsdk/core/mavsdk_impl.hpp
tpayne-censystech: Agreed. Dropped it. update_configuration() takes a plain std::mutex and calls set_configuration_locked().

cpp/src/mavsdk/core/mavsdk_impl.cpp
tpayne-censystech: Agreed and refactored that way: one lifetime CallEvery in the constructor; maybe_send_heartbeats() checks policy (always_send or connected system) and watchdog state, then sends or returns. No start/stop of the send cookie on connect/disconnect/config.
Watchdog: enable -> NeedsFeed; feed -> Armed with deadline from now; expiry -> NeedsFeed. A feed is only valid until that deadline.

tpayne-censystech:
Thanks for the review @julianoes. Addressed in the latest commit.

julianoes:
If you don't have any objections, I'm getting this in @tpayne-censystech

tpayne-censystech:
> If you don't have any objections, I'm getting this in @tpayne-censystech

No objections!

PR #2906, opened 2026-06-22
py: redirect all debug output to stderr

bansiesta:
Redirects all diagnostic `print()` output in the Python SDK to `stderr`, so callers can use `stdout` for their own program output without capturing SDK noise.

Changes:
- `cmavsdk_loader.py` — library search/load messages
- `logging.py` — log callback error
- All 27 generated plugin files — callback error messages
- `templates/file.py.j2` — so future codegen picks this up automatically


JonasVautherin (approved):

julianoes:
Don't need this.

PR #2910, opened 2026-06-22
core: change log output from stdout to stderr

bansiesta:
Redirects all MAVSDK log output (`LogDebug`, `LogInfo`, `LogWarn`, `LogErr`) from `stdout` to `stderr`, so users can freely use `stdout` for their own program output without capturing SDK noise.

Changes:
- `log.cpp`: `std::cout` → `std::cerr` throughout (ANSI color codes + log lines); `STD_OUTPUT_HANDLE` → `STD_ERROR_HANDLE` on Windows
- `log_callback.hpp`: update doc comment to say `stderr`

### DIFF
--- a/cpp/src/mavsdk/core/include/mavsdk/log_callback.hpp
+++ b/cpp/src/mavsdk/core/include/mavsdk/log_callback.hpp
@@ -11,7 +11,7 @@ namespace mavsdk::log {
 enum class Level : int { Debug = 0, Info = 1, Warn = 2, Err = 3 };
 
 /** @brief User-defined callback for logging. Returning true from this callback
- * prevents default mavsdk`s logging to stdout. Returning false keeps it.
+ * prevents default mavsdk`s logging to stderr. Returning false keeps it.
  */
 using Callback =
     std::function<bool(Level level, const std::string& message, const std::string& file, int line)>;
--- a/cpp/src/mavsdk/core/log.cpp
+++ b/cpp/src/mavsdk/core/log.cpp
@@ -48,7 +48,7 @@ void log::subscribe(const log::Callback& callback)
 MAVSDK_TEST_EXPORT void set_color(Color color)
 {
 #if defined(WINDOWS)
-    HANDLE handle = GetStdHandle(STD_OUTPUT_HANDLE);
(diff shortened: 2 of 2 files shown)


julianoes (approved):
I think I'd want this. What do you think @JonasVautherin? I think it's ok to change towards v4.

JonasVautherin:
Yeah I guess if people want that. One thing though is that it will then pollute stderr 🙈.

I like how it works in the Java world with SLF4J where I can set the debug level and redirect the streams into different files if I want to, but I don't know of an equivalent in C++.

julianoes:
No one asked for it actually, but I wanted to make a small example where I print some json that then feeds into `jq`. And of course `jq` choked on the debug output.

JonasVautherin:
I wonder if the clean solution would be something like spdlog or if we don't want to add a dependency, our own logging facade that users can implement and do whatever they want with our logs:

```
struct ILogger {
    virtual void info(std::string_view msg) = 0;
    virtual void warn(std::string_view msg) = 0;
    virtual void error(std::string_view msg) = 0;
    ...
};

// Consumer injects their logger
void mavsdk::set_logger(std::shared_ptr<ILogger> logger);
```

So that it's actually flexible: users can choose the log level they want and what they want to do with them (print to stdout, redirect to a file, ...).

And by having a default implementation, I guess we can make it such that it doesn't change the current behaviour?

julianoes:
We already have that:

https://github.com/mavlink/MAVSDK/blob/main/cpp/examples/log_callback/log_callback.cpp

JonasVautherin:
Oh right. And that doesn't solve your issue with your app?

julianoes:
It would solve it indeed. I was just thinking the default to stderr would be "more correct", so we might as well change before v4.

JonasVautherin:
Your call. I personally usually expect errors to show on stderr, not debug or info outputs. Maybe a "more correct" default would be to disable logging entirely by default, as libraries should probably not spontaneously log stuff, and let the users explicitly enable logging if they want it?

julianoes:
I agree that that off by default should have been the default but changing it now seems very confusing to users, I think.

PR #2912, opened 2026-06-26
Join sending_tracking_status thread before destruction

JoC2000:
Issue:
Destructor does not join the live threat created by `start_sending_tracking_status()`, causing std::terminate whenever a `SET_MESSAGE_INTERVAL` starts it.

Fix:
Join the thread before unregistering the plugin


JonasVautherin (approved):

julianoes (approved):
Cheers!

PR #2913, opened 2026-07-01
core: send param value change notification in the correct protocol

bastianhjaeger:
## Problem

`MavlinkParameterServer::provide_server_param()` emits a change notification when an
already-provided parameter's value is updated. It tagged that broadcast with
`_last_extended` — the protocol (extended vs non-extended) of the *last received
request*. So the message type of a value-change notification depended on unrelated
prior traffic: an update could go out as `PARAM_EXT_VALUE` even though the parameter
is served over the classic protocol (or vice-versa), and a client using the other
protocol never sees the change.

## Fix

Choose the broadcast protocol from the parameter itself via `ParamValue::needs_extended()`,
matching how `MavlinkParameterCache` already decides which parameters belong to the
non-extended set:

- representable in the classic protocol → `PARAM_VALUE`
- otherwise → `PARAM_EXT_VALUE`

`_last_extended` is now unused and removed.

## Testing

Verified with a server component that exposes numeric parameters over the classic
protocol. Updating an already-provided value via `provide_param_*`:

- **before:** broadcast as `PARAM_EXT_VALUE`, missed by a non-extended GCS
- **after:** broadcast as `PARAM_VALUE` with the updated value, received correctly (and
  no stray `PARAM_EXT_VALUE`)

Confirmed on the wire for both the initial value and subsequent external changes.

### DIFF
--- a/cpp/src/mavsdk/core/mavlink_parameter_server.cpp
+++ b/cpp/src/mavsdk/core/mavlink_parameter_server.cpp
@@ -99,20 +99,32 @@ MavlinkParameterServer::provide_server_param(const std::string& name, const Para
                 case MavlinkParameterCache::UpdateExistingParamResult::Ok:
                     find_and_call_subscriptions_value_changed(name, param_value);
                     {
-                        auto new_work = std::make_shared<WorkItem>(
-                            name,
-                            param_value,
-                            WorkItemValue{
-                                std::numeric_limits<std::uint16_t>::max(),
-                                std::numeric_limits<std::uint16_t>::max(),
-                                _last_extended});
-                        asio::post(_io_context, [this, new_work]() {
-                            const bool was_empty = _work_queue.empty();
-                            _work_queue.push_back(new_work);
-                            if (was_empty) {
-                                do_work();
-                            }
-                        });
+                        // Notify of the value change in every protocol the parameter is
+                        // exposed over, so subscribers on either protocol see it:
+                        // PARAM_EXT_VALUE always, plus PARAM_VALUE for parameters that are
+                        // also served over the non-extended protocol. Previously this was
+                        // keyed off the last received request, so a client using the other
+                        // protocol could miss the update.
+                        const auto enqueue_value = [this, &name, &param_value](bool extended) {
+                            auto new_work = std::make_shared<WorkItem>(
+                                name,
+                                param_value,
+                                WorkItemValue{
+                                    std::numeric_limits<std::uint16_t>::max(),
+                                    std::numeric_limits<std::uint16_t>::max(),
+                                    extended});
+                            asio::post(_io_context, [this, new_work]() {
+                                const bool was_empty = _work_queue.empty();
+                                _work_queue.push_back(new_work);
+                                if (was_empty) {
+                                    do_work();
+                                }
+                            });
+                        };
+                        enqueue_value(true);
+                        if (!param_value.needs_extended()) {
+                            enqueue_value(false);
+                        }
                     }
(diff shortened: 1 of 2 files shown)


julianoes (commented):
What about sending it out over extended only if there has been traffic/requests on extended before? And sending it on non-extended, again if such traffic/request has been seen before?

I'm not sure sending out extended messages in any case is quite correct.

bastianhjaeger:
Good point — agreed that broadcasting `PARAM_EXT_VALUE` unconditionally isn't right when nobody's using the extended protocol. Let me refine it to track what's actually been seen.

Plan: replace the old single `_last_extended` with two sticky flags, `_seen_extended` / `_seen_non_extended`, set whenever a request of the respective protocol arrives. On a value change,
emit only on protocols that are both representable for that parameter and have been seen:

```
const bool any_seen = _seen_extended || _seen_non_extended;
if (!any_seen) {
    // Nothing seen yet — emit on every protocol the param supports.
    enqueue_value(true);                          // extended: always representable
    if (!param_value.needs_extended()) enqueue_value(false);
} else {
    if (_seen_extended)                                     enqueue_value(true);
    if (_seen_non_extended && !param_value.needs_extended()) enqueue_value(false);
}
```

A couple of consequences I want to flag so we agree on the intended behavior:

1. Needs-extended parameter, only non-extended traffic seen → no notification. I think that's correct: there's no extended client to receive it, and such a parameter isn't visible over the classic protocol anyway (the cache keeps it out of the non-extended set). So nothing is lost. Shout if you'd rather it still go out.
2. Before any request is seen, I fall back to emitting on every representable protocol rather than nothing, to avoid a regression vs. the old always-notify behavior. In practice there's usually no client connected yet and a late-joining GCS reads current values via `REQUEST_(EXT_)LIST` on connect, so this window rarely matters — but it keeps the "a change always
notifies whoever could be listening" guarantee.
3. Passive listeners (a component that only consumes value broadcasts but never sends a request) won't flip a "seen" flag, so strict gating wouldn't reach them. The pre-traffic fallback covers them only until some active client sends a request. I don't think that's worth complicating the logic for — real extended clients send `PARAM_EXT_REQUEST_LIST` — but let me know if you'd weight it differently.

Does this match what you had in mind?

julianoes:
Might as well Claude Code it myself.

bastianhjaeger:
Sry. Why close it?

bastianhjaeger:
I might got Claude to help, but the feature is still a valid one to request.

julianoes:
I appreciate the contribution but I don't feel like arguing with AI via a pull request is very efficient, so at that point I can have the conversation locally and get it done much quicker.

bastianhjaeger:
I do not want to say that I didn't use AI to get into MAVSDK again and get points ajd pointers. Nut the text was typed and I really thought putting more words to it would help.
You are more into the code base, but my last touchings point was long ago.

If you want to keep this closed and tackle it on your own, I will have to accept, but I really want a fix. If you want, we can also hop on a call tomorrow or the day after. Not sure what might be our best time with Germany and New Zealand.

julianoes:
I've pushed this one instead #2915. I appreciate your help and sorry I was a bit grumpy, but seeing a wall of AI text makes me feel like the reading and thinking is left to me, while you just had to hit copy paste.

bastianhjaeger:
Thanks for the "sorry" section. I get your point of AI test overflow, but I really put time into locating and pushing this.

julianoes:
Appreciate your fix and helping here!

PR #2915, opened 2026-07-06
core: notify param value changes on every used protocol

julianoes:
Alternative fix to https://github.com/mavlink/MAVSDK/pull/2913.

If you don't mind, it would be very helpful if you could test this @bastianhjaeger.


bastianhjaeger:
LGTM. Tested in simulation and hardware. 

When do you estimate a release with this and #2916 included?

julianoes:
Awesome, thanks for testing. v4 is still a few weeks out. There are a few pieces I need to figure out before it goes out. Otherwise, we'd have to backport it to v3, but probably easier to just charge forward and get v4 out.

PR #2917, opened 2026-07-12
core: deprecate old MAVLink interception API and MavlinkPassthrough for v4

bansiesta:
Marks the old MAVLink APIs as deprecated ahead of v4:

- `Mavsdk::intercept_incoming_messages_async()` — use `subscribe_incoming_messages_json()` instead
- `Mavsdk::intercept_outgoing_messages_async()` — use `subscribe_outgoing_messages_json()` instead
- `MavlinkPassthrough` class — use `MavlinkDirect` instead

The JSON-based replacements were added in #2638, which stated the old API would be replaced "by the next major release". MavlinkPassthrough already had a docstring warning; this adds the machine-readable `[[deprecated]]` attribute so compilers emit warnings at call sites.

### DIFF
--- a/cpp/src/mavsdk/plugins/mavlink_passthrough/include/plugins/mavlink_passthrough/mavlink_passthrough.hpp
+++ b/cpp/src/mavsdk/plugins/mavlink_passthrough/include/plugins/mavlink_passthrough/mavlink_passthrough.hpp
@@ -31,7 +31,7 @@ class MavlinkPassthroughImpl;
  *             - Custom message support via XML loading
  *             - Better language wrapper integration
  */
-class DEPRECATED MAVSDK_PUBLIC MavlinkPassthrough : public PluginBase {
+class MAVSDK_PUBLIC MavlinkPassthrough : public PluginBase {
 public:
     /**
      * @brief Constructor. Creates the plugin for a specific System.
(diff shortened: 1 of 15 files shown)


cpp/src/mavsdk/plugins/mavlink_passthrough/include/plugins/mavlink_passthrough/mavlink_passthrough.hpp
```diff
@@ -24,15 +24,14 @@ class MavlinkPassthroughImpl;
...
- *          - Better language wrapper integration
+ * @deprecated MavlinkPassthrough is deprecated as of MAVSDK v4.
+ *             Use MavlinkDirect instead.
+ *             MavlinkDirect provides enhanced functionality including:
+ *             - Runtime message parsing with JSON field representation
+ *             - Custom message support via XML loading
+ *             - Better language wrapper integration
  */
-class MAVSDK_PUBLIC MavlinkPassthrough : public PluginBase {
+class [[deprecated("Use MavlinkDirect instead")]] MAVSDK_PUBLIC MavlinkPassthrough : public PluginBase {
```
julianoes: Can we use the `DEPRECATED` define?

bansiesta: Done, switched to `DEPRECATED`.

julianoes (dismissed):

julianoes (approved):

PR #2918, opened 2026-07-12
core: switch JSON library from jsoncpp to nlohmann/json

julianoes:
## What

Replace the compiled **jsoncpp** dependency with the header-only **nlohmann/json** (v3.11.3, already present transitively via `libevents`).

## Why

- One header-only JSON library instead of a compiled `ExternalProject` to build and link.
- Unifies the two directions: previously jsoncpp was used only for parsing, while serialization was hand-rolled `std::ostringstream` concatenation.
- Fixes a latent bug: the old MavlinkDirect serializer did **not escape strings**. nlohmann's `dump()` escapes correctly and renders `NaN`/`Inf` as `null`.

## How

- **Exception safety** (`-fno-exceptions`): parse via the non-throwing `json::parse(..., allow_exceptions=false)` + `is_discarded()`, with predicate-guarded access so nlohmann never hits its `abort()` path.
- **Serialize** (`libmav_receiver`): reworked onto `nlohmann::ordered_json` + `dump()`, preserving field order (plain `json` would sort keys).
- Migrated `libmav_conversions`, `mission_import`, `mavlink_component_metadata`, `component_metadata_server_impl`, and the system tests.
- Updated build wiring (new `third_party/nlohmann_json`, removed `jsoncpp` + `Findjsoncpp.cmake`), packaging (`.pc`, CMake config, `debian/control` — header-only, so runtime dep dropped), CI, docs, and the example.

## Testing

- `build-debug` and `build-server-debug` both build cleanly.
- Unit tests: 291 passed (includes mission import `.plan` fixtures + component metadata).
- System tests: MavlinkDirect / ComponentMetadata / Intercept / ParamChange / RequestMessage all pass (round-trip serialize↔parse).
- mavsdk_server unit tests: 338 passed.


cpp/third_party/nlohmann_json/CMakeLists.txt
```diff
@@ -22,15 +22,15 @@ if(IOS)
...
 
-ExternalProject_Add(
-    jsoncpp
-    GIT_REPOSITORY https://github.com/open-source-parsers/jsoncpp
-    GIT_TAG 1.9.6
-    PREFIX jsoncpp
+ExternalProject_add(
+    nlohmann_json
+    GIT_REPOSITORY https://github.com/nlohmann/json
+    GIT_TAG v3.12.0
```
JonasVautherin: Nitpick: when possible, I prefer not using GIT here, it's slightly more efficient:

```
URL https://github.com/nlohmann/json/archive/refs/tags/v3.12.0.tar.gz
```

JonasVautherin (dismissed):
Just a small nitpick, feel free to ignore 🙈

cpp/third_party/nlohmann_json/CMakeLists.txt
julianoes: Ah yes, agreed!

Issue #259, opened 2018-02-09
Add support to include Doxygen tags \code and \endcode


shakthi-prashanth-m:
To improve including code snippets in docs, it is most suitable to use Doxygen tags: [\code](http://www.stack.nl/~dimitri/doxygen/manual/commands.html#cmdcode) & [\endcode](http://www.stack.nl/~dimitri/doxygen/manual/commands.html#cmdendcode).

Added PR which use these Doxygen tags: ([dronecore/DroneCore/#258](https://github.com/dronecore/DroneCore/pull/258)).

```
$ ./generate_docs.sh
......
Generating: /home/shakthi/karma/fork/DroneCore/install/docs/xml/classdronecore_1_1_follow_me.xml (class xml)
Unhanded tag in markdown_any_tag: programlisting
Traceback (most recent call last):
  File "/home/shakthi/karma/fork/DroneCore/generate_markdown_from_doxygen_xml.py", line 1074, in <module>
    markdown_string=currentClass.markdown()
  File "/home/shakthi/karma/fork/DroneCore/generate_markdown_from_doxygen_xml.py", line 957, in markdown
    outputstring+=the_function.markdown()
  File "/home/shakthi/karma/fork/DroneCore/generate_markdown_from_doxygen_xml.py", line 683, in markdown
    functionstring+='\n\n%s' % markdown_any_tag(self.detaileddescription).strip()
  File "/home/shakthi/karma/fork/DroneCore/generate_markdown_from_doxygen_xml.py", line 117, in markdown_any_tag
    child_text += markdown_any_tag(child,html=html,para=para) #.strip()
  File "/home/shakthi/karma/fork/DroneCore/generate_markdown_from_doxygen_xml.py", line 122, in markdown_any_tag
    child_text+=str(ET.tostring(child),'utf-8')
TypeError: str() takes at most 1 argument (2 given)
```

hamishwillee:
Implemented/discussed in #258

PR #2922, opened 2026-07-13
docs: fix issues from MAVSDK-docs tracker

bansiesta:
Fixes open issues from https://github.com/mavlink/MAVSDK-docs/issues applied to the in-tree docs, plus new Camera and Gimbal guide pages.

## Test plan
- [x] Review commits
- [ ] Build docs locally: `cd docs && npm install && npm run dev`

### DIFF
--- a/docs/en/cpp/contributing/plugins.md
+++ b/docs/en/cpp/contributing/plugins.md
@@ -182,7 +182,7 @@ Once the proto file has been created, you can generate all files required for th
 
 1. Run the configure step to prepare the tools required:
    ```
-   cmake -DBUILD_MAVSDK_SERVER=ON -Bbuild/default -S.
+   cmake -DBUILD_MAVSDK_SERVER=ON -B build/default -S.
    ```
 1. Install `protoc_gen_mavsdk` which is required for the auto-generation:
    ```
--- a/docs/en/cpp/contributing/plugins.md
+++ b/docs/en/cpp/contributing/plugins.md
@@ -182,7 +182,7 @@ Once the proto file has been created, you can generate all files required for th
 
 1. Run the configure step to prepare the tools required:
    ```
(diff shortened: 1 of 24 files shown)


docs/en/cpp/contributing/plugins.md
```diff
@@ -182,7 +182,7 @@ Once the proto file has been created, you can generate all files required for th
 
 1. Run the configure step to prepare the tools required:
    ```
-   cmake -DBUILD_MAVSDK_SERVER=ON -Bbuild/default -S.
```
julianoes, line 185: Should we only do this for `-B` or also for all other args? Can you check common practice?

julianoes (commented):
Good catches!

docs/en/cpp/contributing/plugins.md
bansiesta: CMake docs do show `-D <var>=<value>` with a space (e.g. in the `-P` script mode synopsis), so it is valid. But `-DVAR=val` (no space) is deeply entrenched CMake convention — virtually universal in practice across the ecosystem. Changing it would make the docs look non-standard without a clarity benefit.

The original issue (#259) was specifically about `-Bbuild` where the no-space form genuinely looks like a typo and `-B <path>` is the clearly preferred modern style (added in CMake 3.13). The `-D` situation is different: there's no ambiguity, it's just convention.

Recommendation: leave `-DVAR=val` as-is, only `-B` and `-S` get spaces.

julianoes (dismissed):

julianoes (approved):

hamishwillee:
Sorry, just saw requested review. If you (really) want a specific review in MAVSDK I am happy to give it, but you are best off pinging me directly because I'm likely to miss the request and not make it a priority.

julianoes:
No worries, these are a few obvious fixes/improvements and gimbal/camera as you requested many years ago, very brief, but better something than nothing.

PR #2925, opened 2026-07-14
core: fix buffer overflows, deadlocks, and races in core

julianoes:
## Summary

An audit of `src/mavsdk/core/` for buffer overflows, deadlocks, and other memory-safety / concurrency bugs. Each fix is its own commit. Build is clean and all unit tests (291) and system tests (95) pass.

## Memory safety (remotely reachable)

- **FTP `_data_as_string` OOB read** — scanned the fixed 239-byte `data[]` ignoring `payload.size`, and `max_data_length - start` (uint8_t − size_t) underflowed; a `CMD_RENAME` with a NUL-free field read far past the buffer. Now bounded by `payload.size`.
- **`PARAM_EXT_SET` custom value OOB read** — used `strlen()` on the non-terminated `char[128]`; switched to `strnlen()`.
- **Metadata uri `std::stoi` → `terminate()`** — a `COMPONENT_METADATA` uri with `comp=` > INT_MAX aborted under `-fno-exceptions`; parse with `strtol` + range check.
- **`fopen` NULL deref** in `download_file_to_path` — unwritable path caused a NULL `FILE*` deref in libcurl's writer / `fclose`.
- **FTP burst offset huge allocation** — a crafted burst `offset` triggered a ~4 GB zero-fill vector; reject offsets past the file size.
- **`getpwuid()` NULL deref** in `get_cache_directory` when `HOME` is unset and the uid has no passwd entry.
- **`inet_ntop` result unchecked** in `resolve_hostname_to_ip` (could read uninitialized buffer / return empty as success).

## Concurrency (races, deadlocks, use-after-free)

- **`MavlinkRequestMessage` UAF** — the destructor didn't cancel pending `TimeoutHandler` cookies, so a timeout could fire on freed memory after teardown.
- **FTP burst-thread deadlock** — `_reset()` / `_work_burst` joined the burst thread while holding `_mutex`, which the worker needs; made `burst_stop` atomic and the worker `try_lock`.
- **Param subscription callbacks under `_all_params_mutex`** — a callback re-entering a server API would self-deadlock; post the notification off the lock (matching `provide_server_param`).
- **`_configuration` inconsistent locking** — written under one mutex, read under another or none; now guarded by a dedicated leaf mutex.
- **`_system_ids` unlocked insert** on the libmav receive path (raced the locked send-path reader).
- **UDP `send_to` vs socket close** race during teardown — close now takes `_remote_mutex`.
- **Callback watchdog captured a loop-local by reference** (debug-only UAF) — capture by value.
- **FTP target ids** made atomic (written on the message thread, read on the burst thread).
- **Unsynchronized statustext / param-changed handler lists** — guarded with a mutex, snapshot-under-lock so callbacks never run while holding it.

## Correctness

- **`unregister_statustext_handler`** used the single-iterator `erase(remove_if(...))` idiom → UB when the cookie isn't present.
- **`MavlinkRequestMessage::handle_any_message`** used `&&` where it needed `||`, so an unrelated message could complete the wrong request.
- **`PARAM_ERROR` never sent** for missing/wrong-type non-extended `PARAM_SET` (`return` instead of `break`).
- **Non-extended parameter view** kept full-set indices, so `param_by_index` / broadcasts were inconsistent with `count()`; reindex contiguously.
- **Mission server accepted duplicate/out-of-order items** unbounded; guard on `seq == _next_sequence` like the client.
- **`_work_write`** checked `ifstream.fail()` after `ofstream.seekp()`.
- **`cli_arg`** accepted trailing garbage in port/baudrate (`from_chars` didn't require full consumption).

## Not addressed (intentionally)

Two lower-confidence, design-level items were left out to avoid a destabilizing change:
- `on_io_thread()` relying on asio's `running_in_this_thread()` across the DSO boundary (possible self-deadlock in `CallbackList::exec`/`drain`).
- `stop()` on the connection classes self-deadlocking if ever driven from the io thread.

## Testing

- `cmake --build build-debug` — clean
- `unit_tests_runner` — 291 passed
- `system_tests_runner` — 95 passed

### DIFF
--- a/cpp/src/mavsdk/core/cli_arg.cpp
+++ b/cpp/src/mavsdk/core/cli_arg.cpp
@@ -230,7 +230,9 @@ std::optional<int> CliArg::port_from_str(std::string_view str)
     int value;
     auto [ptr, ec] = std::from_chars(str.data(), str.data() + str.size(), value);
 
-    if (static_cast<bool>(ec) || value < 1 || value > 65535) {
+    // from_chars stops at the first non-digit and reports success, so also require
+    // that the whole string was consumed to reject inputs like "8080abc".
+    if (static_cast<bool>(ec) || ptr != str.data() + str.size() || value < 1 || value > 65535) {
         return {};
     }
     return {value};
@@ -284,7 +286,8 @@ bool CliArg::parse_serial(const std::string_view rest, bool flow_control_enabled
     auto [ptr, ec] =
         std::from_chars(baudrate_str.data(), baudrate_str.data() + baudrate_str.size(), value);
 
-    if (static_cast<bool>(ec)) {
+    // Reject trailing garbage (e.g. "57600xyz"); from_chars would otherwise accept it.
+    if (static_cast<bool>(ec) || ptr != baudrate_str.data() + baudrate_str.size()) {
         return {};
     }
 
--- a/cpp/src/mavsdk/core/cli_arg_test.cpp
+++ b/cpp/src/mavsdk/core/cli_arg_test.cpp
@@ -400,3 +400,29 @@ TEST(CliArg, RawConnectionWrong)
     EXPECT_FALSE(ca.parse("raw://localhost"));
     EXPECT_FALSE(ca.parse("raw://127.0.0.1"));
 }
+
+TEST(CliArg, PortTrailingGarbageRejected)
+{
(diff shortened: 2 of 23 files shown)


julianoes (commented):
Reviewed, looks good.

PR #2937, opened 2026-07-16
docs: fix unbalanced parenthesis in cruise speed section

Bartok9:
The **Get/Set Cruise Speed** section in the taking off / landing guide opens a parenthetical `(and other [AUTO Flight Modes](...)` that is never closed, so the sentence reads with a dangling open paren.

This adds the missing closing parenthesis right after the link:

```
... *Takeoff* (and other [AUTO Flight Modes](.../px4_basic_concepts#flight-modes)) using the following methods:
```

Docs-only, no code changes. Target link verified reachable (HTTP 200).


hamishwillee (approved):
Thanks!

PR #2939, opened 2026-07-16
docs: remove duplicated word in system information guide

Bartok9:
Typo fix in the system information guide: 'This was a a change inherited from mavlink' -> 'This was a change inherited from mavlink'. Docs-only.


hamishwillee (approved):
Thanks! 

Just FWIW - if you are doing a general search for problems like this then best to batch docs fixes. If however you just saw this in passing and decided to fix it, then great!

Bartok9:
Thanks for merging, and noted! These were spotted in passing rather than a bulk sweep, but I'll batch related docs fixes into a single PR next time to keep your review queue tidy. Appreciate the guidance.

PR #2941, opened 2026-07-17
docs: modernize PX4 tree name and QGC install link in examples index

Bartok9:
## Summary
In the C++ examples index, clarify that SITL commands run from the **PX4-Autopilot** tree (not the legacy “Firmware” moniker alone) and point the QGroundControl install link at the current user-guide path (aligned with `docs/en/cpp/quickstart.md`).

## Motivation
Readers cloning PX4 today land in `PX4-Autopilot`; the master-prefixed QGC docs URL is unnecessary when a stable `/en/` path already works.

## Verification
- Docs-only
- QGC URL returns HTTP 200

AI-assisted; human-reviewed.

### DIFF
--- a/docs/en/cpp/examples/index.md
+++ b/docs/en/cpp/examples/index.md
@@ -62,7 +62,7 @@ After running a standard installation, a simulation can be started from the PX4-
 You can use *QGroundControl* to connect to PX4 and observe vehicle movement and behaviour while the examples are running.
 *QGroundControl* will automatically connect to the PX4 simulation as soon as it is started.
 
-See [QGroundControl > Download and Install](https://docs.qgroundcontrol.com/en/getting_started/download_and_install.html) for information about setting up *QGroundControl* on your platform.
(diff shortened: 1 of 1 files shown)


docs/en/cpp/examples/index.md
```diff
@@ -62,7 +62,7 @@ After running a standard installation, a simulation can be started from the PX4
 You can use *QGroundControl* to connect to PX4 and observe vehicle movement and behaviour while the examples are running.
 *QGroundControl* will automatically connect to the PX4 simulation as soon as it is started.
 
-See [QGroundControl > Download and Install](https://docs.qgroundcontrol.com/master/en/qgc-user-guide/getting_started/download_and_install.html) for information about setting up *QGroundControl* on your platform.
+See [QGroundControl > Download and Install](https://docs.qgroundcontrol.com/en/getting_started/download_and_install.html) for information about setting up *QGroundControl* on your platform.
```
julianoes: While this reads nicer, the previous link is actually what is resolved eventually.

Bartok9:
Good catch, thanks @julianoes — reverted the QGC link back to the original `master/en/qgc-user-guide/...` URL since that's the one that resolves canonically. Kept only the Firmware → PX4-Autopilot directory-name modernization. 🙏

julianoes (approved):

PR #2945, opened 2026-07-17
examples: fix sniffer header comment grammar

Bartok9:
## Summary
Header comment: “Example how to…” → “Example of how to…” for the sniffer example.

## Verification
Comment-only.

AI-assisted; human-reviewed.


JonasVautherin:
This kind of behaviour is the reason why many repos explicitly refuse AI-assisted "contributions". This is just spam.

PR #2958, opened 2026-07-18
docs(geometry): point geo inspiration link at PX4-Autopilot

Bartok9:
## Summary
- `geometry.hpp` credited `PX4/ecl` `geo/geo.cpp`.
- Retarget to `PX4-Autopilot` `src/lib/geo/geo.cpp` on main.

## Motivation
ecl was absorbed into PX4-Autopilot; the old tree link is a confusing dead-end for implementors.

## Verification
- New blob URL 200
- Comment-only in public geometry header


Bartok9:
Pushed docs markdown sync so  matches `geometry.hpp` (PX4-Autopilot geo.cpp). Not a proto-generated header.

Bartok9:

PR #2959, opened 2026-07-18
docs(core): update px4_custom_mode.h source attribution URL

Bartok9:
## Summary
- `px4_custom_mode.hpp` still cited `github.com/PX4/Firmware/...`.
- Point the “copied from” note at `PX4-Autopilot` main commander header.

## Motivation
PX4 renamed the primary tree years ago; attribution should match today’s sources.

## Verification
- New URL 200
- Comment-only; no enum/value changes


Bartok9:

PR #2962, opened 2026-07-19
docs: correct internal plugin log.hpp path in dev logging guide

Bartok9:
## Summary
Update the plugin/test logging guide to the real `log.hpp` location (the old `src/core/log.h` blob 404s).

## Verification
`cpp/src/mavsdk/core/log.hpp` exists on main; `log.h` does not.

Docs-only; internal logging API note only.


Bartok9:

PR #2970, opened 2026-07-19
docs(examples): align offboard velocity source link label with filename

Bartok9:
## Summary
The offboard velocity guide labeled the example source `offboard_velocity.cpp` while the tree file is `offboard.cpp`. Match label to the repository filename (href was already correct).


Bartok9:

PR #2973, opened 2026-07-20
test(core): extend base64 encode/decode edge coverage

Bartok9:
## Why
`base64_test` only covered a single full 0–255 round-trip. Metadata pipelines and FTP-ish callers often hit empty buffers, short remnants (padding), and invalid alphabet input.

## What
Adds unit tests for:
- empty encode → empty string; empty decode → empty vector
- 1- and 2-byte payloads (padding `=` / `==`)
- known short alphabet vectors (`Man` / `Ma` / `M`)
- invalid characters leave decoder with empty/partial-safe behavior (no crash)

No production code change.

## Verification
- Added cases in `cpp/src/mavsdk/core/base64_test.cpp`
- Intent: build `unit_tests` target / CI Build and Test

AI-assisted; human-reviewed.

### DIFF
--- a/cpp/src/mavsdk/core/base64_test.cpp
+++ b/cpp/src/mavsdk/core/base64_test.cpp
@@ -34,3 +34,40 @@ TEST(Base64, Basic)
 
     EXPECT_EQ(vec, vec_back);
 }
+
+TEST(Base64, Empty)
+{
+    std::vector<uint8_t> empty{};
+    EXPECT_EQ(base64_encode(empty), "");
+    EXPECT_TRUE(base64_decode("").empty());
+}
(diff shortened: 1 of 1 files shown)


JonasVautherin:
@Bartok9 which model are you running?

Bartok9:
Hi @JonasVautherin — for scoping/review we run **Grok 4.5** (via OpenRouter / xAI). Implementation lands through **Cursor Composer 2.5 Standard**. Happy to adjust anything on this PR if a human pass would prefer different edges or naming.

Thanks for taking a look!

julianoes (approved):

PR #2974, opened 2026-07-20
test(core): add Crc32 unit coverage for MAVLink polynomial

Bartok9:
## Summary
`Crc32` (Gary Brown / MAVLink flavor: init 0, no final XOR) had **no** dedicated unit tests despite shipping in core.

## Changes
- New `crc32_test.cpp`: empty, known vector `123456789` → `0x2dfd2d88`, incremental vs one-shot, single `0xFF`.
- Register source in core `UNIT_TEST_SOURCES`.

No production logic change.

## Verification
Vectors cross-checked against an independent Python replay of the same table polynomial.

### DIFF
--- a/cpp/src/mavsdk/core/CMakeLists.txt
+++ b/cpp/src/mavsdk/core/CMakeLists.txt
@@ -283,6 +283,7 @@ list(APPEND UNIT_TEST_SOURCES
     ${PROJECT_SOURCE_DIR}/mavsdk/core/mavlink_parameter_cache_test.cpp
     ${PROJECT_SOURCE_DIR}/mavsdk/core/string_utils_test.cpp
     ${PROJECT_SOURCE_DIR}/mavsdk/core/base64_test.cpp
+    ${PROJECT_SOURCE_DIR}/mavsdk/core/crc32_test.cpp
 )
 
 if (NOT BUILD_WITHOUT_CURL)
--- a/cpp/src/mavsdk/core/crc32.hpp
+++ b/cpp/src/mavsdk/core/crc32.hpp
@@ -1,13 +1,14 @@
 #pragma once
 
 #include <cstdint>
+#include "mavsdk_export.h"
 
 namespace mavsdk {
 
 // For more information about the CRC algorithm used, check the comment in the
 // source file.
 
-class Crc32 {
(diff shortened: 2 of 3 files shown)


julianoes (commented):
Doesn't the mavlink library itself have tests for it already?

Bartok9:
Good question! This targets MAVSDK's own `mavsdk::Crc32` class in `core/crc32.{hpp,cpp}` — a small independent implementation of the MAVLink X.25 CRC (poly 0x1021), not the upstream mavlink C library's checksum. That header currently has no direct unit coverage in this repo, so a regression in our own copy (e.g. a bad accumulate or init value) wouldn't be caught here. The test pins the known-answer results for the MAVLink polynomial plus incremental-vs-oneshot equivalence and the null/zero-length edge case. Happy to drop it if you'd rather rely solely on the mavlink lib's tests, but the value is guarding *this* class against local drift.

Bartok9:
Follow-up with a correction and a clearer answer, @julianoes — sorry for the imprecision in my earlier note.

**Short answer:** the mavlink C library does *not* test this code (and does not even implement the same CRC).

| | mavlink `c_library_v2` `checksum.h` | MAVSDK `mavsdk::Crc32` |
|---|---|---|
| Width | **CRC-16** (MCRF4XX / X.25-style for *message* checksums) | **CRC-32** (Gary S. Brown table, poly effectively `0xEDB88320`) |
| Init | `0xFFFF` | `0` |
| Final XOR | (X.25 validate path) | none |
| Used for | Framing every MAVLink packet | MAVLink **FTP** (and friends) file checksums in MAVSDK (`mavlink_ftp_client/server`, component metadata) |

Upstream mavlink ships the CRC-16 helpers inline in headers; I could not find dedicated unit tests of that either, but more importantly **it is a different algorithm entirely**. What this PR covers is MAVSDK’s own local CRC-32 copy that FTP relies on — regressions here would not be caught by anything in the mavlink submodule.

Happy to close this out if you’d rather not land standalone CRC unit tests, or to slim the suite to just the known-answer vector. Entirely your call — thanks for checking.

julianoes:
Oh, ok, so this is for the FTP implementation. Ok thanks.

julianoes (approved):

PR #2976, opened 2026-07-20
test(core): cover NaN-aware Quaternion and EulerAngle equality

Bartok9:
## Summary
`operator==` for `Quaternion` / `EulerAngle` is documented as **NaN-aware** (two NaN components compare equal) but had no tests.

## Changes
Adds gtests for equal finite, equal NaN components, and mismatch cases.

## Why it matters
Telemetry often surfaces NaN for unknown fields; equality helpers are used in comparisons/tests and must stay intentional.


julianoes (approved):
Comparing nan is usually a bit of a foot gun, and nan as one of the components doesn't make much sense, but sure.

PR #3000, opened 2026-07-22
test(core): cover remaining Ardu flight modes and VTOL dispatch

Bartok9:
## Summary

- Fill ArduCopter/Plane/Rover mode gaps from earlier mapping tests.\n- Cover VTOL-as-plane, boat-as-rover, Unknown→PX4 fallback, and unknown PX4 submodes.

## Motivation

Adds focused unit coverage for pure helpers steading regression detection without production behavior change.

## Verification

- `flight_mode_extra_test.cpp` registered in core CMake unit test list
- Offline review of assertions against existing APIs
- No flight/arm/geofence paths touched

---

**Agent-Owner:** `sera` · **Platform:** hermes · **Claim-TTL:** 24h

### DIFF
--- a/cpp/src/mavsdk/core/flight_mode_test.cpp
+++ b/cpp/src/mavsdk/core/flight_mode_test.cpp
@@ -154,3 +154,124 @@ TEST(FlightMode, CustomModeDispatch)
             static_cast<uint32_t>(ardupilot::RoverMode::Hold)),
         FlightMode::Hold);
 }
+
+TEST(FlightMode, ArduCopterExtraModes)
+{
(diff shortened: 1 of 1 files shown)


Bartok9:
Preemptively addressing the same style note as on #3002 / #2995:

- Removed `flight_mode_extra_test.cpp` and its CMake registration.
- Folded the new cases into the existing `flight_mode_test.cpp`.

No production code change — tests only. Happy to adjust further if you'd rather drop any redundant case.

julianoes (approved):

PR #3003, opened 2026-07-22
test(core): cover geometry mid-latitude east scale and south

Bartok9:
## Summary

- East scale at 60° latitude (cos factor), west round-trip, and pure-south offset.

## Motivation

Adds focused unit coverage for pure helpers steading regression detection without production behavior change.

## Verification

- `geometry_extra_test.cpp` registered in core CMake unit test list
- Offline review of assertions against existing APIs
- No flight/arm/geofence paths touched

---

**Agent-Owner:** `sera` · **Platform:** hermes · **Claim-TTL:** 24h

PR #3002, opened 2026-07-22
test(core): cover math constrain and deg/rad identity

Bartok9:
## Summary

- Cover `constrain` inside/outside bounds and deg↔rad identity helpers used by Euler conversion.

## Motivation

Adds focused unit coverage for pure helpers steading regression detection without production behavior change.

## Verification

- `math_utils_extra_test.cpp` registered in core CMake unit test list
- Offline review of assertions against existing APIs
- No flight/arm/geofence paths touched

---

**Agent-Owner:** `sera` · **Platform:** hermes · **Claim-TTL:** 24h

PR #2999, opened 2026-07-22
test(core): cover Time shift FakeTime and AutopilotTime

Bartok9:
## Summary

- Cover `shift_steady_time_by`, monotonic elapsed_ms/us, FakeTime sleep advance, and AutopilotTime offset via time_in.

## Motivation

Adds focused unit coverage for pure helpers steading regression detection without production behavior change.

## Verification

- `mavsdk_time_extra_test.cpp` registered in core CMake unit test list
- Offline review of assertions against existing APIs
- No flight/arm/geofence paths touched

---

**Agent-Owner:** `sera` · **Platform:** hermes · **Claim-TTL:** 24h

PR #2997, opened 2026-07-22
test(core): cover Autopilot stream operator labels

Bartok9:
## Summary

- Unit coverage for `operator<<(Autopilot)` including known values and unknown fallback.

## Motivation

Adds focused unit coverage for pure students of failure regression without production behavior change.

## Verification

- `autopilot_test.cpp` registered in core CMake unit test list
- Offline review of assertions against existing APIs
- No flight/arm/geofence paths touched

---

**Agent-Owner:** `sera` · **Platform:** hermes · **Claim-TTL:** 24h

PR #3004, opened 2026-07-22
test(core): consolidate unit coverage and param_id clamp

Bartok9:
## Summary

Maintainer request (@julianoes): consolidate multiple open core `test(core)` / helper PRs into **one** clean PR to reduce churn and conflicts.

This single commit (rebased on current `main`) folds in:

| Area | Change |
|------|--------|
| **param_id** | Clamp `param_id_to_message_buffer` length; nullptr-safe `extract_safe_param_id`; unit tests |
| **Vehicle** | Public export of stream / `to_vehicle_from_mav_type` + unit coverage |
| **Autopilot** | Stream operator label unit tests |
| **Time** | Export `AutopilotTime` for tests; shift / FakeTime / AutopilotTime cases |
| **math** | `constrain` + deg/rad identity cases (in existing `math_utils_test.cpp`) |
| **geometry** | Mid-latitude east scale + south (in existing `geometry_test.cpp`) |
| **LockedQueue** | stop unblocks waiters, restart after stop, async push delivery (in existing `locked_queue_test.cpp`) |

New TUs registered in core `CMakeLists.txt`: `mavlink_parameter_helper_test.cpp`, `vehicle_test.cpp`, `autopilot_test.cpp`. Extensions go into **existing** test files where those already existed (no parallel `*_extra_*` translation units).

### Supersedes (closing as duplicates)

- #3003 geometry mid-latitude / south  
- #3002 math constrain / deg-rad  
- #2999 Time shift / FakeTime / AutopilotTime  
- #2997 Autopilot stream labels  
- #2996 Vehicle MAV_TYPE + stream  
- #2995 param_id clamp + helpers  

(This PR was formerly LockedQueue-only #3004; head rewritten to the consolidated branch.)

## Motivation

Reduce PR noise and rebase conflicts while keeping the same regression coverage the individual PRs aimed for.

## Verification

- Re-applied tip content of superseded PRs onto fresh `upstream/main`
- Relies on MAVSDK CI `unit_tests`

---

**Agent-Owner:** `bartok` · **Platform:** openclaw · **Claim-TTL:** 24h
**Claim:** bartok

### DIFF
--- a/cpp/src/mavsdk/core/CMakeLists.txt
+++ b/cpp/src/mavsdk/core/CMakeLists.txt
@@ -287,6 +287,9 @@ list(APPEND UNIT_TEST_SOURCES
     ${PROJECT_SOURCE_DIR}/mavsdk/core/flight_mode_test.cpp
     ${PROJECT_SOURCE_DIR}/mavsdk/core/fs_utils_test.cpp
     ${PROJECT_SOURCE_DIR}/mavsdk/core/connection_result_test.cpp
+    ${PROJECT_SOURCE_DIR}/mavsdk/core/mavlink_parameter_helper_test.cpp
+    ${PROJECT_SOURCE_DIR}/mavsdk/core/vehicle_test.cpp
+    ${PROJECT_SOURCE_DIR}/mavsdk/core/autopilot_test.cpp
 )
 
 if (NOT BUILD_WITHOUT_CURL)
--- a/cpp/src/mavsdk/core/autopilot_test.cpp
+++ b/cpp/src/mavsdk/core/autopilot_test.cpp
@@ -1,0 +1,29 @@
+#include "autopilot.hpp"
+#include <gtest/gtest.h>
+#include <sstream>
+
+using namespace mavsdk;
+
+TEST(Autopilot, StreamOperator)
+{
+    {
+        std::ostringstream oss;
+        oss << Autopilot::Px4;
+        EXPECT_EQ(oss.str(), "Px4");
+    }
+    {
+        std::ostringstream oss;
+        oss << Autopilot::ArduPilot;
+        EXPECT_EQ(oss.str(), "ArduPilot");
+    }
+    {
+        std::ostringstream oss;
+        oss << Autopilot::Unknown;
+        EXPECT_EQ(oss.str(), "Unknown");
+    }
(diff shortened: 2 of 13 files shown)


cpp/src/mavsdk/core/locked_queue_extra_test.cpp
```diff
@@ -0,0 +1,65 @@
...
+        EXPECT_EQ(*item, 7);
+    }
+    EXPECT_EQ(q.size(), 0u);
+}
+
+TEST(LockedQueue, WaitAndPopFrontReceivesPush)
+{
+    LockedQueue<int> q{};
+
+    auto fut = std::async(std::launch::async, [&q]() {
```
julianoes: Did you want to set the policy to deferred, or doesn't matter?

julianoes (commented):
Why are they called "extra" tests?

Bartok9:
Preemptively addressing the same style note as on #3002 / #2995:

- Removed `locked_queue_extra_test.cpp` and its CMake registration.
- Folded the new cases into the existing `locked_queue_test.cpp`.

No production code change — tests only. Happy to adjust further if you'd rather drop any redundant case.

julianoes (dismissed):

Bartok9:
@julianoes on the `std::async` launch policy — intentional `std::launch::async`, not deferred.

The stop / wait_pop cases need a real second thread blocked inside `wait_and_pop_front` *before* we call `stop()` / `push_back()`. With the default/`deferred` policy the lambda would not start until `.get()`, so the waiter would never be parked and the race under test would disappear. Explicit `async` keeps that concurrent.

Happy to switch any of them if you'd rather drive the waiter with a plain `std::thread` instead.

cpp/src/mavsdk/core/locked_queue_extra_test.cpp
Bartok9: Intentional `std::launch::async`, not deferred.

These stop / wait_pop cases need the waiter actually blocked in another thread before `stop()` / `push_back()`. Deferred would not start the lambda until `.get()`, so the race under test would go away. Happy to switch to `std::thread` if you prefer that style.

julianoes (approved):

PR #3006, opened 2026-07-22
docs(python): add a migration page for the mavsdk package rename

julianoes:
From v4 the `mavsdk` name on PyPI refers to the native binding rather than the gRPC wrapper, which continues as `mavsdk-grpc` ([mavlink/MAVSDK-Python#826](https://github.com/mavlink/MAVSDK-Python/pull/826)).

The deprecation notices shipped by MAVSDK-Python need somewhere specific to link to. Until now the only candidate was the Python docs index — which documents the *old* gRPC API, so it was exactly the wrong page to land someone on right after telling them their API is changing.

Adds `en/python/migration.md` covering:

- what changed, and the resulting package/import-name table
- the three options: switch to `mavsdk-grpc`, pin `mavsdk<4`, or move to the native binding
- an API mapping table and side-by-side examples for both the asyncio and synchronous interfaces

Also relabels the S3 API reference link on the Python index as belonging to `mavsdk-grpc`, since it will otherwise compete with the native binding docs.

The page carries a `::: warning` noting the native binding is not on PyPI yet — it ships as `mavsdk` 4.0.0. That paragraph should be dropped when it is.

Verified with `npm run docs:build`: the page renders and appears in the sidebar.

### DIFF
--- a/docs/en/python/migration.md
+++ b/docs/en/python/migration.md
@@ -47,7 +47,7 @@ This keeps working indefinitely, but receives no further updates.
 
 Beyond the API itself:
 
-- **No `grpcio` dependency**, and no `mavsdk_server` subprocess. The package wraps the MAVSDK C library directly.
+- **No `grpcio` dependency**, and no `mavsdk_server` subprocess. The package wraps the MAVSDK C++ library via the C wrapper directly.
 - **Two interfaces.** `mavsdk` exposes a synchronous, callback-based API; `mavsdk.asyncio` exposes an asyncio API close in spirit to MAVSDK-Python. Both come from the same distribution — there is nothing extra to install.
 - **Explicit system discovery.** MAVSDK-Python hid connection and discovery behind `drone.connect()`. The native binding separates connecting from discovering systems, which matters once more than one vehicle is involved.
 - **Plugins are constructed, not attributes.** `drone.action.arm()` becomes `Action(drone).arm()`.
--- a/docs/en/python/migration.md
+++ b/docs/en/python/migration.md
@@ -1,0 +1,162 @@
+# Migrating from MAVSDK-Python
+
+The `mavsdk` package on PyPI is changing hands.
+
+It used to be [MAVSDK-Python](https://github.com/mavlink/MAVSDK-Python): an asyncio wrapper that talks gRPC to a `mavsdk_server` process bundled inside the package. From MAVSDK v4, the `mavsdk` name refers instead to a native binding that calls the MAVSDK C library directly, with no gRPC and no server process.
+
+The two have different APIs. This page explains what to do about it.
+
+::: warning
+The native binding has not been released to PyPI yet — it will ship as `mavsdk` 4.0.0. Until then, `pip install mavsdk` still gives you the gRPC wrapper. Nothing breaks today; this page is here so the change doesn't surprise you later.
+:::
+
+## The short version
+
+| | Before | From v4 |
+|---|---|---|
+| gRPC wrapper | `mavsdk` | `mavsdk-grpc`, imported as `mavsdk_grpc` |
+| Native binding | *did not exist* | `mavsdk`, imported as `mavsdk` |
+
+The gRPC wrapper deliberately installs under a *different import name* from the native binding, so the two can be installed side by side without overwriting each other.
+
+## Which one do I want?
+
+**Keep the gRPC API, keep getting updates** — switch to `mavsdk-grpc`. It continues to be developed, including releases tracking MAVSDK v4. Change your requirements and one import:
+
+```sh
+pip install mavsdk-grpc
+```
+
+```python
+import mavsdk_grpc as mavsdk           # or: from mavsdk_grpc import System
+```
+
+Nothing else in your code needs to change.
+
+**Change nothing at all** — pin the last release under the old name:
+
+```
+mavsdk<4
+```
+
+This keeps working indefinitely, but receives no further updates.
+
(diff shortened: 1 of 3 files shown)


docs/en/python/migration.md
```diff
@@ -0,0 +1,162 @@
...
+
+This keeps working indefinitely, but receives no further updates.
+
+**Move to the native binding** — read on. It is worth it if you want to drop the `grpcio` dependency, avoid shipping and launching a `mavsdk_server` binary, or use the synchronous API.
+
+## What the native binding changes
+
+Beyond the API itself:
+
+- **No `grpcio` dependency**, and no `mavsdk_server` subprocess. The package wraps the MAVSDK C library directly.
```
julianoes:
```suggestion
- **No `grpcio` dependency**, and no `mavsdk_server` subprocess. The package wraps the MAVSDK C++ library via the C wrapper directly.
```

docs/en/python/migration.md
```diff
@@ -0,0 +1,162 @@
...
+```python
+import mavsdk_grpc as mavsdk           # or: from mavsdk_grpc import System
+```
+
+Nothing else in your code needs to change.
+
+**Change nothing at all** — pin the last release under the old name:
+
+```
+mavsdk<4
```
hamishwillee, line 39: This is in what file? I only ask because the less than "looks like" a typo, while I expect it is what you wanted.

docs/en/python/migration.md
```diff
@@ -0,0 +1,162 @@
...
+
+The two have different APIs. This page explains what to do about it.
+
+::: warning
+The native binding has not been released to PyPI yet — it will ship as `mavsdk` 4.0.0. Until then, `pip install mavsdk` still gives you the gRPC wrapper. Nothing breaks today; this page is here so the change doesn't surprise you later.
+:::
+
+## The short version
+
+| | Before | From v4 |
```
hamishwillee, line 15: Prettier file?

docs/en/python/migration.md
```diff
@@ -0,0 +1,162 @@
+# Migrating from MAVSDK-Python
```
hamishwillee, line 1: Maybe 

```suggestion
# Migrating MAVSDK to native Python Bindings
```

docs/en/python/migration.md
julianoes: Ah, that's for the requirements.txt, usually. Good point.

docs/en/python/migration.md
julianoes: Hmm, not convinced. I can't think of a better one though.

docs/en/python/migration.md
```diff
@@ -0,0 +1,162 @@
...
+
+The `mavsdk` package on PyPI is changing hands.
```
julianoes:
```suggestion
The `mavsdk` package on PyPI is changing.
```

It's not changing hands, just the implementation is changing. The maintainers stay.

docs/en/python/migration.md
```diff
@@ -0,0 +1,162 @@
...
+
+It used to be [MAVSDK-Python](https://github.com/mavlink/MAVSDK-Python): an asyncio wrapper that talks gRPC to a `mavsdk_server` process bundled inside the package. From MAVSDK v4, the `mavsdk` name refers instead to a native binding that calls the MAVSDK C library directly, with no gRPC and no server process.
```
julianoes:
```suggestion
It used to be [MAVSDK-Python](https://github.com/mavlink/MAVSDK-Python): an asyncio wrapper that talks gRPC to a `mavsdk_server` process bundled inside the package. From MAVSDK v4, the `mavsdk` name refers instead to a native binding that calls the MAVSDK C++ library directly (via C), with no gRPC and no server process.
```

docs/en/python/migration.md
julianoes: How?

docs/en/python/migration.md
```diff
@@ -0,0 +1,162 @@
...
+**Move to the native binding** — read on. It is worth it if you want to drop the `grpcio` dependency, avoid shipping and launching a `mavsdk_server` binary, or use the synchronous API.
+
+## What the native binding changes
+
+Beyond the API itself:
+
+- **No `grpcio` dependency**, and no `mavsdk_server` subprocess. The package wraps the MAVSDK C++ library via the C wrapper directly.
+- **Two interfaces.** `mavsdk` exposes a synchronous, callback-based API; `mavsdk.asyncio` exposes an asyncio API close in spirit to MAVSDK-Python. Both come from the same distribution — there is nothing extra to install.
+- **Explicit system discovery.** MAVSDK-Python hid connection and discovery behind `drone.connect()`. The native binding separates connecting from discovering systems, which matters once more than one vehicle is involved.
+- **Plugins are constructed, not attributes.** `drone.action.arm()` becomes `Action(drone).arm()`.
```
julianoes, line 53:
```suggestion
- **Plugins are constructed, not attributes.** `drone.action.arm()` becomes `Action(drone).arm()`.
- **Support for multiple systems.**
- **Support for server side plugins.** This allows implementations of things like cameras, or gimbals using plugins such as `CameraServer` and `GimbalServer`.
```

docs/en/python/migration.md
```diff
@@ -0,0 +1,162 @@
...
+pip install mavsdk-grpc
+```
+
+```python
+import mavsdk_grpc as mavsdk           # or: from mavsdk_grpc import System
+```
+
+Nothing else in your code needs to change.
+
+**Change nothing at all** — pin the last release under the old name:
```
julianoes:
```suggestion
**Change nothing at all** — pin the last release under the old name (e.g. in your requirements.txt)
```

JonasVautherin (approved):

PR #3019, opened 2026-07-27
Rework Debian packaging around system dependencies

KBentley57:
## Summary

MAVSDK already has a useful Debian release path alongside its self-contained CMake superbuild. This PR carries that direction through to Debian's normal packaging model: common libraries come from the distribution, package metadata describes the resulting dependencies, and the existing bundled build remains the default everywhere else.

## Packaging changes

- Add the opt-in `MAVSDK_USE_SYSTEM_COMMON_DEPENDENCIES` path for Asio, curl, fmt, liblzma, nlohmann-json, and TinyXML2. Ordinary CMake builds remain unchanged.
- Keep the four MAVSDK-specific sources—MAVLink, libevents, PicoSHA2, and libmavlike—pinned and fetched by the superbuild.
- Declare Debian build requirements up front, use native Python packages instead of a private pip bootstrap, derive runtime dependencies with `dpkg-shlibdeps`, and install into the correct multiarch paths.
- Produce conventional `libmavsdk3`, `libmavsdk-dev`, and detached `libmavsdk3-dbgsym` packages, with corrected CMake/pkg-config metadata for consumers.
- Support Debian 12 and 13 on amd64 and arm64, while retaining Debian 11 as an explicitly labelled legacy path.
- Add a narrowly scoped workaround for [Debian bug #1111197](https://bugs.debian.org/1111197), documented for removal when the affected libpsl transition no longer requires it.

The remaining fetched sources mean this is suitable for MAVSDK's upstream binary releases, but not yet for Debian archive upload. `debian/README.source` records that boundary. The existing licensing file is deliberately left for maintainer review.

## Validation

The complete project matrix passed on the exact PR head, `b2ca97dc5`:

- [Linux, including Debian 11/12/13 amd64 and arm64 packaging](https://github.com/KBentley57/MAVSDK/actions/runs/30273384247)
- [macOS and iOS, including the assembled XCFramework](https://github.com/KBentley57/MAVSDK/actions/runs/30273383860)
- [Windows](https://github.com/KBentley57/MAVSDK/actions/runs/30273383844)
- [Documentation](https://github.com/KBentley57/MAVSDK/actions/runs/30273383899)
- [Fuzzing](https://github.com/KBentley57/MAVSDK/actions/runs/30273383882)

This also exercises the default bundled superbuild, system-dependency builds, CMake consumers, cross-compilation, package installation and contents, Lintian, static pkg-config resolution, and separated debug symbols.


KBentley57:
Temporarily closing this draft while the corrected head is validated in fork CI. I will reopen it once the full matrix is in a reviewable state.

.github/workflows/linux.yml
```diff
@@ -460,14 +460,8 @@ jobs:
             os_name: Ubuntu
             os_version: "24.04"
           - os_type: debian
-            os_name: Debian
+            os_name: Debian (legacy)
```
julianoes, line 463: LTS support for Debian 11 is about to run out. I wonder if we should drop this one.

cpp/debian/changelog
```diff
@@ -0,0 +1,7 @@
+mavsdk (3.17.2-1) unstable; urgency=medium
+
+  * Build common dependencies from Debian packages.
+  * Split the runtime library from the development files.
+  * Generate runtime dependencies from shared-library metadata.
```
julianoes, line 5: This is going to be a pain and forever conflict. We use the version number from the git tag, so we try not to have it in source itself. Could we just link to the github releases or something like that?

cpp/debian/control
```diff
@@ -1,24 +1,44 @@
 Source: mavsdk
-Priority: optional
-Maintainer: hcb <hcb@archaeopteryx>
-Build-Depends: debhelper (>= 11), nlohmann-json3-dev, libcurl4-openssl-dev, libtinyxml2-dev
-Standards-Version: 4.1.3
 Section: libs
-Homepage: https://mavsdk.mavlink.io/main/en/index.html
+Priority: optional
+Maintainer: MAVSDK Development Team <julian@oes.ch>
```
julianoes, line 4: > Maintainer: fullname-email (recommended)
> Should be in the format “Joe Bloggs <jbloggs@foo.com>”, and is
> typically the person who created the package, as opposed to
> the author of the software that was packaged.

So this would be you, not me, right?

cpp/third_party/CMakeLists.txt
```diff
@@ -13,38 +13,47 @@ if (CMAKE_CROSSCOMPILING)
...
-        if(ANDROID)
-            build_target(boringssl)
-        else()
-            build_target(openssl)
-        endif()
 
-        if(NOT IOS)
-            build_target(zlib-ng)
-        endif()
+    if(NOT MAVSDK_USE_SYSTEM_COMMON_DEPENDENCIES)
```
julianoes, line 28: A bit ugly but I see what you're doing. To me this should just be "not-superbuild", and then you have to run some script to prepare the things missing.

.github/workflows/linux.yml
KBentley57: I anticipated dropping it, but I left it in for the CI run.  EOL is at the end of August, and I would say most people have already moved on.

cpp/debian/changelog
KBentley57: I was thinking of a few different ways to do this.  In CMake we could use FetchContent_{Declare, Populate, GetProperties} with specific git tags or hashes.  There's also The GIT_URL and associated features inside ExternalProject_Add, that will essentially accomplish the same thing as the superbuild, fetching the source from a specific hash or tag.

cpp/debian/control
KBentley57: Well I did fill it out with your name :).  In all seriousness, I think if we can get this set up right, it won't take much maintenance.  I wouldn't care to help out in this regard though.

KBentley57:
I wanted to get this out for you guys to review.  I'm open to any comments or suggestions that make it work better with your flow, or anticipated changes.  There's a lot of ways to make it work, and the one thing I don't want to do is add more work for you both.  There's also the matter of the four sources that are still built inline.  If you want to take this further, after testing this out, I could make those into their own debian packages as well, and work towards a true debian stack of packages and dependencies.  

Edit:  

There are a few questions I have about the other dependencies, but I'd rather focus on one improvement at a time.

cpp/third_party/CMakeLists.txt
KBentley57: I had thought about that, but then I realized, some people may not be on debian, and also not want to do the superbuild, etc..  I know it's probably an edge case, but I know many people use red hat and its derivatives.  I suppose I could also look towards packaging for those OS's in the same way, but that would certainly be more code that is new, and less that is changed.

cpp/debian/changelog
julianoes: We already get the git tag in cmake:

https://github.com/mavlink/MAVSDK/blob/99f6c9b4696b8490000d9b94624532f765a96b66/cpp/CMakeLists.txt#L18-L41

So, we should probably make this file a template and fill it out with cmake?

cpp/debian/control
julianoes: I expect this to be my responsibility, so yeah my name is probably right. That being said, the spec says it would be you as you're adding this...not sure how much it matters - probably best if it's just me to keep it consistent indeed.

cpp/third_party/CMakeLists.txt
julianoes: What about a script for Debian that installs/prepare things as required, like the one we had, and then one for Fedora/etc. We used to ship rpms but no one seemed to need/want them, so I dropped it.

I just don't want 3+ ways to build it. It should be either:
- SUPERBUILD - everything taken care of
- not SUPERBUILD - you do whatever you want, in this case grab packages required and install the rest via the system.

If we want to make sure the two ways are not getting out of sync, what about adding a file with env vars that lists the git hashes or tags per repo and then use those strings in either the cmake SUPERBUILD or the debian package tooling script?

KBentley57: Having a go / no go on the superbuild is probably best.  All or nothing.  I think that would be the easiest to manage too.  I'll rework a few things, including the template idea you mentioned above.  I think there might be a use case for redhat like systems, and if we have a debian script, it probably wouldn't be too hard to make the small mods to make the RH work.  I'll have to think about the env vars.  i've been bitten so many times by a dirty environment or cache variables that I'd want to to be foolproof.

julianoes:
Are you blocked by us? Any questions we need to answer? It sounded like you were ok trying the full superbuild or not at all option?

julianoes:
Gentle bump? Still interested in this?

KBentley57:
Hi Julian,

It's still on my radar.  I had a vacation, then a sprint, then the kids'
school starting up, etc.. busy time of the year haha.

PR #3020, opened 2026-07-29
core: heartbeat watchdog (deadman timer), reworked

julianoes:
Alternative to #2895, which this builds on the same feature but is reworked after review by Claude. @tpayne-censystech's commits are preserved.

Includes mavlink/MAVSDK-Proto#421.

## Original description

When enabled, this disables MAVSDK's internally generated HEARTBEAT sending and drives emission from the client via a runtime-configurable deadman timeout. The client refreshes within the timeout to keep heartbeats flowing; if it crashes, disconnects, or stops servicing the path, MAVSDK stops sending, so PX4 sees the GCS/companion component go offline and triggers the configured lost-link failsafe (e.g. RTL). MAVSDK continues to own heartbeat content (sysid, compid, mav_type, etc.); the client only controls whether emission continues. Default behavior is unchanged.

API:

```cpp
Mavsdk::Configuration config(ComponentType::GroundStation);
config.set_heartbeat_watchdog_timeout_s(2.0);   // 0 disables (the default)
Mavsdk mavsdk(config);

mavsdk.feed_heartbeat_watchdog();               // from wherever the app proves it is healthy
```

For gRPC wrappers: `mavsdk_server --heartbeat-watchdog-timeout 2`, or the `SetHeartbeatWatchdogTimeout` / `FeedHeartbeatWatchdog` RPCs.

## What changed relative to #2895

- **Refactored** the heartbeat function to one call_every loop which then checks all conditions before sending (or not) the heartbeat. I think this is cleaner than starting and stopping that loop based on two conditions.

- **The watchdog is its own class.** for easier testing.

- **Minimum timeout is 2 s, not 1 s.** to avoid aliasing.

- **Fixed a mutex issue that could leak a heartbeat.**

**C API and mavsdk_server surface.** The watchdog is only reachable from gRPC wrappers, so the additions to `c/` (the C wrapper over libmavsdk) are dropped.

**Reverted unused bool** the `bool` return added to `TimeoutHandler::refresh()`

**Reverted the instance stuff for `MavlinkChannels`. I don't understand why that would be required.

## Docs

New guide page at `docs/en/cpp/guide/heartbeat_watchdog.md

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "mavsdk-proto"]
 	path = proto
-	url = https://github.com/mavlink/MAVSDK-Proto.git
+	url = https://github.com/tpayne-censystech/MAVSDK-Proto.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "mavsdk-proto"]
 	path = proto
-	url = https://github.com/tpayne-censystech/MAVSDK-Proto.git
+	url = https://github.com/mavlink/MAVSDK-Proto.git
--- a/c/src/cmavsdk/mavsdk.cpp
+++ b/c/src/cmavsdk/mavsdk.cpp
@@ -121,6 +121,16 @@ void mavsdk_configuration_set_always_send_heartbeats(mavsdk_configuration_t conf
     cpp_config->set_always_send_heartbeats(always_send_heartbeats != 0);
 }
 
+double mavsdk_configuration_get_heartbeat_watchdog_timeout_s(mavsdk_configuration_t config) {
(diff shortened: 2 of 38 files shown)


julianoes:
@tpayne-censystech I did an iteration with Claude on this. Are you ok with this? Happy to force push to your branch as well to keep your PR authorship.

tpayne-censystech:
> @tpayne-censystech I did an iteration with Claude on this. Are you ok with this? Happy to force push to your branch as well to keep your PR authorship.

I'd prefer to keep the PR authorship if you don't mind.

julianoes:
Force-pushed to existing PR, closing.

PR #3026, opened 2026-07-31
Various python fixes

JonasVautherin:
* `System` and `ServerComponent` never freed their handle.
* Plugins were pinned forever by `atexit` and outlived `mavsdk_destroy()`. They now  get destroyed with their system.
* Added missing `autopilot_type()` and `vehicle_type()`.

The `c/` changes are needed for this: `mavsdk_system_destroy()` didn't have `CMAVSDK_EXPORT`, so with `-visibility=hidden` it was disappearing.

Note: for the plugins only the template changed, the rest is auto-generated.

### DIFF
--- a/py/mavsdk/templates/file.py.j2
+++ b/py/mavsdk/templates/file.py.j2
@@ -8,7 +8,6 @@
 {{ class_description }}
 """
 
-import atexit
 import ctypes
 
 from typing import Callable, Any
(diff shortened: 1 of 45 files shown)


py/mavsdk/templates/file.py.j2
```diff
@@ -228,7 +227,11 @@ class {{ plugin_name.upper_camel_case }}:
         if not self._handle:
             raise RuntimeError("Failed to create {{ plugin_name.upper_camel_case }} plugin - C function returned null handle")
 
-        atexit.register(self.destroy)
+        {% if is_server %}
```
JonasVautherin, line 230: Registering with the owner replaces `atexit.register(self.destroy)`, which held a strong reference for the whole process (so a plugin could never be collected)

PR #3030, opened 2026-08-02
mavsdk_server: fix Android CI build

JonasVautherin:
This should fix the `mavsdk_server` Android tar and therefore the MAVSDK-Java builds, which have been stuck at 3.16.0.

We will only be able to confirm that with the next MAVSDK release though, when I run MAVSDK-Java against it 👍.


julianoes (approved):
Hm, did we break this somehow?

JonasVautherin:
Yeah, adding `mavsdk_export.h` broke it 👍

PR #3033, opened 2026-08-07
core: add possible "unknown" value for param count

JoC2000:
### Issue:
`enqueue_value_broadcast` filled both the param index and count with 0xFFFF.
Component anounces 65535 parameters. GCS starts requesting for missing params that are out of range.

If you register only 9 new params: 
```
param at 1204 out of bounds (9)
Ignoring request_read message - param index not found: 1204
```
### Fix:
Get actual index and count from `_param_cache`


julianoes (approved):
Ah, thanks!

PR #3038, opened 2026-08-09
Action: add GotoLocationFixedwing

### LINKED ISSUE #2853: Augment goto_location action for planes
Hi, I was wondering if we could add the possibility to set the loiter radius when calling the **goto_location** action.

It is using the **MAV_CMD_DO_REPOSITION (192)** (https://mavlink.io/en/messages/common.html#MAV_CMD_DO_REPOSITION) MAVLink command whose third parameter is "Loiter radius for planes. Positive values only, direction is controlled by Yaw value. A value of zero or NaN is ignored."

I don't know about PX4, but this allows me to fly a plane with ArduPilot. However I would understand that it can make the API a bit ugly.

bansiesta:
## Summary

Adds the `goto_location_fixedwing` action, a fixed-wing-friendly variant of `goto_location` that accepts a `loiter_radius_m` parameter instead of a yaw angle. Under the hood it sends `MAV_CMD_DO_REPOSITION` with param3 = loiter radius (positive: clockwise, negative: counter-clockwise, 0: ignored by the autopilot) and param4 (yaw) = NaN, since yaw isn't meaningful while a fixed-wing aircraft is continuously turning in a loiter.

Mirrors mavlink/MAVSDK-Proto#412.

## Changes

- `cpp/src/mavsdk/plugins/action/action_impl.cpp`/`.hpp`: new `goto_location_fixedwing`/`goto_location_fixedwing_async`, packing `MAV_CMD_DO_REPOSITION` with the loiter radius on param3 and NaN yaw on param4. Reuses the existing PX4/ArduPilot flight-mode gating before sending the command.
- `cpp/src/mavsdk/plugins/action/action.cpp`/`include/plugins/action/action.hpp`: public API wrappers, matching `goto_location`'s pattern.
- `cpp/src/mavsdk/plugins/action/mocks/action_mock.hpp`: mock method for unit tests.
- `cpp/src/mavsdk_server/src/plugins/action/action_service_impl.hpp`: gRPC server translation for the new `GotoLocationFixedwing` RPC.
- `proto` submodule: pinned to MAVSDK-Proto#412's branch head so this compiles against the new `GotoLocationFixedwingRequest`/`Response` messages. **This needs to be updated to MAVSDK-Proto's `main` once #412 merges.**

Closes #2853.

## Test plan

- [ ] MAVSDK-Proto#412 merges and the `proto` submodule pointer here is updated to `main`
- [ ] CI build passes (C++ core + mavsdk_server gRPC layer)
- [ ] Manual SITL check: `action.goto_location_fixedwing(...)` on a fixed-wing airframe holds a loiter of the commanded radius/direction at the target location

I wasn't able to do a full local build/test — the `proto` submodule isn't populated in this environment and MAVSDK-Proto#412 (which defines the new messages) isn't merged yet, so the gRPC layer can't be compiled here. The core plugin changes follow the exact pattern of the neighboring `goto_location` implementation.

### DIFF
--- a/c/src/cmavsdk/plugins/action/action.cpp
+++ b/c/src/cmavsdk/plugins/action/action.cpp
@@ -565,6 +565,50 @@ mavsdk_action_goto_location(
     return translate_result(ret_value);
 }
 
+// GotoLocationFixedwing async
+void mavsdk_action_goto_location_fixedwing_async(
+    mavsdk_action_t action,
+    double latitude_deg,
+    double longitude_deg,
+    float absolute_altitude_m,
+    float loiter_radius_m,
+    mavsdk_action_goto_location_fixedwing_callback_t callback,
+    void* user_data)
+{
+    auto wrapper = reinterpret_cast<mavsdk_action_wrapper*>(action);
+
+    wrapper->cpp_plugin->goto_location_fixedwing_async(
+        latitude_deg,
+        longitude_deg,
+        absolute_altitude_m,
+        loiter_radius_m,
+        [callback, user_data](
+            mavsdk::Action::Result result) {
+                if (callback) {
+                    callback(
+                        translate_result(result),
+                        user_data);
+                }
+        });
+}
+
+
+// GotoLocationFixedwing sync
+mavsdk_action_result_t
+mavsdk_action_goto_location_fixedwing(
+    mavsdk_action_t action,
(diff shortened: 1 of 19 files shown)


julianoes (dismissed):
Ok, let me merge the proto part.

PR #3047, opened 2026-08-30
core: send asynchronously instead of blocking the io thread

julianoes:
## The problem

Reads went async when the connections moved to Asio, but writes did not. Every stream send path still performed a **synchronous** write while occupying the io thread:

| Transport | Where |
|---|---|
| Serial | `asio::write(_serial_port, …)` |
| TCP client | `asio::write(_socket, …)`, inside a handler dispatched to the io thread, caller blocked on `future.get()` |
| TCP server | same, on `_client_socket` |

That one thread also runs message dispatch, every internal timer, and all the plugin protocol state machines, so a single slow or stalled peer stalled all of them. `deliver_message()` holds `MavsdkImpl::_mutex` across the send, so a stalled write additionally blocked `add_any_connection()`, `remove_connection()` and `systems()` on user threads.

A 280-byte MAVLink frame at 57600 baud is already ~48 ms of io thread. TCP backpressure from a peer that stopped reading is unbounded — and as the new test shows, it also makes `~Mavsdk` hang, because the io thread can never be joined.

## The change

Each stream connection owns a bounded `TxQueue` (new `tx_queue.hpp`) drained by an `async_write` chain, one write in flight at a time. `send_raw_bytes()` copies the bytes, enqueues them and returns.

Enqueueing uses `asio::dispatch`, so it runs inline on the io thread — which is where the send path already is — keeping it ordered against writes already queued.

Some things that fell out of it:

- **All socket access is now on the io thread**, so the per-connection `_send_mutex` is gone from all three stream connections. To get there, `TcpClientConnection`'s initial `do_connect()` is now posted onto the io thread rather than running on the caller's.
- **Failure reporting had to be rebuilt**, since a write error is now discovered after the call returned. `Connection::report_send_error()` → `MavsdkImpl::report_connection_error()` routes it to the same `subscribe_connection_errors()` subscribers. Immediately knowable failures (`"Not connected"`) are still returned synchronously, backed by new `_connected` / `_client_connected` / `_port_open` atomics, so `deliver_message()`'s success counting still works as before.
- **Drop policy is oldest-first**, capped at 1024 items (~280 KiB) per connection. For MAVLink the stalest telemetry is the right thing to lose.

## UDP is deliberately left alone

`UdpConnection` is untouched — byte-identical to `main`. I did convert it at first, and CI caught a regression: `Ftp.DownloadBurstBigFile` failed under ThreadSanitizer with `Offset mismatch`, preceded by a run of `Ignore: last: 5, req: 15` warnings.

It reproduces without TSan by running the test on a fully loaded machine (16 spinners on 8 cores):

| | failures |
|---|---|
| queued UDP sends | 6 / 8 |
| `main`, same load | 0 / 8 |

**It was latency, not packet loss.** Two measurements pin that down:

| | result |
|---|---|
| kernel `UdpInErrors` / `UdpRcvbufErrors` delta over 6 failing runs | **0 / 0** |
| queued UDP, FTP timeout raised 0.1 s → 3.0 s, nothing else changed | **0 / 6 failures** |

Nothing was dropped; the client just didn't get its next burst packet within 100 ms.

The reason is how a one-in-flight queue paces sends. Synchronously, `deliver_message()` puts the datagram on the wire inline, before the handler returns, so delivery is paced by how fast the FTP burst thread produces packets. With the queue, only the first datagram goes out immediately (Asio does a speculative non-blocking send); every subsequent one is gated on the previous one's completion handler, because the queue is `busy()`. Delivery is then paced by the **io_context turn rate** — ~210 burst packets become ~210 sequential turns, each competing with all other io_context work. Under CPU contention that comfortably exceeds 100 ms. The FTP client hit its burst timeout and switched to missing-pieces recovery (that is what `last: 5`, i.e. `ReadFile`, means) while burst packets (`req_opcode 15`) were still arriving and being ignored, until the offsets desynced.

There is a second, independent reason not to queue datagrams: when the io thread *is* idle, the queue drains as a tight back-to-back chain with none of `deliver_message()`'s per-message work (libmav parse, JSON interception, tlog, mutex) in between — i.e. it bursts more than the receiver's socket buffer at once, which is a known way to cause drops between PX4 and MAVSDK. The synchronous path accidentally paced itself because that work sat between every send.

None of this applies to TCP or serial, which have real flow control. And UDP never needed the fix in the first place: what makes a stream write block is peer backpressure, and a datagram socket has none — `send_to()` hands the datagram to the kernel and returns.

## Testing

`Connections.StalledLinkDoesNotStallOtherConnections` guards the actual property. The autopilot gets a serial link to a pseudo-terminal nobody drains plus a UDP link to a real ground station; after a short flood the pty buffer is full, and the test checks the ground station still receives heartbeats.

Verified in both directions:

| | heartbeats in 3 s | result |
|---|---|---|
| before this change | 0 | fails, then hangs at teardown — io thread unjoinable |
| after | 3, 3, 3 (3 runs) | passes, ~6.2 s |

A pty is used rather than a loopback TCP peer that stops reading, because the kernel auto-tunes a loopback socket's send buffer into the megabytes — I tried that first and 5000 messages (200 KB) never blocked, so the test caught nothing. A pty buffer is a fixed few kilobytes and does not grow. It's POSIX-only, since only the way the link is wedged is platform-specific, not the behaviour.

`TxQueue` itself is unit-tested directly: bounds, drop-oldest, and that the in-flight buffer survives both eviction and `clear()` (an async write is reading from it).

- system tests: **101/101**, and 0/10 failures for the FTP burst tests under the heavy load described above
- unit tests: **418/418**
- CI green including ThreadSanitizer, AddressSanitizer and UndefinedBehaviour

One CI failure along the way, `Connections.TcpConnectionReconnectionFromServerSide`, was flaky rather than caused by this branch: a re-run of the identical commit passed, and the test allows 2 s for a reconnect driven by a 1 s retry timer plus a 1 Hz heartbeat, which is marginal by construction. Probably worth tightening separately.

## Notes for review

This came out of a broader read of the threading model. A few adjacent things I deliberately did *not* touch here, in rough priority order:

1. `TcpClientConnection::do_connect()` still does a **blocking `resolver.resolve()`**, and on the reconnect path that runs on the io thread once per retry second — an unreachable DNS server freezes the SDK for the resolver timeout, repeatedly. Wants `async_resolve`.
2. `TimeoutHandler` / `CallEveryHandler` copy due callbacks out under the lock and invoke them **outside** it, while `remove()` is callable from any thread (e.g. `LogFilesImpl::deinit()`). Copy → `remove()` returns → owner torn down → callback fires. Use-after-free.
3. Two conventions coexist for unregister-on-destroy: blocking (`MavlinkRequestMessage`, `MavlinkParameterClient`, `SystemImpl`) vs posted (`MavlinkCommandSender`, `Timesync`, `Ping`). The posted one leaves a window; it's safe today only because those objects die after the io_context stops.
4. `MavsdkImpl::send_message()` returns an unconditional `true` — it did before this PR too, but it's worth deciding whether that `bool` should go away or be documented as "queued, not sent".

Happy to do these as follow-ups.

https://claude.ai/code/session_01CMVJeAkDnTNSvfb3uogDDW

### DIFF
--- a/cpp/src/mavsdk/core/CMakeLists.txt
+++ b/cpp/src/mavsdk/core/CMakeLists.txt
@@ -281,6 +281,7 @@ list(APPEND UNIT_TEST_SOURCES
     ${PROJECT_SOURCE_DIR}/mavsdk/core/mavlink_statustext_handler_test.cpp
     ${PROJECT_SOURCE_DIR}/mavsdk/core/ringbuffer_test.cpp
     ${PROJECT_SOURCE_DIR}/mavsdk/core/timeout_handler_test.cpp
+    ${PROJECT_SOURCE_DIR}/mavsdk/core/tx_queue_test.cpp
     ${PROJECT_SOURCE_DIR}/mavsdk/core/unittests_main.cpp
     ${PROJECT_SOURCE_DIR}/mavsdk/core/mavlink_parameter_cache_test.cpp
     ${PROJECT_SOURCE_DIR}/mavsdk/core/string_utils_test.cpp
--- a/cpp/src/mavsdk/core/connection.cpp
+++ b/cpp/src/mavsdk/core/connection.cpp
@@ -114,6 +114,11 @@ void Connection::receive_message(
     _receiver_callback(result, message, connection);
 }
 
+void Connection::report_send_error(const std::string& message)
+{
+    _mavsdk_impl.report_connection_error(message, _handle);
+}
+
 bool Connection::should_forward_messages() const
 {
     return _forwarding_option == ForwardingOption::ForwardingOn;
--- a/cpp/src/mavsdk/core/connection.hpp
+++ b/cpp/src/mavsdk/core/connection.hpp
@@ -40,6 +40,11 @@ public:
     bool should_forward_messages() const;
     static unsigned forwarding_connections_count();
 
+    // Called by MavsdkImpl::add_connection() while the connection is being registered,
+    // before anything can send through it. Lets the connection attribute an
+    // asynchronous send failure to itself; see report_send_error().
+    void set_handle(Mavsdk::ConnectionHandle handle) { _handle = handle; }
+
     // Access to libmav receiver for message creation.
     // Returns a shared_ptr so the caller can safely use the object even if
     // stop_libmav_receiver() is called concurrently on another thread.
@@ -63,6 +68,16 @@ protected:
     void stop_libmav_receiver();
     void receive_libmav_message(const Mavsdk::MavlinkMessage& message, Connection* connection);
 
+    // Report a send failure that was only discovered asynchronously, i.e. after
+    // send_message()/send_raw_bytes() already returned. Reaches the same
+    // Mavsdk::subscribe_connection_errors() subscribers as a synchronous failure.
+    // Must be called on the io_context thread.
+    void report_send_error(const std::string& message);
+
+    // How many datagrams/chunks a connection queues for sending before the oldest is
+    // dropped. At the MAVLink maximum of 280 bytes this caps a stalled link at ~280 KiB.
+    static constexpr std::size_t MAX_TX_QUEUE_ITEMS = 1024;
+
     ReceiverCallback _receiver_callback{};
     LibmavReceiverCallback _libmav_receiver_callback{};
     MavsdkImpl& _mavsdk_impl; // For thread-safe MessageSet access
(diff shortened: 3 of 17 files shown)


fallenmi (approved):
Reviewed the materially changed exact head `1def7c4be624cc91fd59b356c3044cfe80c231ca` and its current merge with `main`. The previous head was excluded before review because ThreadSanitizer and style/proto checks were failing; the new corrective commit restores UDP's synchronous datagram path and clears those regressions while preserving the stream-transport fix.

The undrained-PTY oracle reproduces the original isolation failure on the exact base implementation: after the serial link fills, the unrelated UDP path receives zero heartbeats and the old process cannot shut down cleanly. Exact head/current merge receives three heartbeats in each of three runs and exits normally. The TxQueue suite passed 500/500, both TCP reconnect directions pass, and the full current exact-head matrix is green: 60 successful checks with only two expected skipped publishing jobs, including ASan, TSan, UBSan, and style/proto.

Disclosure: I used OpenAI Codex to assist with this review under the account owner's authorization; I independently validated the materially changed commit history, exact code, current merge, policy, interactions, and test results reported above.

julianoes (commented):
Reviewed, I think it looks sane.

PR #3048, opened 2026-08-30
core: resolve TCP client hostnames asynchronously

julianoes:
> Stacked on #3047 — please review that one first. Base is `async-writes`, so the diff here is just this change.

Finding 3.2 from the threading review.

## The problem

`TcpClientConnection::do_connect()` called `resolver.resolve()`, which runs `getaddrinfo()` **on the io thread**.

That would be tolerable as a one-off at startup, which is what the comment there claimed:

```cpp
// Resolve the remote hostname/IP synchronously. This is called rarely (once at startup,
// then on each reconnect) so blocking briefly here is acceptable.
```

But the reconnect path calls it again **every second** for as long as the peer is unreachable. A DNS server that is itself unreachable then freezes the whole SDK — message dispatch, every internal timer, every other connection — for the resolver timeout, and again a second later, indefinitely.

## The change

`async_resolve()`, which hands the lookup to Asio's internal resolver thread and delivers the result back on the io thread.

Two details worth a reviewer's attention:

**`resolver::cancel()` does not interrupt an in-flight lookup.** It only expires the cancellation token, so a `getaddrinfo()` already running on Asio's resolver thread runs to completion and the handler is *still* invoked afterwards — possibly after `stop()` has returned and the connection has been destroyed. That is safe here only because the handler checks `ec == asio::error::operation_aborted` first and `||` short-circuits before `_stopping` dereferences `this`. Reading `ec` on a dead object is fine; touching a member would not be. Since that ordering is load-bearing and easy to "tidy up" into a bug, it is spelled out in a comment at the check.

**`_reconnect_timer.cancel()` moved into the posted lambda in `stop()`.** It was being called on the caller's thread — the same kind of cross-thread access to an Asio object that the rest of `stop()` exists to avoid. `_resolver.cancel()` joins it there.

## Testing

`Connections.UnresolvableTcpHostDoesNotDisturbOtherConnections` points a TCP client at a `.invalid` name (RFC 2606, guaranteed never to resolve) alongside a working UDP link, and checks the UDP link keeps delivering heartbeats. The retry timer fires several times in the test window, so the resolve-failure path is exercised repeatedly rather than once, and teardown at the end exercises cancelling a resolve that may be in flight.

**What this does not prove**, stated plainly because it matters: the case that motivates the change is a DNS server that *hangs*, whereas `.invalid` answers NXDOMAIN quickly. Making a lookup actually hang needs an unreachable resolver, which a test cannot arrange portably. So this test guards the new code path and the cancel, not the non-blocking property itself. The test says so in a comment.

- system tests: **102/102**
- unit tests: **418/418**
- clang-format clean

## Next in the stack

3.3 (`TimeoutHandler` / `CallEveryHandler` can fire a callback after `remove()` returned — the one memory-safety bug in the review), then 3.4 (uniform unregister-on-destroy), then 3.6 + 3.8.

https://claude.ai/code/session_01CMVJeAkDnTNSvfb3uogDDW

### DIFF
--- a/cpp/src/mavsdk/core/tcp_client_connection.cpp
+++ b/cpp/src/mavsdk/core/tcp_client_connection.cpp
@@ -31,6 +31,7 @@ TcpClientConnection::TcpClientConnection(
     _remote_ip(std::move(remote_ip)),
     _remote_port_number(remote_port),
     _socket(mavsdk_impl.io_context()),
+    _resolver(mavsdk_impl.io_context()),
     _reconnect_timer(mavsdk_impl.io_context())
 {}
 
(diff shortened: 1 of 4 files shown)


julianoes (commented):
Looking at the added system test, this should be good.

PR #3049, opened 2026-08-30
core: make timer removal actually stop the callback

julianoes:
Part 3.3 and 3.4 of an Asio/threading review. Stacked on #3047 (async writes) and the async-resolve branch.

`TimeoutHandler::run_once()` and `CallEveryHandler::run_once()` copied the due callbacks out under the lock and then invoked them outside it. Between those two steps `remove()` — callable from any thread — could return, letting the caller tear down the object the callback had captured, and the callback would then run on the io thread anyway. Removing a due entry from within another callback of the same batch did not stop it either, since it had already been copied out.

Both handlers now park the due entries in a member list instead of a local, so `remove()` cancels them right up to the moment they are invoked, and they gain a `remove_blocking()` that additionally waits when the callback is executing right now. Once that returns, the callback is neither running nor going to run.

`remove()` keeps the non-blocking behaviour on purpose: several call sites remove a cookie while holding a lock the callback itself takes (Offboard, FollowMe, LogStreaming), and waiting there would deadlock.

Destructors and `deinit()`/`disable()` paths switch to the blocking variants, and where they held such a lock the removal is moved out of it. That let two sleeps go — the ones that were standing in for the missing guarantee:

- `CameraImpl::deinit()`'s 100 ms *"wait briefly for call_every lambdas that may already be mid-flight"*
- `TelemetryImpl::disable()`'s 200 ms sleep, next to its `FIXME` about exactly this race

The remaining destructors that unregistered message handlers via the posted `unregister_all()` now use `unregister_all_blocking()` too, so "my destructor stops callbacks" means the same thing everywhere.

### Note for reviewers

Cookie `0` is never handed out by `add()`, and callers keep a default-initialised cookie around for a timer they may never have started. `remove_blocking()` also uses `0` for "nothing executing", so it has to bail out on that value before waiting — without the guard the very first system test deadlocks. There is a regression test for both handlers.

### Testing

### DIFF
--- a/cpp/src/mavsdk/core/call_every_handler.cpp
+++ b/cpp/src/mavsdk/core/call_every_handler.cpp
@@ -2,7 +2,6 @@
 
 #include <utility>
 #include <algorithm>
-#include <vector>
 
 namespace mavsdk {
 
@@ -18,7 +17,7 @@ CallEveryHandler::Cookie CallEveryHandler::add(std::function<void()> callback, d
     _time.shift_steady_time_by(before, -interval_s - 0.001);
     new_entry.last_time = before;
     new_entry.interval_s = interval_s;
-    std::lock_guard<std::recursive_mutex> lock(_mutex);
+    std::lock_guard<std::mutex> lock(_mutex);
     new_entry.cookie = _next_cookie++;
     _entries.push_back(new_entry);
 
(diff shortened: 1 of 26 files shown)


julianoes (commented):
I kept bumping in this. Hopefully, that's now done right.

PR #3050, opened 2026-08-30
core: keep every receive path on the io thread

julianoes:
Parts 3.6, 3.7, 3.8 and 3.10 of an Asio/threading review. Stacked on #3049.

`RawConnection` was the one exception to the "everything happens on the io thread" invariant. `pass_received_raw_bytes()` called straight into the parsers on whatever thread the user handed the bytes in on, and `_mavlink_receiver` / `_libmav_receiver` hold parser state with no synchronisation at all, so two threads feeding raw bytes concurrently corrupted it. `receive()` now posts the bytes onto the io_context, and `stop()` drains the io_context first so no posted handler can outlive the connection.

With that hole closed, the receive path can `dispatch()` instead of `post()`: socket and serial connections already parse on the io thread, so this saves a deferred turn plus a `mavlink_message_t` copy per received message, and still posts for anything that does arrive from elsewhere.

`send_message()` deliberately keeps `post()` — dispatching would let an io-thread send jump ahead of user-thread sends already queued. Its return value is now documented as "queued", not "sent", since that is all it has been able to mean since sending became asynchronous.

Also:

- `server_component_by_id_with_lock()` filled in a null entry for an existing component id but then fell through and appended a **second** entry for the same id.
- `set_callback_executor()` started and joined `_process_user_callbacks_thread` with nothing guarding it, so two concurrent calls raced; and it ran the user's executor with `_callback_executor_mutex` held, so a slow executor blocked the io thread inside the lock. The executor is now held as a `shared_ptr`, taken under the lock and invoked outside it.
- `Connection` gained `io_context()` and `drain_io_context()` helpers, replacing the `static_cast` of `socket.get_executor().context()` back to an `io_context&` in each transport. Each `stop()` asserts it is not running on the io thread — it posts and waits, so it would wait for itself.
- `MavsdkImpl::on_io_thread()` records the io thread's own id rather than relying on asio's `running_in_this_thread()`, whose thread-local is per-DSO here.
- Comments for why `_systems` is walked without `_mutex`, and for `TcpServerConnection` serving one client at a time.

### Testing

### DIFF
--- a/cpp/src/mavsdk/core/connection.cpp
+++ b/cpp/src/mavsdk/core/connection.cpp
@@ -1,5 +1,8 @@
 #include "connection.hpp"
 
+#include <asio/post.hpp>
+#include <cassert>
+#include <future>
 #include <memory>
 #include <utility>
 #include "mavsdk_impl.hpp"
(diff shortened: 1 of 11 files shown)


julianoes (commented):
Looks consistent with the rest now.

PR #3051, opened 2026-08-30
core: replace the FTP burst thread with a timer chain

julianoes:
Part 3.9 of an Asio/threading review, plus the write-up. Stacked on #3050.

A burst read spun up a dedicated `std::thread` per session that looped on `try_to_lock(_mutex)` + `yield()` until it got the lock, then sent burst packets as fast as it could. That burns a core while it waits, and it was the last place in `core/` that wrote MAVLink from a thread other than the io thread — which is also why `burst_stop` had to be atomic and why `_target_system_id` / `_target_component_id` are atomics.

`_work_burst()` already runs on the io thread, so the burst can just chain a zero-delay `steady_timer` there instead: one packet per io_context turn, so everything else — including the FTP message that terminates the burst — still gets a turn in between. Stopping is `_burst_timer.cancel()` rather than setting a flag and joining, which also takes the join out of `_reset()`, where it was called with `_mutex` held.

The lock is still taken with `try_to_lock`, but a failure now re-arms the timer for 1 ms instead of spinning, so a user thread holding `_mutex` cannot block the io thread either.

### `core/THREADING.md`

Also adds a short document writing down what the per-file comments have been saying individually: the two threads, the "shared state is io-thread-only, post to mutate" invariant, `post` vs `dispatch`, which unregister variant guarantees what, the lock order, and the teardown ordering the rest depends on.

The intent is that a new contributor can tell when they are about to break one of these, rather than having to reconstruct the rules from comments spread over a dozen files.

### Testing

426/426 unit tests, 102/102 system tests, including the full `Ftp.*` suite (burst downloads, lossy burst downloads, stop-and-retry).

### DIFF
--- a/cpp/src/mavsdk/core/THREADING.md
+++ b/cpp/src/mavsdk/core/THREADING.md
@@ -1,0 +1,133 @@
+# Threading in libmavsdk core
+
+This is the one place that states the rules the core relies on. The per-file comments explain
+individual decisions; this explains the shape they add up to. If you are adding code to
+`core/` or to a plugin impl, read this first.
+
+## The threads
+
+| Thread | Created in | What runs on it |
+|---|---|---|
+| **io thread** | `MavsdkImpl` ctor (`mavsdk_impl.cpp`) | `asio::io_context::run()`. All socket and serial I/O, MAVLink parsing, message routing, plugin protocol state machines, and every internal timer. |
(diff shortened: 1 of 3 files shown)


julianoes (commented):
That's better. I didn't like the ad-hoc thread needed.

PR #3052, opened 2026-08-30
core: arm the timer poll for the next deadline instead of ticking

julianoes:
Part 3.5 of an Asio/threading review. Stacked on #3051.

`TimeoutHandler` and `CallEveryHandler` were driven by a 5 ms tick that ran whether or not either of them had anything due, so an idle instance woke 200 times a second and no timeout could fire more precisely than the tick.

Both now expose `next_deadline()`, and `MavsdkImpl` arms `_timers_poll_timer` for whichever of them is due first. `add()` and `refresh()` call back into `wake_timers_poll()` when they move the earliest deadline earlier, so a newly added short timeout does not have to wait out the current wait. The wait is capped at `MAX_TIMERS_POLL_WAIT` (100 ms) so a wakeup that somehow does not arrive can only make a timer late, never stall it.

### The interesting part

Doing this exposed that the 5 ms tick was load-bearing for discovery latency.

The 1 Hz heartbeat entry is registered in the `MavsdkImpl` constructor and is due straight away, but `maybe_send_heartbeats()` returns without sending when there is nothing connected yet — and that still consumes the tick, pushing the first heartbeat out by a full second. The old 5 ms delay before the first `run_once()` happened to give the application enough time to call `add_any_connection()` first. Firing at the deadline instead lost that race every time, and added ~1 s to **every** system test (252 s → 338 s for the suite).

So `add_connection()` now marks the heartbeat entry due again (`CallEveryHandler::call_soon()`) once there is a connection to send on. That removes the dependency on the accidental grace window, and means the first heartbeat no longer depends on how long the application took between constructing `Mavsdk` and adding a connection. Suite timing is back to baseline (259 s, matching the branch below).

This is the one behaviour change in the stack, so it is worth a second opinion — happy to drop it and floor the wait at 5 ms instead, which keeps the idle-wakeup win but gives up the sub-5 ms timeout resolution.

### Not included

The two `do_work()` chains (`ServerComponentImpl` and `SystemImpl`) still poll at a fixed 10 ms. Making those event-driven means giving every work queue an "enqueued" signal and its own retry deadline, which is a bigger change; noted in `core/THREADING.md` for a later PR.

### Testing

### DIFF
--- a/cpp/src/mavsdk/core/THREADING.md
+++ b/cpp/src/mavsdk/core/THREADING.md
@@ -76,16 +76,23 @@ in `process_message()`. Use the posted `unregister_all()` there.
 
 Everything time-based is driven from the io thread:
 
-- `MavsdkImpl::schedule_timers_poll()` — every 5 ms, runs `TimeoutHandler::run_once()` and
-  `CallEveryHandler::run_once()`.
+- `MavsdkImpl::schedule_timers_poll()` — runs `TimeoutHandler::run_once()` and
+  `CallEveryHandler::run_once()`. **Deadline-driven**: it asks both handlers for their next
+  deadline and arms for that, so a timeout fires at its deadline rather than at the next
+  tick after it. `add()` and `refresh()` call back into `MavsdkImpl::wake_timers_poll()`
+  when they move the earliest deadline earlier, which re-arms from whatever thread they were
+  called on. The wait is capped at `MAX_TIMERS_POLL_WAIT` so that a wakeup we somehow miss
+  can only make a timer late, never stall it.
 - `MavsdkImpl::schedule_do_work()` — every 10 ms, `ServerComponentImpl::do_work()` for every
   server component.
 - `SystemImpl::schedule_system_work()` — every 10 ms when connected, 100 ms otherwise; runs
   `do_work()` on the command sender, timesync, mission transfer, FTP and parameter clients,
   plus the 5 s ping.
 
-These are fixed-cadence polls rather than deadline-driven timers, so timeout resolution is
-capped at the 5 ms tick and an idle instance still wakes up regularly. That is a known
+The two `do_work()` chains are still fixed-cadence polls that run whether or not their work
+queues have anything in them. Making them event-driven means giving every work queue
+(command sender, parameter clients, mission transfer, FTP) a "something was enqueued" signal
+and its own retry deadline, which is a larger change than the timer one above. Known
 trade-off, not an invariant.
 
 ## Locks that remain
--- a/cpp/src/mavsdk/core/call_every_handler.cpp
+++ b/cpp/src/mavsdk/core/call_every_handler.cpp
@@ -17,11 +17,22 @@ CallEveryHandler::Cookie CallEveryHandler::add(std::function<void()> callback, d
     _time.shift_steady_time_by(before, -interval_s - 0.001);
     new_entry.last_time = before;
     new_entry.interval_s = interval_s;
-    std::lock_guard<std::mutex> lock(_mutex);
-    new_entry.cookie = _next_cookie++;
-    _entries.push_back(new_entry);
 
-    return new_entry.cookie;
+    Cookie cookie;
+    {
+        std::lock_guard<std::mutex> lock(_mutex);
+        new_entry.cookie = _next_cookie++;
+        _entries.push_back(new_entry);
+        cookie = new_entry.cookie;
+    }
+
+    // last_time was deliberately shifted into the past above, so this entry is due right
+    // away and the owner's timer needs to fire now rather than at its current deadline.
+    if (_wakeup_callback) {
+        _wakeup_callback();
+    }
+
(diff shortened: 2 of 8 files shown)


fallenmi (changes requested):
[P1] Re-arm the timer poll when a recurring interval is shortened

I reviewed the single-commit stacked diff from exact base `6e0affcb70f0a16fc3158f10b29ec7fbc2b0023f` to exact head `287d1730bbf5a96bec20218a6e3d96fbde1b04ef`; this finding applies to that diff only, not to the seven commits below it.

`CallEveryHandler::change()` updates `interval_s` under `_mutex` but never invokes `_wakeup_callback`. With this PR, `MavsdkImpl::_timers_poll_timer` may therefore remain armed for the old deadline (or the 100 ms safety poll) even though `next_deadline()` moved earlier. Once that late poll finally runs, `run_once()` advances `last_time` by only one interval, so the zero-delay re-arms then deliver the missed intervals as a callback burst.

I compiled an exact-source oracle against `call_every_handler.cpp` and `mavsdk_time.cpp`, changing a just-fired recurring callback from 1 s to 10 ms. The PR base's former 5 ms polling model delivered the first changed callback at 10.1 ms, with one callback per poll. Exact head `287d1730bbf5a96bec20218a6e3d96fbde1b04ef` and GitHub's current merge produced `change_wakeups=0`, first delivered at 100.05 ms, and then delivered 10 callbacks in the same scheduler wake. The result was byte-identical across 50 runs on both head and merge, and the merge run also reproduced under ASan/UBSan.

Please have `change()` compare the earliest deadline before and after the interval update and, after releasing the lock, wake the owner when the deadline moved earlier. A scheduler-level regression for shortening an existing interval would also prevent the current handler-only tests from missing this path.

Disclosure: I used OpenAI Codex to assist this review; I verified the cited source paths and ran the commands and results described above.

julianoes:
Good catch — fixed in c8305d76b.

You're right that `change()` was the gap. I wired the wakeup into `add()`, `refresh()` and `call_soon()` but missed the one mutator that shortens an existing interval, so the timer stayed parked on the old deadline until the `MAX_TIMERS_POLL_WAIT` cap, and `run_once()`'s one-interval-per-pass catch-up then delivered the backlog at zero delay. `change()` now compares the earliest deadline before and after the update and wakes the owner when it moved earlier.

I also fixed the `set_wakeup_callback()` comment, which claimed `reset()` wakes. It doesn't — `reset()` only ever pushes a deadline further out, so there is nothing to re-arm for.

On the test: rather than a scheduler-level one I added `CallEveryHandler.WakeupOnlyWhenDeadlineMovesEarlier`, which pins the contract for every mutator — `add()`, a shortening `change()` and `call_soon()` wake; a lengthening `change()`, `reset()` and `remove()` don't. That's the invariant the scheduler depends on, and it fails with the `change()` fix removed (verified). I first wrote a timing-based test as well and deleted it: it passed with and without the fix, because the handler side was always correct — the defect was purely that the owner's timer never got re-armed.

One thing I've deliberately left alone: the catch-up burst itself. `run_once()` advancing `last_time` by a single interval per pass predates this PR — with the old 5 ms tick it just spread the same backlog one-per-poll. With the wakeup fixed there's no missed deadline to catch up on, so I'd rather not change cadence semantics for every existing caller inside a threading PR. Happy to do it separately if you think it's worth it.

427/427 unit and 102/102 system tests.

julianoes:
Pushed 6a3d9199c, which fixes the `Camera.Mode` failure CI hit on this PR (Debian 11, 10 s timeout at `camera_mode.cpp:61`).

`subscribe_on_new_system()` checked "is a system already connected?" synchronously on the caller's thread, but `CallbackListImpl::subscribe()` only *posts* its list insertion — as its own comment says, the mutation lands on the next io turn. So when the io thread is inside the very handler that completes discovery:

```
user thread   is_any_system_connected() -> false, insertion posted
io thread     (still in the earlier handler) system connects
io thread     notify_on_discover() -> iterates a list without our callback
io thread     the posted insertion finally lands
```

The notification is gone, nothing sends another, and the subscriber waits forever for a system that is already connected.

Posting the catch-up check puts it behind the insertion in the same post FIFO, so it cannot observe a connection established while the subscription was in flight.

The race predates this branch, but it needs a system to connect inside that small window, and the heartbeat nudge in this PR made it far more likely: discovery now lands right about when a test subscribes.

Note the failure was intermittent — a re-run of the same job passed — so this is reasoning plus a green suite rather than a reproduction. 427/427 unit and 102/102 system tests.

Separately, the Debian 12 segfault on the earlier run of this PR was a *different* problem: heap corruption, reached through `Camera::Camera()` -> `operator new`. I could not reproduce it (40 ASan iterations of the camera tests, 25 RelWithDebInfo runs, 20 pinned to a single core, all clean) and the re-run passed. #3055 adds CI instrumentation so the next occurrence prints a usable stack instead of a bare "Segmentation fault (core dumped)".

PR #3054, opened 2026-08-30
core: drive the work queues from events instead of polling

julianoes:
The follow-up promised in #3052. Stacked on it.

The protocol handlers were driven by two fixed-cadence timers that ran whether or not anything was queued:

- `ServerComponentImpl::do_work()` every 10 ms
- `SystemImpl`'s chain every 10 ms connected / 100 ms otherwise — command sender, timesync, mission transfer, FTP and parameter clients, plus the 5 s ping

Both are gone.

### It was mostly already event-driven

This turned out much smaller than expected. Every enqueue path already did `if (was_empty) do_work();` inside its posted lambda, and the FTP client, the parameter client and the parameter server already posted `do_work()` wherever they retired an item. Only two things actually needed the poll:

**`MavlinkMissionTransferClient` / `...Server`** only retire the front item once it reports `is_done()`, and nothing told them when that became true — the 10 ms poll was what noticed. `WorkItem::set_done()` now notifies the owner.

Two things worth knowing about that notification, both commented at the definition:
- it does **not** take `_mutex`, because every caller reaches it through `callback_and_reset()`, which already holds it (I got this wrong first and deadlocked the mission-transfer unit tests)
- it must only **post**, never call `do_work()` inline — retiring the item drops the last reference to the object we are standing on

**`MavlinkCommandSender::do_work()`** walks the whole queue rather than just the front item, and skips a command while another with the same command id is in flight. So its enqueue kick is now unconditional, and `receive_command_ack()` / `receive_timeout()` post one when they retire an item that may have been blocking another.

### The genuinely periodic bits

Timesync and the ping were the only real periodic work in those loops, and they are now `CallEveryHandler` entries — which are deadline-driven after #3052, so they cost nothing while idle. `Timesync::enable()` keeps its old idempotence explicitly, since `System::enable_timesync()` is public API and registering the tick twice would double the rate.

### Measured

An idle instance holding one unused `udpin` connection, counting context switches across all threads over 10 s:

| | wakeups/s |
|---|---|
| `async-resolve` (before this series) | 291.4 |
| deadline-driven timers (#3052) | 108.5 |
| this PR | 10.3 |

The residual 10/s is `MAX_TIMERS_POLL_WAIT`, the 100 ms safety cap on the timer poll from #3052. If we end up trusting the wakeup contract, raising that cap is where the next order of magnitude is.

### The risk, stated plainly

There is no periodic sweep left to cover for a missing kick. If a new work queue — or a new path that retires an item — forgets to run `do_work()`, the symptom is work sitting in the queue forever rather than being a bit late. `THREADING.md` says so explicitly.

I considered keeping a slow (1 Hz) safety-net poll so a missed kick degrades to latency instead of a stall. I left it out because a hidden 1 s stall is worse to debug than an obvious hang, but I'm happy to add it if you'd rather have the belt and braces.

### Testing

427/427 unit tests, 102/102 system tests. Suite runtime unchanged (258 s vs 258 s on the branch below), which is the main signal that nothing quietly fell back to a slower path.

### DIFF
--- a/cpp/src/mavsdk/core/THREADING.md
+++ b/cpp/src/mavsdk/core/THREADING.md
@@ -83,17 +83,35 @@ Everything time-based is driven from the io thread:
   when they move the earliest deadline earlier, which re-arms from whatever thread they were
   called on. The wait is capped at `MAX_TIMERS_POLL_WAIT` so that a wakeup we somehow miss
   can only make a timer late, never stall it.
-- `MavsdkImpl::schedule_do_work()` — every 10 ms, `ServerComponentImpl::do_work()` for every
-  server component.
-- `SystemImpl::schedule_system_work()` — every 10 ms when connected, 100 ms otherwise; runs
-  `do_work()` on the command sender, timesync, mission transfer, FTP and parameter clients,
-  plus the 5 s ping.
+- `SystemImpl::schedule_ping()` — the 5 s ping, as a `CallEveryHandler` entry.
+- `Timesync` registers its own `CallEveryHandler` entry in `enable()`.
 
-The two `do_work()` chains are still fixed-cadence polls that run whether or not their work
-queues have anything in them. Making them event-driven means giving every work queue
-(command sender, parameter clients, mission transfer, FTP) a "something was enqueued" signal
-and its own retry deadline, which is a larger change than the timer one above. Known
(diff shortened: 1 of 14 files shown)


julianoes (commented):
Okidok.

PR #3055, opened 2026-08-30
CI: print a symbolized backtrace for core dumps from failing tests

julianoes:
Independent of the threading stack (#3049–#3054), based on `main` so it can go in on its own.

### Why

A crash in CI currently looks like this, and nothing more:

```
[ RUN      ] Camera.DefinitionUncompressed
...
##[error]Process completed with exit code 139.
```

The core and the test binaries *are* uploaded as artifacts, which is good — but turning them into a stack means downloading both, reconstructing the container's libraries into a sysroot (the core references paths and glibc/libstdc++ versions the host does not have), and driving gdb by hand. I just did that for a Debian 12 failure and it took a while.

### What

On failure, install gdb and print `thread apply all bt full` for every `/tmp/core.*` the run left behind. Artifacts are still uploaded for deeper digging; this makes the common case readable straight from the log.

The script also dumps the crashing thread's registers and raw stack, which covers the case gdb's unwinder cannot. When a crash is a jump through a corrupted pointer, frame #0 is `?? ()` and there is nothing to unwind from — but the return addresses are still sitting on the stack, and gdb symbolizes them in the `x/64a` output:

```
#0  0x000000025912aae4 in ?? ()
=== crashing thread: registers and raw stack ===
0x7ffe132ccbe0: ... <mavsdk::Camera::Camera(std::shared_ptr<mavsdk::System>)+88>
0x7ffe132ccc10: ... <Camera_DefinitionUncompressed_Test::TestBody()+1910>
```

That is what identified the recent Debian 12 crash as heap corruption reached through `Camera::Camera()` → `operator new`, rather than a fault in the constructor itself. Without it there is no way to tell those apart from the log.

### MALLOC_PERTURB_

Also set for the unit and system test runs in the same two jobs. glibc then fills freed memory with a fixed byte pattern, so a use-after-free read tends to fail loudly and near the bug instead of quietly working most of the time.

It costs nothing at build time and, unlike `MALLOC_CHECK_`, does not change the heap layout — so it should not hide the very corruption it is meant to surface. It may well surface pre-existing latent issues; that is the point, but worth knowing before merging.

### Scope

Only the two jobs that already collect core dumps are covered (`ubuntu-superbuild` and `deb-package`). Enabling it elsewhere is a matter of adding the same "configure core dumps" step those two have — happy to widen it if you want more chances to catch a rare crash.

### Testing

Ran the script against the actual core dump and binaries from https://github.com/mavlink/MAVSDK/actions/runs/33327086338/job/99299064821, and it reproduces the stack above. Also checked the no-core-dumps path exits 0 quietly, so the step is harmless on a non-crash failure.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -295,11 +295,18 @@ jobs:
         run: cmake --build examples/build -j$(nproc)
       - name: unit tests
         run: ./build/release/src/unit_tests/unit_tests_runner
+        env:
+          MALLOC_PERTURB_: "165"
       - name: system tests
         run: ./build/release/src/system_tests/system_tests_runner
         timeout-minutes: 15
+        env:
+          MALLOC_PERTURB_: "165"
       - name: test (mavsdk_server)
         run: ./build/release/src/mavsdk_server/test/unit_tests_mavsdk_server
+      - name: analyze core dumps
+        if: failure()
+        run: tools/analyze_core_dumps.sh build/release
       - name: upload core dumps
         if: failure()
         uses: actions/upload-artifact@v7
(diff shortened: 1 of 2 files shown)


julianoes (commented):
Nice nice.

PR #3056, opened 2026-08-31
core: stop the io thread calling into plugins user code is destroying

julianoes:
Stacked on #3054. This is the remaining structural hole from the threading review: the one place where the io thread holds a bare pointer into memory that *user code* owns.

### The problem

`SystemImpl` keeps plugins as `std::vector<PluginImplBase*>`. They are owned by user code (`Telemetry telemetry{system};`) and destroyed on whatever thread the user drops them on — while the io thread calls `enable()` on every entry when the system connects and `disable()` when it times out. Nothing serialized those two things.

### Reproducing it first

`system_tests/plugin_lifetime_churn.cpp` creates and destroys a `Telemetry` plugin in a loop while a second thread flaps the link — injected heartbeats over a `raw://` connection plus a 50 ms heartbeat timeout give ~46 connect/disconnect transitions across ~500 plugin lifetimes in five seconds.

Against the unfixed code, ThreadSanitizer reports two races on `TelemetryImpl::_homepos_cookie`:

1. `register_plugin()` calling `enable()` **on the caller's thread**, against the io thread's `set_disconnected() → disable()`
2. the io thread's `set_connected() → enable()`, against the **main thread inside `~TelemetryImpl → unregister_plugin()`**

The second is the use-after-free, caught one step before it lands.

### The fix

`_plugin_impls`, and every `enable()`/`disable()` call, now belong to the io thread. The two ends are deliberately asymmetric:

- **`register_plugin()` posts** the insertion and the `if (_connected) enable()` that goes with it. Posting rather than waiting means a plugin constructor can never block on the io thread; the only cost is that `enable()` lands an io turn later. Testing `_connected` *there* rather than on the caller's thread also makes it atomic against `set_connected()`/`set_disconnected()` — done on the caller's thread, a plugin registering just as the system connects could be enabled twice (by both paths) or not at all (by neither). Neither of those is a crash, which is why they'd have been unpleasant to find.
- **`unregister_plugin()` posts the removal and waits**, then calls `disable()`/`deinit()`. Once the removal has run, the io thread cannot reach the plugin and any `enable()`/`disable()` it had started has finished, because the removal is serialized behind them on that same thread.

`set_connected()`/`set_disconnected()` then iterate directly, so `_plugin_impls_mutex` goes away — along with the `_plugin_impls_mutex → plugin-lock` nesting `set_disconnected()` used to create.

No plugin signatures change, so none of the 38 plugins needed editing. `ServerComponentImpl::register_plugin` keeps no registry at all (it just calls `init()`/`deinit()`), so the server side is unaffected.

### Worth a second opinion

- I chose **post** over **post-and-wait** for registration, which means a plugin is enabled one io turn after its constructor returns rather than synchronously. I checked that no plugin's `enable()` body blocks, so either variant is deadlock-free today; posting removes the possibility that a future one could. If you'd rather keep construction strictly synchronous, waiting is a one-line change.
- `unregister_plugin()` now waits on the io thread. That is the same rule as every other `*_blocking` in the codebase — the caller must not hold a lock the io thread needs — and plugin destructors are already subject to it via `deinit()`. But it moves the wait earlier, before `disable()`.
- Ordering changed from disable-then-erase to erase-then-disable. I believe that is strictly more correct, but it is a behaviour change.

### Testing

- New test: **2 TSan races before, 0 after**.
- Full system suite under TSan: **103/103, 0 warnings**.
- Plain: 103/103 system, 427/427 unit.

`core/THREADING.md` gains a "Plugin lifetime" section explaining why the asymmetry exists, so the next person doesn't "simplify" it back.

### DIFF
--- a/cpp/src/mavsdk/core/THREADING.md
+++ b/cpp/src/mavsdk/core/THREADING.md
@@ -31,6 +31,9 @@ That is what removes most of the locking. Concretely:
   already on the io thread.
 - The work queues in `MavlinkCommandSender`, the mission transfer classes, the FTP client and
   the parameter clients are io-thread-only; enqueuing from elsewhere is posted.
+- `SystemImpl::_plugin_impls` is io-thread-only, and so is every `enable()`/`disable()` call.
+  See **Plugin lifetime** below -- this one is load-bearing, because the entries are raw
+  pointers to objects that user code owns and destroys.
 
 `MavlinkMessageHandler` and `MavlinkParameterSubscription` each have a debug-only
 `note_*_thread()` that records the accessing `std::thread::id` and asserts if it ever changes.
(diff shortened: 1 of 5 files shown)


julianoes (commented):
I trust the tests.

PR #3057, opened 2026-08-31
core: abort with an explanation when something outlives its Mavsdk

julianoes:
Stacked on #3056. This is hole 2 from the threading review: `System`, `ServerComponent` and every plugin hold a raw reference back into `MavsdkImpl`, nothing said so, and nothing checked.

### Why abort rather than just log

I'd normally argue against a library killing the host process. The argument doesn't hold here, and the reason is specific:

`~SystemImpl` reaches back into `MavsdkImpl` *by itself* — `remove_call_every_blocking()`, `unregister_timeout_handler_blocking()`. So this isn't only about *using* a leaked `System`; merely letting one be destroyed after its `Mavsdk` is already a use-after-free. There is no "held it but never touched it" case that works today.

That means aborting breaks no working program. It only converts programs that would have segfaulted somewhere unrelated later into ones that stop immediately, at the point where we can still name the mistake:

```
Mavsdk is being destroyed while things it handed out are still alive:
  - 1 System(s) still referenced (from systems() or first_autopilot())
A System, a ServerComponent or a plugin must not outlive the Mavsdk instance.
They hold a reference back into it, so using one afterwards -- or merely letting it
be destroyed -- is a use-after-free.
Keep the Mavsdk instance alive at least as long as everything obtained from it.
Aborting now: continuing would crash later and much less clearly.
```

### The false positive this nearly shipped with

The first run tripped on `Mavsdk.version` in the unit tests. `_default_server_component` points at an entry that is *also* in `_server_components`, so `use_count()` is legitimately 2 with zero user references. The check now compares against a baseline that accounts for our own references.

Worth stating plainly: had I gone straight to `abort()` without running the suite, this would have killed every MAVSDK program on shutdown. The whole system suite now runs clean through the check across ~100 different construction shapes, which is the real evidence the baseline is right.

### Testing both directions

- `system_tests/mavsdk_outlived.cpp` proves the check *fires*, for a leaked `System` and a leaked `ServerComponent`. Death tests in `threadsafe` style (forking a threaded program isn't safe), and no `ASSERT_*` inside the death statement — that expands to a `return`, which gtest rejects outright.
- The rest of the suite proves it *doesn't* fire when it shouldn't, which for a check like this is the more important half.

105/105 system tests, 105/105 under TSan with 0 warnings, 427/427 unit.

### Documentation

The rule is now on `systems()`, `first_autopilot()` and `server_component()` in the public header, plus an "Object lifetime" section in `core/THREADING.md`.

### Incidental find, not fixed here

Getting the System death test working turned up that `SystemImpl::init()` registers its HEARTBEAT handler with a *posted* registration — so the very first heartbeat, the one that creates the system, arrives before that system's own handler is in the table and doesn't mark it connected. The second one does.

Harmless at 1 Hz and long-standing, but it means discovery is structurally one heartbeat slower than it looks. Fixing it would mean making that registration synchronous, which cuts against the posted-mutation invariant, so I left it alone — happy to look at it separately if you think it's worth it.

### DIFF
--- a/cpp/src/mavsdk/core/THREADING.md
+++ b/cpp/src/mavsdk/core/THREADING.md
@@ -144,6 +144,21 @@ use-after-free waiting to happen.
 **If you add anything else the io thread reaches through a raw pointer into user-owned memory,
 it needs the same treatment.**
 
+## Object lifetime
+
+`System`, `ServerComponent` and every plugin hold a reference back into `MavsdkImpl`, so none
+of them may outlive the `Mavsdk` instance they came from. That is not just a rule about
+*using* them afterwards: `~SystemImpl` itself reaches back in (`remove_call_every_blocking()`,
+`unregister_timeout_handler_blocking()`), so a leaked `System` is a use-after-free even if it
+is never touched again.
+
+Since there is no harmless version of getting it wrong, `~MavsdkImpl` looks for outstanding
+references and aborts with an explanation rather than leaving a segfault to surface elsewhere
+later — see `MavsdkImpl::abort_if_references_outlive_us()` and
+`system_tests/mavsdk_outlived.cpp`. Note that a couple of those references are our own
+(`_default_server_component` points at an entry that is also in `_server_components`), so the
+check compares against a baseline rather than against 1.
+
 ## Locks that remain
 
 Locks are for state genuinely reachable from user threads. The order is documented where the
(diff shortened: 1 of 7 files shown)


julianoes (commented):
Makes sense, better cleanly abort rather than segfault.

PR #3058, opened 2026-08-31
core: deliver the first heartbeat to the system it just created

julianoes:
Stacked on #3057. This is the incidental find from that PR, fixed.

### The problem

`SystemImpl::init()` runs on the io thread, from `MavsdkImpl::process_message()`, while that is handling the very message which caused the system to be created. Its handler registrations went through the usual **posted** `register_one()`, so they landed an io turn later — after the current message had already been dispatched.

The heartbeat that discovered the system was therefore never delivered *to* that system. Nothing called `set_connected()`, and discovery waited for the next heartbeat: a full interval, typically a second, for nothing.

### The fix

Register the three handlers synchronously. `register_one_on_io_thread()` mirrors the `unregister_all_on_io_thread()` that already exists for the same reason in reverse, and carries the same two constraints, asserted and documented:

- must run on the io thread
- never from inside a message callback, since growing the table underneath `process_message()`'s loop would invalidate its iteration

Neither is a problem here: there is exactly one `System` construction site and it is on the io thread, and the table being written belongs to the system being created, so nothing is iterating it yet.

### Testing

`system_tests/first_heartbeat.cpp` injects exactly one heartbeat and expects a connected system. **Without the fix it times out at 2087 ms; with it, it passes in 85 ms.**

The effect is visible across the whole suite too — with discovery no longer costing an extra heartbeat interval every time:

| | system tests |
|---|---|
| before | ~260 s (105 tests) |
| after | **173 s** (106 tests) |

106/106 plain, 106/106 under TSan with **0 warnings**, 427/427 unit.

The TSan run is the one that matters here, since this replaces a posted table mutation with a direct one — exactly what the posted design exists to avoid. The `_processing` assert and `note_table_thread()` cover the constraints, and the sanitizer agrees.

### DIFF
--- a/cpp/src/mavsdk/core/mavlink_message_handler.cpp
+++ b/cpp/src/mavsdk/core/mavlink_message_handler.cpp
@@ -43,6 +43,15 @@ void MavlinkMessageHandler::register_one_impl(
     });
 }
 
+void MavlinkMessageHandler::register_one_on_io_thread(
+    uint16_t msg_id, const Callback& callback, const void* cookie)
+{
+    // See the header for why this exists and when it is allowed.
+    assert(!_processing);
+    note_table_thread();
+    _table.push_back(Entry{msg_id, {}, callback, cookie});
+}
+
 void MavlinkMessageHandler::unregister_one(uint16_t msg_id, const void* cookie)
 {
     unregister_impl({msg_id}, cookie);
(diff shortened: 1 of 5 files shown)


julianoes (commented):
Seems like a sensible optimization, and it helps running the system tests quicker.

PR #3059, opened 2026-08-31
core: never drop the user callbacks somebody is waiting for

julianoes:
Stacked on #3058. This is hole 3 from the threading review — and investigating it turned out to make it more serious than "lossy by design".

### It's a liveness bug, not just dropped telemetry

Everything the user sees arrives through one bounded queue, and **blocking API calls depend on it**:

```cpp
Action::Result ActionImpl::arm() const
{
    auto prom = std::promise<Action::Result>();
    auto fut = prom.get_future();
    arm_async([&prom](Action::Result result) { prom.set_value(result); });
    return fut.get();                                  // no timeout
}
```

`arm_async` → `send_command_async` → `MavlinkCommandSender::call_callback` → `call_user_callback` → **the bounded queue**. So once the queue was full, the command result was dropped, the promise was never satisfied, and `arm()` never returned.

And the queue overflows precisely when the application is busy — which is when it's most likely to be issuing commands.

`system_tests/callback_queue_overflow.cpp` reproduces it with a deliberately slow position subscriber, a publisher keeping the queue saturated, and `arm()` on its own thread so a regression fails rather than hangs the suite:

| | result |
|---|---|
| before | `arm() never returned: its result callback was dropped from the user callback queue` |
| after | 2104 positions published, 111 delivered, **`arm()` succeeded** |

### Two classes

| | Use | Dropped? |
|---|---|---|
| `call_user_callback()` | Anything somebody may be waiting for: command results, mission and parameter completions. | **Never** — one per user action, so they can't run the queue away. |
| `call_user_callback_droppable()` | A subscription delivering a stream, where the newest value is what matters. | Yes, past the bound. |

Must-deliver is the **default**, so the ~230 existing call sites are safe without being touched and only high-rate paths opt in. In Telemetry that's 37 stream deliveries marked droppable, leaving the two `get_gps_global_origin` one-shots as must-deliver.

When the bound is reached the **oldest** droppable entry is evicted rather than the new one refused, so a struggling subscriber sees fresh data with gaps instead of falling steadily further behind. If the queue somehow holds nothing droppable, the new stream value gives way rather than displacing a completion.

### Drops are now visible

Counted, and reported at most every five seconds with a running total.

The old logging did the opposite of what was useful: a warning on **every** enqueue once the queue passed ten (up to eighty-nine lines in a burst), a single error at exactly ninety-nine, then silent dropping from a hundred. Loudest before anything was lost, quiet once it was.

### Testing

107/107 system tests, 107/107 under TSan with **0 warnings**, 427/427 unit.

### Deliberately not included

A public getter for the drop count. It would be genuinely useful, but it is a public API commitment, so it seemed better to ask than to slip it in — say the word and I'll add it.

Also worth its own conversation: `fut.get()` with no timeout means *any* reason a callback fails to arrive is a permanent hang, not just this one. This PR removes the reason we know about; it doesn't make the pattern robust.

### DIFF
--- a/cpp/src/mavsdk/core/THREADING.md
+++ b/cpp/src/mavsdk/core/THREADING.md
@@ -144,6 +144,31 @@ use-after-free waiting to happen.
 **If you add anything else the io thread reaches through a raw pointer into user-owned memory,
 it needs the same treatment.**
 
+## User callbacks: two classes
+
+Everything the user sees arrives through one bounded queue drained by the user-callback
+thread, and *blocking API calls depend on it*. `Action::arm()` waits on a promise that is
+only satisfied when the command-result callback runs, so a result that gets dropped is not a
+gap in data -- it is `arm()` never returning.
+
+So the queue distinguishes two kinds of work:
+
+| | Use | Dropped? |
+|---|---|---|
+| `call_user_callback()` | Anything somebody may be waiting for: command results, mission and parameter completions, one-shot requests. | **Never.** These are one per user action, so they cannot run the queue away. |
+| `call_user_callback_droppable()` | A subscription delivering a stream, where the newest value is what matters. | Yes, once `MAX_DROPPABLE_USER_CALLBACKS` of them are queued. |
+
+Must-deliver is the default, so a new call site is safe until somebody deliberately opts into
+dropping. When the droppable bound is reached, the **oldest** droppable entry is evicted
+rather than the new one refused, so a subscriber that cannot keep up sees fresh data with
+gaps instead of falling further and further behind.
+
+Drops are counted and reported at most every few seconds with a running total, rather than
+per drop -- whatever causes drops causes a great many of them.
+
+`system_tests/callback_queue_overflow.cpp` pins the property that matters: a slow subscriber
+must not stop a blocking call from returning.
+
 ## Object lifetime
 
 `System`, `ServerComponent` and every plugin hold a reference back into `MavsdkImpl`, so none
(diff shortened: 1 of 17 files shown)


julianoes:
Added `022b6dffc`, following the suggestion that progress callbacks should be droppable too.

The first pass marked Telemetry's streams droppable and left everything else must-deliver. That was the safe default, but it left progress updates on the must-deliver path — and they are streams as much as telemetry is. A large FTP download emits them constantly, so a subscriber falling behind could pin the queue with progress reports: the exact failure this PR set out to prevent, reached by a different route.

The distinction turns out to be **per delivery, not per callback**:

- **Progress-only lambdas** — Mission and MissionRaw report progress through a *separate* callback from the one carrying the result, so those are simply droppable.
- **Mixed paths** — Calibration, FTP and LogFiles deliver both through one path, so those sites choose at runtime on `Result::Next`. That is the API's own notion of an intermediate update, not one I invented: `MavlinkFtpClient::ClientResult::Next` is documented as *"Intermediate message showing progress"*.

### The one that looks like progress and isn't

`MissionImpl`'s download completion passes `Mission::Result::Next` with the downloaded mission attached, and then calls the callback **again with the final result from the same queued lambda**. Marking that droppable would hang `download_mission()` — precisely what this PR fixes. It stays must-deliver, and it is worth a look during review since it is the trap in this change.

`CameraImpl` needed nothing: it returns early on `ClientResult::Next` and never forwards FTP progress to a user callback at all.

107/107 system tests (30 of them covering the FTP, mission, log-files and calibration paths touched here), 427/427 unit.

PR #3063, opened 2026-09-01
Generate Requires.private only for static pkg-config installs

### LINKED ISSUE #3018: mavsdk.pc forces bundled private dependencies on shared-library consumers
### Problem

The generated `mavsdk.pc` declares implementation dependencies as pkg-config private requirements:

```pkgconfig
Requires.private: libcurl jsoncpp tinyxml2
```

(Current `main` uses `nlohmann_json` instead of `jsoncpp`, but retains the same structure.)

For the prebuilt shared-library packages, these dependencies are bundled into `libmavsdk.so`. They are not exposed through MAVSDK's installed public headers and are not dynamic dependencies of the installed shared object. Nevertheless, pkg-config attempts to resolve the private modules when compiler flags are requested.

On a clean Debian 13 ARM64 system with MAVSDK 3.17.2 installed:

```console
$ pkg-config --modversion mavsdk
3.17.2

$ pkg-config --libs mavsdk
-lmavsdk

$ pkg-config --cflags mavsdk
Package libcurl was not found in the pkg-config search path.
Package 'libcurl', required by 'mavsdk', not found
Package 'jsoncpp', required by 'mavsdk', not found
Package 'tinyxml2', required by 'mavsdk', not found
```

This prevents the conventional command from working:

```sh
c++ example.cpp $(pkg-config --cflags --libs mavsdk)
```

CMake consumers using `find_package(MAVSDK)` build and run successfully on the same system.

### Package inspection

I reproduced this while producing and testing a Debian 13 ARM64 build from the v3.17.2 tag, and also confirmed that the official v3.17.2 Debian 13 AMD64 package contains the same `Requires.private` line.

For the ARM64 shared object:

```console
$ readelf -d /usr/lib/libmavsdk.so.3.17.2 | grep NEEDED
Shared library: [libstdc++.so.6]
Shared library: [libm.so.6]
Shared library: [libgcc_s.so.1]
Shared library: [libc.so.6]
```

The installed MAVSDK public headers do not include curl, jsoncpp, or tinyxml2 headers. The package supplies the shared library and no static `libmavsdk.a`.

### Expected behavior

Installing the shared-library development package should be sufficient for:

```sh
pkg-config --cflags --libs mavsdk
```

Private implementation dependencies already incorporated into `libmavsdk.so` should not require their development packages on the consumer system.

### Suggested fix

Remove `Requires.private` from `mavsdk.pc` for shared-library builds/packages, or generate it conditionally only for a static installation whose final consumer genuinely needs those external archives.

This is related to #2475, which covers the separate hard-coded CI-prefix problem in the generated pkg-config file.

yurekami:
## Summary

- emit `Requires.private` only for static MAVSDK builds
- omit curl from static metadata when `BUILD_WITHOUT_CURL=ON`
- add a CMake-only validator wired into CTest

## Why

Shared-library `mavsdk.pc` files advertised bundled implementation dependencies, forcing downstream `pkg-config` consumers to resolve packages they do not link directly. Static consumers still need those private dependencies. This fixes #3018.

## Verification

- shared valid/invalid metadata cases
- static with curl and static without curl cases
- validator path containing spaces
- `git show --check`

The full third-party superbuild was not completed; the focused install-metadata matrix passed.

Fixes #3018

### DIFF
--- a/cpp/src/CMakeLists.txt
+++ b/cpp/src/CMakeLists.txt
@@ -109,8 +109,29 @@ configure_file(MAVSDKConfig.cmake.in
     "${PROJECT_BINARY_DIR}${CMAKE_FILES_DIRECTORY}/MAVSDKConfig.cmake" @ONLY)
 
 # Pkg-config
+set(MAVSDK_PC_REQUIRES_PRIVATE "")
+if(NOT BUILD_SHARED_LIBS)
+    set(_mavsdk_pc_requires_private "nlohmann_json tinyxml2")
+    if(NOT BUILD_WITHOUT_CURL)
+        set(_mavsdk_pc_requires_private "libcurl ${_mavsdk_pc_requires_private}")
+    endif()
+
+    set(MAVSDK_PC_REQUIRES_PRIVATE "Requires.private: ${_mavsdk_pc_requires_private}")
+endif()
+
 configure_file(mavsdk.pc.in
     "${PROJECT_BINARY_DIR}/mavsdk.pc" @ONLY)
+
+if(BUILD_TESTING)
+    add_test(
+        NAME mavsdk_pkg_config_metadata
+        COMMAND ${CMAKE_COMMAND}
+            -DBUILD_SHARED_LIBS=${BUILD_SHARED_LIBS}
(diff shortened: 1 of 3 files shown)


julianoes:
Why did you close it?

PR #3073, opened 2026-09-11
Reduce a vehicle-supplied camera definition path to a file name

nicholasaleks:
cam_definition_uri arrives in CAMERA_INFORMATION from whatever is on the MAVLink link, and the handler is registered unconditionally at construction, so it is acted on with no operator involvement. For an mftp:// or mavlinkftp:// uri, strip_prefix() removes the scheme and nothing else, so "../" survives, and std::filesystem::operator/ does not normalise a right-hand side that starts with "..". _tmp_download_path / downloaded_filename therefore resolves outside the download directory.

That path is then used for real work: std::filesystem::remove() on the .xz branch, and _file_cache->insert(), which renames the file into the cache for any extension. A uri of mftp://../../root/Documents/mission.plan is enough to move an unrelated file out of existence.

It is also simply the wrong path. MavlinkFtpClient::download_start() writes to local_folder / fs::path(remote_path).filename(), so the downloaded file is at the basename, not at the traversed path this code rebuilds. Today any uri with a directory component already fails to load for that reason. Applying .filename() here makes the two agree and keeps the result inside _tmp_download_path.

Related to: https://github.com/nicholasaleks/infected-drones/tree/main/MAVSDK-02_camera_cam_definition_ssrf_traversal


julianoes:
I think you're right! Thanks.

julianoes (dismissed):

julianoes:
Please run `cpp/tools/fix_style.py cpp`.

Edit: done.

nicholasaleks:
Is this good for approval / merge?

PR #3074, opened 2026-09-11
Bound the decompressed size of an .xz camera definition

nicholasaleks:
inflateLZMAFile() is reached with input that comes off the MAVLink link. A CAMERA_INFORMATION cam_definition_uri ending in .xz is fetched and handed straight to the decoder (camera_impl.cpp, both the http and the mftp branch), and so is a COMPONENT_METADATA uri ending in .xz or .lzma (mavlink_component_metadata.cpp). None of that needs operator involvement: a heartbeat from an unknown component id is enough to make MAVSDK ask for CAMERA_INFORMATION by itself.

The decoder is initialised with a memory limit of UINT64_MAX and the write loop keeps no running total, so the amount written is whatever the sender's stream decodes to. 156 KB of .xz expands to 1 GiB on disk at preset 9e, and the result is then moved into the file cache, whose only bound is 50 entries.

Cap the output at 64 MiB, which is three orders of magnitude above any real camera definition or component metadata document, and delete the partial file when the decode fails. Without that second part a failed decode leaves its output behind: neither this function nor the error branches in the callers remove it, so a stream that is truncated after a gigabyte still costs a gigabyte, permanently.

Related to: https://github.com/nicholasaleks/infected-drones/tree/main/MAVSDK-03_lzma_decompression_bomb


julianoes (approved):
Good idea, thanks!

PR #3081, opened 2026-09-14
fix: correct UTF-8 filename handling for FTP server on Windows

bansiesta:
## Summary

Fixes the pre-existing Windows CI failure in `FtpUnicode.NameTooLongIsRejected` and `FtpUnicode.ListDirectoryWithNameTooLongForPayload` (confirmed failing on `main` independent of any open PR — not something introduced by #3078, which is where this was first noticed).

**Root cause:** `std::filesystem::path` constructed from / converted to a plain `std::string` uses the OS-native narrow encoding — UTF-8 on POSIX, but the ANSI codepage on Windows. FTP filenames arrive UTF-8 on the wire, so on Windows this mangled multi-byte sequences (e.g. the emoji used in the test fixtures). POSIX builds never surface this since narrow already equals UTF-8 there.

**Fix:** round-trip filenames through `std::u8string`/`char8_t` (project targets C++20) instead of the implicit narrow conversion, at every point a filename crosses the wire-protocol boundary in `mavlink_ftp_server.cpp` (path construction/decomposition in list/open/create/rename/remove/CRC handlers). This is a no-op on POSIX, so no `#ifdef` needed. Applied the identical fix to the `ftp_unicode.cpp` test fixture, which had the same bug in its own local-disk path construction.

Diagnostic-only `.string()` calls used purely for log messages, and the FTP-client-plugin-boundary local-path arguments in the test fixture (`upload`/`download`/`are_files_identical`), were left unchanged as out of scope for this fix.

## Test plan

- [x] Built `mavsdk` core library and `system_tests_runner` cleanly on Linux (g++ 12, C++20).
- [x] Ran the full `Ftp*` system test suite locally (29 tests) — all pass, including the two tests this PR targets. (Bug doesn't reproduce on POSIX, but this confirms no regression from the refactor.)
- [ ] Windows CI (no Windows/MSVC toolchain available in this environment to verify directly — relying on this PR's CI run).


julianoes:
Looks like the openssl download is broken, ignoring that.

PR #3084, opened 2026-09-15
mavlink_direct: reject out-of-range target system/component IDs

bansiesta:
## Summary
- `MavlinkMessage.target_system_id`/`target_component_id` are `uint32` in the API, but MAVLink system/component IDs are still 1 byte on the wire until sysid32 support lands. Values over 255 were silently truncated by `static_cast<uint8_t>` (e.g. 256 -> 0) instead of being rejected.
- Both `send_message()` paths (`mavlink_direct` client, `mavlink_direct_server`) now return `Result::InvalidField` and log an error for out-of-range target IDs, instead of truncating silently.

## Test plan
- [x] New system test `MavlinkDirect.TargetIdsOutOfRangeRejected` covers both the client and server send paths (out-of-range rejected, in-range still succeeds).
- [x] Full local `MavlinkDirect*`/`MavlinkDirectServer*` suite (21 tests) passes.
- [x] `clang-format` clean on all changed files.


bansiesta:
Folded into #3082 (this commit is now part of that PR).

PR #3086, opened 2026-09-15
system_tests: parse raw bytes with our own buffers in RawBytes test

julianoes:
## Problem

`RawBytes.SendReceive` parses the bytes from `subscribe_raw_bytes_to_be_sent()` with `mavlink_parse_char(MAVLINK_COMM_0, ...)`. That uses the global receive buffer and status of channel 0.

The callback runs on the io thread, while the autopilot's `ServerComponentImpl` checks out a channel in its constructor -- the first free one, 0 -- and packs its messages on it, e.g. `publish_position()` from the test thread. So the test and the library modify the same channel status concurrently. `heartbeat_watchdog.cpp` already works around the same problem for packing with `TestMavlinkPackChannel`.

Found while reviewing the system tests for #3085.

## Fix

Each raw bytes callback carries one whole message, so parse with buffers local to the callback via `mavlink_frame_char_buffer()`, the same way `MavlinkReceiver` does. No channel state is touched any more.

Note: this touches the same lambda as #3085, so whichever merges second needs a small rebase.

## Test plan

- [x] `RawBytes.*` and `HeartbeatWatchdog.*` pass locally


julianoes (commented):
Makes sense!

PR #3088, opened 2026-09-15
generate: fix repeated/message param handling in Python binding templates

### LINKED ISSUE #3087: Auto-generated Python wrappers for MissionRaw are incorrect
I've attempted to use MissionRaw from both `py/mavsdk` and `py/aiomavsdk` and found the following issues:

1. `py/mavsdk` [upload_mission](https://github.com/mavlink/MAVSDK/blob/62bb5cf590e4d069177dad4d65ed25fb13d89f9c/py/mavsdk/mavsdk/plugins/mission_raw/mission_raw.py#L439) is missing a size parameter when calling the underlying C implementation
2. `py/mavsdk` [upload_mission_with_progress_async](https://github.com/mavlink/MAVSDK/blob/62bb5cf590e4d069177dad4d65ed25fb13d89f9c/py/mavsdk/mavsdk/plugins/mission_raw/mission_raw.py#L471) does not convert mission_plan to C struct
3. `py/aiomavsdk` [upload_mission_with_progress](https://github.com/mavlink/MAVSDK/blob/62bb5cf590e4d069177dad4d65ed25fb13d89f9c/py/aiomavsdk/aiomavsdk/plugins/mission_raw/mission_raw.py#L76) calls underlying implementation in `py/mavsdk` with incorrect number of parameters

Did some investigation regarding this, it's related to how types are interrogated to create bindings. Some things are just omitted. Noting this down as an issue, might get some time to resolve this myself.

As a workaround, mission can be uploaded via `upload_mission_with_progress_async` found in `py/aiomavsdk` until this issue is resolved.

bansiesta:
## Summary

Fixes #3087. The bug was in the code-generation templates, not the generated files, per the issue reporter's hunch — I traced it to two gaps in the Jinja2 templates and regenerated all affected bindings via the project's own generation scripts (no hand-edited generated files).

- `py/mavsdk/templates/file.py.j2`: blocking (`is_sync`) calls built C call arguments for repeated (list) params without converting them to a `ctypes` array + separate size argument — this is Bug 1 (`upload_mission` missing size param). Async calls skipped both the array conversion for repeated params **and** `.to_c_struct()` for message-type params — this is Bug 2 (`upload_mission_with_progress_async` not converting `mission_plan`).
- `py/aiomavsdk/templates/file.py.j2`: the `is_finite` async wrapper (used for progress-reporting methods) dropped all method params entirely when forwarding to the underlying sync plugin's `_async` call — this is Bug 3 (`upload_mission_with_progress` in `py/aiomavsdk` calling through with the wrong argument count).

## Scope

Since the fix is in the shared templates, regenerating turned up the same classes of bugs beyond `MissionRaw`:

- `py/mavsdk`: `camera`, `component_metadata_server`, `geofence`, `log_files`, `mission`, `mission_raw`, `tune`
- `py/aiomavsdk`: `ftp`, `log_files`, `mission`, `mission_raw`

All other plugins were regenerated and produced byte-identical output, confirming the fix only touches the affected code paths.

## Test plan

- [x] Regenerated all Python bindings via `py/mavsdk/tools/generate_from_protos.bash` and `py/aiomavsdk/tools/generate_from_protos.bash` (the project's real generation scripts), confirmed via `git diff --stat` that only plugins with repeated or message-type params changed.
- [x] Stubbed the ctypes C library layer and directly invoked `MissionRaw.upload_mission`, `upload_mission_async`, and `upload_mission_with_progress_async`, confirming the argument count/types now match the C header signatures (`mission_raw.h`) exactly.
- [x] Stubbed the same layer for `aiomavsdk`'s `MissionRawAsync.upload_mission_with_progress` and confirmed `mission_plan` is now correctly forwarded through to the underlying C call (previously dropped).
- [ ] No real `libcmavsdk.so` was available in this environment to run the full integration test suite — CI should cover that.


julianoes (approved):

julianoes (approved):

julianoes (approved):
Looks correct.

PR #3089, opened 2026-09-15
py: add end-to-end mission upload system tests

bansiesta:
Stacked on #3088 — please merge that one first.

#3088 fixed the generated Python bindings for `MissionRaw`, but nothing in CI would have caught those bugs in the first place: the generated code imports fine and the signatures look plausible. What was missing is a test that actually puts a message on the wire.

## What this adds

`py/test/` — end-to-end tests that bring up two MAVSDK instances in-process, a ground station and an autopilot linked over local UDP, upload a raw mission from one to the other, and assert **item-by-item on what `MissionRawServer` actually received**. Nothing is mocked. Both wrappers are covered against the same scenario:

| test | exercises |
| --- | --- |
| `test_upload_mission` | `mavsdk`, blocking upload |
| `test_upload_mission_with_progress` | `mavsdk`, progress callback |
| `test_upload_mission_async` | `aiomavsdk`, blocking upload |
| `test_upload_mission_with_progress_async` | `aiomavsdk`, progress stream |

Checked out against the bindings as generated *before* #3088, all four fail — e.g. `TypeError: MissionRaw.upload_mission_with_progress_async() missing 1 required positional argument: 'callback'`. So they are a real regression test for #3087, not just a smoke test.

They're meant to read as a template for further Python tests, so `py/test/README.md` documents the build and the two things the tests do on purpose (register server plugins *before* the connection is added; assert on what the server received, not on the result code). `conftest.py` provides a `connected_pair` fixture that most new tests should be able to reuse. Ports are picked per-test rather than hard-coded. No `pytest-asyncio` dependency — the async tests use `asyncio.run` and an async context manager.

New CI job `Ubuntu 24.04 (Python system tests)` in `linux.yml`. It builds C++ static and the C bindings shared, which makes `libcmavsdk.so` self-contained so the Python loader needs no `LD_LIBRARY_PATH`.

## Two bugs the tests turned up

Both are in separate commits, so they're easy to review independently.

**1. Segfault on the most natural possible usage** (`py: keep the owner alive while a plugin holds it`)

```python
mission_raw = MissionRaw(mavsdk.first_autopilot(10.0))     # segfault on first call
server = MissionRawServer(mavsdk.server_component())       # same
```

`System._plugins` and `ServerComponent._plugins` are weak sets, and `Mavsdk._children` is weak too, so the temporary `System` / `ServerComponent` is collected as soon as the plugin constructor returns. Its `destroy()` then walks the weak plugin set — which still holds the very much alive plugin — and nulls its handle. Fixed by giving the plugin a strong reference back to its owner; the owner's side stays weak, so this isn't a cycle.

**2. `aiomavsdk` server-side subscriptions silently dropped every update** (`generate: pass the result through aiomavsdk stream subscriptions`)

For a subscription that reports a result alongside its data, the sync layer calls `callback(result, data, user_data)` but the generated asyncio wrapper declared `def callback(data, _user_data)`. Every update raised `TypeError` inside the sync wrapper, which logs and moves on, so the async generator simply never yielded. The template now takes `method.has_result` into account and yields `Tuple[<Plugin>Result, T]`. Regeneration touches only the eight server-side subscriptions in `action_server` and `mission_raw_server`; every other plugin comes out byte-for-byte identical.

## One footgun left deliberately alone

Generated `unsubscribe_*` methods don't guard against being called after `destroy()`, so unsubscribing during interpreter shutdown after the `Mavsdk` instance is gone segfaults. The hand-written `Mavsdk.unsubscribe_on_new_system` already does `if not self._handle: return`. I worked around it with teardown ordering in the test rather than changing generated code here, to keep this PR's scope to the tests — happy to do it as a follow-up if you want the guard in the template.


py/test/test_mission_raw_upload.py
```diff
@@ -0,0 +1,218 @@
+"""End-to-end mission upload tests for the Python bindings.
+
+Each test runs a real client/server pair over a local UDP link: one MAVSDK
+instance configured as a ground station uploads a raw mission, another
+configured as an autopilot receives it through ``MissionRawServer``. Both the
+sync (``mavsdk``) and the asyncio (``aiomavsdk``) wrappers are covered, and the
+mission is checked item-by-item on the server side rather than only asserting
+on the upload result.
```
julianoes: Please split the tests up into one test for mavsdk and one for aiomavsdk.

py/test/conftest.py
```diff
@@ -0,0 +1,94 @@
+"""Shared fixtures for the Python system tests.
+
+The tests import ``mavsdk`` and ``aiomavsdk`` straight out of the source tree,
+so that they exercise the bindings as generated in this repo rather than
+whatever happens to be installed. ``libcmavsdk.so`` still has to be built and
+placed in ``py/mavsdk/mavsdk/lib/`` first — see ``py/test/README.md``.
```
julianoes: Hang on, can't you install the dependencies using `pip install -e ...` and then use it normally? I hate to have this complicated extra machinery.

py/test/conftest.py
```diff
@@ -0,0 +1,94 @@
+"""Shared fixtures for the Python system tests.
```
julianoes: I would like to avoid having shared fixtures and instead have the system test standalone, so it is clear and without layers of abstraction added for testing. I don't want  test setup in one file and actual testing in another.

julianoes (changes requested):

julianoes (dismissed):

julianoes (approved):
Fixes end to end tests, so presumably right.

PR #3090, opened 2026-09-15
fix unsubscribe after destroy in the C bindings

bansiesta:
Stacked on #3089 (which is stacked on #3088). Review the last three commits; the ones below belong to those PRs.

#3089 left a footgun in the teardown path, and the Python side was only where it was easy to notice. The actual bug is in the C bindings and affects every consumer of them.

### The bug

`mavsdk_<plugin>_unsubscribe_<method>` took the handle at face value:

```c
auto* cpp_handle = reinterpret_cast<...Handle*>(handle);
{
    std::lock_guard<std::mutex> lock(wrapper->handles_mutex);
    auto& vec = wrapper->..._handles;
    vec.erase(std::remove(vec.begin(), vec.end(), cpp_handle), vec.end());
}
wrapper->cpp_plugin->unsubscribe_...(std::move(*cpp_handle));
delete cpp_handle;
```

`mavsdk_<plugin>_destroy` already unsubscribes and deletes every handle in that vector, so a caller who unsubscribes afterwards — or unsubscribes the same handle twice — dereferences and deletes freed memory. It reproduces as `free(): double free detected in tcache 2`. And since a destroyed wrapper leaves the caller holding a null plugin pointer, the `wrapper->handles_mutex` access on the line before is a null dereference: a plain segfault.

Neither case is exotic. Unsubscribing during teardown is the normal thing to do, and teardown is exactly when the plugin may already be gone.

### The fix

1. **`c/`** — guard both pointers against null, and look the handle up in the wrapper's own vector before touching it. If it is not there, the subscription is already gone and the call is a no-op. Made in `c/templates/file.cpp.j2` and regenerated; the five hand-written unsubscribes in `c/src/cmavsdk/mavsdk.cpp` got the same null guards.
2. **`py/mavsdk`** — `unsubscribe_*` returns early when the plugin handle is null, so the trip into C is not made at all. Template change, regenerated.
3. **`py/test/test_teardown.py`** — four scenarios covering the orders in which Python can drop a handle into a C++ object: unsubscribe after destroy, unsubscribe twice, an async subscription outliving its `Mavsdk` instance, and a plugin collected at interpreter shutdown.

Three of those four crashed before this branch (two segfaults, one double free); the fourth is there because interpreter-shutdown ordering is what makes the other three easy to hit by accident. Because a regression here is a crash rather than a failed assertion, each scenario runs in its own interpreter and is judged by its exit code — otherwise one bad commit takes the whole pytest run down with it.

Verified by reverting the two fix commits, rebuilding, and re-running the suite: 3 failed, 5 passed. With them: 8 passed.


py/mavsdk/templates/file.py.j2
```diff
@@ -343,7 +343,15 @@ class {{ plugin_name.upper_camel_case }}:
 
     {% if method.is_stream and not method.is_finite %}
     def unsubscribe_{{ method.name.lower_snake_case }}(self, handle: ctypes.c_void_p):
-        """Unsubscribe from {{ method.name.lower_snake_case }}"""
+        """Unsubscribe from {{ method.name.lower_snake_case }}
+
+        A no-op once this plugin is destroyed: destroying it already tore every
+        subscription down. Teardown reaches here in that order routinely -- a
+        generator unwinding from a ``finally`` block, or a __del__ running at
+        interpreter shutdown, after the owning Mavsdk instance is gone.
```
julianoes: This is way to verbose. Tone it down please.

julianoes (dismissed):

julianoes (approved):
Fairly defensive but I think ok.

PR #3092, opened 2026-09-16
ci: reduce and fix the third_party build caches

julianoes:
The repo sits at the 10 GB Actions cache limit (9.97 GB across 87 entries), so `third_party` caches are continuously evicted and rebuilt, and a miss means a full superbuild. In run 35029034320 the `macOS 15 framework x64` job spent over 50 minutes in `configure` because its entry had been evicted minutes before the run started, against 11-13 minutes for the arm64 jobs that hit.

### Share the macOS `third_party` cache between framework and non-framework builds

The four macOS jobs keyed on `matrix.name`, so `macOS 15 x64` and `macOS 15 framework x64` each kept a private ~358 MB copy of an identical dependency tree, and likewise for the arm64 pair. `MACOS_FRAMEWORK` is read in three lines of `src/mavsdk_server/src/CMakeLists.txt` and nowhere under `third_party/`, so keying on `matrix.architecture` takes 4 entries down to 2 (~713 MB back) and lets each pair warm the other's cache.

All four jobs stay as they are: `iOS-XCFramework` needs both frameworks for the fat `macos-arm64_x86_64` slice. Both jobs of a pair may race to save the same key; `actions/cache` warns and skips the loser.

### Fix the `hashFiles` path in the kotlin cache keys

Both kotlin keys hashed `./third_party/**`, but it lives at `cpp/third_party`. `hashFiles()` resolves against `GITHUB_WORKSPACE` and ignores `defaults.run.working-directory`, so the glob matched nothing and returned an empty string — the keys were literally `kotlin-linux-x86_64--1`. Those entries never invalidated, so `jni-desktop` and `jni-android` kept restoring a stale dependency tree after `cpp/third_party` changed. Every other workflow already uses `./cpp/third_party/**`. Suffix bumped to `-2`.

### Notes

- The first run after merging is cold on the changed keys, so expect one round of full superbuilds.
- The superseded `macOS-macOS 15 *-3` and `kotlin-*--1` entries become orphans once this lands and can be deleted for ~1.5 GB.
- Not included: guarding cache saves on `pull_request`. Caches written by a PR are scoped to `refs/pull/N/merge` and readable only by that PR, yet still count against the shared 10 GB. Larger diff, left for a follow-up.


julianoes (commented):
Ok!

PR #3094, opened 2026-09-16
core/ftp: apply the seq check to burst gap-fill reads

julianoes:
## Summary

Burst FTP downloads fail over a telemetry radio. A 128 KiB file from PX4 over a 57600 SiK died at roughly 10% of the way through, every time, with `ProtocolError`. With this change the same download completes byte-identical, three runs out of three, at the default 500 ms timeout.

## Problem

#2850 added a strict `seq_number` check so a stale reply cannot be mistaken for the answer to the current request, and exempted burst transfers, since the server sends many packets with incrementing `seq_number` for one `CMD_BURST_READ_FILE`.

The exemption keys off the type of the work item rather than the opcode of the response. A burst download does not only receive burst packets: on timeout it falls back to `request_next_rest()`, which issues plain `CMD_READ_FILE` reads to fill gaps, and those are ordinary request/response pairs with a predictable `seq_number`. They were therefore left without the check #2850 added, which is the stale-packet corruption it set out to prevent. A late reply then no longer matches `missing_data.front()` and the transfer is failed outright instead of the packet being discarded.

It cannot happen over USB, where the round trip is far shorter than the timeout. Over a radio it is routine: the measured round trip between bursts was around 487 ms against the 500 ms default.

## Solution

Key the exemption off the opcode of the response. The logs then show four to nine stale replies rejected per run, each of which previously ended the transfer.

`Ftp.DownloadBigFileLossy` does not cover this. It drops packets but does not delay them, so a request never times out with a reply still in flight; reproducing it needs simulated latency rather than loss. I have not added such a test.

Tested against PX4 on an FMU-v6C over a 57600 SiK radio, and the existing 30 FTP system tests pass. Note that the radio side also needs a PX4-side fix to stop the burst overrunning the link, which is not in this PR; this change is what makes the transfer survive rather than abort.

The two example commits are independent and can be dropped: `ftp_client` demonstrated the lockstep path rather than burst, which hides this entirely, and `params` could pick a SiK radio's own system instead of the autopilot and then report that it found none.


julianoes (commented):
Yes, reviewed.

PR #3096, opened 2026-09-17
py: raise a typed error per plugin

julianoes:
Stacked on #3007 (targets its branch); retarget to `main` once that is merged.

A failed request in the native Python binding raised a bare `Exception("arm failed: 8")`: there was no type to catch, and the result showed up as a number because `IntEnum` formats as `int` on Python 3.11+. The asyncio plugins defined and documented `<Plugin>Error`, but never raised it.

## Changes

- Each plugin has a `<Plugin>Error`, derived from `mavsdk.MavsdkError`, with `result`, `origin` and `params` attributes. The sync plugins raise it, and the asyncio plugins re-export the same class, so one `except ActionError` covers both APIs. The message reads like MAVSDK-Python's: `COMMAND_DENIED; origin: arm(); params: ()`.
- The asyncio progress streams (calibration, FTP, log download, mission upload/download) raise the error on failure instead of yielding the failed result and stopping. They only yield `NEXT` and `SUCCESS`. The calibration example is updated.
- Fix sync methods returning an enum: they called a lowercase name that does not exist, so `Telemetry.landed_state()`, `vtol_state()`, `flight_mode()` and `Camera.mode()` raised `NameError`.
- The quickstart and migration page name the error types.

The plugin changes are all regenerated from the templates.

## Tested

Against an in-process `Mavsdk` configured as autopilot:
- `arm()` raises `ActionError` with `result == ActionResult.TIMEOUT`, sync and async.
- `calibrate_gyro()` raises `CalibrationError` (`UNSUPPORTED`).
- `Telemetry.landed_state()` returns `LandedState.UNKNOWN`.
- The 8 tests in `py/test` pass.


julianoes (commented):
I think this makes sense.

PR #3042, opened 2026-08-24
shell: allow selecting SERIAL_CONTROL device

Bartok9:
## Summary
Expose SERIAL_CONTROL device selection on shell `Send`/`Receive` (default remains shell).

## Linked
- Fixes #2107
- Depends on mavlink/MAVSDK-Proto#427 (temporary submodule tip until that merges)

## Notes
- Per maintainer: temporary/wrong proto submodule until proto PR merges, then bump here.
- Nested `Receive` message for protoc-gen-mavsdk (single non-Result stream field).
- C++ plugin + mavsdk_server bindings regenerated against that tip.

PR #3097, opened 2026-09-17
v4: C++ migration docs, MAVLink C headers opt-in

julianoes:
Prepares the C++ side of v4 for users upgrading from v3.

## Migration docs

Adds a v4 section to `docs/en/cpp/api_changes.md` covering what breaks when moving from v3 to v4:

- `.h` → `.hpp` headers
- building and linking: C++ standard, `find_package(MAVSDK 4)`, soname, static dependencies, the `cpp/` source directory, dropped packages
- the 32-bit system ID API (IDs above 255 still abort)
- Telemetry `home()`, Shell `send()`/receive, Info `get_flight_information()`, Ftp `ListDirectoryData`
- enums whose numeric values changed with #3078
- the MAVLink C header opt-in and `MavType` (below)
- behaviour changes: abort when a `System`/plugin outlives its `Mavsdk`, the callback drop policy, stderr logging, discovery/connection timing, MavlinkDirect JSON

Also fixes guide pages that no longer matched: logging goes to stderr and uses `log_callback.hpp`, apps need C++17 rather than C++11, and the user callback queue messages changed.

The C++ snippets on the page compile with `-std=c++17` against the current headers.

## MAVLink C headers are opt-in

Including `mavsdk.hpp` no longer pulls in the MAVLink C headers.

- New `MavType` enum, a copy of `MAV_TYPE` with the same values, used by `Configuration::get/set_mav_type()` (was `uint8_t`) and `to_vehicle_from_mav_type()` (was `MAV_TYPE`). `mav_type.cpp` checks every value against MAVLink at compile time and fails if MAVLink adds a type.
- The default component IDs are spelled out as numbers, checked against `MAV_COMP_ID_*` in `mavsdk.cpp`.
- `MavlinkPassthrough` and `intercept_incoming/outgoing_messages_async()` are only available with `MAVSDK_ENABLE_MAVLINK_C_API`, which also brings back the MAVLink C headers. Including `mavlink_passthrough.hpp` without it is an `#error` pointing to MavlinkDirect.
- The library, system tests and fuzzer define it for themselves; Doxygen defines it so the reference keeps documenting those APIs.
- The C binding keeps its `uint8_t` and converts.

Verified against an install of this branch: a C++17 TU including every public header except `mavlink_passthrough.hpp` contains no MAVLink identifiers after preprocessing; with the define, MavlinkPassthrough and `intercept_*` compile. Unit tests (430), the intercept system tests, the mavsdk_server build and the C binding build pass locally. The fuzzer was not built locally.

The regenerated API reference also adds the `Shell::Receive` page that #3042 did not commit, which should fix the docs check on `main`.


julianoes (commented):
Nice nice.

PR #3100, opened 2026-09-18
core: unregister mission work items on the io thread

bansiesta:
## Problem

`Mission.DownloadCancellation` intermittently aborts on the Windows `x86-Debug` leg:

```
Assertion failed: previous_id == std::thread::id() || previous_id == this_id || _io_context.stopped(),
file cpp\src\mavsdk\core\mavlink_message_handler.cpp, line 190
```

The `MavlinkMessageHandler` table is confined to the io thread, and `note_table_thread()` asserts on that. Mission work items break the confinement on cancellation:

- `MissionImpl::cancel_mission_download()` locks the `weak_ptr` it kept for the transfer, so the cancelling thread (a user thread, here the user-callback thread) holds a strong reference to the `WorkItem`.
- The item finishes, and `do_work()` — on the io thread — pops it off `_work_queue`, dropping the queue's reference.
- The cancelling thread now holds the *last* reference, so `~DownloadWorkItem` runs there and calls `_message_handler.unregister_all_on_io_thread(this)` off the io thread.

Debug builds hit the assert; release builds race on the handler table instead. It is timing-dependent, which is why only one Windows leg went red.

## Fix

Unregister the item's message callbacks when it is *retired* in `do_work()`, which always runs on the io thread, rather than in the destructor. The destructor still calls `unregister_messages()` for items that never made it through the queue (e.g. torn down while still queued), and the call is idempotent, so retired items are unaffected.

Applied to both `MavlinkMissionTransferClient` and `MavlinkMissionTransferServer`, which have the same queue shape.

No plugin-side change is needed: the ~13 `.lock()` sites across `mission`, `mission_raw` and `geofence` can keep handing strong references to user threads.

## Test

`MavlinkMissionTransferClientTest.DownloadMissionCancelledFromAnotherThreadIsRetiredSafely` reproduces it directly: cancel from the test thread, let the queue retire the item, then destroy the last reference on a separate thread. It needs a work guard so the `io_context` is not `stopped()`, which is the assert's escape hatch and the state a polled test context would otherwise be left in.

Without the fix it aborts on the assert; with it, it passes.

## Verification

- `unit_tests_runner`: 428 passed (the one failure is `Curl.DownloadFile_ProgressFeedback_COULDNT_RESOLVE_HOST`, which needs DNS this sandbox does not have).
- `system_tests_runner --gtest_filter='Mission.*Cancellation'`: 3 consecutive runs green.
- clang-format 19 clean.


julianoes (approved):
LGTM

PR #3102, opened 2026-09-19
build: use the pinned MAVLink generator on desktop and iOS

YoussGm3o8:
MAVLink header generation currently upgrades Python packages during the build, so the generated headers can depend on packages outside the pinned source graph. Use the pymavlink submodule already recorded by the MAVLink revision instead, and apply one shared patch on desktop and iOS. This also avoids overlapping pip-removal hunks when combining the common and iOS patches.

Adds `cpp/tools/test_mavlink_generator.py`: it clones a supplied local MAVLink checkout and its initialized pymavlink submodule, creates a Python environment without pip, applies the patch, generates ArduPilot C headers, and checks the resulting headers. XML validation remains available when lxml is installed; the no-package regression deliberately exercises generation without it.

Validation: the regression passed on Windows at MAVLink d6a7eeaf43319ce6da19a1973ca40180a4210643; replacing the common patch with upstream main's version made it fail at the pip requirement. Python lint passed. The same dependency patch passed the downstream standalone Windows build and its 414 unit, 114 system, and 33 ArduCopter compatibility tests; downstream Apple/Linux hosted qualification is still running.

Run: `python cpp/tools/test_mavlink_generator.py /path/to/recursive/mavlink-checkout`.

This is an isolated build-reproducibility contribution from the ArduPilot compatibility work tracked in #1568. It does not claim to resolve that broader support issue or add new vehicle behavior.

Replaces #3101 after renaming the source branch; the commit is unchanged.


julianoes (approved):

YoussGm3o8:
Not sure why that test fails, ran it on my fork and seems to have passed. Rerunning it might make it pass?

julianoes:
Unrelated, yes.

PR #3103, opened 2026-09-19
geofence: validate geometry before starting mission transfer

YoussGm3o8:
Reject malformed geofence geometry before converting it into mission items.

The geofence plugin now validates input before upload and returns `InvalidArgument` for:

* polygons with fewer than three points
* nonfinite or out-of-range latitude/longitude values
* nonfinite or nonpositive circle radii
* geofence plans exceeding the 16-bit mission-item count

Validation is performed before mission items are assembled or transferred.

The change also adds unit coverage for the validation rules through the existing `unit_tests_runner`. The tests cover malformed polygons, item-count overflow, invalid coordinates, invalid circle radii/centers, and valid coordinate boundary values.

The validation helper is exposed to the unit-test target using MAVSDK's existing `MAVSDK_TEST_EXPORT` mechanism.

Run the focused tests with:

```bash
./unit_tests_runner --gtest_filter=GeofenceValidation.*
```

Existing geofence upload/download behavior for valid inputs is unchanged.


julianoes (commented):
I think the test would be easier as a unit test, no?

And thanks.

YoussGm3o8:
> I think the test would be easier as a unit test, no?
> 
> And thanks.

Good point, thanks. I made the unit test and removed the old one.

julianoes (dismissed):

julianoes:
The CI issue looks real.

PR #3104, opened 2026-09-20
ci: upload Windows archives from their actual directory

YoussGm3o8:
The Windows workflow creates ZIPs under `cpp/` through its default run directory, while action inputs resolve from the repository root. Combined uploads therefore report success with a missing-file warning and publish no archive; the tagged release paths have the same mismatch.

Use the actual `cpp/` paths for combined uploads and both ZIP release steps, and fail when a combined archive is missing.

The failure is independently observed in [fork run 35474561893](https://github.com/YoussGm3o8/MAVSDK/actions/runs/35474561893): all combined-package jobs completed, but the artifact API contains none of `mavsdk-windows-x86`, `mavsdk-windows-x64` or `mavsdk-windows-arm64`.

With the fix, all 14 jobs in the full Windows workflow [run 35481870500](https://github.com/YoussGm3o8/MAVSDK/actions/runs/35481870500) pass on the first attempt, including Debug and RelWithDebInfo example builds for all three combined packages. The artifact API independently confirms all three nonempty archives; the downloaded x64 ZIP contains both libraries, headers and CMake configuration. `git diff --check` passes. Tag-triggered release publication was not executed.


julianoes (approved):
Nice catch.

PR #3105, opened 2026-09-20
core: match request message results to the request that sent them

bansiesta:
Fixes the flaky `SystemTest.RequestMessageRapid` (e.g. [this x64-RelWithDebInfo run](https://github.com/mavlink/MAVSDK/actions/runs/35474428943/job/106048118750)), which fails with `Telemetry::Result::Timeout`.

### The bug

`MavlinkRequestMessage` matched command results and timeouts back to a work item by **message id alone**. That is not an identity.

A request completes as soon as the requested message arrives. When the autopilot is streaming that message, this is normally *before* the `COMMAND_ACK` for the command that asked for it — so the work item is already erased, and the next request for the same message id is already in flight when the ack turns up. The ack then lands on that innocent successor:

- non-`Success` result → `++it->retries` on a command that request never sent;
- `Success` → `it->timeout_cookie = _timeout_handler.add(...)` over the top of the existing cookie, leaking a timer that can no longer be cancelled. A second later it fires `handle_timeout()`, which also matched by message id, and bumps the retry counter of whichever request is live at that moment.

Five such hits and a request that had nothing wrong with it exceeds `RETRIES` and reports `Timeout`. Because the hits are near-instant rather than spaced by real 1 s timeouts, this fits comfortably inside the ~2 s of hammering the failing CI run had — five genuine timeouts could not have.

### The fix

Give each request a unique id and capture it in the command callback. That callback is already one-per-command, so this needs no plumbing on the `MavlinkCommandSender` side — the identity was there all along, it was just being thrown away in favour of `message_id`.

Also capture the attempt number (the retry count at send time) so a result belonging to an attempt we already gave up on cannot be mistaken for the current one, and `remove()` the old cookie before re-arming.

Results that match nothing are now simply dropped — which is what the comment in `handle_any_message` already claimed happened.

`handle_any_message` is unchanged: msgid + compid is the right match there, since the response genuinely carries no other identity.

### Verification

The clean signal is a work item receiving a **second `Success` ack while still at zero retries** — such an item has only ever sent one command, so a second ack is provably cross-talk from an already-completed request.

| | before | after |
|---|---|---|
| 20 runs under heavy CPU load | 509 | 0 |
| ~190 runs, 24 concurrent processes | 354 | 0 |

`SystemTest.RequestMessageRapid` is green 25/25 locally, and the rest of the system tests are unaffected (the four `Camera.*` failures and the 31 `MissionRaw`/import unit-test failures in my sandbox reproduce identically on unpatched `main` — missing fixture files).

The test's header comment and assertion message described the deferred-unregister race that #2894 already fixed, so they are updated to describe what the test actually catches now.


julianoes (commented):
I think that's right.

julianoes (approved):

PR #3106, opened 2026-09-20
core: make FTP downloads survive a bad link

julianoes:
### Problem

Downloading a file over a telemetry link is more fragile than it needs to be.

The client gives up after 10 timeouts in a row, which at the default 500 ms
timeout is 5 seconds of silence — a fade or a busy autopilot swallows that
routinely, and an otherwise healthy transfer is thrown away. When it does give
up, the caller is told nothing about how far it got.

Burst downloads make it worse. One timeout turns the rest of the file into
239-byte reads, one round trip each, and while those are in flight every burst
packet still arriving is discarded — including the data being re-requested. On
top of that, the server's burst reads and its plain reads share one stream
position, so a gap-filling read lands the burst's next packet on the wrong
bytes under the right offset: silent corruption, and once the position runs off
the end the whole session NAKs.

### Solution

Give up after 60 timeouts of silence rather than 10, back off between retries
up to 8 times the timeout, and report the bytes transferred when a transfer
fails. Both budgets follow the configured timeout, so the default of 500 ms
means 30 seconds of silence and at most 4 seconds between retries. An answer we
discard still shows the link works, so it stops the backoff without counting as
progress.

For bursts, take data from wherever it comes: every packet says where it
belongs, so burst packets and read replies go through one path that writes at
the offset and updates what is still missing. Holes are filled at a part
boundary, where the server is idle and no read can race a running burst, and a
stalled burst resumes as a burst from the lowest missing offset instead of
being demoted to reads. `NAK(EOF)` during a burst means the part is done, which
is how PX4 ends one.

The server now seeks before every burst read, so an interleaved read cannot
corrupt it, and sends bursts in 35 KB parts like PX4 does, so the client has
part boundaries to work with.

### Testing

Over a 57600 baud SiK link to PX4: a 25 second outage in the middle of an 81 KB
download is survived where it used to fail, and an outage past the budget ends
with `Burst download timed out after 21931 of 81192 bytes` instead of a bare
timeout.

New system tests cover a burst spanning several parts, a burst stalled mid
transfer, a read interleaved into a live burst (which fails without the server
fix, with corrupted data), and a burst ending in `NAK(EOF)`. All 119 system
tests and 431 unit tests pass.


julianoes:
Fable review:

  Worth fixing

  1. Late burst packets after completion each send another TERMINATE. In download_burst_continue the "complete" check runs for every absorbed packet,
     including when work.last_opcode is already CMD_TERMINATE_SESSION.
     - Packets still in flight then each call download_burst_end() again with a fresh seq number. This happens, for example, when the last part was resumed
       from a hole and the server re-streams data we already have.
     - The ACK to the first TERMINATE then fails the seq check. The transfer only ends well because the later TERMINATEs get NAK(InvalidSession), which the
       last commit turns into success, or because the timeout path reports success.
     - It works, but only by accident. An early if (work.last_opcode == CMD_TERMINATE_SESSION) return true; at the top of download_burst_continue would fix
       it.
  2. Stale or ignored data counts as progress. work->note_progress() runs in the DownloadBurstItem ACK branch before download_burst_continue decides whether
     to drop the packet.
     - That covers "before the file is open", "past the file size", and duplicate bytes. A peer that keeps streaming junk therefore keeps resetting the 30 s
       give-up budget.
     - The comment on note_link_alive() says the budget should keep running in these cases. I'd move note_progress() to where bytes are actually accepted, or
       where a hole shrinks or current_offset grows.
  3. The retry backoff carries over to the next transfer. _retry_timeout_s is a client member, not part of Work. When a transfer gives up, the value stays set
     at up to 8x the timeout, so the first request of the next work item starts with a 4 s timeout. Reset it when popping the work item or in do_work(), or
     move it into Work.
  4. Server: a NAK inside a burst never ends the burst. The new "seek failed" path in _make_burst_packet, like the existing "read failed" path, returns
     without setting burst_complete.
     - _send_burst_packet() then returns false and the server sends a NAK every 2 ms until the session is reset. The read-failure half predates this branch,
       but this branch adds a second way in.
     - Having _send_burst_packet return true on RSP_NAK would fix both.

  Minor

  - Hole limit: once missing_data.size() > MAX_MISSING_RANGES, every incoming burst packet drives request_burst_next, which sends a new burst request from the
    first hole until the count drops back under 64. It still makes progress, but at one request per packet. It needs more than 64 separate losses within one
    part, so it's rare.
  - Stale burst_complete: a stale packet or NAK(EOF) from an abandoned burst passes the drives_next check, since only the opcode is compared. It can start the
    hole reads while the new burst is still running, which is the race the parts design was meant to avoid. This is harmless because all data is absorbed.
  - Session dropped mid-transfer: NAK(InvalidSession) is only forgiven on TERMINATE. If a server drops the session during an outage shorter than the 30 s
    budget, the next READ or BURST still fails the transfer. Reopening and resuming would be the fix, probably as a follow-up.
  - Overflow: payload.offset + kBurstPartSize in _work_burst can overflow uint32_t near 4 GB. The EOF check above it makes this theoretical.
  - Includes: <algorithm> is included twice in mavlink_ftp_client.cpp.

  I only skimmed the outline of the test file: the test list and its std::call_once captures, which are via shared_ptr. I did not read the tests themselves

